PR #72, opened 2026-05-27
Add visual feedback when Inspector is updated

ericwinger:
This change provides better visual indications when an inspect-it is done so it's more clear that something happened if you're unfamiliar with where to look for the inspector pane. 

Raises the Inspector panel and returns focus to the active editor, shows an amber status bar item and activity-bar badge for 5 seconds on each inspect-it.


ericwinger:
Note that tests were run and passed but gci tests were not run as there is still an outstanding issue related to the port used for gs64ldi in my environment.

PR #80, opened 2026-06-01
Use a modal dialog when asking users to debug

### LINKED ISSUE #68: Displaying workspace expression that raises an error and not closing the notifier window "permanently" removes menu options
@jgfoster - Question for you near bottom. 

In the workspace, Display it on this expression: 
```
Array with: 1 with: 2 with: 3 with: 4 with: 5 with: 6
```
This will open a notifier window in the lower right corner of VS Code, as expected. 
```
GemStone error: a MessageNotUnderstood occurred (error 2010), a Array class does not understand  #'with:with:with:with:with:with:'
```

Now, do not close that window or debug. The menu items in the workspace context menu disappear.  And if you hit the hot key Ctrl-K D on a new expression, you'll get the default VS Code behavior. This is really confusing because the menus disappear for the entire time that notifier is open. I could even get a new session and the menu items were still gone. 

Screen capture demonstration: https://github.com/user-attachments/assets/9b3a9a18-efc0-4e3b-9d07-3171534198b8

-------

Claude's explanation: 
**What was happening:** When GciTsNbPoll (a native GCI call) threw an exception inside a setTimeout callback, JavaScript's event loop couldn't route it back to the Promise. The Promise hung forever, the await pollForResult() in execute() never resumed, and the finally block that resets gemstone.executing to false never ran. That left the context key permanently true, which disabled all the !gemstone.executing keybindings and menus — and VS Code's default Ctrl+K D ("Compare with Saved", which shows "(in file)") took over.

Suggested possible fix.  _(Eric was unsure this was desirable never having seen it before)_ 
**The fix** ([codeExecutor.ts:390-448](vscode-webview://1lfsnpdkjf5usqae89vpj90qkp762ho7ob7g8t22ngne1fb555ll/client/src/codeExecutor.ts#L390-L448)): Wrapped the entire doPoll body in a try/catch so any thrown exception — whether on the first call or a later setTimeout-dispatched call — properly calls reject(e) and settles the Promise. That allows execute()'s finally block to run and reset gemstone.executing.

Another fix is to make the warning dialog modal forcing the user to take an action.  @jgfoster Thoughts on a modal dialog when a code execution error arises? 

After discussions with Claude, it suggested doing both.  

Image

npapagna:
Previously, the debug prompt was non-modal. While the prompt was open, the editor remained in the executing state, causing menu items to stay disabled until the user made a choice.

Make the prompt modal so users must explicitly choose whether to start debugging before interacting with the editor, preventing inconsistent UI state.

TODO: I'll remove the duplicated code to debug an expression in a follow-up pr.

Closes #68 

https://github.com/user-attachments/assets/d70f51b1-d7d3-4d98-b405-625b3e1d6bc6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to the **GemStone Smalltalk** extension will be documented i
 
 ## [Unreleased]
 
+### Changed
+
+  - Use a modal dialog when asking whether to debug an error raised during Display It, Inspect It, or Execute It. This prevents workspaces from becoming stuck in the executing state if the prompt is not answered. 
+
 ### Fixed
 
 - **Databases panel running/stopped indicators are now tied to the correct version.** With two versions installed (e.g. 3.6.2 and 3.7.5) that share a stone/NetLDI name, starting one stone lit up the Stone/NetLDI nodes under *both* versions — and trying to stop the wrong one failed with an incompatible-version error. `ProcessManager.isStoneRunning` / `isNetldiRunning` now match the gslist process version against the database's configured version (`versionsMatch`, prefix-tolerant so a gslist `3.7.4` still matches a `3.7.4.3` install), so each database reflects only its own running processes. The same version guard is applied to the delete / replace-extent safety checks.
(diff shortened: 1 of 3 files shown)


hernanwilkinson:
What is the difference between "Dismiss" and "Cancel" in the modal dialog? Why not just "Cancel" ?

PR #90, opened 2026-06-09
Fix search filter lost on column repopulation

npapagna:
When the server loads new data into a column — for example, selecting a dictionary loads its classes — any active search query should be re-applied to the new items. This broke silently whenever the column reloaded before the user had ever typed in the filter box: the hook that triggers the re-apply only existed after first use.

The filter now finds its list independently of initialization order, so the active query is always re-applied correctly after a reload.

### DIFF
--- a/client/src/systemBrowser.ts
+++ b/client/src/systemBrowser.ts
@@ -1952,7 +1952,7 @@ export class SystemBrowser {
         listEl.appendChild(div);
       }
       
-      listEl.refreshFilter();
+      ListFilter.refreshFilterOf(listEl);
     }
 
     function selectItemInColumn(listEl, value) {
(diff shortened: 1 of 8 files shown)


client/src/systemBrowser.ts
```diff
@@ -1952,7 +1952,7 @@ export class SystemBrowser {
         listEl.appendChild(div);
       }
       
-      listEl.refreshFilter();
+      ListFilter.refreshFilterOf(listEl);
```
npapagna, line 1955: @jgfoster this change is embedded in the html. I was thinking we could follow the web component approach and extract a list web component so we can test it in isolation. That seems like a big change, we can address it in a separate PR if that's the direction we want to go.

PR #104, opened 2026-06-12
Auto-open new compiled methods

npapagna:
Previously, after compiling a new method from the new-method editor,
the tab remained open with no clear path to the resulting method.

https://github.com/user-attachments/assets/e77f84f3-bcc8-46f0-996a-809f22c59618

Now Jasper automatically opens the newly compiled method in a permanent
tab and closes the new-method placeholder. The success message also
shows the actual selector (e.g. "Compiled method Array>>#foo").

https://github.com/user-attachments/assets/14ce0c03-1ab8-488a-b24a-68c610c07741

### DIFF
--- a/client/src/gemstoneFileSystemProvider.ts
+++ b/client/src/gemstoneFileSystemProvider.ts
@@ -110,16 +110,23 @@ export function buildNewMethodUri(
   category: string,
   environmentId: number,
 ): vscode.Uri {
+  assertIsValidUriPath('Dictionary name', dictName);
+  assertIsValidUriPath('Class name', className);
+  assertIsValidUriPath('Method category name', category);
+  
   const side = isMeta ? 'class' : 'instance';
-  const envQuery = environmentId > 0 ? `?env=${environmentId}` : '';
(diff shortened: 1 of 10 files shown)


client/src/gemstoneFileSystemProvider.ts
```diff
@@ -292,6 +314,32 @@ export class GemStoneFileSystemProvider implements vscode.FileSystemProvider {
     }
   }
 
+  private compileMethod(parsedMethodUri: ParsedNewMethodUri, sourceCode: string, session: ActiveSession) : vscode.Uri {
+    const result = queries.compileMethod(
+        session, parsedMethodUri.className, parsedMethodUri.isMeta, parsedMethodUri.category, sourceCode, parsedMethodUri.environmentId,
+    );
+    const selector = result.split('>> ')[1]?.trim();
```
npapagna, line 321: This should be refactord in a follow-up pr, so it returns the selector that was compiled. The change is big, as it spreads across tools, the MCP server and Jasper itself.

Issue #118, opened 2026-06-17
Would like GCI tests to use env variables ($GEMSTONE, etc) and not to hard code stone/netldi


ericwinger:
I haven't yet been able to run the Jasper GCI tests and today investigated why.  A lot of the tests have hardcoded stone & netldi paths. 

Example: (gciBrowser.tests.ts) 
const STONE_NRS = '!tcp@localhost#server!gs64stone';
const GEM_NRS = '!tcp@localhost#netldi:50377#task!gemnetobject';

A couple possible solutions, in order of preference: 

1) Be able to login to a server from jasper and just run the gci tests with that stone/netldi. 

2) Be able to set env variables and source them before running them in a terminal.

ericwinger:
Moved to gitlab internal work item: 

https://code.gemtalksystems.com/GemStone/grail/-/work_items/37

PR #120, opened 2026-06-17
Add method override indicators (▲/▼) with clickable hierarchy navigation

ericwinger:
## Summary

Adds a per-selector override indicator to the System Browser method list:

- **▲** — the method overrides a superclass implementation
- **▼** — the method is overridden in a subclass
- both glyphs appear when both apply

The indicators are **clickable**: ▲ opens the superclass-chain implementations, ▼ opens the subclass overrides, reusing the existing method-results UI.

Work item: https://code.gemtalksystems.com/GemStone/grail/-/work_items/21

## How it works

- **One round trip.** The override bits are computed server-side and piggyback on the existing method-list query (`getClassEnvironments`) — each selector token is prefixed with a 0–3 bitmask. No extra server call to show the arrows.
- **Full-chain detection.** `whichClassIncludesSelector:` walks the entire superclass chain (not just the immediate parent) and `allSubclasses` covers all descendants, so an override three levels up/down is still detected.
- **Click navigation** uses a new `hierarchyImplementorsOf` query (up = superclass chain, down = all subclasses), correctly targeting the instance or class side.

## Incidental bug fix

Fixes a latent bug in the shared method-search serialization: a class's home dictionary is now resolved by its **own name key** (`k = v name asSymbol`). Previously any dictionary that merely referenced the class could win — e.g. Python's `#object -> Object` alias sorts ahead of `Globals`, so navigating to `Object` selected the wrong dictionary and failed. This also corrects **Find Implementors / Senders / References** for aliased classes.

## Testability refactor

The webview's arrow rendering and click dispatch were extracted into a standalone `methodListView.js` (mirroring `listFilter.js`) so they can be unit-tested in jsdom instead of being trapped in an inline `<script>` string.

## Tests

**31 new tests**, full client suite green (**1772 passing**):

| Area | Coverage |
|---|---|
| Parser | bit decode/strip, zero-bit omitted, instance **and** class side |
| `hierarchyImplementorsOf` query | up/down walks, instance vs meta targeting, escaping, row parsing |
| Override-bit selection | per-side filtering, ALL-METHODS aggregation, class-side toggle |
| ListFilter | arrows preserved across matched/unmatched filter renders |
| Click forwarding | `showHierarchyImpls` → `gemstone.hierarchyImplementorsOf` with correct context (both sides) |
| Glyph rendering | bit 1→▲, 2→▼, 3→▲▼, click → correct message |

## Not covered (by design)

- **End-to-end chain *depth*** (e.g. an override two levels up resolving correctly against a live image) needs a GCI smoke test against a running stone — deferred.
- The `gemstone.hierarchyImplementorsOf` **command body** (env-loop + dedup + results display) isn't unit-tested; there's no `activate()` command-capture harness in `extension.test.ts`, and it mirrors the existing untested `implementorsOfSelector`. Its novel logic (the query) and the browser→command forwarding are both tested.

## Manual verification

Computed override bits and both query directions were exercised against a live GemStone session during development. Recommend a clean F5 to confirm the glyphs render and arrow-clicks navigate (including the `Collection` → ▲ on `auditInformation` → `Object` case that exposed the dictionary bug).

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -23,9 +23,10 @@ mcp-server/node_modules/**
 !node_modules/koffi/build/**
 
 client/src/**
-# listFilter.js is read at runtime via fs.readFileSync and injected into the webview as a <script> tag.
-# It must ship as a standalone file — it cannot be compiled into the bundle.
+# listFilter.js and methodListView.js are read at runtime via fs.readFileSync and injected into the
+# webview as <script> tags. They must ship as standalone files — they cannot be compiled into the bundle.
 !client/src/listFilter.js
+!client/src/methodListView.js
 server/src/**
 mcp-server/src/**
 
--- a/client/src/__tests__/listFilter.test.ts
+++ b/client/src/__tests__/listFilter.test.ts
@@ -546,4 +546,71 @@ describe('ListFilter', () => {
       vi.useRealTimers();
     });
   });
+
+  // The method list prepends override-indicator arrows (▲/▼) as leading
+  // children of each item. ListFilter rewrites item text on every render, so it
+  // must preserve those arrows rather than wipe them.
+  describe('override-arrow preservation', () => {
+    function makeListWithArrows(
+      id: string, specs: Array<{ selector: string; dirs: Array<'up' | 'down'> }>,
+    ): HTMLDivElement {
+      const list = document.createElement('div');
+      list.id = id;
+      for (const { selector, dirs } of specs) {
+        const div = document.createElement('div');
+        div.className = 'item';
+        div.dataset.value = selector;
+        for (const dir of dirs) {
+          const span = document.createElement('span');
+          span.className = 'override-arrow ' + dir;
+          span.textContent = dir === 'up' ? '▲' : '▼';
+          div.appendChild(span);
+        }
+        div.appendChild(document.createTextNode(selector));
+        list.appendChild(div);
+      }
+      document.body.appendChild(list);
+      return list;
+    }
+
+    it('keeps the arrow and rebuilds selector text on an unmatched (empty) render', () => {
+      makeListWithArrows('items', [{ selector: 'size', dirs: ['up'] }]);
+      const filter = makeFilter('items');
+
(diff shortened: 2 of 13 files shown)


npapagna:
Hi @ericwinger!

About this:

> End-to-end chain depth (e.g. an override two levels up resolving correctly against a live image) needs a GCI smoke test against a running stone — deferred.

I’m currently working on this as part of my UTF-8 fixes, so we should soon have something in place that can be used for that.

ericwinger:
> Hi @ericwinger!
> 
> About this:
> 
> > End-to-end chain depth (e.g. an override two levels up resolving correctly against a live image) needs a GCI smoke test against a running stone — deferred.
> 
> I’m currently working on this as part of my UTF-8 fixes, so we should soon have something in place that can be used for that.

@npapagna Great!   I've held off putting in any gci tests that use a live stone until I can get them to run easily from VS Code or with a script.  (https://github.com/jgfoster/Jasper/issues/118)  

As an aside, I was looking at a number of the Claude-generated mock tests and lamented that they are sometimes rather lacking (e.g. test that we put a method selector anywhere in a string). So being able to easily add and run gci tests that get real values from the server would be wonderful.

npapagna:
> > Hi @ericwinger!
> > About this:
> > > End-to-end chain depth (e.g. an override two levels up resolving correctly against a live image) needs a GCI smoke test against a running stone — deferred.
> > 
> > 
> > I’m currently working on this as part of my UTF-8 fixes, so we should soon have something in place that can be used for that.
> 
> @npapagna Great! I've held off putting in any gci tests that use a live stone until I can get them to run easily from VS Code or with a script. (#118)
> 
> As an aside, I was looking at a number of the Claude-generated mock tests and lamented that they are sometimes rather lacking (e.g. test that we put a method selector anywhere in a string). So being able to easily add and run gci tests that get real values from the server would be wonderful.

Yes, @ericwinger , I noticed that too. I think having a way to test against a real GemStone instance will go a long way toward improving these tests and strengthening the overall test suite.

PR #121, opened 2026-06-18
Open new class definitions when created from the new-class template or from an existing definition

npapagna:
## Summary

 Before this PR, when you saved a class definition — either creating a new class or updating an existing one — the editor didn't do much to acknowledge it. The class would
  compile successfully in GemStone, but you'd be left looking at the same tab you started from, with no clear indication of where the newly compiled class lived.

https://github.com/user-attachments/assets/3bbc0c98-db1d-477f-965d-2e7aa84ed59a

  After this PR, saving a class definition now behaves the same way saving a method does: the editor automatically navigates you to the compiled class's definition tab. When
  creating a new class, the placeholder tab gives way to the real definition. When updating an existing one, the definition tab reflects the result. The workflow feels complete
  rather than leaving you to find your way there manually.

https://github.com/user-attachments/assets/c8484e1f-140b-422a-ac9b-fb216c05c1c5

## Changes

- Extracts `compileClassDefinition` into a dedicated private method on `GemStoneFileSystemProvider`, mirroring the existing `compileMethod` pattern
- Fires `onDefinitionCompiled` after both new-class and existing-class-definition saves, so the extension can open the compiled class's definition tab and close the placeholder
- Moves export mirror sync (`syncClass`) into each compile function so it uses the name returned by GemStone rather than the URI — handles class renames correctly
- Renames `MethodCompiledEvent` → `DefinitionCompiledEvent` for accuracy

## Pending (in FUP PRs)
* [System browser is not refreshed when classes are created](https://code.gemtalksystems.com/GemStone/grail/-/work_items/31)
* [Open the class definition tab below the system browser](https://code.gemtalksystems.com/GemStone/grail/-/work_items/32)
* [Changes in the dictionary in the class definition are not reflected in the URIs](https://code.gemtalksystems.com/GemStone/grail/-/work_items/33)

## Test plan

- [ ] `npx vitest run` passes (1745 tests)
- [ ] Save a new-class definition → placeholder tab closes, new class definition tab opens
- [ ] Save an existing class definition → definition tab refreshes; both old and new tabs open if class name changed (old class still exists in GemStone)

### DIFF
--- a/client/src/__tests__/extension.test.ts
+++ b/client/src/__tests__/extension.test.ts
@@ -203,7 +203,7 @@ describe('handleMethodCompiled', () => {
     (vscode.window.tabGroups as any).all = [];
   });
 
-  it('opens the new uri and then closes the previous uri when isNewMethod is true', async () => {
+  it('opens the new uri and closes the previous tab when the previous URI is a template', async () => {
     const document = { uri, getText: vi.fn(() => '') } as unknown as vscode.TextDocument;
     vi.mocked(vscode.workspace.openTextDocument).mockResolvedValue(document);
     vi.mocked(vscode.window.showTextDocument).mockResolvedValue({} as vscode.TextEditor);
@@ -211,7 +211,7 @@ describe('handleMethodCompiled', () => {
     (vscode.window.tabGroups as any).all = [{ tabs: [previousTab] }];
     vi.mocked(vscode.window.tabGroups.close).mockResolvedValue(undefined as never);
 
-    await extension.handleMethodCompiled({ uri, previousUri, isNewMethod: true });
+    await extension.handleMethodCompiled({ uri, previousUri, previousUriIsTemplate: true });
 
     expect(vscode.workspace.openTextDocument).toHaveBeenCalledWith(uri);
     expect(vscode.window.tabGroups.close).toHaveBeenCalledWith(previousTab);
@@ -222,20 +222,20 @@ describe('handleMethodCompiled', () => {
   it('does nothing when uri equals previousUri (selector unchanged)', async () => {
     (vscode.window.tabGroups as any).all = [];
 
-    await extension.handleMethodCompiled({ uri, previousUri: uri, isNewMethod: false });
+    await extension.handleMethodCompiled({ uri, previousUri: uri, previousUriIsTemplate: false });
 
     expect(vscode.workspace.openTextDocument).not.toHaveBeenCalled();
     expect(vscode.window.tabGroups.close).not.toHaveBeenCalled();
   });
 
-  it('opens the new uri but does not close the previous tab when isNewMethod is false (selector changed)', async () => {
+  it('opens the new uri but does not close the previous tab when the previous URI is not a template', async () => {
     const document = { uri, getText: vi.fn(() => '') } as unknown as vscode.TextDocument;
     vi.mocked(vscode.workspace.openTextDocument).mockResolvedValue(document);
     vi.mocked(vscode.window.showTextDocument).mockResolvedValue({} as vscode.TextEditor);
     const previousTab = { input: new vscode.TabInputText(previousUri) } as unknown as vscode.Tab;
     (vscode.window.tabGroups as any).all = [{ tabs: [previousTab] }];
 
-    await extension.handleMethodCompiled({ uri, previousUri, isNewMethod: false });
+    await extension.handleMethodCompiled({ uri, previousUri, previousUriIsTemplate: false });
 
     expect(vscode.workspace.openTextDocument).toHaveBeenCalledWith(uri);
     expect(vscode.window.tabGroups.close).not.toHaveBeenCalled();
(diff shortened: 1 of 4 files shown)


client/src/gemstoneFileSystemProvider.ts
```diff
@@ -335,9 +342,29 @@ export class GemStoneFileSystemProvider implements vscode.FileSystemProvider {
...
 
+  private compileClassDefinition(uri: vscode.Uri, parsed: ParsedNewClassUri | ParsedDefinitionUri, source: string, session: ActiveSession) {
+    const className = queries.compileClassDefinition(session, source);
+
+    const message = parsed.kind === 'new-class'
+      ? `Class created: ${className}`
+      : `Class definition updated for ${className}`;
+    vscode.window.showInformationMessage(message);
+
+    void this.exportManager?.syncClass(session, parsed.dictName, className);
```
MatiasFernandez, line 356: 💭 The old code had a comment explaining the mirror sync ("Keep the local .gemstone mirror in step… for a new class the name isn't in the URI, so fall back to a debounced full re-sync"). Now that we sync eagerly with the class name GemStone returns — which is the real improvement here, since it's what makes renames sync to the right place — that rationale is gone. Could we keep a short comment on this `syncClass` call so the next reader knows *why* we use the returned `className` rather than `parsed.className`/the URI?

client/src/gemstoneFileSystemProvider.ts
```diff
@@ -335,9 +342,29 @@ export class GemStoneFileSystemProvider implements vscode.FileSystemProvider {
...
+
+    const definitionUri = buildClassDefinitionUri(parsed.sessionId, parsed.dictName, className);
```
MatiasFernandez, line 358: 💭 `syncClass` and `buildClassDefinitionUri` both take `parsed.dictName` (from the URI) but the GemStone-returned `className`. So we now handle a *class* rename correctly, but if the definition's `inDictionary:` differs from the URI's dictionary, both the mirror sync and the opened definition tab point at the wrong dictionary. I see this is already in your Pending list ("Changes in the dictionary in the class definition are not reflected in the URIs") — just confirming that's the deliberately-deferred case and not something you intended to cover here?

client/src/gemstoneFileSystemProvider.ts
```diff
@@ -141,6 +151,12 @@ export interface MethodCompiledEvent{
   isNewMethod: boolean;
 }
 
+export interface ClassDefinitionCompiledEvent {
```
MatiasFernandez, line 154: nit: `MethodCompiledEvent` names its flag `isNewMethod`; this one uses `isNew`. Consider `isNewClass` for symmetry with the sibling event — minor, but it keeps the two parallel.

client/src/__tests__/extension.test.ts
```diff
@@ -242,6 +242,59 @@ describe('handleMethodCompiled', () => {
...
+  beforeEach(() => {
+    vi.restoreAllMocks();
+    vi.mocked(vscode.workspace.openTextDocument).mockReset();
+    vi.mocked(vscode.window.showTextDocument).mockReset();
+    vi.mocked(vscode.window.tabGroups.close).mockReset();
+    vi.mocked(vscode.window.showErrorMessage).mockResolvedValue(undefined);
+    (vscode.window.tabGroups as any).all = [];
+  });
+
+  it('opens the definition uri when uri differs from previousUri', async () => {
```
MatiasFernandez, line 258: 💭 This test passes `isNew: true`, which also exercises the close branch (it just no-ops because `tabGroups.all` is `[]`). Since the dedicated close test below already covers `isNew: true`, would it isolate this case better to use `isNew: false` here? Then each test asserts exactly one behavior.

client/src/__tests__/gemstoneFileSystemProvider.test.ts
```diff
@@ -211,10 +212,154 @@ describe('GemStoneFileSystemProvider', () => {
...
+      provider.writeFile(newClassUri, encode("Object subclass: 'MyClass'"), { create: true, overwrite: true });
+
+      expect(listener).toHaveBeenCalledWith(
+        expect.arrayContaining([
+          expect.objectContaining({ type: 1, uri: newClassUri }),
+        ]),
+      );
+    });
+
+    it('does not fire onClassDefinitionCompiled and sets a diagnostic when new-class compilation throws', async () => {
```
MatiasFernandez, line 261: 💭 Nice that this asserts the listener doesn't fire and the diagnostic is set. Since `compileClassDefinition` throwing happens *before* the `syncClass` call, would it be worth also asserting `em.syncClass` was **not** called on failure? That locks in that a failed compile never desyncs the local mirror — which is the subtle invariant this refactor relies on.

npapagna: Thanks for the catch!

client/src/__tests__/extension.test.ts
npapagna: Thanks for the close read! I see the appeal of keeping each test to one behavior, and I want to explain why I kept `isNew: true` here intentionally.

The scenario this test models is "open the compiled URI when it differs from the previous one" — and both `isNew: true` and `isNew: false` are genuine real-world triggers for that path: creating a class from the new-class template and saving an existing definition that renames the class both produce a URI change. Using `isNew: false` because it currently no-ops the close branch is an implementation detail rather than a semantic choice — if the `isNew: false` path ever gains its own behavior, the test would silently be exercising a different scenario than intended.

The dedicated close test below already pins down the `isNew: true` + close interaction in full, so there's no missing coverage.

client/src/gemstoneFileSystemProvider.ts
npapagna: Thanks for the catch! Added a comment explaining both scenarios.

client/src/gemstoneFileSystemProvider.ts
npapagna: Yes, exactly — it's tracked in the Pending section of this PR. Fixing it properly means compileClassDefinition needs to return the dictionary name alongside the class name,
but it has quite a few callers, so it's more than a one-liner. Since the issue predates this change I'll tackle it in a follow-up PR to keep this one focused.

client/src/gemstoneFileSystemProvider.ts
npapagna: Just pushed a commit to unify this and clarify the language used. Thanks!

MatiasFernandez (commented):
LGTM, but I'll defer to James

PR #125, opened 2026-06-25
Add GemStone integration test infrastructure

npapagna:
## Summary

- Adds shell scripts to download, install, and manage a local GemStone instance for integration testing (`gs-install.sh`, `gs-start.sh`, `gs-stop.sh`, `gs-reset-extent.sh`, `gs-create-test-env-file.sh`, `gs-test-setup.sh`, `gs-config.sh`)
- Adds `useIntegrationTest` vitest helper that manages GCI session lifecycle (login, per-test begin/abort, logout) and a first integration test for `GciLibrary`
- CI health-check workflow runs a matrix job over the versions in `client/.gemstone-integration-releases.json` (currently 3.6.1, 3.6.8, 3.7.2, 3.7.4.3, 3.7.5)

## Test plan

- [ ] `npm run test:setup` completes without errors on Linux or macOS Apple Silicon
- [ ] `npm test` passes (integration tests connect to the running stone)
- [ ] Re-running `npm run test:setup` is a no-op (idempotent)
- [ ] CI health-check workflow passes for all matrix versions
- [ ] Adding a new entry to `.gemstone-integration-releases.json` causes CI to pick it up automatically


client/src/__tests__/gciLibrary.test.ts
```diff
@@ -0,0 +1,24 @@
+import {describe, expect, it} from 'vitest';
+import {GciLibrary} from '../gciLibrary';
+import {useIntegrationTest} from './useIntegrationTest';
+import {OOP_ILLEGAL, OOP_NIL, OOP_TRUE} from "../gciConstants";
+
+describe('GciLibrary', () => {
```
MatiasFernandez, line 6: 💭 Before this PR there was a deliberate split: GCI/stone-requiring tests ran on their own (`test:gci` + `vitest.gci.config.ts`) and were excluded from the default `vitest.config.ts` (`exclude: ['src/__tests__/gci/**']`), so plain `npm test` stayed hermetic. With this change the new integration test sits at the top level of `src/__tests__/`, so the default config picks it up (`include: ['src/**/__tests__/**/*.test.ts']`) — effectively merging stone-requiring and stone-free tests into a single suite/command, and wiring `.test.env` into the shared config does the same.

One concrete consequence: `npm test` now needs a running stone + `.test.env`, so GitLab CI's `test` stage (`npm ci && npm test`, no stone) and anyone on Windows / Intel macOS — or who just hasn't run `npm run test:setup` — will hit a `beforeAll` throw on `new GciLibrary(process.env.TEST_GCI_LIBRARY_PATH!)` (the env var is `undefined`).

What's your vision here going forward — a single unified suite where the stone-dependent tests skip gracefully when there's no env, or keeping the two tiers separate like today? Mostly want to make sure the merge is intentional rather than a side effect of where the file landed.

client/src/__tests__/useIntegrationTest.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    // Wrap each test in a GCI transaction. Always abort (never commit) so
+    // database changes from one test don't leak into the next.
+    beforeEach(() => {
+        gciLibrary.GciTsBegin(session);
+    });
+
+    afterEach(() => {
+        gciLibrary.GciTsAbort(session);
+    });
```
MatiasFernandez, line 76: 💭 The login path carefully checks `err.number !== 0` and throws, but `GciTsBegin` and `GciTsAbort` both return `{ success, err }` and we ignore it here. The whole point of this helper is isolation — "always abort so changes from one test don't leak into the next." If `GciTsBegin` silently fails, the test runs without an open transaction; if `GciTsAbort` silently fails, changes *do* leak into the next test — the exact thing this comment promises won't happen. Should begin/abort assert on `err` the same way login does, so a failed rollback surfaces loudly instead of quietly corrupting the next test?

client/src/__tests__/useIntegrationTest.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export function useIntegrationTest(callback: (gciLibrary: GciLibrary, session: unknown) => void) {
+    let gciLibrary: GciLibrary;
+    let session: unknown;
+
+    // gciLibrary and session are created inside a beforeAll hook, so they don't
+    // exist at call time. The callback fires at the end of that hook, letting
+    // callers assign the values into their own variables before any test runs.
+    beforeAll(() => {
+        gciLibrary = new GciLibrary(process.env.TEST_GCI_LIBRARY_PATH!);
```
MatiasFernandez: 💭 The existing `gci/` integration tests read `GCI_LIBRARY_PATH`, while this new helper reads `TEST_GCI_LIBRARY_PATH` (plus the `TEST_GS_*` vars from `.test.env`). So we now have two integration-test harnesses with two different env-var schemes pointing at the same kind of thing. Is the plan to migrate the `gci/` suite onto this `test:setup` + `useIntegrationTest` infra eventually? If so, a note in CONTRIBUTING that the older mechanism is being superseded would help; if both are meant to coexist, the divergent env-var names will probably trip people up.

client/bin/gemstone-integration-versions.js
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env node
+
```
MatiasFernandez, line 10: 💭 Could you add a short description explaining the purpose of this script?

client/bin/gs-install.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Usage: ./gs-install.sh <version>
+#
+# Downloads and installs GemStone/S 64 Bit into tmp/gemstone/. Both the
```
MatiasFernandez, line 6: Nit:

> Both the downloaded archive and the extracted installation are cached

Are you talking about CI here? Then it may be worth being explicit like this:

```
Both the downloaded archive and the extracted installation are cached in CI using GitHub actions caching.
```

Because the caching aspect is not part of this script on its own.

client/src/__tests__/useIntegrationTest.ts
```diff
@@ -0,0 +1,53 @@
+import {GciLibrary} from "../gciLibrary";
+import {afterAll, afterEach, beforeAll, beforeEach} from "vitest";
+
+type UseIntegrationTestCallback = (gciLibraryToUse: GciLibrary, sessionToUse: unknown) => void;
+
+export function useIntegrationTest(callback: UseIntegrationTestCallback) {
```
MatiasFernandez, line 35: 💭 Could you add a bit of JS doc here? I think it would help agents to choose using this function more often, and also help the casual reader to understand the purpose quickly

client/src/__tests__/useIntegrationTest.ts
```diff
@@ -0,0 +1,53 @@
...
+    let gciLibrary: GciLibrary;
+    let session: unknown;
+
+    // gciLibrary and session are created inside a beforeAll hook, so they don't
+    // exist at call time. The callback fires at the end of that hook, letting
+    // callers assign the values into their own variables before any test runs.
+    beforeAll(() => {
```
MatiasFernandez, line 42: 💭 Since `beforeAll` runs once per file, and vitest runs test files in parallel workers by default, every integration-test file that uses this helper will create its own GciLibrary + GciTsLogin — so once there's more than one such file, we'll have N sessions logging into the stone concurrently, one per file.

It's harmless today (only one file, and every test aborts so sessions stay transaction-isolated), but it could bite later.

The existing gci/ suite already guards against exactly this — `vitest.gci.config.ts` sets `fileParallelism: false` `(+ maxConcurrency: 5)` — but the default `vitest.config.ts` these tests currently run under sets neither. Feels like another reason to put the integration tests on their own config rather than the default one — what do you think?

client/vitest.config.ts
```diff
@@ -1,3 +1,6 @@
+import dotenv from 'dotenv'
+dotenv.config({ path: '.test.env' })
```
MatiasFernandez: 💭 I would suggest adding 

```javascript
dotenv.config({ path: 'test.local', override: true })
```

Which is a standard practice to allow for local overrides. You'll need to add that to `.gitignore` too.

MatiasFernandez: 💭 I think the standard naming for `.env` files scoped by environment type is to call it `.env.test` rather than `.test.env`

MatiasFernandez (commented):
I left you a few comments/questions @npapagna. Great work! 🎉

client/src/__tests__/gciLibrary.test.ts
npapagna: Yes, this is intentional:
* It looks like tests are fast enough to run them, at least from my preliminary experiments.
* I forgot to add a validation: if the .test.env file is missing, i was planning to raise a descriptive error asking users if they forgot to run `npm run test:setup`
* In the future, I'd expect for the object that wraps GCI to live in its own Typescript project and be used as a dependency. In that context, all tests are going to be "integration" tests.

client/src/__tests__/useIntegrationTest.ts
npapagna: Yes, that's a good point. I was planning to do a thorough tests of basic functions (login, logout, being, abort), but these are uses all across the codebase and changing their signature (to throw instead of returning an error code) made this PR way longer that I'd like.

We can do the checks in place atm, and clean them up when these changes are implemented.

client/src/__tests__/useIntegrationTest.ts
npapagna: I understand that the current gci tests are not working (based on something that @jgfoster mentioned). 
Trying to make them work was a bigger scope than what I needed so I chose to isolate this machinery from the existing one. 
Also, `GCI_LIBRARY_PATH` could already be set (AFAIK that's a well known variable) so I chose names that revealed the intent in which they were used.

client/bin/gs-install.sh
npapagna: This is talking about the script's behavior. It won't download or re-install if files are already present.
It is not related to GitHub.

client/src/__tests__/useIntegrationTest.ts
npapagna: I feel the limitation will come from the GemStone side (if any).
What kind of issues do you have in mind?

client/vitest.config.ts
npapagna: Do we really have an use case for it? I don't want to add stuff we don't really need.

npapagna: Yes, nice catch!

client/src/__tests__/useIntegrationTest.ts
MatiasFernandez: yeah, makes sense. An inline check is enough for now to ensure tests are robust and a global cleanup can be done later

client/vitest.config.ts
MatiasFernandez: Eric asked for a way to tweak the env under which GCI tests run. I didn't check what problem he was trying to solve, but that was the main inspiration for it.

client/src/__tests__/useIntegrationTest.ts
MatiasFernandez: GemStone doesn't forbid concurrent sessions — the thread-safe GCI is built for them. The only hard rule is one in-flight call per `GciSession`; separate sessions are independent. So this isn't a GCI-correctness issue. The risk is test isolation + environment, introduced by vitest, not by `gciLibrary.ts`: the default `forks` pool runs files in parallel, and `useIntegrationTest` logs in once per file, so N files = N sessions into the same stone, same user, at once.

**No DB isolation between files (my main worry).** `forks` isolates process memory, which is all vitest assumes files share — but the shared state here is the stone, which it doesn't isolate. So the isolation is illusory. The `begin`/`abort` bracket doesn't close this: it only refreshes *this* session's view and drops *this* session's uncommitted writes — it does nothing about another session's *committed* writes. Intra-session mechanism, inter-session leak. Example: file A commits a 4th entry to a fixture (its own commit, the code under test's, or seeded data); file B asserts the fixture has 3. B's next `begin` picks up A's commit and B fails — passes alone, coin-flip in parallel, fails in whichever file is running when the commit lands, not the one that caused it. Same hazard for class/method compilation, which is global persistent state.

I guess this risk can only occur if we ever allow a commit happen in a test

client/vitest.config.ts
npapagna: That sounds really good to me! Thanks!

client/src/__tests__/useIntegrationTest.ts
MatiasFernandez: As a side-note, regarding test concurrency settings:

**`maxConcurrency` is the wrong knob.** It only caps `test.concurrent` tests *within* a file; it doesn't parallelize plain `it`/`test` and doesn't govern file-level parallelism (`fileParallelism`/`maxWorkers`). The session count is driven purely by file-level parallelism, so it doesn't touch any of this.

**Latent footgun.** All tests in a file share one `beforeAll` session. If in the future someone marks them as `concurrent` they will point at that single session — breaking the one-in-flight-call rule. Synchronous koffi bindings mask call overlap today (so the first symptom is `begin`/`abort` interleaving across tests), but the non-blocking GCI (`GciTsNb*`, already bound) or koffi `.async()` makes it real corruption. 

That's a risk for the future tho, but may be worth documenting it in the vitest config, so in the future someone doesn't try enabling test concurrency. For example, manually disabling testConcurrency + a comment above it

MatiasFernandez: Adding context here that we discussed in a call:

Whenever tests share a session, there are risks associated with shared state, even when tests run sequentially. Those risks are far worse if we ever introduce test concurrency.

Using one session per test is safe, but slower. And if we enable test concurrency, we'll hit the session limit quickly.

## Two options for CI

1. **Prioritize correctness over performance** → one session per test.
2. **Prioritize performance over correctness** → a shared session + no concurrent tests. We think the code won't use `SessionTemps` or similar and We trust no one will add concurrent tests in the future (and we document the risks of the shared session in the helper to orient agents).

In either case, the decision needs to be revisited over time. If we prioritize correctness and things get very slow or costly, that becomes a problem. If we prioritize speed but start seeing test flakiness, that's also a problem.

I'm honestly not sure which I prefer. But what I do know is that for option 2: if vitest isn't running tests in a random order, what we'll get is not just flakiness — we'll get **silently invalid tests**, which is the worst possible outcome.

Option 1 looks more like an **open-door decision** — easy to revert. We can see the impact quickly: when tests start getting slow, we revisit.

Option 2 looks more like a **closed-door decision** — not because it's hard to revert technically, but because if the decision turns out to be wrong, the cost can be hidden:
- Invalid tests you don't notice are invalid, especially since we're not even randomizing test order.
- Time wasted chasing flakiness whose root cause is unclear.

**Separately, regardless of which option we choose:** we should be randomizing vitest test order regardless — to catch any state-sharing between tests that we're not aware of.

npapagna: I believe we could reify this as a login strategy
* login-per-test: used in CI only. Slower but catches tests that depend on each other.
* login-per-test-suite: configured locally by default, faster but has the potential issues if misused (catched by login-per-test in CI)

MatiasFernandez: I like that. And I'm ok if you decide to do that in a separate PR because this is not yet including real tests, and I don't want to keep extending the scope of this work forever.

jgfoster:
I'm working through this and need a bit of help understanding it. I see a couple things that seem to have prompted it. 

First, "Eric asked for a way to tweak the env under which GCI tests run." I vaguely recall Eric pointing out that I have everything set up to use `50377` as the NetLDI port and this isn't very friendly to people who have a different setup. (I run only one NetLDI and it is on my laptop so doesn't conflict with anything else. The convention at GemTalk is that each users has a separate NetLDI so multiple users can run on a shared machine. I believe that when Eric runs Jasper for development, he can configure the NetLDI port to match his environment, but if he tries to run the GCI tests it doesn't work because of the hard-coded port. We should certainly make it easier for Eric to run the GCI tests!

Second, "the current gci tests are not working (based on something that @jgfoster mentioned)." I had forgotten that issue and went back to look at things. You are right that `package.json:1276` does not run the GCI tests. I tried running them by hand and saw a number of failures (including `GCI_LIBRARY_PATH` was not set). I believe that the GCI tests are the only ones that actually require a running system (am I right?); the rest are run against "mocks" (as unit tests should do). The GCI tests are really "integration" tests, and would require much more infrastructure to run in a pipeline (perhaps a worthy goal, but not done now). When I set `GCI_LIBRARY_PATH` to `~/Documents/GemStone/GemStone64Bit3.7.5-arm64.Darwin/lib/libgcits-3.7.5-64.dylib`, I still got an error. That error has been [fixed](https://github.com/jgfoster/Jasper/commit/9e0d2dd9e980f369acff5746b3d3727478270951), so now the GCI tests work (as long as I have a Stone and NetLDI).

So for now I think it would be useful to have the GCI tests run more easily, but I'd like to better understand the purpose of this merge request. Jasper already has tools for downloading and running GemStone. Are the shell scripts for automated tests? Manual tests? End users? I see some discussion that suggest answers to my questions, but it would be helpful to have it summarized for me. (And I do believe that getting the GCI tests into CI would be good, particularly since I didn't know they weren't being run and there were errors when I finally did get them running!)

npapagna:
> I'm working through this and need a bit of help understanding it. I see a couple things that seem to have prompted it.
> 
> First, "Eric asked for a way to tweak the env under which GCI tests run." I vaguely recall Eric pointing out that I have everything set up to use `50377` as the NetLDI port and this isn't very friendly to people who have a different setup. (I run only one NetLDI and it is on my laptop so doesn't conflict with anything else. The convention at GemTalk is that each users has a separate NetLDI so multiple users can run on a shared machine. I believe that when Eric runs Jasper for development, he can configure the NetLDI port to match his environment, but if he tries to run the GCI tests it doesn't work because of the hard-coded port. We should certainly make it easier for Eric to run the GCI tests!
> 
> Second, "the current gci tests are not working (based on something that @jgfoster mentioned)." I had forgotten that issue and went back to look at things. You are right that `package.json:1276` does not run the GCI tests. I tried running them by hand and saw a number of failures (including `GCI_LIBRARY_PATH` was not set). I believe that the GCI tests are the only ones that actually require a running system (am I right?); the rest are run against "mocks" (as unit tests should do). The GCI tests are really "integration" tests, and would require much more infrastructure to run in a pipeline (perhaps a worthy goal, but not done now). When I set `GCI_LIBRARY_PATH` to `~/Documents/GemStone/GemStone64Bit3.7.5-arm64.Darwin/lib/libgcits-3.7.5-64.dylib`, I still got an error. That error has been [fixed](https://github.com/jgfoster/Jasper/commit/9e0d2dd9e980f369acff5746b3d3727478270951), so now the GCI tests work (as long as I have a Stone and NetLDI).
> 
> So for now I think it would be useful to have the GCI tests run more easily, but I'd like to better understand the purpose of this merge request. Jasper already has tools for downloading and running GemStone. Are the shell scripts for automated tests? Manual tests? End users? I see some discussion that suggest answers to my questions, but it would be helpful to have it summarized for me. (And I do believe that getting the GCI tests into CI would be good, particularly since I didn't know they weren't being run and there were errors when I finally did get them running!)

Thanks for the detailed context, @jgfoster. Let me clarify the goal of this PR first, then address your specific questions.

What this PR is for: This is purely a development tool, not something end users interact with. The goal is to be able to run GCI integration tests automatically in CI, against multiple GemStone versions, with a setup that any developer can also run locally with a single command. The motivation is concrete: I've been finding serious bugs in the GCI layer that only show up against a real running instance, and without CI coverage there's no safety net for that. You can see it running here: https://github.com/jgfoster/Jasper/actions/runs/28257187131?pr=125

On the shell scripts vs. VersionManager: Jasper already has code to manage GemStone downloads, but it's coupled to the VSCode runtime and can't be called from CI without significant refactoring. The scripts are the simplest path to get the job done without taking on that scope.

On Eric's NetLDI port: That's exactly what this PR addresses. The scripts connect by NetLDI name by default, but the NRS strings written to .env.test support both name and port. To use a different setup, a developer can create a .env.test.local alongside it to override any value. It's gitignored and takes precedence automatically.

On the GCI tests: The tests weren't wired into npm test, which made them easy to overlook. This PR adds test:setup to give every developer a single command to get a working GemStone instance configured for tests. Getting the existing GCI tests into CI is a natural follow-up once this infrastructure is in place, and I'm happy to take that on as a separate PR after this is merged.

Happy to walk through any part of it if it would help.

jgfoster:
That is excellent. And now I see that there are tests running in the pipeline; thanks! My only remaining question is whether it is ready to merge. Since the tests pass I'm going to follow @MatiasFernandez's suggestion that other things can come in a separate PR. Thanks!

PR #132, opened 2026-06-30
Test infrastructure improvements and dev docs

npapagna:
## Summary

**Running tests locally depended on your machine setup.** If you already had GemStone installed (i.e. `GEMSTONE_GLOBAL_DIR` was set in your shell), the integration test harness would abort instead of run. The harness now saves, overrides, and restores that variable around each suite, so tests work regardless of local environment.

**Tests always ran in the same order.** That made hidden ordering dependencies invisible — a test could accidentally rely on side effects from a previous one and nobody would notice until something broke unexpectedly. All three workspaces now shuffle test order on every run, surfacing these dependencies early. Vitest prints the seed at the top of each run so any failure can be reproduced exactly. The `test:shuffle:*` opt-in scripts are removed since they're no longer needed.

**Getting started required reading past a lot of noise.** CONTRIBUTING.md mixed step-by-step setup instructions with edge-case reference material, making it hard to follow for a first-time contributor. It now opens with a clear numbered setup flow and surfaces things a new developer needs to know upfront: that `test:setup` only works on Linux and macOS (Apple Silicon), that re-running it is destructive, and how `test:gci` differs from `npm test`. Advanced topics (custom GemStone instance, CI matrix) are moved to their own sections.

**Claude Code had no orientation to the codebase.** A new CLAUDE.md gives Claude Code a complete picture of the repo: workspace layout, key subsystems, testing conventions, and architecture notes. AI-assisted development now works much better out of the box.

## Test plan

- [ ] Run `npm run compile && npm test` — full suite passes.
- [ ] Run tests with `GEMSTONE_GLOBAL_DIR` set in your shell — confirm the suite no longer aborts.
- [ ] Run `npm test` twice — confirm test output shows a different file order each time, with a seed printed at the top.


ericwinger:
@jgfoster "I" have reviewed the pull request and checked out the branch & verified having GEMSTONE_GLOBAL_DIR set no longer inhibits the test run and that I can remove the shuffle from my pre-push file and still get shuffled tests. I think it's ok to merge pending @MatiasFernandez 's analysis. As soon as it's merged, I'll make the changes in my environment. 

@npapagna Claude noted this minor bug. Could be worth tweaking: 

If the developer had no GEMSTONE_GLOBAL_DIR set originally (originalGemstoneGlobalDir === undefined), teardown leaves process.env.GEMSTONE_GLOBAL_DIR === "undefined" (the literal 4-letter string) instead of removing it. To truly restore the "unset" state you have to delete it:

```
function restoreGemstoneGlobalDir() {
    if (originalGemstoneGlobalDir === undefined) {
        delete process.env.GEMSTONE_GLOBAL_DIR;
    } else {
        process.env.GEMSTONE_GLOBAL_DIR = originalGemstoneGlobalDir;
    }
}
```

CLAUDE.md
```diff
@@ -0,0 +1,97 @@
...
+- **Utilities** (`utils/`) — `DocumentManager` (LSP text-document lifecycle), `WorkspaceIndex` (cross-file symbol index), `ScopeAnalyzer`, `AstUtils`
+- Format-specific parsers: `tonel/tonelParser.ts` and `topaz/topazParser.ts`
+
+### mcp-server workspace (`mcp-server/src/`)
+
+A standalone Node.js process that can run as stdio, SSE, or proxy transport. `tools.ts` registers MCP tools; `mcpSession.ts` wraps a GCI session for AI tool calls. Used when Claude Code or Claude Desktop connects directly (as opposed to the in-extension MCP server).
+
+## Tests
+
+`client/__tests__/CLAUDE.md` contains the authoritative test authoring rules — read it before writing or editing tests.
```
MatiasFernandez: This path is missing `src/` — the file actually lives at `client/src/__tests__/CLAUDE.md` (everywhere else in this PR the test paths are under `client/src/__tests__/`). Since the whole point of this doc is to orient Claude Code, an agent following this pointer would hit a dead path on its first hop into the test rules. Mind fixing it to `client/src/__tests__/CLAUDE.md`?

client/src/__tests__/useIntegrationTest.ts
```diff
@@ -120,4 +131,16 @@ export function useIntegrationTest(callback: UseIntegrationTestCallback) {
...
+    function configureGemstoneGlobalDir() {
+        // Vite only exposes VITE_-prefixed variables to test code. GemStone
+        // expects GEMSTONE_GLOBAL_DIR (no prefix), so we copy the VITE_ variant
+        // over for the suite's duration and restore the original value afterward.
+        originalGemstoneGlobalDir = process.env.GEMSTONE_GLOBAL_DIR;
+        process.env.GEMSTONE_GLOBAL_DIR = process.env.VITE_GEMSTONE_GLOBAL_DIR;
+    }
+
+    function restoreGemstoneGlobalDir() {
+        process.env.GEMSTONE_GLOBAL_DIR = originalGemstoneGlobalDir;
```
MatiasFernandez: 💭 When `GEMSTONE_GLOBAL_DIR` wasn't set in the shell, `originalGemstoneGlobalDir` is `undefined`, and `process.env.GEMSTONE_GLOBAL_DIR = undefined` stores the literal string `"undefined"` rather than unsetting the variable — Node coerces every `process.env` value to a string. So on a clean machine (the case this PR is meant to enable) the "restore" leaves it as `"undefined"` instead of putting it back the way it was.

The practical impact is small since each suite re-sets it before login, but it doesn't quite match the "restore the original value afterward" intent in the doc comment above. Worth guarding the unset case?

```ts
function restoreGemstoneGlobalDir() {
    if (originalGemstoneGlobalDir === undefined) {
        delete process.env.GEMSTONE_GLOBAL_DIR;
    } else {
        process.env.GEMSTONE_GLOBAL_DIR = originalGemstoneGlobalDir;
    }
}
```

CONTRIBUTING.md
```diff
@@ -1,44 +1,51 @@
...
    ```sh
-   npm test            # unit suites + one integration test (gciLibrary.test.ts)
-   npm run test:gci    # the rest of the GCI integration suite (client/src/__tests__/gci/**)
+   npm run test:setup
    ```
 
-   `npm test` excludes `client/src/__tests__/gci/**` — run those with `test:gci`. It does, however, include the one integration test `gciLibrary.test.ts` (it lives outside the `gci/` folder), which still needs a running stone.
+   This provisions a local GemStone instance and writes `.env.test` with the connection details. To pin a specific version instead of the latest, pass it as an argument: `npm run test:setup -- 3.7.2`.
+
+   **Supported on Linux and macOS (Apple Silicon) only — Windows is not yet supported.**
```
MatiasFernandez, line 40: 💭 Since `test:setup` is Linux/macOS-only, Windows contributors can't provision a stone at all — so for them `npm test` will always have one failing file (`gciLibrary.test.ts`, the lone integration test in the default run). The old "Running tests without a stone" section explained that this single failure is expected and not a regression; this PR removes it entirely.

Dropping the `GEMSTONE_GLOBAL_DIR`-abort note makes sense now that the harness handles that variable, but should we keep a short "what to expect without a provisioned stone" line so a Windows contributor (or anyone who skips step 4) doesn't read that one failure as a broken checkout?

package.json
MatiasFernandez: 💭 I think there's room for improvement in a separate PR about the workflow to handle the lifecycle of the test gemstone server. 

Running `npm run test:setup` works well for the initial dev env setup and automatically starts the gemstone "test" server. But restarting the computer won't boot it again, and you'll get failures when trying to run the tests. It could be great to have package.json commands to handle the lifecycle of the installed version, for example, to start/stop the server. 

In a project with, for example, PostgreSQL. We can either have the service installed as a OS service that starts automatically, or you can start it with docker-compose, etc. I know there are scripts to deal with the start/stop, but I think the UX is not the easiest one. That's why I think having `test:server:start` could be useful.

client/src/__tests__/useIntegrationTest.ts
MatiasFernandez: That was a Claude comment, but apart from that, I'm wondering if we actually need to restore anything. You can't affect the original shell ENV var through modifying `process.env`. It just includes a copy of the original env during the lifetime of this process (the node process running the tests). Given this is a setup that's handled at the beginning of running the tests, what's the cost of not reverting to "the original value"?

CLAUDE.md
```diff
@@ -0,0 +1,97 @@
+# CLAUDE.md
```
MatiasFernandez, line 1: 💭 I think this is a good start, but I have a few ideas to improve it. I'm ok merging as-is now, and I'll open a PR with tweaks later

package.json
npapagna: yes, we could surface `npm run test:server:start` and `stop`. We have the scripts.

I feel it's much simpler to have one script to do it all.
If you never installed the gemstone test instance, `npm run test:setup` will do that for you.
If it's stopped for any reason (as you mentioned, rebooting is a great example) or if it is non-responsive, `npm run test:setup` will fix that for you using a single command.

Remember that we still have access to these commands in bin/ (though they might not be as accessible as having them in the npm script).

What's the use case you have in mind?

client/src/__tests__/useIntegrationTest.ts
jgfoster: It sounds like setting the value at the beginning of the test with a comment that it applies to this process only, not the original shell ENV, would be sufficient. Then remove the restore code altogether.

CONTRIBUTING.md
npapagna: Good catch! When integration tests fail without a provisioned stone, the error message is pretty self-explanatory:                                                            

```                                                                                                                                                                                
  -----------------------------------------------------------------------------------------                                                                                     
  Integration test initialization failed.                                                                                                                                       
                                                                                                                                                                                
  A common cause is a missing or misconfigured test environment.                                                                                                                
                                                                                                                                                                                
  If you haven't already, try running `npm run test:setup`.                                                                                                                     
  This installs the required test GemStone instance and creates or updates the `.env.test` file.                                                                                
                                                                                                                                                                                
  If the environment is already set up, refer to the original error below for more details.                                                                                     
  -----------------------------------------------------------------------------------------                                                                                     
```
                                                                                                                                                                                
 As for Windows, `npm run test:setup ` doesn't support it yet (this is the first iteration of the integration test infrastructure), so Windows devs won't be able to provision a stone natively. WSL might work, though I haven't tested it. That said, we don't have anyone on Windows at the moment, and I'd rather not add documentation for a workaround that doesn't exist yet. If/when a Windows contributor shows up, we can tackle that together then.                                                                             
                                                                                                                                                                                
My instinct is to keep things minimal for now and encourage people to actually run the tests rather than skip them. Happy to revisit if there's a concrete use case I'm missing!

client/src/__tests__/useIntegrationTest.ts
npapagna: I was telling @MatiasFernandez during a call that probably the best thing would be to restore the env var. 
Chances are low, but if we're trying out ideas that need to read from `process.env.GEMSTONE_GLOBAL_DIR`, if we do that after a `useIntegrationTest` call, then the _expected_  `process.env.GEMSTONE_GLOBAL_DIR` value will be left overriden by `useIntegrationTest`.

Is a bit more code, but I feel doing a proper cleanup is the right thing to do here to avoid masking bugs.

CLAUDE.md
npapagna: Great catch! Thanks!

package.json
npapagna: Ended up refactoring things a bit (and updating docs). Now we have:

* npm run test:server:start
* npm run test:server:stop
* npm run test:server:list

See commit for more info.

client/src/__tests__/useIntegrationTest.ts
MatiasFernandez: We discussed this IRL, but now I understand the advantage of having this, and the new approach I think it's correct and makes sense.

MatiasFernandez (approved):
All good from my point of view. I tested this locally and it worked great, the new commands for stopping and listing are very helpful.

I delegate to James for the final confirmation before merging, but I think this is good

PR #133, opened 2026-07-01
GCI tests: use .env.test connection instead of hardcoded stone/netldi (work item #37)

ericwinger:
## What & why

The on-demand GCI test suite (`client/src/__tests__/gci/**`) hardcoded the Stone and NetLDI (`gs64stone` / `netldi:50377` / `gs64ldi`), so it only ran against one developer's local stone — anyone else's run silently connected to the wrong stone (or nothing) and cascaded into skips. This makes the suite read its connection the same way the automatic integration tests do, so `npm run test:server:start` is all you need to run it.

Addresses [grail #37](https://code.gemtalksystems.com/GemStone/grail/-/work_items/37).

## Changes

- **Shared connection config** — new `client/src/__tests__/gci/gciTestConfig.ts` sources the connection from the `VITE_GEMSTONE_*` vars in `.env.test` (written by `npm run test:server:start`), with `GS_*` / `GCI_LIBRARY_PATH` shell fallback and a `requireEnv()` helper that fails fast with an actionable message. It also copies `VITE_GEMSTONE_GLOBAL_DIR` → `GEMSTONE_GLOBAL_DIR` so the GCI library can locate the provisioned stone (this was the actual cause of the login failures). All hardcoded stone/netldi names are removed from the 18 gci test files and `queryHarness.ts`.

- **Vitest projects** — the on-demand suite is now a project named `gci` in `client/vitest.config.ts`, alongside `unit`. `npm test` runs `--project unit` and `npm run test:gci` runs `--project gci`, so the GCI tests are **discoverable/runnable in the VS Code Testing panel** (tagged `[gci]`) but stay **out of the automatic `npm test`** run. The separate `vitest.gci.config.ts` is removed.

- **Image-agnostic cap test** — `querySunit.smoke.test.ts` no longer assumes the image has >100 `TestCase` subclasses; it asserts the "too many to run" throw only when the class count exceeds `MAX_RUN_CLASSES`, and otherwise runs the available classes. `MAX_RUN_CLASSES` is exported for the test.

- **Trace-log hygiene** — the `GciTsGemTrace` test turns tracing on, which makes the GCI library write `gci<pid>trace.log` into the working directory. These are now removed on a clean run but **kept on failure** for debugging (`globalSetup` + `setupFiles` + an on-disk marker). `gci*trace.log` and `.gci-trace-keep` are gitignored.

- **Unit test** — added `removeTraceLogs` coverage (`client/src/__tests__/gciTraceGlobalSetup.test.ts`), runs in the default suite with no stone.

- **Docs** — updated `CLAUDE.md` to describe the `gci` project and `.env.test` wiring.

## Testing

- `npm test` (no stone-specific setup beyond a provisioned test stone): server 320 / client 2301 / mcp 95 — green.
- `npm run test:gci`: 21 files / 222 tests — green, auto-targeting the provisioned stone.
- Verified trace-log cleanup end-to-end: clean run leaves 0, a failing run keeps the log, and the next clean run clears the stale log + marker.

## Notes

- No product behavior change: the only non-test edit is exporting `MAX_RUN_CLASSES` (visibility only); the cap logic is already unit-tested in `sunitQueries.test.ts`.
- CI is unaffected — it doesn't run the on-demand suite.


client/src/__tests__/gci/gciTestConfig.ts
```diff
@@ -0,0 +1,79 @@
...
+// These tests talk to a live GemStone, exactly like the automatic GCI tests
+// (see ../useIntegrationTest.ts), so they read their connection from the same
+// place: the VITE_GEMSTONE_* variables that vitest loads out of .env.test,
+// which `npm run test:server:start` generates. Once a test stone is
+// provisioned, `npm run test:gci` targets it with no extra setup.
+//
+// Plain GS_* / GCI_LIBRARY_PATH shell variables are honored as a fallback, so a
+// custom stone can be targeted without touching .env.test. There are no
+// hardcoded Stone or NetLDI names — every value comes from the environment, and
+// a missing value fails fast with an actionable message.
```
npapagna, line 12: It looks like the GS_* / GCI_LIBRARY_PATH fallback is meant to support a specific setup beyond `npm run test:server:start`. What does that workflow look like?

client/vitest.config.ts
```diff
@@ -1,16 +1,44 @@
...
+        test: {
+          name: 'unit',
+          include: ['src/**/__tests__/**/*.test.ts'],
+          exclude: ['src/__tests__/gci/**'],
+          setupFiles: [
+            'src/__tests__/vitest.windowSetup.cjs',
+            'src/__tests__/vitest.uriSetup.ts',
+          ],
+          sequence: {
+            shuffle: true,
```
npapagna, line 79: The gci project doesn't shuffle test order, was that intentional?

client/src/__tests__/gci/querySunit.smoke.test.ts
```diff
@@ -173,13 +173,25 @@ describe('SUnit queries (live GCI)', () => {
...
+    // The blocking-call guard, end-to-end. The no-args path runs every
+    // discovered TestCase subclass through one synchronous, un-interruptible
+    // GCI call — before the cap, THIS is the call that ran the entire image and
+    // hung the smoke run. Correct behavior depends on image size, so the test
+    // asks the live stone how many classes there are and checks the matching
+    // guarantee: a large image (over the cap) must fail fast with a "narrow it"
+    // error rather than wedge the session; a small image (within the cap, e.g.
+    // a freshly provisioned test stone) simply runs what's there and returns
+    // results.
+    it('limits how many test classes a single run executes', () => {
```
npapagna, line 185: Nice!

client/package.json
```diff
@@ -15,7 +15,8 @@
     "vscode-uri": "^3.1.0"
   },
   "scripts": {
-    "test": "vitest run",
+    "//test": "Runs only the 'unit' vitest project (see vitest.config.ts). The 'gci' project is on-demand (needs a live stone) and is excluded here on purpose; run it with `npm run test:gci`. Without --project, vitest would run both projects.",
```
npapagna: What's the thinking behind keeping the gci project out of `npm test`?
Now that these are running reliably against the test stone (btw that is awesome!), what would it take to get them into the default run?

client/vitest.config.ts
```diff
@@ -1,16 +1,44 @@
...
   test: {
-    include: ['src/**/__tests__/**/*.test.ts'],
-    exclude: ['src/__tests__/gci/**'],
-    setupFiles: [
-      'src/__tests__/vitest.windowSetup.cjs',
-      'src/__tests__/vitest.uriSetup.ts'
+    projects: [
+      {
+        test: {
+          name: 'unit',
```
npapagna: Nit: What name fits better than 'unit' for a project that also runs integration tests? Wondering if something like 'default' would be clearer.

client/vitest.config.ts
```diff
@@ -1,16 +1,44 @@
...
+      {
+        test: {
+          name: 'gci',
+          include: ['src/__tests__/gci/**/*.test.ts'],
+          maxConcurrency: 5,
+          fileParallelism: false,
+          // Removes gci<pid>trace.log written by the GciTsGemTrace test on a
+          // clean run; keeps them when a test failed (for debugging).
+          globalSetup: ['./src/__tests__/gci/gciTraceGlobalSetup.ts'],
+          setupFiles: ['./src/__tests__/gci/gciTraceCleanup.ts'],
```
npapagna, line 97: The trace log setup and cleanup are wired in as global setup/setupFiles for the whole gci project, but only one test actually calls GciTsGemTrace (if I got that right). What was the thinking there, could that test handle its own cleanup?

package.json
```diff
@@ -1299,7 +1299,8 @@
     "test:server": "cd server && npx vitest run",
     "test:client": "npm test --workspace client",
     "test:mcp": "cd mcp-server && npx vitest run",
-    "test:gci": "cd client && npx vitest run --config vitest.gci.config.ts",
+    "//test:gci": "Runs only the on-demand 'gci' vitest project (client/vitest.config.ts); needs a live test stone (npm run test:server:start). `npm test` runs the 'unit' project and excludes these.",
+    "test:gci": "cd client && npx vitest run --project gci",
```
npapagna, line 1303: Now that these run reliably against the provisioned stone, what would it take to add them to the CI run in .github/workflows/health-check.yml?

npapagna (commented):
LGTM 🚀 Having the GCI tests running against the provisioned stone is a great step forward!                                                                          

There's some overlap with `useIntegrationTest` that I'll take care of on my end when I'm done with the changes I have in progress there, no action needed from you.          

Left a couple of questions inline, nothing blocking. One follow-up I'd love to get to: adding these to the CI run in .github/workflows/health-check.yml.  

I'll defer to @jgfoster for the final pass.

Thanks @ericwinger!

client/src/__tests__/gci/gciTestConfig.ts
ericwinger: It would allow (with a little work)  for the current env vars to be used if they are set. So you could run against your current development stone (might be nice if you're doing server work and want to test jaspers gci test against it) .  With a little more work, I'd like to be logged into a stone and just be able to say 'run gci tests against thist stone that I'm logged into in a new session'.  We could add a work item for those two things if this doesn't get us far enough.

client/vitest.config.ts
ericwinger: That's an oversight.  I'll change that.

client/package.json
ericwinger: This was my call, but I'm not beholden to it.  I got the sense that we wanted two levels of testing - an always-fast smoke test like npm test does now with the one gci test.  And, a second level deeper set of tests against a live server which might run slower (as we add many more gci tests). 

If everyone thinks we should roll all the gci tests into npm test, and require that a live stone is always present, then I'm happy to change this.

client/vitest.config.ts
ericwinger: default it is!

client/vitest.config.ts
ericwinger: I'm going to defer to claude for this answer as we had some problems getting this to run.  Here's Claude's explanation. We could try to give that test it's own session but may not be the most reliable. (the overall goal is to help keep our test running environment from building up cruft) 

"Good question — it's intentional. It can't easily live in that one test for two reasons. First, the trace log is flushed on session close, which for our shared session happens at worker exit, after that test's own afterAll has run — so a cleanup inside the test fires too early to catch the file. That's why cleanup runs in every file's afterAll; the last one reliably clears the late-flushed log. Second, keep-on-failure needs a run-wide signal, and vitest isolates each file's module context, so an in-memory flag won't cross files — hence the on-disk marker. Localizing it would mean giving the test its own session it closes itself and only keeping the log on that test's own failure; I'd rather keep the current wiring, which is documented in the cleanup files."

package.json
ericwinger: Yeah, I can do that.

client/vitest.config.ts
ericwinger: And shuffling did catch a genuine test order bug!

ericwinger:
These were good suggestions @npapagna . I've pushed a new commit to address the shuffle oversight, print the seed for reproducibility,  fix a couple shuffle-discovered bugs, rename unit -> default, add run on-demand GCI tests on CI health check. 

Let me know if you find anything else. 

@jgfoster I think it's ready for review & merge.

ericwinger:
Well, actually. Maybe we aren't good.  Got some health check failures to review.

ericwinger:
@jgfoster - I reverted the failing change to the CI health check and all is healthy. 
@npapagna - I opened this work item to address putting on demand gci tests into the health check - https://code.gemtalksystems.com/GemStone/grail/-/work_items/56

client/src/__tests__/gci/gciTestConfig.ts
npapagna: This is really helpful, thanks for clarifying. I didn't know about that use case.

client/vitest.config.ts
npapagna: Thanks!

PR #138, opened 2026-07-02
Gate Enhanced Inspector support to GemStone 3.7.5+

ericwinger:
## Summary

The Enhanced Inspector support is now limited to **GemStone 3.7.5 and later**. On older servers Jasper keeps the classic Inspector and never offers to install the enhanced support. The gate is a single semantic version check, so future releases (3.7.6, 4.0, …) are admitted automatically with no list to maintain.

## Why we exclude pre-3.7.5 servers

The vendored GToolkit/RSR payload cannot deliver a working Enhanced Inspector below 3.7.5. There are **two independent reasons**, which together set the floor at 3.7.5:

### 1. Pre-3.7.0 — the payload won't file in (RSR requirement)

`RemoteServiceReplication`'s `RsrAsyncMournHandler` references `GcFinalizeNotification`, a GemStone **kernel class introduced in 3.7.0**. On 3.6.x that symbol does not exist, so filing in the payload fails to compile (`CompileError 1001, undefined symbol GcFinalizeNotification`) and the install aborts. In other words, RSR itself requires GemStone ≥ 3.7.0 — it is a 3.7-era component (3.7.x even ships RSR as a bundled project).

### 2. 3.7.0–3.7.4 — it installs, but shows no views (Unicode/byte string mismatch)

On these servers the payload files in cleanly, but the inspector comes up empty. The cause is a **server-side** string-encoding behavior, not the payload files (the same files work on 3.7.5):

- On GemStone before 3.7.5, string literals in GCI-compiled queries compile as **Unicode** strings.
- The payload builds its view dictionaries with **byte-`String`** keys.
- These servers **raise `ArgumentError` on a mixed Unicode-vs-byte `String =` comparison**, so the inspector's view lookup (`... at: 'views'`) throws. The error is caught and swallowed, and the inspector renders no views.
- GemStone 3.7.5 compiles those literals as byte `String`, so the comparison is byte-vs-byte and the lookup succeeds.

Because reason (1) blocks everything below 3.7.0 and reason (2) blocks 3.7.0 through 3.7.4, the first release where the full pipeline (install **and** views) works is **3.7.5**.

We deliberately chose to gate on version rather than paper over the platform behavior in the queries (e.g. coercing every dictionary key to a byte `String`): that touches many comparison sites, is fragile, and would only mask a server limitation. Gating is the honest, maintainable fix.

## What changed

- **`supportsEnhancedInspector(stoneVersion)` / `ENHANCED_INSPECTOR_MIN_VERSION`** (`enhancedInspectorInstall.ts`) — the single source of truth. Uses the semantic `compareGemStoneVersions`, extracts the leading numeric token from the raw `GciTsVersion` string (which carries a trailing build suffix), and pads a short `major.minor` form so a future `"4.0"` compares cleanly.
- **Three gates:**
  - `refreshEnhancedInspectorAvailable` — the routing latch returns `false` without even probing on < 3.7.5, so a server that already has the classes (e.g. a shared stone) still falls back to the classic Inspector.
  - `maybeOfferEnhancedInspectorInstall` — no connect-time offer or auto-install.
  - `runInstallEnhancedInspector` — reports "requires GemStone 3.7.5 or later" and stops.
- **`gemstone.enhancedInspectorSupported` context key**, driven by session selection, hides the *Install Enhanced Inspector Support* command in the Command Palette on unsupported servers.
- **Docs/UX** — the walkthrough, the walkthrough step, and the `enhancedInspector.autoInstall` setting description all note the 3.7.5+ requirement.

## Testing

- Version-gate coverage across **unsupported** (3.6.2 / 3.7.0 / 3.7.2) and **supported** (3.7.5 / 3.7.6 / 4.0, including short `major.minor` and build-suffix forms) versions, at every gate: helper, routing latch, offer, auto-install, and manual command.
- The on-demand GCI routing smoke test now uses the stone's **real** version and asserts the gated routing decision holds across the whole CI version matrix (not just 3.7.5).
- `npm run compile` clean; full suite green (server / client / mcp) and the on-demand GCI suite green against a live 3.7.5 stone.

### DIFF
--- a/client/src/__tests__/enhancedInspectorAvailability.test.ts
+++ b/client/src/__tests__/enhancedInspectorAvailability.test.ts
@@ -9,10 +9,11 @@ vi.mock('vscode', () => ({
 import { ActiveSession } from '../sessionManager';
 import { GemStoneLogin } from '../loginTypes';
 import * as queries from '../browserQueries';
+import { refreshEnhancedInspectorAvailable } from '../enhancedInspectorAvailability';
 
 const noErr = { number: 0, message: '', context: 0n, category: 0, fatal: false, argCount: 0, exceptionObj: 0n, args: [] };
 
-function createMockSession(executeFetchData = ''): ActiveSession {
+function createMockSession(executeFetchData = '', stoneVersion = '3.7.5'): ActiveSession {
   const mockGci = {
     GciTsResolveSymbol: vi.fn(() => ({ result: 1000n, err: { ...noErr } })),
     GciTsExecuteFetchBytes: vi.fn(() => ({ data: executeFetchData, err: { ...noErr } })),
@@ -24,7 +25,7 @@ function createMockSession(executeFetchData = ''): ActiveSession {
     gci: mockGci as unknown as ActiveSession['gci'],
     handle: {},
     login: { label: 'Test' } as GemStoneLogin,
-    stoneVersion: '3.7.2',
+    stoneVersion,
   };
 }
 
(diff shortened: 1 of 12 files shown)


jgfoster:
Hm.. This will be an argument against doing development on 3.6.2!

PR #139, opened 2026-07-03
Improve .gitignore and CLAUDE.md context files

MatiasFernandez:
## Summary

Two independent changes bundled together, both aimed at making Claude Code sessions in this repo more effective:

### 1. `.gitignore`: ignore `.claude/` by default, opt in the shared bits

Previously the entire `.claude/` directory was gitignored, which meant any Claude Code config could only ever be a personal, local artifact — there was no way to share rules, settings, or skills with the team via source control.

But most of what lands under `.claude/` is genuinely personal/machine-local runtime state — locks, checkpoints, mailbox, routines, daemon/registry state, and so on. Committing all of it by default would be noise. So rather than un-ignoring everything, this PR keeps `.claude/` ignored by default and **explicitly un-ignores (allowlists) only the curated dirs/files meant to be shared**:

```gitignore
.claude/*
!.claude/rules/
!.claude/skills/
!.claude/commands/
!.claude/agents/
!.claude/output-styles/
!.claude/settings.json
# Personal overrides — never commit even though they sit in a shared tree:
CLAUDE.local.md
.claude/settings.local.json
```

This means nothing new under `.claude/` gets committed by accident — it stays ignored until someone deliberately adds a `!` line for it. **That opt-in should only happen once the setting, tool, or practice is designed to be shared with the team** — experimental rules, skills, or commands you're still trying out stay ignored (or go in `settings.local.json` / a personal path) until they are ready to be shared.

This unblocks the CLAUDE.md restructuring below (rules and per-workspace context files need to be committable).

### 2. Context files: apply progressive disclosure to avoid context pollution

The root `CLAUDE.md` had grown into a full architecture reference — a per-subsystem breakdown of the client, server, and mcp-server workspaces, GCI/native library notes, and detailed test-authoring conventions — all loaded into every agent's context on every session, whether or not the task touched those areas.

The restructuring keeps the root `CLAUDE.md` as a lean top-level map (commands, high-level workspace layout, done-gate) and pushes subsystem-specific detail down to where it's only loaded when relevant. Detail landed in two kinds of places:

**Per-workspace `CLAUDE.md` files** (auto-load for their subtree):
- `client/CLAUDE.md` — the client architecture map, the `queries/` + webview-scripts conventions, and the single-test-file command.
- `server/CLAUDE.md` — the LSP server architecture note.

**Path-scoped rule files** under `.claude/rules/`, each with `paths:` frontmatter so it only loads when matching files are in play:
- `.claude/rules/tests.md` — generic, repo-wide test-authoring conventions (naming, three-part structure), scoped to `**/*.test.ts` and `**/__tests__/**` so they load for any test file anywhere.
- `.claude/rules/client/tests.md` — client-only test material (VS Code API mock, `QueryExecutor`, `systemBrowser` fire-and-forget, seeded ordering, integration tests), scoped to the client workspace.
- `.claude/rules/client/gci.md` — GCI/native-library notes, scoped to `gciLibrary.ts` and the GCI test suite.

The mindset behind all of this: an agent's context window is a scarce, shared resource, and instructions that aren't relevant to the current task are pure noise that crowds out the things that are. Rather than trying to write one document that anticipates every task, structure context so the right slice loads on demand for the task at hand, and let the top-level file stay a map rather than the territory. This is guided by:
- https://claude.com/blog/steering-claude-code-skills-hooks-rules-subagents-and-more — Claude Code's own mechanisms (skills, hooks, rules, subagents) for progressive, task-scoped context loading instead of one monolithic file.
- https://arxiv.org/abs/2602.11988v2 — on context pollution/rot in long agent contexts and why unrelated instructions degrade performance on the task at hand.
- https://www.humanlayer.dev/blog/writing-a-good-claude-md — practical guidance on keeping CLAUDE.md a concise, high-signal map rather than an exhaustive reference.

## Test plan

- [x] Open a fresh Claude Code session in this repo and confirm the root `CLAUDE.md` loads lean (no GCI/class-sync/MCP detail) unless the relevant files are touched
- [x] Work in the `client/` or `server/` workspace and confirm the respective `client/CLAUDE.md` / `server/CLAUDE.md` loads
- [x] Touch `client/src/gciLibrary.ts` and confirm `.claude/rules/client/gci.md` auto-loads
- [x] Touch a file under `client/src/sync/**` or `mcp-server/**` and confirm the relevant header/inline comment is present in the code itself (no rule file to auto-load for these anymore — that knowledge now always ships with the file)
- [x] Touch a test file (e.g. under `client/src/__tests__/**`) and confirm the path-scoped `.claude/rules/tests.md` (and `.claude/rules/client/tests.md` for client tests) auto-loads
- [x] `git status` shows `.claude/settings.local.json` (if present) as ignored and the allowlisted `.claude/rules/*.md` as tracked; confirm a new un-allowlisted file under `.claude/` stays ignored


.gitignore
```diff
@@ -3,16 +3,22 @@ out/
...
 gemstone
 .gemstone/
 tsconfig.tsbuildinfo
 .gemstone
 
+# Claude Code: personal/machine-local overrides, not shared with the team.
+# Other files are considered globally-shareable to the team, so don't use them for personal overrides.
+CLAUDE.local.md
+.claude/settings.local.json
+.claude/worktrees/
```
npapagna: 🤖 _Claude Code finding:_ This allowlist doesn't cover all personal Claude Code runtime files. My checkout's `.claude/` currently has `scheduled_tasks.lock`, which is only kept out of git by this machine's personal `.git/info/exclude` (which also excludes `agent-registry.json`, `checkpoints/`, `mailbox/`, `routines/.state/`, `assistant-daemon-state.json`, etc.) — none of which are in this shared `.gitignore`. On a fresh clone or CI checkout without that personal exclude file, `git add .claude/` would stage local runtime state that should stay personal.

.claude/rules/transcript.md
```diff
@@ -0,0 +1,16 @@
...
+  - "client/src/smalltalkNotebookController.ts"
+---
+
+# Transcript sink (`transcriptSink.ts`)
+
+Jade-style server-side Transcript. At login a small `JasperTranscriptSink` class is compiled into the session (never committed; held via `SessionTemps`, like JadeServer) and installed at `SessionTemps #TranscriptStream_SessionStream` — the stream GemStone's `TranscriptStreamPortable` actually writes to. (An earlier wrapper keyed `SessionTemps #Transcript`, which no supported version consults, so Transcript output was silently lost.)
+
+Two modes:
+
+- **live** (Execute/Display/Inspect It, notebook cells) — each write reaches the client mid-execution as a ClientForwarder send: the in-flight nb call returns GCI error 2336, `settleNbResult` (`nbRunner.ts`) appends the text to the "GemStone Transcript" output channel (revealed on every write), and resumes via `GciTsContinueWithAsync` — a koffi worker-thread binding, so a long run after a write never blocks the extension host.
```
npapagna: 🤖 _Claude Code finding:_ `settleNbResult` is actually defined in `client/src/transcriptSink.ts`, not `nbRunner.ts` — worth fixing so an agent tracing the live-Transcript resume path doesn't end up in the wrong file.

.claude/rules/transcript.md
```diff
@@ -0,0 +1,16 @@
+---
+paths:
+  - "client/src/transcriptSink.ts"
+  - "client/src/codeExecutor.ts"
+  - "client/src/nbRunner.ts"
+  - "client/src/smalltalkNotebookController.ts"
```
npapagna: 🤖 _Claude Code finding:_ This path list misses a couple of files the prose above is actually about: `client/src/transcriptChannel.ts` (implements the "revealed on every write" output-channel behavior this rule documents) and `debuggerPanel.ts`/`debugQueries.ts` (the rule explicitly calls out "debugger stepping" as a buffer-mode consumer, but only `nbRunner.ts` — which both of those files import — is listed, not the files themselves). Editing those files directly won't auto-load this rule.

.claude/rules/class-sync.md
```diff
@@ -0,0 +1,10 @@
+---
+paths:
+  - "client/src/sync/**"
```
npapagna: 🤖 _Claude Code finding:_ This rule's own first sentence says the sync engine is "used by `exportManager.ts`", but `exportManager.ts` lives outside `client/src/sync/` so it isn't covered by this glob — editing the primary consumer of the sync engine won't load this rule.

CLAUDE.md
```diff
@@ -19,82 +19,28 @@ nvm use                  # activate pinned Node version
...
-
-## GemStone documentation
+## Architecture
 
-`docs/3.7/` contains the GCI header files (`gcits.hf`, `gci.ht`, `gcicmn.ht`, `gcits.ht`) — the authoritative reference for GCI function signatures, struct layouts, and constants. Consult these when working with `gciLibrary.ts` or any GCI call.
+`client/src/extension.ts` registers all commands, tree views, and language features, starts the LSP client and MCP socket server, and manages sessions. Major client subsystems: GCI bridge (`gciLibrary.ts`), views (`systemBrowser.ts`, `globalsBrowser.ts`, `enhancedInspector.ts`, `debuggerPanel.ts`, `classBrowser.ts`), class sync (`sync/`), sessions (`sessionManager.ts`, `codeExecutor.ts`), Transcript sink (`transcriptSink.ts` — server-side Transcript with live streaming), infrastructure (`versionManager.ts`, `databaseManager.ts`, `processManager.ts`, `sysadminStorage.ts`), virtual FS (`gemstoneFileSystemProvider.ts`), debugger (`gemstoneDebugSession.ts`, `debugQueries.ts`), WSL support (`wslBridge.ts`, `wslFs.ts`). The `server/` LSP parses Topaz (`.gs`/`.tpz`), Tonel (`.st`), and GemStone Smalltalk (`.gst`) into an AST with per-feature services under `server/src/services/`.
```
npapagna: 🤖 _Claude Code finding:_ The old CLAUDE.md's `server/` workspace section (lexer/parser dirs, the full LSP services list, `utils/` — `DocumentManager`/`WorkspaceIndex`/`ScopeAnalyzer`/`AstUtils`, the Tonel/Topaz parser filenames) doesn't seem to have been relocated anywhere — there's no `.claude/rules/server.md` alongside the other five new rule files, and this line only keeps "per-feature services under `server/src/services/`". Dropped intentionally, or does it need a home too?

.claude/rules/client/tests.md
```diff
@@ -0,0 +1,33 @@
+---
+paths:
+  - "client/**/*.test.ts"
+  - "client/src/**/__tests__/**"
```
npapagna, line 4: 🤖 _Claude Code finding:_ Neither this glob nor `tests.md`'s `**/__tests__/**` matches `client/src/__mocks__/vscode.ts` (it's under `__mocks__`, not `__tests__`) — but this file's content (the VS Code API mock, `__resetConfig`/`__setConfig`) is centrally about that exact file. Editing the mock directly won't load this rule.

.claude/rules/mcp.md
```diff
@@ -0,0 +1,11 @@
+---
+paths:
+  - "client/src/mcp*.ts"
```
npapagna: 🤖 _Claude Code finding:_ Heads up, `client/src/mcp*.ts` currently matches 6 files, not just the 3 named below — `mcpZodErrorMap.ts`, `mcpServerTreeProvider.ts`, `mcpOwnerSidecar.ts` also match but aren't mentioned in the rule body. Lower severity since it's over-inclusion rather than a missed load, but worth tightening if those files aren't meant to pull this rule in.

npapagna (commented):
Liked a lot the cleaner separation of Claude files! 👏 
I think we're good content-wise (expect for the findings).

🤖 Claude Code review — I (Claude, via gh api) ran a multi-angle automated pass over this diff (correctness, removed-behavior, glob/gitignore correctness, reuse/altitude/conventions) at Nico's request and left the inline comments below. All findings in this review are AI-generated and worth a human sanity-check before acting on them.

Summary: mostly cases where a rule's own prose names a file its paths: frontmatter doesn't cover, plus one gitignore gap.

Nico's hunch (by Claude):
On the file-placement question raised in review feedback: I'd lean toward colocating rules with the code they describe (e.g. mcp-server/CLAUDE.md, or a rule living inside the subsystem's own directory) rather than centralizing everything under root .claude/rules/. Several of the inline findings below are exactly the failure mode that centralizing invites: a rule names a file (exportManager.ts, transcriptChannel.ts, debuggerPanel.ts, __mocks__/vscode.ts) that its own glob doesn't match, so editing that file won't load the rule documenting it. When a rule lives inside the directory it governs, a renamed or added file is much more likely to land in the same review/PR as the doc update; a path-scoped glob living in a separate tree can drift silently with no other signal to catch it. That's not an argument against the split itself — the progressive-disclosure goal here is sound — just a vote for colocation as the thing that makes this class of drift structurally harder to introduce, rather than something we have to catch by memory later.

Nico's hunch (by Claude):
This ties back to the point above — maybe the real question is how we organize the files that implement a given feature in the first place. If everything for a feature already lives under one folder (e.g. client/src/classSync/ for class sync), the .md rule documenting it has a natural, obvious home right there — no separate colocation decision to make, and no drift risk between the rule's glob and the folder it actually describes.

.gitignore
MatiasFernandez: Good catch, I'm going to fix this

.claude/rules/transcript.md
MatiasFernandez: Good catch. This is new from my latest rebase, and I think the contents of the rule are actually duplicates of knowledge the agent can already mostly discover by reading the code. I'll document what's missing within the code and remove the rule.

.claude/rules/transcript.md
MatiasFernandez: Same as above, the overall advice from the sources I listed above is that we should not restate what the agent can already discover easily. On top of that, the lack of structure on this repo makes the rules super fragile because they need to list all files individually. I removed the rule and captured any non-obvious knowledge within "transcriptSink.ts"

.claude/rules/class-sync.md
MatiasFernandez: Same as above regarding removing duplicate information in rules.

CLAUDE.md
MatiasFernandez: This is intentional, auto-generated CLAUDE.md's are too verbose by default, and the sources I consulted recommend skipping details the agent can discover quickly by just listing the files in the project. Once the project gets a better structure (with folders grouping functionality/sub-components), we may introduce some details here and there, but for now, it's better to be less verbose. If in the future we see an agent misinterpreting the project, we can introduce specific guidance, but in this case, the new version should be good enough

.claude/rules/client/tests.md
MatiasFernandez: The contents of this rule are more meant to be used when writing tests than when touching `client/src/__mocks__/vscode.ts` so I think I wouldn't change this. Is there anything within this rule you think should be read when touching the `vscode.ts` file?

.claude/rules/mcp.md
MatiasFernandez: Same as I mentioned for the other rules. I realized the contents of this rule were not that valuable, and having to maintain a list of files is cumbersome. I moved the valuable knowledge into per-file documentation and deleted the rule

MatiasFernandez:
Thanks for the review, @npapagna. I replied to all the comments and addressed them. Regarding the location of rule files:

I looked into moving the `gci` and `tests` rules down into the `client/` workspace so they'd live next to the code they document. Short answer: we can't do it the way I hoped, but there's a clean alternative. There are two mechanisms available, and it's a trade-off between them.

| | **Path-scoped rules** (`.claude/rules/*.md`) | **Reference doc + pointer in a submodule `CLAUDE.md`** |
|---|---|---|
| How it loads | Deterministically, when a file matching the rule's `paths:` glob is read/edited | The agent reads the linked doc only if it judges it relevant |
| Where it can live | **Root `.claude/` only** — nested `client/.claude/rules/` is *not* honored | The `CLAUDE.md` lives inside the submodule/workspace |
| Guarantee | Always loads on a path match (deterministic) | Advisory — we point at it, the agent decides |
| Context cost | Zero until a relevant path is touched | The pointer line is always in the submodule's `CLAUDE.md`; the doc itself loads only if the agent opens it |
| Main downside | Can't sit next to the code it documents | Not guaranteed to be read |

I actually tested nesting a `client/.claude/rules/` folder — it's silently ignored, so option 1 really is root-only.

**So the two realistic options are:**

1. **Keep the rules at the root `.claude/rules/`.** Loading stays deterministic (they fire exactly when a matching `client/` file is touched), at the cost of the files sitting at the repo root, away from the code they describe.

2. **Move the content into a reference doc (e.g. `docs/…`) and list it from `client/CLAUDE.md`.** The knowledge lives next to the code and the pointer only costs context when working in `client/`, but loading is advisory — we recommend the doc, the agent decides whether to read it. I used this approach effectively in a smaller project, but I'm not sure how well it will work here.

Option 2 favors file organization over determinism; option 1 favors determinism over colocated rules. Happy to go either way — flagging the trade-off so we choose deliberately.

.claude/rules/client/tests.md
npapagna: If that's the intent of this rule, then I believe we're fine. Thanks.

npapagna:
Thanks @MatiasFernandez! 

> Thanks for the review, @npapagna. I replied to all the comments and addressed them. Regarding the location of rule files:
> 
> I looked into moving the `gci` and `tests` rules down into the `client/` workspace so they'd live next to the code they document. Short answer: we can't do it the way I hoped, but there's a clean alternative. There are two mechanisms available, and it's a trade-off between them.
> 
> **Path-scoped rules** (`.claude/rules/*.md`)	**Reference doc + pointer in a submodule `CLAUDE.md`**
> How it loads	Deterministically, when a file matching the rule's `paths:` glob is read/edited	The agent reads the linked doc only if it judges it relevant
> Where it can live	**Root `.claude/` only** — nested `client/.claude/rules/` is _not_ honored	The `CLAUDE.md` lives inside the submodule/workspace
> Guarantee	Always loads on a path match (deterministic)	Advisory — we point at it, the agent decides
> Context cost	Zero until a relevant path is touched	The pointer line is always in the submodule's `CLAUDE.md`; the doc itself loads only if the agent opens it
> Main downside	Can't sit next to the code it documents	Not guaranteed to be read
> I actually tested nesting a `client/.claude/rules/` folder — it's silently ignored, so option 1 really is root-only.
> 
> **So the two realistic options are:**
> 
> 1. **Keep the rules at the root `.claude/rules/`.** Loading stays deterministic (they fire exactly when a matching `client/` file is touched), at the cost of the files sitting at the repo root, away from the code they describe.
> 2. **Move the content into a reference doc (e.g. `docs/…`) and list it from `client/CLAUDE.md`.** The knowledge lives next to the code and the pointer only costs context when working in `client/`, but loading is advisory — we recommend the doc, the agent decides whether to read it. I used this approach effectively in a smaller project, but I'm not sure how well it will work here.
> 
> Option 2 favors file organization over determinism; option 1 favors determinism over colocated rules. Happy to go either way — flagging the trade-off so we choose deliberately.

Option 1 seems to be the most simple thing we can do atm to make Claude more efficient.

LGTM!

npapagna (approved):

PR #140, opened 2026-07-03
Fix: hierarchy click should only open Class definition, not the source file

MatiasFernandez:
## Summary

Clicking a class in the System Browser's hierarchy view opened both the Class Definition panel *and* the on-disk `.gs` Topaz cache file for that class. This shoved a redundant read-only tab in front of whatever the user was actively editing. The column-view click already only opened the definition — the hierarchy click should behave the same way.

## Changes

- Removed the extra `openClassFile` call from the hierarchy-click handler in `client/src/systemBrowser.ts`, so it now only calls `applyClassSelection`, matching column-click behavior.
- Updated surrounding comments to reflect that both the column click and hierarchy click are definition-only.
- Added a regression test in `client/src/__tests__/systemBrowser.test.ts` asserting that selecting a hierarchy class does not trigger `showTextDocument`.

Addresses Gitlab#36.

## Test plan

- [x] `npm run compile && npm test` passes
- [x] Manually tested in the extension: clicking a class in the hierarchy view now only opens/refreshes the Class Definition panel and no longer opens a redundant source file tab, confirming the fix works as intended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to the **GemStone Smalltalk** extension will be documented i
 
 ## [Unreleased]
 
+### Fixed
+
+- **System Browser: clicking a class in the hierarchy view no longer reopens its source file.** The hierarchy click already refreshed the Class Definition pane; it also called `openClassFile`, shoving a file in front of whatever the user was editing. That call is removed, matching the column click's definition-only behavior.
+
 ## [1.7.7] - 2026-07-07
 
 ### Added
--- a/client/src/__tests__/systemBrowser.test.ts
+++ b/client/src/__tests__/systemBrowser.test.ts
@@ -1064,23 +1064,6 @@ describe('SystemBrowser', () => {
       );
     });
 
-    // Regression: hierarchy click opened the definition panel AND the
-    // on-disk .gs Topaz cache file, shoving a redundant read-only tab in
-    // front of whatever the user was editing. Column click only opens the
-    // definition; hierarchy click should behave the same way.
-    it('does not open a file when selecting a hierarchy class', async () => {
-      messageHandler({ command: 'toggleViewMode', mode: 'hierarchy' });
-      vi.mocked(fs.existsSync).mockReturnValue(true);
-      vi.mocked(window.showTextDocument).mockClear();
-
-      messageHandler({ command: 'selectHierarchyClass', className: 'Array' });
-      // The redundant .gs reveal (when present) fires from an unawaited async
-      // call, so flush the microtask/macrotask queue before asserting.
-      await new Promise(resolve => setTimeout(resolve, 0));
-
-      expect(window.showTextDocument).not.toHaveBeenCalled();
-    });
-
     it('resolves correct dictionary for hierarchy class from different dict', () => {
       messageHandler({ command: 'toggleViewMode', mode: 'hierarchy' });
       vi.mocked(mockPanel.webview.postMessage).mockClear();
--- a/client/src/__tests__/systemBrowser.test.ts
+++ b/client/src/__tests__/systemBrowser.test.ts
@@ -1064,6 +1064,23 @@ describe('SystemBrowser', () => {
       );
     });
 
+    // Regression: hierarchy click opened the definition panel AND the
+    // on-disk .gs Topaz cache file, shoving a redundant read-only tab in
+    // front of whatever the user was editing. Column click only opens the
+    // definition; hierarchy click should behave the same way.
+    it('does not open a file when selecting a hierarchy class', async () => {
+      messageHandler({ command: 'toggleViewMode', mode: 'hierarchy' });
+      vi.mocked(fs.existsSync).mockReturnValue(true);
+      vi.mocked(window.showTextDocument).mockClear();
+
+      messageHandler({ command: 'selectHierarchyClass', className: 'Array' });
+      // The redundant .gs reveal (when present) fires from an unawaited async
+      // call, so flush the microtask/macrotask queue before asserting.
+      await new Promise(resolve => setTimeout(resolve, 0));
+
+      expect(window.showTextDocument).not.toHaveBeenCalled();
+    });
+
     it('resolves correct dictionary for hierarchy class from different dict', () => {
       messageHandler({ command: 'toggleViewMode', mode: 'hierarchy' });
       vi.mocked(mockPanel.webview.postMessage).mockClear();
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ All notable changes to the **GemStone Smalltalk** extension will be documented i
 ### Changed
 
 - **Integration test default version is now the oldest supported release.** `npm run test:server:start` (and the `--oldest` flag on `gemstone-integration-versions.js`) now defaults to the oldest version in `.gemstone-integration-releases.json` instead of the latest, ensuring new releases do not silently drop the minimum-version bar.
+- **Enhanced Inspector now requires GemStone 3.7.5 or later.** Below 3.7.5 the vendored payload either can't file in (pre-3.7.0 kernel gap) or files in but shows no views (a Unicode/byte-String comparison mismatch on 3.7.0/3.7.2), so the feature is now gated off entirely on older stones instead of silently failing: no install offer, no auto-install, and the **Install Enhanced Inspector Support** command is hidden from the palette. The version check is semantic (`supportsEnhancedInspector`), so future pass without needing a maintained list. Walkthrough and settings docs now call out the 3.7.5 requirement. ([#138](https://github.com/jgfoster/Jasper/pull/138))
```
npapagna: Did this leak from another branch?

client/src/systemBrowser.ts
```diff
@@ -1106,9 +1106,9 @@ export class SystemBrowser {
     this.state.selectedDictIndex = dictIndex;
     // Route through the canonical class-selection helper so the Class
     // Definition panel updates too — the earlier inline mutations did
-    // not, so a hierarchy-view click left Class Definition stale.
+    // not, so a hierarchy-view click left Class Definition stale. This
+    // mirrors the column click: definition-only, no source-file reveal.
     this.applyClassSelection(className, false, true);
-    this.openClassFile(className);
```
npapagna, line 1248: This one is called from `handleNavigateTo` and `handleSelectMethod` as well. Should we take care of them too?

client/src/__tests__/systemBrowser.test.ts
```diff
@@ -872,6 +872,23 @@ describe('SystemBrowser', () => {
       );
     });
 
+    // Regression: hierarchy click opened the definition panel AND the
+    // on-disk .gs Topaz cache file, shoving a redundant read-only tab in
+    // front of whatever the user was editing. Column click only opens the
+    // definition; hierarchy click should behave the same way.
+    it('does not open a file when selecting a hierarchy class', async () => {
```
npapagna: [Claude] Nit: per this repo's test-naming convention, this reads as scenario-specific rather than as a durable guarantee. Something like `'selecting a class from the hierarchy view leaves the editor untouched'` would survive a future rename of `selectHierarchyClass`/the click-handling mechanics.

client/src/__tests__/systemBrowser.test.ts
```diff
@@ -872,6 +872,23 @@ describe('SystemBrowser', () => {
...
+      messageHandler({ command: 'toggleViewMode', mode: 'hierarchy' });
+      vi.mocked(fs.existsSync).mockReturnValue(true);
```
npapagna: [Claude] Since the fix deletes the only code path that would reach `openClassFile` from this handler, `fs.existsSync` can never be reached here — this mock (and the `showTextDocument.mockClear()` two lines down) can't affect the outcome. Looks copy-pasted from a sibling test where it was meaningful; worth dropping so a future reader doesn't assume it's load-bearing.

client/src/__tests__/systemBrowser.test.ts
```diff
@@ -872,6 +872,23 @@ describe('SystemBrowser', () => {
...
+      vi.mocked(window.showTextDocument).mockClear();
+
+      messageHandler({ command: 'selectHierarchyClass', className: 'Array' });
+      // The redundant .gs reveal (when present) fires from an unawaited async
+      // call, so flush the microtask/macrotask queue before asserting.
+      await new Promise(resolve => setTimeout(resolve, 0));
```
npapagna: [Claude] This file already has an established idiom for flushing an async post-selection effect a bit further down: `await vi.waitFor(() => expect(...).toHaveBeenCalled())`. That's stronger/more diagnostic than a raw `setTimeout(resolve, 0)`, which only flushes one tick and fails less informatively if the timing is ever different.

client/src/__tests__/systemBrowser.test.ts
npapagna: I don't think we actually need this test. It's checking for the absence of a call, and that behavior only exists as leftover from the bug we're fixing. If we'd been TDDing the hierarchy click from scratch, we'd never have arrived at a test like this one, since there was never a real reason for it to open a file in the first place. Would you consider dropping it, unless you'd rather keep it as a regression guard specifically for this bug?

npapagna (commented):
Thanks a lot @MatiasFernandez!

Left a couple of comments.

CHANGELOG.md
MatiasFernandez: I need to re-rebase. When I rebased it it was a change that was not released but was not included in the changelog. This change is already released, so I need to fix this CHANGELOG one more time

client/src/systemBrowser.ts
MatiasFernandez: I think those are correct because in those cases, we actually want to open the method implementation. My fix is when opening a class; in that case, we don't want to open the code of the class, just the definition. Does that make sense?

client/src/__tests__/systemBrowser.test.ts
MatiasFernandez: You are right, I'll address this

jgfoster:
For what it's worth, I'm becoming less confident that we need the cache of files outside the repository, so we might "solve" this by removing the external mirror/cache. In any case, I'm not merging yet since it looks like it is still in process.

client/src/__tests__/systemBrowser.test.ts
MatiasFernandez: I think that makes sense. I'm removing an extra line that shouldn't be there, and the rest of the functionality is well-covered. This test helped me ensure the change was the correct one, but now it is just mocking unrealistic logic that does not exist anymore. Once the code is deleted, this is not even a deterministic way to prevent future similar bugs in the future because it's tied to one specific implementation. I'm going to delete this

npapagna (approved):
LGTM!

npapagna:

PR #141, opened 2026-07-03
Clean up npm test plumbing across workspaces

MatiasFernandez:
## Goal

Bring the monorepo's `npm test` plumbing closer to a conventional npm-workspaces setup: make the three workspaces (`client`, `server`, `mcp-server`) run tests consistently, remove a latent divergence between how the root and the workspaces invoked their tests, and drop a Node flag that is now a no-op. Verified `npm test` still passes across all three workspaces on the supported toolchain (Node pinned to 24.15.0 via `.nvmrc`).

## What changed

**`server/package.json`** — the workspace had no `scripts` block at all, which is why the root had to `cd server && npx vitest run` as a workaround. Added a standard `"test": "vitest run"`, so it can be driven like any other workspace.

**`package.json`** — the root `test` script hand-chained three legs with `&&`, each invoking their workspace differently (`cd server && npx vitest run`, `npm test --workspace client`, `cd mcp-server && npx vitest run`). Now:
- `test` → `npm run test --workspaces`, the documented mechanism for running a script across all workspaces.
- `test:server` / `test:mcp` → `npm test --workspace <name>`, consistent with the existing `test:client`, so all per-workspace legs route through the workspace's own script.

**Behavior change:** the old `&&`-chained root `test` was fail-fast — it stopped at the first workspace to fail, leaving the remaining workspaces' tests unrun. `npm run test --workspaces` runs the script in every workspace regardless of earlier failures, then exits non-zero if any of them failed. So `npm test` now always reports results from all three workspaces instead of stopping early on the first failure.
If in the future we think this is not the ideal solution, it's easy to change it back.

**`mcp-server/package.json` + `package.json`** — the mcp-server test script carried `NODE_OPTIONS=--experimental-require-module` to let CommonJS `require()` the ESM-only `@modelcontextprotocol/sdk`. On Node 24 that flag is a no-op: `require(ESM)` has been enabled by default since Node v23.0.0 / v22.12.0 / v20.19.0. The inline `NODE_OPTIONS=...` syntax is also POSIX-shell-only (breaks on Windows `cmd.exe`). So:
- Dropped the flag; the script is now just `"vitest run"` (matching `server`).
- Added `"engines": { "node": ">=22.12" }` to the root to make explicit the Node floor the flag used to paper over.

Removing the flag also eliminates the divergence class at its root: with no special flag, the root and workspace invocations can no longer differ.

## Verification

- `npm test` passes across all three workspaces.
- `npm run test:mcp` → 95/95 pass, exit 0, no `ExperimentalWarning`, no `ERR_REQUIRE_ESM`, with the flag removed on Node 24.15.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
     "url": "https://github.com/jgfoster/Jasper/issues"
   },
   "engines": {
-    "vscode": "^1.85.0"
+    "vscode": "^1.85.0",
+    "node": ">=22.12"
   },
   "icon": "resources/gemstone-icon.png",
   "categories": [
@@ -1337,13 +1338,13 @@
     "compile:mcp": "tsc -p mcp-server/tsconfig.json",
     "bundle": "node esbuild.mjs",
     "watch": "tsc -b -w client/tsconfig.json server/tsconfig.json",
-    "test": "npm run test:server && npm run test:client && npm run test:mcp",
+    "test": "npm run test --workspaces",
     "test:server:start": "npm run test:server:start --workspace client",
     "test:server:stop": "npm run test:server:stop --workspace client",
     "test:server:list": "npm run test:server:list --workspace client",
-    "test:server": "cd server && npx vitest run",
+    "test:server": "npm test --workspace server",
     "test:client": "npm test --workspace client",
-    "test:mcp": "cd mcp-server && npx vitest run",
+    "test:mcp": "npm test --workspace mcp-server",
     "//test:gci": "Runs only the on-demand 'gci' vitest project (client/vitest.config.ts); needs a live test stone (npm run test:server:start). `npm test` runs the 'default' project and excludes these.",
     "test:gci": "cd client && npx vitest run --project gci",
     "package": "vsce package",
--- a/mcp-server/package.json
+++ b/mcp-server/package.json
@@ -5,7 +5,7 @@
   "description": "MCP server for GemStone database interaction",
   "main": "out/index.js",
   "scripts": {
-    "test": "NODE_OPTIONS=--experimental-require-module npx vitest run"
+    "test": "vitest run"
   },
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.12.1",
--- a/server/package.json
+++ b/server/package.json
@@ -2,6 +2,9 @@
   "name": "gemstone-smalltalk-server",
   "version": "0.1.0",
   "private": true,
+  "scripts": {
+    "test": "vitest run"
+  },
   "dependencies": {
     "esbuild": "^0.28.1",
     "vscode-languageserver": "^9.0.0",


package.json
```diff
@@ -1315,13 +1316,13 @@
     "compile:mcp": "tsc -p mcp-server/tsconfig.json",
     "bundle": "node esbuild.mjs",
     "watch": "tsc -b -w client/tsconfig.json server/tsconfig.json",
-    "test": "npm run test:server && npm run test:client && npm run test:mcp",
+    "test": "npm run test --workspaces --if-present",
```
npapagna: Found by Claude Code: this isn't quite the "behavioral no-op" described in the PR summary. `npm run <script> --workspaces` runs every workspace's script and *continues even if an earlier workspace's script fails* (only the aggregate exit code reflects a failure at the end), whereas the old `test:server && test:client && test:mcp` stopped at the first failing workspace. It also changes execution order to the `workspaces` array order (client → server → mcp-server) instead of the previous server → client → mcp order. Net effect: a failing `server` suite no longer short-circuits the `client`/`mcp-server` runs — CI takes longer to surface a failure and produces more mixed output, even though the final exit code is still non-zero. Worth calling out explicitly (or restoring fail-fast) if that ordering/short-circuit behavior was relied upon.

npapagna: Before: test:server && test:client && test:mcp — stops at the first failing workspace.

After: npm run test --workspaces --if-present — always runs all three workspaces , even if an earlier one already failed.

Running everything regardless of earlier failures is a reasonable default. But worth considering: say project B depends on project A. If A's tests fail because something there is fundamentally broken, B's tests may fail too — for a confusing, secondary reason — and CI still burns time running them.

Locally: just a few extra seconds, no big deal.
On CI: those extra minutes come out of the shared CI minutes budget, on a run that was likely doomed anyway.

Worth an explicit call: keep continue-on-failure (simpler, complete failure reports) or restore fail-fast so a broken server skips client/mcp-server and saves CI time.

npapagna: One small thing worth flagging: with --if-present, if we ever add a new workspace that doesn't have a test script yet, npm test --workspaces will just silently skip it — including on CI. We might think we're covering it because it's in the workspaces list, but we actually wouldn't be.

What do you think about dropping --if-present? My guess is every workspace will end up with some form of testing anyway, and for the rare case where one genuinely has nothing to test, an explicit no-op "test": "echo \"no tests\" && exit 0" script would make that a visible, documented choice rather than something that quietly falls through the cracks.

npapagna (commented):
This is great @MatiasFernandez!
Thanks.

package.json
MatiasFernandez: > this isn't quite the "behavioral no-op"

That's true, I fixed the PR description. I think this new approach is good enough for now. In the future, if we see a CI that is too slow or expensive, we can change the approach.

> What do you think about dropping --if-present?

I think that's ok. I dropped it, if we need it in the future we can add it back

npapagna (approved):
Thanks @MatiasFernandez!

PR #142, opened 2026-07-06
Find Method in Class: fix and rename picker

MatiasFernandez:
## Summary
Fixes Gitlab#55

Fixes the "Find Method…" command (`gemstone.findMethod`), which was broken when invoked with no class selected in the System Browser: it fell back to a plain text input box, and picking a method from the resulting list would fail to navigate to or open it.

- Replaced the text-input fallback with the same filterable class quick-pick used by **Find Class**. The command now always shows this picker.
- When a class *is* selected in the System Browser, that class is pre-highlighted as the default active item — pressing Enter immediately reproduces the old scoped-to-selection behavior, while the full list is still right there to pick a different class instead.
- Renamed the command from **"Find Method…"** to **"Find Method in Class…"** (id changed from `gemstone.findMethod` to `gemstone.findMethodInClass`) since the command is scoped to a single class, to distinguish it from a possible future cross-class method search.
- Extracted the handler out of the giant `activate()` function in `extension.ts` into its own module, `client/src/commands/findMethodInClass.ts`, along with a dedicated test file.
- Reused the shared `buildMethodUri` helper for the "open the method" fallback instead of hand-rolling the `gemstone://` URI (applied to the sibling `showMethodResults` handler in `extension.ts` too), and folded the duplicated progress-notification wrappers into one helper. Besides deduplication this fixes a latent bug: keyword selectors were percent-encoded (`:` → `%3A`), yielding a different URI than the rest of the extension and opening the same method in a second, duplicate editor tab.

## Decisions

- **Command id rename is a breaking change** for anyone with a custom keybinding referencing `gemstone.findMethod` — the built-in `Ctrl/Cmd+K M` chord was updated and is unaffected. Called out explicitly in the CHANGELOG.
- Used `vscode.window.createQuickPick` instead of `showQuickPick` for the class picker specifically because pre-highlighting a default item via `activeItems` while leaving the filter text empty isn't possible with the simpler `showQuickPick` API.

## Testing

Tested manually in a running extension host against a live GemStone session:
- Find Method in Class with no class selected in the System Browser
- Find Method in Class with a class selected — confirmed it's pre-highlighted, and Enter reproduces the old scoped behavior
- Picking a different class from the list, then a method, and confirming navigation/open both when a System Browser is open and when it isn't (falls back to opening the `gemstone://` document)

Recording:

https://github.com/user-attachments/assets/d0cc5490-d6bd-4273-9a9c-d8a5ef773f50

### DIFF
--- a/client/src/__mocks__/vscode.ts
+++ b/client/src/__mocks__/vscode.ts
@@ -226,6 +226,43 @@ export const window = {
   })),
   showInputBox: vi.fn(),
   showQuickPick: vi.fn(),
+  // Controllable low-level QuickPick. Each call returns a fresh instance whose
+  // registered handlers can be fired from a test via `__accept()` / `__hide()`
+  // (or `hide()`), after inspecting/setting `items`, `activeItems`, and
+  // `selectedItems`. Grab the created instance with
+  // `vi.mocked(vscode.window.createQuickPick).mock.results.at(-1).value`.
+  createQuickPick: vi.fn(() => {
+    const acceptHandlers: Array<() => void | Promise<void>> = [];
+    const hideHandlers: Array<() => void> = [];
+    const qp: Record<string, unknown> = {
+      title: '',
+      placeholder: '',
+      value: '',
+      items: [] as unknown[],
+      selectedItems: [] as unknown[],
+      activeItems: [] as unknown[],
+      enabled: true,
+      busy: false,
+      matchOnDescription: false,
+      matchOnDetail: false,
+      onDidAccept: vi.fn((h: () => void | Promise<void>) => {
+        acceptHandlers.push(h);
+        return { dispose: vi.fn() };
+      }),
+      onDidHide: vi.fn((h: () => void) => {
+        hideHandlers.push(h);
+        return { dispose: vi.fn() };
+      }),
+      show: vi.fn(),
+      hide: vi.fn(() => hideHandlers.forEach((h) => h())),
+      dispose: vi.fn(),
+      __accept: async () => {
+        for (const h of acceptHandlers) await h();
+      },
+      __hide: () => hideHandlers.forEach((h) => h()),
+    };
+    return qp;
+  }),
   showOpenDialog: vi.fn(),
   showSaveDialog: vi.fn(),
   tabGroups: {
--- a/client/src/__mocks__/vscode.ts
+++ b/client/src/__mocks__/vscode.ts
@@ -234,6 +234,16 @@ export const window = {
   createQuickPick: vi.fn(() => {
     let onAccept: (() => void | Promise<void>) | undefined;
     let onHide: (() => void) | undefined;
+    let visible = false;
+    // Mirror real VS Code: hide()/dispose() only fire onDidHide while the input
+    // is still visible, and at most once — dispose() cascades to a hide if the
+    // input hasn't been hidden yet, so `dispose()` after a still-visible show()
+    // fires onDidHide (which a disposed-then-nothing input does not re-fire).
+    const fireHide = () => {
+      if (!visible) return;
+      visible = false;
+      onHide?.();
+    };
     const qp: Record<string, unknown> = {
       title: '',
       placeholder: '',
@@ -253,13 +263,15 @@ export const window = {
         onHide = h;
         return { dispose: vi.fn() };
       }),
-      show: vi.fn(),
-      hide: vi.fn(() => onHide?.()),
-      dispose: vi.fn(),
+      show: vi.fn(() => {
+        visible = true;
+      }),
+      hide: vi.fn(() => fireHide()),
+      dispose: vi.fn(() => fireHide()),
       __accept: async () => {
         await onAccept?.();
       },
-      __hide: () => onHide?.(),
+      __hide: () => fireHide(),
     };
     return qp;
   }),
--- a/client/src/__mocks__/vscode.ts
+++ b/client/src/__mocks__/vscode.ts
@@ -232,8 +232,8 @@ export const window = {
   // `selectedItems`. Grab the created instance with
   // `vi.mocked(vscode.window.createQuickPick).mock.results.at(-1).value`.
   createQuickPick: vi.fn(() => {
-    const acceptHandlers: Array<() => void | Promise<void>> = [];
-    const hideHandlers: Array<() => void> = [];
+    let onAccept: (() => void | Promise<void>) | undefined;
+    let onHide: (() => void) | undefined;
     const qp: Record<string, unknown> = {
       title: '',
       placeholder: '',
@@ -246,20 +246,20 @@ export const window = {
       matchOnDescription: false,
       matchOnDetail: false,
       onDidAccept: vi.fn((h: () => void | Promise<void>) => {
-        acceptHandlers.push(h);
+        onAccept = h;
         return { dispose: vi.fn() };
       }),
       onDidHide: vi.fn((h: () => void) => {
-        hideHandlers.push(h);
+        onHide = h;
         return { dispose: vi.fn() };
       }),
       show: vi.fn(),
-      hide: vi.fn(() => hideHandlers.forEach((h) => h())),
+      hide: vi.fn(() => onHide?.()),
       dispose: vi.fn(),
       __accept: async () => {
-        for (const h of acceptHandlers) await h();
+        await onAccept?.();
       },
-      __hide: () => hideHandlers.forEach((h) => h()),
+      __hide: () => onHide?.(),
     };
     return qp;
   }),
--- a/client/src/commands/findMethodInClass.ts
+++ b/client/src/commands/findMethodInClass.ts
@@ -2,9 +2,32 @@ import * as vscode from 'vscode';
 import * as queries from '../browserQueries';
 import { SessionManager } from '../sessionManager';
 import { SystemBrowser } from '../systemBrowser';
+import { buildMethodUri } from '../gemstoneFileSystemProvider';
 
 interface ClassPickItem extends vscode.QuickPickItem {
   entry: queries.ClassNameEntry;
+}
+
+/**
+ * Run `fn` behind a modal progress notification, surfacing any thrown error as
+ * an error message. Returns `undefined` when the work failed. `fn` may be
+ * synchronous — the notification still shows for the duration of the call.
+ */
+async function withLoadingProgress<T>(
+  title: string,
+  failLabel: string,
+  fn: () => T,
+): Promise<T | undefined> {
+  try {
+    return await vscode.window.withProgress(
+      { location: vscode.ProgressLocation.Notification, title, cancellable: false },
+      () => Promise.resolve(fn()),
+    );
+  } catch (e: unknown) {
+    const msg = e instanceof Error ? e.message : String(e);
+    vscode.window.showErrorMessage(`${failLabel}: ${msg}`);
+    return undefined;
+  }
 }
 
 /**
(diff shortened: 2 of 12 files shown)


client/src/commands/findMethodInClass.ts
```diff
@@ -0,0 +1,118 @@
...
+  // pre-highlighted via activeItems without also filtering the list.
+  const pickedClass = await new Promise<ClassPickItem | undefined>(resolve => {
+    const qp = vscode.window.createQuickPick<ClassPickItem>();
+    qp.items = classItems;
+    qp.matchOnDescription = true;
+    qp.placeholder = 'Type to find a class…';
+    if (current) {
+      const preselected = classItems.find(
+        i => i.entry.className === current.className && i.entry.dictName === current.dictName,
+      );
```
npapagna, line 36: > 🤖 **AI-assisted comment** — drafted by Claude (Claude Code) during an automated review, posted on behalf of @npapagna. Not written directly by npapagna.

`SystemBrowser.getSelectedClassName` only returns `{ dictName, className }`, and this match compares those as strings. But two dictionaries in the SymbolList can share a name (see the disambiguation comment on `gemstone.findClass`'s fallback URI in `extension.ts`, which uses `dictIndex` for exactly this reason). If that happens and the browser's current selection is a class in the "other" dictionary, this can pick the wrong `ClassNameEntry` to pre-highlight — pressing Enter (the "reproduces the old scoped behavior" default this PR advertises) then loads methods for / navigates to the wrong class, silently.

`ClassNameEntry` already carries `dictIndex`, and `getSelectedClassName`'s internal state (`browser.state.selectedDictIndex`) has it too — it's just not returned. Extending `getSelectedClassName`'s return type to include `dictIndex` and matching on that here would close the gap. It's the only caller of `getSelectedClassName`, so the return-type change should be low-risk.

client/src/commands/findMethodInClass.ts
```diff
@@ -0,0 +1,118 @@
...
+  if (!session) return;
+
+  const current = SystemBrowser.getSelectedClassName(session.id);
+
+  const entries = await withLoadingProgress(
+    'Loading class list…',
+    'Failed to load classes',
+    () => queries.getAllClassNames(session),
+  );
+  if (!entries) return;
```
npapagna: > 🤖 **AI-assisted comment** — drafted by Claude (Claude Code) during an automated review, posted on behalf of @npapagna. Not written directly by npapagna.

This only guards the "the load threw" case (`withLoadingProgress` returns `undefined` on error). An empty array is truthy in JS, so if `getAllClassNames` succeeds but legitimately returns zero classes, this falls through to an empty, unexplained QuickPick instead of a message — unlike the method-list path below (`if (methods.length === 0) { showInformationMessage(...); return; }`). Suggest mirroring that here, e.g. `if (!entries || entries.length === 0) { ...; return; }`.

client/src/commands/findMethodInClass.ts
```diff
@@ -0,0 +1,118 @@
...
+interface ClassPickItem extends vscode.QuickPickItem {
+  entry: queries.ClassNameEntry;
+}
+
+/**
+ * Run `fn` behind a modal progress notification, surfacing any thrown error as
+ * an error message. Returns `undefined` when the work failed. `fn` may be
+ * synchronous — the notification still shows for the duration of the call.
+ */
+async function withLoadingProgress<T>(
```
npapagna: > 🤖 **AI-assisted comment** — drafted by Claude (Claude Code) during an automated review, posted on behalf of @npapagna. Not written directly by npapagna.

Nit: `gemstone.findClass` (`extension.ts`, right where the old `findMethod` handler used to live) already inlines this exact try/withProgress/catch/showErrorMessage shape for the same `getAllClassNames` call, plus the identical `{label, description, entry}` item mapping a few lines later. Since this PR is already extracting `findMethodInClass` into its own module, might be worth pulling `withLoadingProgress` (and the class-item mapper) out as a shared, exported helper and pointing `findClass` at it too, rather than leaving two copies of the same logic to keep in sync. Not blocking.

client/src/commands/findMethodInClass.ts
```diff
@@ -0,0 +1,118 @@
...
+  const picked = await vscode.window.showQuickPick(items, {
+    placeHolder: `Type to find a method in ${className}…`,
+    matchOnDescription: true,
+  });
+  if (!picked) return;
+
+  const result: queries.MethodSearchResult = { dictName, className, ...picked.method };
+
+  if (!SystemBrowser.navigateTo(session.id, result)) {
+    const uri = buildMethodUri({ kind: 'method', sessionId: session.id, ...result, environmentId: 0 });
```
npapagna, line 80: > 🤖 **AI-assisted comment** — drafted by Claude (Claude Code) during an automated review, posted on behalf of @npapagna. Not written directly by npapagna.

`buildMethodUri` (in `gemstoneFileSystemProvider.ts`) calls `assertIsValidUriPath` on the selector, which throws `Error("Selector must not contain '/': ...")` for any selector containing a literal `/`. `/` and `//` are ordinary Smalltalk binary selectors (division, integer division on `Integer`/`Number`/`Float`) — `getMethodList` will happily return them verbatim from `cls sortedSelectorsIn: cat`. If a user picks one of these via "Find Method in Class" with no System Browser panel open for the session (so `SystemBrowser.navigateTo` returns `false` and we fall into this branch), this throws synchronously and uncaught — there's no try/catch here, unlike the two `withLoadingProgress`-wrapped calls earlier in this same function. The old (removed) handler built this same fallback URI via `vscode.Uri.parse(...encodeURIComponent(selector)...)`, which safely percent-encoded `/` to `%2F` and never threw.

Might be worth having `buildMethodUri` percent-encode `/` within a segment instead of rejecting it, or wrapping this call in the same try/catch pattern as the loads above.

client/src/extension.ts
```diff
@@ -851,15 +852,7 @@ export function activate(context: vscode.ExtensionContext) {
...
-      const side = r.isMeta ? 'class' : 'instance';
-      const uri = vscode.Uri.parse(
-        `gemstone://${session.id}` +
-        `/${encodeURIComponent(r.dictName)}` +
-        `/${encodeURIComponent(r.className)}` +
-        `/${side}` +
-        `/${encodeURIComponent(r.category)}` +
-        `/${encodeURIComponent(r.selector)}`
-      );
+      const uri = buildMethodUri({ kind: 'method', sessionId: session.id, ...r, environmentId: 0 });
```
npapagna, line 925: > 🤖 **AI-assisted comment** — drafted by Claude (Claude Code) during an automated review, posted on behalf of @npapagna. Not written directly by npapagna.

Same regression as the fallback in the new `findMethodInClass.ts`, but this call site is reachable from more places: `showMethodResults` backs Senders of, Implementors of, Hierarchy Implementors of, and References to, none of which wrap this in a try/catch. The old code here built the URI manually with `encodeURIComponent`, which tolerated `/` fine; `buildMethodUri`'s `assertIsValidUriPath` throws for it. Running e.g. "Senders of #/" with no System Browser panel open now throws instead of opening the file. Worth fixing in one place (`buildMethodUri`) rather than two.

client/src/__mocks__/vscode.ts
```diff
@@ -225,6 +225,43 @@ export const window = {
   })),
   showInputBox: vi.fn(),
   showQuickPick: vi.fn(),
+  // Controllable low-level QuickPick. Each call returns a fresh instance whose
+  // registered handlers can be fired from a test via `__accept()` / `__hide()`
+  // (or `hide()`), after inspecting/setting `items`, `activeItems`, and
+  // `selectedItems`. Grab the created instance with
+  // `vi.mocked(vscode.window.createQuickPick).mock.results.at(-1).value`.
+  createQuickPick: vi.fn(() => {
+    const acceptHandlers: Array<() => void | Promise<void>> = [];
```
npapagna: > 🤖 **AI-assisted comment** — drafted by Claude (Claude Code) during an automated review, posted on behalf of @npapagna. Not written directly by npapagna.

Minor, non-blocking: `acceptHandlers`/`hideHandlers` are arrays (multi-subscriber pub/sub), but every real caller (`findMethodInClass.ts`, `enhancedInspectorCommand.ts`) only ever registers one `onDidAccept` and one `onDidHide` handler per `QuickPick` instance — a single-shot create → subscribe once → show → resolve pattern. A single nullable handler field per event (e.g. `let onAccept: (() => void | Promise<void>) | undefined`) would be simpler with no loss of coverage for anything actually exercised today.

client/src/__mocks__/vscode.ts
```diff
@@ -225,6 +225,43 @@ export const window = {
...
+        acceptHandlers.push(h);
+        return { dispose: vi.fn() };
+      }),
+      onDidHide: vi.fn((h: () => void) => {
+        hideHandlers.push(h);
+        return { dispose: vi.fn() };
+      }),
+      show: vi.fn(),
+      hide: vi.fn(() => hideHandlers.forEach((h) => h())),
+      dispose: vi.fn(),
```
npapagna: > 🤖 **AI-assisted comment** — drafted by Claude (Claude Code) during an automated review, posted on behalf of @npapagna. Not written directly by npapagna.

Minor, non-blocking: real VS Code's `QuickInput.dispose()` hides the input first if it's still visible, which fires `onDidHide` (per the `@types/vscode` doc comments). This mock's `dispose` doesn't call `hideHandlers`, so no test can exercise "onDidHide fires right after dispose()". Harmless today — `findMethodInClass.ts`'s `onDidAccept` handler calls `resolve(...)` before `dispose()`, and a native Promise only settles once, so the `onDidHide` handler's `resolve(undefined)` would be a no-op even if it did fire. But if `onDidHide` ever grows a real side effect (telemetry, resetting other state), it would silently double-fire in production while this test suite stays green, since the mock can't reproduce the dispose-cascades-to-hide behavior to catch it.

npapagna (commented):
Thanks @MatiasFernandez!
Gave it a try and worked fine for me!
Left some comments about stuff discussed/reviewed with Claude.

npapagna:
I think we are good to go (defer to @jgfoster) once the issues have been fixed/addressed/discarded.

client/src/commands/findMethodInClass.ts
MatiasFernandez: Fixed

client/src/extension.ts
MatiasFernandez: Fixed

client/src/commands/findMethodInClass.ts
MatiasFernandez: Fixed

client/src/commands/findMethodInClass.ts
MatiasFernandez: I explored this a bit and concluded that this can't be reproduced easily. `getAllClassNames` walks `System myUserProfile symbolList`, which always contains `Globals` with hundreds of kernel classes. No normal scope yields zero classes. Do you have any scenario in mind that would reproduce this issue?

client/src/commands/findMethodInClass.ts
MatiasFernandez: Fixed

client/src/__mocks__/vscode.ts
MatiasFernandez: Fixed

client/src/__mocks__/vscode.ts
MatiasFernandez: Fixed

client/src/commands/findMethodInClass.ts
npapagna: I don't have any, just surfaced Claude's finding. I think we're good.

npapagna:
LGTM! 🚀

npapagna (approved):

PR #144, opened 2026-07-07
WIP: Rowan Package support

srbaker:
An interface to browse and load Rowan packages.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,22 @@
+# The image runs `npm ci` itself, so never ship host artifacts — especially
+# native modules built for macOS and multi-gigabyte GemStone downloads.
+**/node_modules
+client/tmp
+.vscode-test
+**/out
+*.vsix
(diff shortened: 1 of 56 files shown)


MatiasFernandez:
@srbaker I've been trying to give some structure to the project to keep it manageable. I saw you sded some of the files into a Rowan folder, but I would still give it another twist. For example, adding a __tests__ folder within queries/rowan to colocate the tests there instead of in the top level __tests__ folder. And also group the files that are lingering in the src folder into a "rowan". This is not the definitive structure I want to push for, but is a first step.

Also, is "rowan" enough to describe this feature/module or "rowan-browser" is more precise? I'm ok with "rowan" tho, just wanted to know what you think

jgfoster:
I'm going to merge this so it can be published and people can play with it. I expect that this will be extensively revised regarding features, UI/UX, and internal code layout.

PR #145, opened 2026-07-07
GciLibrary: add ergonomic layer over raw GCI calls

npapagna:
## Summary

- Adds an ergonomic layer on top of `GciLibrary`'s raw `GciTsXxx` FFI wrappers: `login`, `logout`, `execute`, `resolveSymbol`, `createString`, plus PureExportSet/UserGlobals/SessionTemps helpers. Each throws a new `GciLibraryError` instead of returning a raw `{success, err}`/`{result, err}` pair.
- Rewrites the integration-test harness (`useIntegrationTest`) to use these helpers, and fixes a real per-test cleanup gap in the process (see below).

**Before:** every call site that needed to log in, execute code, or resolve a symbol called the raw `GciTsXxx` functions directly and had to manually unwrap and check GCI's C-style `{success, err}` convention by hand. In the integration-test harness specifically, `afterEach` only called `GciTsAbort` — which rolls back the transaction but doesn't touch the session's PureExportSet (pinned OOPs) or SessionTemps (session-scoped temp dictionary), since those live outside the transaction boundary. `GciTsReleaseAllObjs` already existed as a raw wrapper, but its only caller anywhere in the codebase was its own dedicated unit test — nothing cleared PureExportSet or SessionTemps between tests in a shared way, so a test that retained objects or wrote SessionTemps entries could leak that state into the next test in the same file. If a test author cared about this, they had to clean up manually.

**After:** `GciLibrary` exposes a second, integration-tested tier of methods that wrap the raw calls and throw `GciLibraryError` on failure. The test harness now calls a new `resetNonTransactionalSessionState` after every test's `afterEach`, which releases every PureExportSet oop, resets SessionTemps, and clears the new per-session `Utf8`-class-oop cache — automatically, for every test using this harness, with no per-test opt-in required. It also fails every subsequent test in a describe block immediately if that cleanup itself fails, rather than risk leaking state into later tests under this project's shuffled test order.

**New capabilities** (didn't exist before this PR):

- **Catchable, structured errors.** `GciLibraryError` lets callers write `instanceof GciLibraryError` to tell "GCI operation failed" apart from any other bug. Previously a raw call returned a `{success, err}` struct (nothing to catch), and the one place that already threw manually threw a plain `Error`, indistinguishable from any other exception.
- **Tests can exercise login failure and logout, mid-suite.** `GciTestContext` now exposes `login(options)` and `logout()` to individual tests, with the harness auto-recovering if a test logs out and a later one needs a session. Previously `useIntegrationTest` logged in exactly once in `beforeAll` for the whole file, so there was no supported way to test invalid credentials or logout behavior.
- **A regression-detection primitive for oop leaks.** `didPureExportSetGainOnlyOopsProvidedBy` snapshots the PureExportSet before/after an operation and asserts it changed by exactly the oops declared — nothing extra, nothing missing. There was no shared way to make that assertion before.

**Not used yet:** this layer isn't wired into any production call site in this PR — `sessionManager.ts`, `debugQueries.ts`, and others still call the raw GCI functions directly. The next PR will use these primitives to implement the UTF8 fix (production code currently calls `execute` with no UTF8 class oop, so it can't handle non-Latin1 source) and, as a side effect, will start actually benefiting from `execute()`'s Utf8-class-oop caching. This PR is just the tested groundwork it depends on.

## Test plan

- [x] `npm run compile`
- [x] `npm test` against a live GemStone test stone (31/31 new tests passing)

### DIFF
--- a/client/src/__tests__/gciLibrary.test.ts
+++ b/client/src/__tests__/gciLibrary.test.ts
@@ -32,7 +32,7 @@ describe('GciLibrary', () => {
      * The callback receives the Smalltalk snippet as its argument so it can
      * embed it in any expression under test (e.g. pass it to `execute`).
      */
-    function expectToThrowExpectedGciLibraryError(callback: (throwOopsError: string) => unknown) {
+    function expectToThrowExpectedGciLibraryError(callback: (signalExpectedErrorExpression: string) => unknown) {
         expectToThrowGciLibraryError(
             () => callback(`self error: 'oops'`),
             'a UserDefinedError occurred (error 2318), reason:halt, oops');
@@ -108,8 +108,8 @@ describe('GciLibrary', () => {
         });
 
         it('throws when the expression signals an error', () => {
-            expectToThrowExpectedGciLibraryError(throwExpectedErrorExpression => {
-                gciLibrary.execute(session, throwExpectedErrorExpression);
+            expectToThrowExpectedGciLibraryError(signalExpectedErrorExpression => {
+                gciLibrary.execute(session, signalExpectedErrorExpression);
             });
         });
 
@@ -132,8 +132,8 @@ describe('GciLibrary', () => {
         })
 
         it('throws when the evaluated code signals an error', () => {
-            expectToThrowExpectedGciLibraryError(throwOopsError => {
-                gciLibrary.executeDiscardingResult(session, throwOopsError);
+            expectToThrowExpectedGciLibraryError(signalExpectedErrorExpression => {
+                gciLibrary.executeDiscardingResult(session, signalExpectedErrorExpression);
             });
         })
 
--- a/client/src/__tests__/gciLibrary.test.ts
+++ b/client/src/__tests__/gciLibrary.test.ts
@@ -115,6 +115,30 @@ describe('GciLibrary', () => {
 
     });
 
+    describe('evaluating expressions for effect only, discarding the result', () => {
+
+        it('evaluates an expression', () => {
+            const key = gciLibrary.nextKey();
+
+            gciLibrary.executeDiscardingResult(session, `UserGlobals at: ${key} put: true`);
+
+            expectUserGlobalsToInclude(key, true);
+        })
+
+        it('does not retain the discarded result in the PureExportSet', () => {
+            expectPureExportSetToStayUnchanged(() => {
+                gciLibrary.executeDiscardingResult(session, 'Object new');
+            });
+        })
+
+        it('throws when the evaluated code signals an error', () => {
+            expectToThrowExpectedGciLibraryError(throwOopsError => {
+                gciLibrary.executeDiscardingResult(session, throwOopsError);
+            });
+        })
+
+    });
+
     describe('creating strings', () => {
 
         it('creates a String object from the given contents', () => {
--- a/client/src/__tests__/gciLibrary.test.ts
+++ b/client/src/__tests__/gciLibrary.test.ts
@@ -76,6 +76,79 @@ describe('GciLibrary', () => {
     function expectUserGlobalsToInclude(key: string, shouldBeIncluded: boolean) {
         expect(gciLibrary.isIncludedInUserGlobals(session, key)).toBe(shouldBeIncluded);
     }
+
+    /**
+     * Asserts that the session's PureExportSet gains at least one new oop
+     * across `callback`, per `shouldGrow`.
+     *
+     * @param shouldGrow - Whether the PureExportSet is expected to gain a new oop.
+     * @param callback - The operation to observe.
+     */
+    function expectPureExportSetToGrow(shouldGrow: boolean, callback: () => unknown) {
+        expect(gciLibrary.didPureExportSetGrow(session, callback)).toBe(shouldGrow);
+    }
+
+    /** Spies on `resolveSymbol` for the duration of `callback`, then restores it. */
+    function spyOnResolveSymbol(callback: (spy: Mock<typeof GciLibrary.prototype.resolveSymbol>) => void) {
+        const spy =  vi.spyOn(gciLibrary, 'resolveSymbol');
+
+        try{
+            callback(spy);
+        } finally{
+            spy.mockRestore();
+        }
+    }
+
+    /**
+     * Asserts a fresh symbol lookup happens for `sessionToUse`.
+     *
+     * Checks the looked-up session with `toBe`, not `toHaveBeenCalledWith`
+     * -- koffi's session pointers have no enumerable properties, so
+     * vitest's deep equality can't tell two different sessions apart and
+     * would pass regardless of which one was actually used.
+     *
+     * @param sessionToUse - The session expected to require a fresh lookup; defaults to the shared `session`.
+     */
+    function expectUtf8OopToResolveViaSymbolLookup(sessionToUse: unknown = session) {
+        spyOnResolveSymbol(resolveSymbolSpy => {
+            gciLibrary.utf8ClassOop(sessionToUse);
+
+            expect(resolveSymbolSpy).toHaveBeenCalledTimes(1);
+            const [calledSession, calledSymbol] = resolveSymbolSpy.mock.calls[0];
+            expect(calledSession).toBe(sessionToUse);
+            expect(calledSymbol).toBe('Utf8');
+        });
+    }
+
+    /** Asserts `session`'s already-cached Utf8 oop is reused, without a fresh symbol lookup. */
+    function expectUtf8OopToBeCached() {
+        spyOnResolveSymbol(resolveSymbolSpy => {
+            gciLibrary.utf8ClassOop(session);
+
+            expect(resolveSymbolSpy).not.toHaveBeenCalled();
+        });
+    }
+
+    /**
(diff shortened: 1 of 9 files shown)


client/src/gciLibrary.ts
```diff
@@ -1794,4 +1830,488 @@ export class GciLibrary {
...
+  }
+
+  // ---------------------------------------------------------------------
+  // UserGlobals management
+  // ---------------------------------------------------------------------
+
+  private uniqueKeyCounter = 0n;
+
+  /**
+   * Returns a Smalltalk symbol literal (e.g. `#Key_12345`) guaranteed not to
```
MatiasFernandez: 💭 The doc says the key is "guaranteed not to already be in use as a key," but `nextUniqueKey` is just a monotonic counter scoped to this `GciLibrary` instance — it has no knowledge of the target dictionary, so a key like `#Key_0` could in principle collide with something a previous run committed to `UserGlobals`. The real uniqueness enforcement lives in `storeInUniqueUserGlobalsKey` (the `includesKey:` check that errors on collision). Should we soften this to describe what the counter actually gives you (unique per instance) and let `storeInUniqueUserGlobalsKey`'s doc carry the collision guarantee?

client/src/__tests__/useIntegrationTest.ts
```diff
@@ -73,38 +86,82 @@ export function useIntegrationTest(callback: UseIntegrationTestCallback) {
...
             process.env.VITE_GEMSTONE_STONE_NRS!,
-            null,
-            null,
-            false,
             process.env.VITE_GEMSTONE_GEM_NRS!,
-            process.env.VITE_GEMSTONE_USER!,
-            process.env.VITE_GEMSTONE_PASSWORD!,
-            0,
-            0,
+            options?.user || process.env.VITE_GEMSTONE_USER!,
```
MatiasFernandez: 💭 Since the whole point of the `user` option is to test login with different (e.g. invalid) credentials, is `||` a bit of a footgun here? An empty-string `user: ''` — a plausible invalid-credential case — would silently fall back to `VITE_GEMSTONE_USER` and log in successfully, so a test written around it would pass for the wrong reason. `options?.user ?? process.env.VITE_GEMSTONE_USER!` would only fall through on `undefined`. What do you think?

client/src/gciLibrary.ts
```diff
@@ -1794,4 +1830,488 @@ export class GciLibrary {
...
+
+  /**
+   * Releases the session's cached `Utf8` class oop (see {@link utf8ClassOop})
+   * from the PureExportSet and clears it from the cache. A no-op if nothing
+   * is cached for `session`.
+   *
+   * @param session - The GemStone session to operate in.
+   * @throws {GciLibraryError} If releasing the cached oop fails.
+   */
+  public releaseCachedSymbolOops(session: unknown) {
```
MatiasFernandez: nit: The name reads as plural and generic ("...SymbolOops"), but this only ever releases the single cached `Utf8` class oop — the cache is `Map<session, bigint>`, one oop per session. Is the generic name anticipating more cached symbols down the line, or would something like `releaseCachedUtf8Oop` match what it does today?

client/src/__tests__/gciLibrary.test.ts
```diff
@@ -1,24 +1,374 @@
...
+
+            expect(firstKey).not.toBe(secondKey);
+            expectOopToBeTrue(gciLibrary.execute(session, `(UserGlobals at: ${firstKey}) = 1`));
+            expectOopToBeTrue(gciLibrary.execute(session, `(UserGlobals at: ${secondKey}) = 2`));
+        });
     });
     
-    it('executes a Smalltalk expression and returns its result', () => {
-        const {result: utf8Oop} = gciLibrary.GciTsResolveSymbol(session, 'Utf8', OOP_NIL);
+    describe('SessionTemps management', () =>{
```
MatiasFernandez, line 347: 💭 `executeDiscardingResult`'s whole reason to exist is the PureExportSet contract — the `. nil` trick so it doesn't retain a result oop (per its doc). It looks like that's the one `execute` variant whose PureExportSet-neutrality isn't asserted directly; it's only covered transitively via the `resetSessionTemps` test. Given how carefully the rest of the PR pins PureExportSet accounting, is it worth a direct `expectPureExportSetToStayUnchanged(() => gciLibrary.executeDiscardingResult(session, 'Object new'))`?

client/src/__tests__/vitest.customErrorMatchers.ts
```diff
@@ -0,0 +1,66 @@
+import {expect} from 'vitest';
+
+expect.extend({
+    toThrowExactly(callback: () => unknown, expectedError: Error) {
```
MatiasFernandez, line 11: 💭 could we add documentation for the custom matchers? If in future work you see that Claude misses the existence of these matchers when they should have used them, we can add a short mention to the test rules file. I would wait to confirm the agent is not discovering these automatically, just in case

client/src/gciLibrary.ts
```diff
@@ -148,6 +150,37 @@ function toBigInt(value: number | bigint): bigint {
...
+ *   symbol, following the struct/pointer patterns already established below.
+ * - The remaining methods (grouped into labeled sections further down: Session
+ *   lifecycle, Code execution, Symbol resolution & Utf8 caching, Object
+ *   lifecycle & PureExportSet, UserGlobals/SessionTemps management, OOP
+ *   predicates) are an ergonomic layer on top: they call one or more
+ *   `GciTsXxx` methods, throw {@link GciLibraryError} on failure instead of
+ *   returning a `{success, err}`/`{result, err}` pair, and give the raw calls
+ *   memorable names and typed parameters.
+ *
+ * Recurring GemStone vocabulary used throughout this file's JSDoc:
```
MatiasFernandez, line 169: 👏  I like the idea of clarifying vocabulary. At some point, we could collect these into a glossary in the docs section.

client/src/gciLibrary.ts
```diff
@@ -1794,4 +1830,488 @@ export class GciLibrary {
...
+  public createString(session: unknown, contents: string){
+    const {result, err} = this.GciTsNewString(session, contents);
+
+    this.throwOnIllegalOop(result, err);
+
+    return result;
+  }
+
+  /** Returns the cached `Utf8` class oop for `session`, or `undefined` if none is cached. */
+  public cachedUtf8OopFor(session: unknown) {
```
MatiasFernandez: 💭 is this one public just for testing purposes?

client/src/gciLibrary.ts
```diff
@@ -1794,4 +1830,488 @@ export class GciLibrary {
...
+
+  // ---------------------------------------------------------------------
+  // OOP predicates
+  // ---------------------------------------------------------------------
+
+  /**
+   * Returns whether `oop` is `OOP_ILLEGAL` — the GCI sentinel used to signal
+   * that a call produced no valid result (symbol not found, error, etc.).
+   */
+  private isIllegalOop(oop: bigint) {
```
MatiasFernandez, line 2385: 💭 Are you planning on reifying OOPs as objects at some point, so these could be methods on the OOP object instead of passing around bigints?

MatiasFernandez (commented):
Great work @npapagna I left some questions/comments but nothing major

jgfoster:
While I haven't looked at the code, the description seems to point in a good direction. I'll let you respond to the other comments before merging.

jgfoster (approved):

client/src/gciLibrary.ts
npapagna: Great catch, fixed!

client/src/__tests__/useIntegrationTest.ts
npapagna: `??` is originally what I intended to use. Thanks for bringing this up!
Fixed 🔧

client/src/gciLibrary.ts
npapagna: We can go with `releaseCachedUtf8Oop`. A previous implementation of this PR cached all resolved symbols and I felt that was too much: we don't have strong evidence we need that so I cached only what was needed.
We can generalize this name later if needed.

client/src/__tests__/gciLibrary.test.ts
npapagna: I'm fine with testing this through other methods. Having said that, It's true that this one could be handy on its own, so it makes sense to test it directly.

Fixed! 🔧

client/src/__tests__/vitest.customErrorMatchers.ts
npapagna: Fixed! 🔧

client/src/gciLibrary.ts
npapagna: It was, yes. I made it private and found a way to test the same from the public interface.

client/src/gciLibrary.ts
npapagna: Not really. I did this because these well-known oops are an implementation detail. Probably rare, but a different version could have different values for these oops. That's why I encapsulated these checks.

MatiasFernandez (approved):
LGTM

PR #149, opened 2026-07-13
Lint GitHub Actions workflows: timeouts, actionlint, zizmor

npapagna:
## Before

The Health Check workflow had no job timeouts at all — every job would
run for up to GitHub's 360-minute (6h) default before being killed on a
hang. It also had no static analysis on `.github/workflows/*.yml`: no
syntax/expression checking, and no security scanning for things like
unpinned action references or token handling.

## What this PR does

- Sets `timeout-minutes` by hand on both existing jobs (1 min for
  `prepare`, 5 min for `health-check`), so a hang fails fast instead of
  running for 6 hours.
- Adds `scripts/lint-workflow-timeouts.sh` + `npm run
  lint:github:workflows`, which fails CI if any job — present or future
  — lacks `timeout-minutes`, so that fix can't silently regress the next
  time a job is added.
- Wires in `actionlint` and `zizmor` against every workflow file, on
  every push.

## What we found once linting was actually wired in

Turning on `actionlint` and `zizmor` against this repo's own workflow
immediately surfaced several pre-existing issues:

- **Unpinned actions** (`zizmor: unpinned-uses`) — `checkout`, `setup-node`,
  and `cache` were all referenced by a mutable tag (`@v6`), which can be
  repointed to different code without any change in this repo.
- **Credential persistence** (`zizmor: artipacked`) — `checkout` steps
  didn't set `persist-credentials: false`, so the job token stayed
  configured in the local git config for the rest of the job.
- **No explicit permissions** (`zizmor: excessive-permissions`) — the
  workflow ran with default token permissions instead of a minimal,
  explicit `permissions:` block.
- **Template injection** (`zizmor: template-injection`) — the GemStone
  matrix version was interpolated directly as `${{ matrix.gemstone-version }}`
  into a shell script. That substitution happens as text, before the
  shell parses anything, so an untrusted value could inject arbitrary
  shell commands.
- Two ShellCheck findings (`actionlint`) — unquoted `$GITHUB_OUTPUT` and
  `$(pwd)`, both word-splitting/globbing hazards.

## What's here now

- `lint-workflows` is a job of its own, separate from `prepare` (not
  serialized in front of the GemStone version matrix) so its ~10s of
  Docker image pulls (`actionlint`'s pinned image, `zizmor`'s official
  `zizmor-action`) don't delay the matrix from starting — confirmed
  against an actual run: merging it into `prepare` took `prepare` from
  ~7s to ~17s, all of it attributable to two cold image pulls that cost
  the same either way; keeping it separate pays that cost in parallel
  instead of serially.
- All the findings above are fixed: actions pinned to commit SHA
  (`checkout` at v7.0.0, `setup-node` at v6.4.0, `cache` at v6.1.0),
  `persist-credentials: false` on every checkout, a top-level
  `permissions: contents: read`, and the matrix version now flows
  through an `env:` var instead of direct `${{ }}` interpolation.
- Every changed line has an inline comment naming the specific rule it
  satisfies, so a future edit that reintroduces a finding is easy to
  trace back to why the line looks the way it does.

Verified locally and in a real workflow run: `lint:github:workflows`,
`actionlint`, and `zizmor` all pass clean against the final file.

### DIFF
--- a/.github/workflows/health-check.yml
+++ b/.github/workflows/health-check.yml
@@ -3,30 +3,36 @@ name: Health Check
 on: push
 
 jobs:
-  load-gemstone-integration-versions:
+  prepare:
     timeout-minutes: 1
     runs-on: ubuntu-latest
     outputs:
-      versions: ${{ steps.read.outputs.versions }}
+      versions: ${{ steps.read-gemstone-integration-versions.outputs.versions }}
     steps:
       - uses: actions/checkout@v6
         with:
           sparse-checkout: |
+            .github/workflows
+            /package.json
+            scripts/lint-workflow-timeouts.sh
             client/.gemstone-integration-releases.json
             client/bin/gemstone-integration-versions.js
             client/src/gemStoneVersion.js
           sparse-checkout-cone-mode: false
-      - id: read
+      - name: Lint workflow timeouts
+        run: npm run lint:github:workflows
+      - name: Read GemStone integration versions
+        id: read-gemstone-integration-versions
         run: echo "versions=$(node client/bin/gemstone-integration-versions.js)" >> $GITHUB_OUTPUT
 
   health-check:
     timeout-minutes: 5
     name: GemStone S/64 ${{ matrix.gemstone-version }}
-    needs: load-gemstone-integration-versions
+    needs: prepare
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        gemstone-version: ${{ fromJson(needs.load-gemstone-integration-versions.outputs.versions) }}
+        gemstone-version: ${{ fromJson(needs.prepare.outputs.versions) }}
 
     steps:
       - name: Checkout code
--- a/.github/workflows/health-check.yml
+++ b/.github/workflows/health-check.yml
@@ -4,6 +4,7 @@ on: push
 
 jobs:
   load-gemstone-integration-versions:
+    timeout-minutes: 1
     runs-on: ubuntu-latest
     outputs:
       versions: ${{ steps.read.outputs.versions }}
@@ -19,6 +20,7 @@ jobs:
         run: echo "versions=$(node client/bin/gemstone-integration-versions.js)" >> $GITHUB_OUTPUT
 
   health-check:
+    timeout-minutes: 5
     name: GemStone S/64 ${{ matrix.gemstone-version }}
     needs: load-gemstone-integration-versions
     runs-on: ubuntu-latest
--- a/.github/workflows/health-check.yml
+++ b/.github/workflows/health-check.yml
@@ -61,8 +61,9 @@ jobs:
           persist-credentials: false
       - name: Run actionlint
         # actionlint/shellcheck(SC2046): quote $(pwd) to prevent word
-        # splitting/globbing
-        run: docker run --rm -v "$(pwd)":/repo --workdir /repo rhysd/actionlint:1.7.12 -color
+        # splitting/globbing. Pinned to a digest, not a mutable tag, for the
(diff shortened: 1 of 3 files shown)


.github/workflows/health-check.yml
```diff
@@ -2,47 +2,112 @@ name: Health Check
...
+        # actionlint/shellcheck(SC2046): quote $(pwd) to prevent word
+        # splitting/globbing
+        run: docker run --rm -v "$(pwd)":/repo --workdir /repo rhysd/actionlint:1.7.12 -color
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@192e21d79ab29983730a13d1382995c2307fbcaa # v0.5.7
+        with:
+          advanced-security: false
 
   health-check:
+    timeout-minutes: 5
```
MatiasFernandez, line 73: 💭 Was this 5-minute budget validated against a cache-miss run? On a new `gemstone-version` the `Cache GemStone installation` step misses and the download happens inside this same job, on top of `npm ci`, `npm run compile`, server start, and the full test suite. If a legitimate cold run can approach 5 min, this would surface as intermittent timeout failures rather than catching a real hang. If 5 is comfortably above the observed worst case (cold cache included), ignore me.

scripts/lint-workflow-timeouts.sh
```diff
@@ -0,0 +1,39 @@
...
+missing=0
+
+for file in .github/workflows/*.yml .github/workflows/*.yaml; do
+    [ -e "$file" ] || continue
+
+    while IFS= read -r job; do
+        missing=1
+        echo "  - $file: job \"$job\" has no timeout-minutes"
+    done < <(
+        yq eval '.jobs // {} | to_entries | .[] | select(.value["timeout-minutes"] == null) | .key' "$file"
```
MatiasFernandez: 💭 Since this `yq` call is inside a process substitution (`< <(...)`), its exit status isn't visible to the parent shell, so `set -euo pipefail` won't trip if it fails — the loop just reads zero lines and the file gets treated as having no missing timeouts. So if `yq` ever errors (a malformed workflow, or a contributor who happens to have the *other* `yq` — the Python/jq wrapper — on PATH, where `command -v yq` passes but `eval` isn't understood), the linter reports a clean pass instead of failing. Would capturing into a var first — `jobs=$(yq eval '…' "$file")` and then looping over `$jobs` — be enough to let a yq failure actually fail the script?

.github/workflows/health-check.yml
npapagna: I tried both cached and non-cached! From the commit message that introduced the change:

```
Neither job set timeout-minutes, so a hang (e.g. a stone stuck
waiting on input) would run until GitHub's 360-minute default
killed it. Set explicit ceilings sized from real run history instead.

Checked ~60 matrix-job samples across 12 runs (2026-07-09 to
2026-07-13, warm cache): health-check ran 57-74s per GemStone
version, so its 5-minute timeout leaves 4-5x headroom. Also forced
a cold cache miss (deleted the cache via the GitHub UI) and reran:
worst case rose to 89s, still 3.4x under the limit. The version-list
load job ran 3-8s across the same runs against its 1-minute timeout.
```

npapagna: I also checked disabling `setup-node`'s cache (the npm cache) and the GemStone cache and the workflow took ~2 minutes. I believe we are fine!

scripts/lint-workflow-timeouts.sh
npapagna: Great catch! Thanks a lot.

Fixed 🔧

MatiasFernandez (approved):
LGTM :shipit:

PR #150, opened 2026-07-13
Adjust and align Node/VSCode/Electron/Typescript versions

MatiasFernandez:
## Summary

Jasper declared four unrelated, unvalidated "Node version" knobs. The real runtime floor — dictated by `engines.vscode: ^1.85.0` → Electron → bundled Node 18.15–18.17 (already EOL) — sat *below* the `@types/node@^20` the compiler trusted and *below* the Node used to build/test the repo. This raises the client/server floor to the earliest VS Code that bundles a currently-supported Node LTS (**VS Code 1.101.0 → Node 22.15.1**) and aligns every static knob to that real floor, so "type-checks pass" once again implies "runs on the floor."

- `engines.vscode` `^1.85.0` → `^1.101.0`, `engines.node` `>=22.12` → `>=22.15.1` (root `package.json`)
- `client/@types/vscode` bumped to `^1.101.0` to match
- root `@types/node` bumped to `^22.15`
- `tsconfig.base.json` `lib` updated to match `@tsconfig/node22` (`ES2024` + ESNext helper libs); `target`/`module` unchanged
- root `typescript` devDependency floor raised `^5.3.0` → `^5.7.2`, the minimum version that recognizes the new `ES2024` lib entry (ES2024 support landed in TS 5.7) — a separate, compiler-version concern from the runtime-floor knobs above
- `esbuild.mjs`: add `target: 'node22.15.1'` to the client and server builds (mcp-server intentionally left untouched — separate follow-up)
- CI (`health-check.yml`): dev jobs resolve Node from `.nvmrc` (via `node-version-file`), keeping the broad GemStone matrix on the common recent-VS-Code Node; a new `prepare` job output (`oldest-version`, from `gemstone-integration-versions.js --oldest`) feeds a new matrix `include` entry pinning Node `22.15.1` against the oldest supported GemStone version, as a floor smoke-test that won't go stale as the supported-versions list changes. Job names disambiguate the two jobs sharing that GemStone version.
- Docs split in two: `CONTRIBUTING.md` gets a short "Supported VS Code & Node versions" pointer section (current floor + link) and its stale CI note is fixed to reflect the dev/floor matrix split; the full floor-selection policy, the coordinated knob list, and the raise-the-floor steps now live in a new `docs/how-to/raising-the-version-floor.md`
- `.claude/rules/version-pins.md`: new path-scoped agent guardrail over the coordinated pin files, pointing at the how-to doc so future changes to any one knob get a warning to read the policy first rather than bumping it in isolation
- `CHANGELOG.md`: two entries under `[Unreleased]` — the VS Code/Node floor bump, and the `typescript` devDependency floor raise

Scope is client + server only — they share one floor since `server/` is forked in-process via `TransportKind.ipc`/`process.execPath`. `mcp-server/` hardening (loud version guard, absolute node path resolution, its own `engines`/CI floor, esbuild target, per-workspace `@types/node` split) is deferred to a separate follow-up ticket.

Ref: Gitlab#57

## Test plan
- [x] `npm run compile` clean under the narrowed `@types/node@22` / `@types/vscode@1.101` / `lib: ES2024` / `typescript@^5.7.2`
- [x] `npm test` against a local test stone
- [x] CI (`health-check` workflow) green across all 6 jobs: 5 dev jobs (`.nvmrc` Node × each GemStone version) + 1 floor smoke-test (Node `22.15.1` × GemStone `3.6.2`, the current oldest supported version)

### DIFF
--- a/.claude/rules/version-pins.md
+++ b/.claude/rules/version-pins.md
@@ -1,0 +1,12 @@
+---
+paths:
+  - "package.json"
+  - "client/package.json"
+  - "tsconfig.base.json"
+  - "esbuild.mjs"
+  - ".github/workflows/health-check.yml"
+---
+
+# VS Code / Node version pins
+
+These paths together encode one coordinated VS Code/Node/TypeScript version floor — not independent knobs. Never change one in isolation. Read [docs/how-to/raising-the-version-floor.md](../../docs/how-to/raising-the-version-floor.md) for the full coordinated-file list and the TypeScript devDependency note before touching any of them.
--- a/.github/workflows/health-check.yml
+++ b/.github/workflows/health-check.yml
@@ -13,6 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       versions: ${{ steps.read-gemstone-integration-versions.outputs.versions }}
+      oldest-version: ${{ steps.read-gemstone-integration-versions.outputs.oldest-version }}
     steps:
       # zizmor(unpinned-uses): actions must be pinned to a commit SHA, not
       # a mutable tag like @v6
@@ -42,7 +43,9 @@ jobs:
         id: read-gemstone-integration-versions
         # actionlint/shellcheck(SC2086): quote $GITHUB_OUTPUT to prevent
         # word splitting/globbing
-        run: echo "versions=$(node client/bin/gemstone-integration-versions.js)" >> "$GITHUB_OUTPUT"
+        run: |
+          echo "versions=$(node client/bin/gemstone-integration-versions.js)" >> "$GITHUB_OUTPUT"
+          echo "oldest-version=$(node client/bin/gemstone-integration-versions.js --oldest)" >> "$GITHUB_OUTPUT"
 
   lint-workflows:
     name: Lint GitHub Actions workflows
@@ -71,12 +74,22 @@ jobs:
 
   health-check:
     timeout-minutes: 5
-    name: GemStone S/64 ${{ matrix.gemstone-version }}
+    name: GemStone S/64 ${{ matrix.gemstone-version }}${{ matrix.node-version && format(' (Node {0})', matrix.node-version) || '' }}
     needs: prepare
     runs-on: ubuntu-latest
     strategy:
       matrix:
         gemstone-version: ${{ fromJson(needs.prepare.outputs.versions) }}
+        node-version: ['']            # empty ⇒ dev jobs resolve Node from .nvmrc (node-version-file); floor include overrides
+        # node-version's declared axis value is only '' (which falls back to
+        # the .nvmrc Node version); '22.15.1' isn't one of the declared
+        # values, so GitHub can't merge this into an existing combination —
+        # it adds an additional job pinned to the oldest GemStone version
+        # and the Node floor, as a smoke test to verify tests run on the
+        # Node floor too. This is to support older VS Code versions.
+        include:
+          - gemstone-version: ${{ needs.prepare.outputs.oldest-version }}
+            node-version: '22.15.1'   # pin the floor for the smoke-test
 
     steps:
       - name: Checkout code
@@ -92,7 +105,10 @@ jobs:
         # not a mutable tag like @v6
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: '22.15.1'
+          # Dev jobs leave node-version empty → Node resolved from .nvmrc.
+          # The floor include job sets node-version, which takes precedence.
+          node-version: ${{ matrix.node-version }}
+          node-version-file: '.nvmrc'
       - name: Install dependencies
         run: npm ci
       - name: Compile
--- a/.github/workflows/health-check.yml
+++ b/.github/workflows/health-check.yml
@@ -92,7 +92,7 @@ jobs:
         # not a mutable tag like @v6
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version-file: '.nvmrc'
+          node-version: '22.15.1'
       - name: Install dependencies
         run: npm ci
       - name: Compile
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,6 +41,12 @@ Follow these steps in order after cloning the repo:
 
    > **Destructive:** re-running `test:server:start` stops any running stone, resets the database to a pristine state, and overwrites `.env.test`.
 
+## Supported VS Code & Node versions
+
+**Current floor:** VS Code `1.101.0` → bundled Node `22.15.1`.
+
+See [docs/how-to/raising-the-version-floor.md](docs/how-to/raising-the-version-floor.md) for the policy behind this floor and the steps to raise it.
+
 ## Build and test
 
 - Build: `npm run compile`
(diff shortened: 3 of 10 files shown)


tsconfig.base.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "ES2022",
     "module": "commonjs",
-    "lib": ["ES2022"],
+    "lib": ["ES2024", "ESNext.Array", "ESNext.Collection", "ESNext.Iterator"],
```
npapagna, line 5: This `lib` bump does not actually guarantee compatibility with the `typescript` devDependency range, which this PR leaves at `^5.3.0` (package.json).

I checked the actual published `typescript` npm tarballs: the earliest version shipping all four of these lib files is **5.7.2** (`ESNext.Collection` landed in 5.4.2, `ESNext.Array` in 5.5.2, `ESNext.Iterator` in 5.6.2, `ES2024` in 5.7.2). `^5.3.0` permits resolving anywhere from 5.3.0 up to <6.0.0, so plenty of valid resolutions (5.3.x-5.6.x) are missing one or more of these lib definitions.

It is not broken today only because `package-lock.json` pins the resolved version to 5.9.3. But if the lockfile is ever regenerated against this declared range (fresh `npm install`, a different registry mirror, etc.), `tsc` would fail with `TS6046: Cannot find lib definition for 'es2024'`, breaking `npm run compile` and `npm test`.

Given this PR's own `.claude/rules/version-pins.md` establishes a "coordinated set" of knobs that must move together with the Node/VS Code floor, it seems worth adding the TypeScript compiler's own minimum version to that list and bumping the devDependency to something like `^5.7.2`, so the declared range matches what this `lib` array actually requires.

MatiasFernandez: fixed

npapagna, line 5: 🤖 Claude: `mcp-server/tsconfig.json` extends this base with no `lib`/`target` override, so it silently inherits the widened ES2024/Iterator-helper/Set-collection surface. `esbuild.mjs`'s `target: 'node22.15.1'` bump and the `engines.node` floor were both scoped to client+server only (per the PR description), and `mcp-server/package.json` declares no `engines` field at all — meanwhile `docs/mcp-server.md` documents mcp-server as launched via a bare `node` resolved from the *client machine's* PATH, not VS Code's bundled Node. If shared code (or `mcp-server/src`) later adopts one of these newly-inherited APIs (`Object.groupBy`, `Iterator.prototype.map/take/drop`, `Set.prototype.union/intersection`, etc.), it'll type-check and bundle cleanly yet can throw at runtime on an older Node — the exact "type-check passes but doesn't imply it runs" gap this PR's own CHANGELOG says it's fixing, reintroduced for mcp-server.

I see the PR description already flags mcp-server hardening as a deferred follow-up — worth making sure that follow-up scopes for the *widened* lib surface (ES2024 vs the prior ES2022), not just the pre-existing coupling.

esbuild.mjs
```diff
@@ -18,6 +19,7 @@ await esbuild.build({
   format: 'cjs',
   external: ['vscode', 'koffi'],
   sourcemap: true,
+  target: 'node22.15.1',
 });
 
 await esbuild.build({
```
npapagna, line 25: 🤖 Claude: this third `build()` call (mcp-server) is the one that doesn't get `target: 'node22.15.1'`, unlike the two above — see the tsconfig.base.json comment for why that matters now that the shared `lib` array was also widened for this workspace.

.github/workflows/health-check.yml
```diff
@@ -92,6 +105,9 @@ jobs:
         # not a mutable tag like @v6
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
+          # Dev jobs leave node-version empty → Node resolved from .nvmrc.
+          # The floor include job sets node-version, which takes precedence.
+          node-version: ${{ matrix.node-version }}
```
npapagna, line 110: 🤖 Claude: for the floor smoke-test job, `matrix.node-version` is `'22.15.1'` (non-empty) while `node-version-file: '.nvmrc'` below is also always set. `actions/setup-node@48b55a0` (v6.4.0)'s `resolveVersionInput()` logs `core.warning('Both node-version and node-version-file inputs are specified, only node-version will be used')` whenever both are truthy — so this one job will always emit that warning in its logs (harmless, but noisy, and could bury a real future warning in the same job).

CONTRIBUTING.md
```diff
@@ -41,6 +41,12 @@ Follow these steps in order after cloning the repo:
 
    > **Destructive:** re-running `test:server:start` stops any running stone, resets the database to a pristine state, and overwrites `.env.test`.
 
+## Supported VS Code & Node versions
+
+**Current floor:** VS Code `1.101.0` → bundled Node `22.15.1`.
```
npapagna, line 46: 🤖 Claude: this "Current floor" sentence is duplicated verbatim in `docs/how-to/raising-the-version-floor.md`'s opening line. That doc's own "How to raise it" checklist (step 4: "Update the 'Current floor' line at the top of this document") only reminds the maintainer to update its own copy — not this one. A future floor-raise that follows the checklist literally will update the how-to doc but leave this line stale, giving the repo two contradictory floor statements.

.claude/rules/version-pins.md
```diff
@@ -0,0 +1,12 @@
...
+  - "package.json"
+  - "client/package.json"
+  - "tsconfig.base.json"
+  - "esbuild.mjs"
+  - ".github/workflows/health-check.yml"
+---
+
+# VS Code / Node version pins
+
+These paths together encode one coordinated VS Code/Node/TypeScript version floor — not independent knobs. Never change one in isolation. Read [docs/how-to/raising-the-version-floor.md](../../docs/how-to/raising-the-version-floor.md) for the full coordinated-file list and the TypeScript devDependency note before touching any of them.
```
npapagna, line 12: 🤖 Claude: this "coordinated set, don't change in isolation" invariant is enforced only via this markdown agent-rule + the how-to doc — nothing in the repo (script, lint, CI job) actually parses these files and asserts they encode the same floor. The repo already has precedent for automating exactly this kind of guard (`scripts/lint-workflow-timeouts.sh`, wired into CI via `npm run lint:github:workflows` in the `prepare` job) — a human editor, or any AI session without this specific rule loaded, gets no automated feedback if they break the coordination.

npapagna (approved):
Great work @MatiasFernandez!
Left a couple of findings. I think they're expected, so I'm approving this one.

tsconfig.base.json
MatiasFernandez: The node version for `mcp-server` is a different story; I'm still deciding how we should tackle it. I'll leave that as follow-up work

esbuild.mjs
MatiasFernandez: same as above, I'm deferring mcp-server version tweaks for a follow-up

.github/workflows/health-check.yml
MatiasFernandez: I couldn't find a simple way to remove this; there were alternatives, but they required reimplementing logic that is already defined in the `version-file:` feature. So I preferred the warning instead of a complex solution for this.

CONTRIBUTING.md
MatiasFernandez: This one is hard to fix; both documents fulfill a different role, and I think this small repetition isn't that bad. Updating the version with the help of an agent would catch this

.claude/rules/version-pins.md
MatiasFernandez: I'm working on adding a linter, I'll investigate if we can have a rule to enforce this

PR #151, opened 2026-07-13
Fix flaky PureExportSet membership check on dead oops

npapagna:
isOopIncludedInPureExportSet resolved the checked oop via `Object objectForOop:` before testing membership. If that oop no longer referenced a live object (e.g. already GC'd), the resolve itself raised an ArgumentError instead of just reporting "not found" -- causing sporadic CI failures that looked like a real bug.

Now it compares each live PureExportSet member's own oop against the target instead of resolving the target into an object, so a dead oop just reads as absent. Added a regression test using the oop of `nil`, a documented case of an oop that's never entered in the PureExportSet.

Also moved `nilOop()` into the OOP predicates section with the rest of its kin (it had landed at the end of the class with no home or doc comment) and updated the class-level doc comment's list of ergonomic sections, which had drifted out of sync with the actual sections.

### DIFF
--- a/client/src/__tests__/gciLibrary.test.ts
+++ b/client/src/__tests__/gciLibrary.test.ts
@@ -494,6 +494,10 @@ describe('GciLibrary', () => {
             expectPureExportSetToIncludeOop(false, oopToRelease);
         })
 
+        it('does not include an oop that was never stored', () => {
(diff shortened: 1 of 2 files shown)


MatiasFernandez (approved):
LGTM :shipit:

PR #152, opened 2026-07-14
Add paged UTF-8 string fetching to GciLibrary

npapagna:
## Summary
- Adds `executeAndRelease` (evaluate code, auto-release the result oop around a callback) and `executeAndFetchString` (evaluate code and fetch its result as a decoded UTF-8 string, paged out via `GciTsFetchBytes` in `FETCH_STRING_PAGE_SIZE_BYTES` chunks)
- Extracts the existing "resolve-then-release" pattern in `resolveSymbol` into a shared `releaseAfterUse`/`safelyReleaseObject` helper pair
- `executeAndFetchString` isn't called anywhere yet — it's landing ahead of its use, with its own test coverage. A follow-up PR will wire it in to fix UTF-8/long-string truncation issues elsewhere in Jasper

## Test plan
- [x] `npm run compile`
- [x] `npm test` (all 3225 client / 320 server / 95 mcp-server tests pass)
- [x] New `gciLibrary.test.ts` coverage for `executeAndRelease` and `executeAndFetchString`, including a page-boundary multi-byte-character split test, verified against the real GCI library with a running test stone

### DIFF
--- a/client/src/__tests__/gciLibrary.test.ts
+++ b/client/src/__tests__/gciLibrary.test.ts
@@ -149,7 +149,36 @@ describe('GciLibrary', () => {
     function expectPureExportSetToIncludeOop(shouldBeIncluded: boolean, oop: bigint) {
         expect(gciLibrary.isOopIncludedInPureExportSet(session, oop)).toBe(shouldBeIncluded);
     }
-    
+
+    /**
+     * Forces the next `releaseObject` call to throw for the duration of
+     * `callback`, then restores it.
+     *
+     * @param callback - The operation to run while `releaseObject` is rigged to fail.
+     */
+    function simulateReleaseObjectFailure(callback: () => void) {
+        const spy = vi.spyOn(gciLibrary, 'releaseObject').mockImplementationOnce(() => {
+            throw GciLibraryError.withMessage('Simulated releaseObject failure');
+        });
+
+        try {
+            callback();
+        } finally {
+            spy.mockRestore();
+        }
+    }
+
+    /**
+     * Asserts that evaluating `codeToEvaluate` and fetching its result as a
+     * string yields `expectedResult`.
+     *
+     * @param codeToEvaluate - Smalltalk source to evaluate.
+     * @param expectedResult - The string the evaluated result is expected to decode to.
+     */
+    function expectEvaluatedStringToBe(codeToEvaluate: string, expectedResult: string) {
+        expect(gciLibrary.executeAndFetchString(session, codeToEvaluate)).toBe(expectedResult);
+    }
+
     describe('evaluating expressions', () => {
 
         it('returns the result of evaluating an expression', () => {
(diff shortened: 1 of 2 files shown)


client/src/gciLibrary.ts
```diff
@@ -2473,5 +2529,86 @@ export class GciLibrary {
...
+   * additionally hand back a `*utf8String` oop that the caller must track
+   * and release, for no benefit.
+   *
+   * @param session - The GemStone session to operate in.
+   * @param stringOop - The oop of a `Utf8` (or byte-object) instance to fetch.
+   * @param pageSize - The maximum number of bytes to fetch per GCI round-trip.
+   * @returns The fetched bytes, decoded as a UTF-8 JS string.
+   * @throws {GciLibraryError} If the underlying GCI call fails.
+   */
+  private fetchUtf8String(session: unknown, stringOop: bigint, pageSize: number) {
```
MatiasFernandez, line 2591: 💭 `fetchUtf8String` is private and its only caller always passes `FETCH_STRING_PAGE_SIZE_BYTES`, so `pageSize` is effectively fixed. The boundary tests exercise the paging edges by allocating full 256KB+ strings — which works, but is a fair bit of allocation per case. Since the param is already here, was the idea to let tests drive the boundaries with a small page size instead (and skip the big allocations)? Or if it's not meant to vary, would reading the constant directly here be simpler? Either way no objection — just checking the intent.

client/src/gciLibrary.ts
```diff
@@ -2115,6 +2135,42 @@ export class GciLibrary {
...
+   * matters to the caller.
+   *
+   * @param session - The GemStone session to operate in.
+   * @param oop - The oop to release.
+   */
+  private safelyReleaseObject(session: unknown, oop: bigint) {
+    try {
+      this.releaseObject(session, oop);
+    } catch (releaseError) {
+      console.warn(`Failed to release oop ${oop}:`, releaseError);
```
MatiasFernandez, line 2151: 💭 Nice extraction. One small thing: the old `resolveSymbol` release-failure warning named the symbol (`Failed to release symbol name oop for '${symbolName}'`), and now that it goes through this shared helper it's just `Failed to release oop ${oop}`. For the generic callers that's the right message, but for `resolveSymbol` we lose the symbol-name context that would help pin down which release failed. Worth threading through, or is the oop enough? Not blocking — it's only a warn-level log for a rare failure.

client/src/__tests__/gciLibrary.test.ts
```diff
@@ -609,5 +805,5 @@ describe('GciLibrary', () => {
                 })
             })
         });
-    })
+})
```
jgfoster: I think the old formatting was better.

jgfoster (approved):
One cosmetic formatting issue. Otherwise this looks nice!

client/src/gciLibrary.ts
npapagna: I believe the oop is enough: if the release ever fails, `Object objectForOop: oop` against the same repository resolves it straight back to the symbol (or whatever it was), since a release failure means it's still reachable. The context isn't lost, just one step away instead of inline in the log. I feel it's not worth threading the name through for a rare warn-level case.

client/src/gciLibrary.ts
npapagna: Good question. Honestly, neither of those, on purpose. I left pageSize as a plain parameter without committing to either direction, because I think there's a bigger design question hiding underneath it, and I didn't want to bake in an answer before we'd thought it through: should executeAndFetchString stay a low-level primitive where callers pick the page size, or should it become a higher-level (Jasper-ish) method that just picks a sensible size for you?                                                           

Tied into that is the tests point you raised. I'd like to eventually derive some real numbers from actual GemStone instances (method/class-definition sizes) to pick sane defaults, but customers' codebases will vary, so I don't think there's one "right" constant. That's the rabbit hole I was trying to stay out of for this PR.                  

Definitely something I want to pick up in a follow-up. Happy to loop you in when I get to it!

client/src/__tests__/gciLibrary.test.ts
npapagna: Oops, that was a slip. Fixed!

client/src/gciLibrary.ts
MatiasFernandez: Makes sense! Let's keep it as-is now and we can continue iterating it in the future

MatiasFernandez (approved):
:shipit:

PR #161, opened 2026-07-16
Reject non-local returns in gciLibrary's code-evaluation helpers

npapagna:
## Summary
- `executeAndFetchString` wraps `code` as `[ code ] value encodeAsUTF8`; a top-level `^` in `code` would non-locally return before that `encodeAsUTF8` send ran, silently skipping the encoding step `fetchUtf8String` relies on to receive a `Utf8` byte object.
- `executeDiscardingResult` has the same shape of bug: it wraps `code` as `[ code ] value. nil` specifically to keep `code`'s own result out of the PureExportSet, but a top-level `^` would skip the trailing `. nil` statement and leak `code`'s real result oop into the PureExportSet forever.
- Both methods now reject `code` containing a non-local return upfront, via a shared `denyIncludesANonLocalReturn` check, with a clear error message telling the caller to use a plain expression instead.

## Test plan
- [x] `npm run compile`
- [x] `npm test` (client workspace `gciLibrary.test.ts`, 81/81 passing, against a live test stone)

### DIFF
--- a/client/src/__tests__/gciLibrary.test.ts
+++ b/client/src/__tests__/gciLibrary.test.ts
@@ -485,7 +485,7 @@ describe('GciLibrary', () => {
 
                     | encodeAsUTF8LiarClass |
                     encodeAsUTF8LiarClass := Object subclass: #EncodeAsUTF8Liar instVarNames: {} inDictionary: UserGlobals.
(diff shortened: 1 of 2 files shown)


npapagna:
Superseded by #162: we actually need to support non-local returns (see PR description).

PR #162, opened 2026-07-17
Support non-local returns in GciLibrary's execute helpers

npapagna:
## Summary

Supersedes #161

- `executeAndFetchString` sends `encodeAsUTF8` as its own `performAndRelease` call on the evaluated result oop, instead of appending it to the evaluated source (`[ code ] value encodeAsUTF8`).
- `executeDiscardingResult` wraps `code` in `[ code ] ensure: [ ^ nil ]`, instead of appending a trailing `nil` statement (`[ code ] value. nil`).
- Both previously assumed the evaluated source would always run to completion and hand control back at the point right after `code`. A non-local return (`^`) inside `code` breaks that assumption: it exits the whole doit immediately, skipping whatever was appended after it in the source string.

## Why this matters

`executeAndFetchString` is going to back most of the query/eval paths in Jasper, so we don't control what code runs through it. Users can (and do) write workspace snippets with an explicit `^` — a non-local return is completely ordinary Smalltalk. Before this fix:

- `executeAndFetchString` would silently skip the `encodeAsUTF8` send, so unusual string encodings (e.g. a UTF-16 result) would reach `GciTsFetchUtf8Bytes`, which doesn't support that encoding, and come back garbled instead of throwing.
- `executeDiscardingResult` would leak the discarded result's oop into the session's `PureExportSet` forever, since callers assume the discard produces `nil` (never tracked) and never release it.

Both bugs were silent, not crashes, so they'd have been easy to miss without an explicit regression test covering the `^` case.

## Test plan

- [x] `npm run compile`
- [x] `npm test` (142 client test files / 3258 tests, 12 server files / 320 tests, 3 mcp-server files / 95 tests — all passing, including two new regression tests exercising a non-local return in each helper)

### DIFF
--- a/client/src/__tests__/gciLibrary.test.ts
+++ b/client/src/__tests__/gciLibrary.test.ts
@@ -600,6 +600,10 @@ describe('GciLibrary', () => {
             });
         });
 
+        it('returns the result of code that uses a non-local return', () => {
+            expectEvaluatedStringToBe(`^ 'a' encodeAsUTF16`, 'a');
+        });
+
     })
 
     describe('UserGlobals management', () => {
--- a/client/src/__tests__/gciLibrary.test.ts
+++ b/client/src/__tests__/gciLibrary.test.ts
@@ -315,6 +315,98 @@ describe('GciLibrary', () => {
 
     });
     
+    describe('sending messages', () => {
+        
+        it ('returns the result of sending a message', () => {
+            const result = gciLibrary.perform(session, gciLibrary.falseOop(), 'not');
+
+            expectOopToBeTrue(result)
+        })
+
+        it ('throws when the selector cannot be resolved', () => {
+            expectToThrowGciLibraryError(
+                () => gciLibrary.perform(session, gciLibrary.falseOop(), 'foo'),
+                'a NameError occurred (error 2404), foo, There is no Symbol with the specified value'
+            )
+        })
+        
+    })
+    
+    describe('sending a message and releasing its result automatically', () => {
+
+        it('passes the resulting oop to the callback', () => {
+            gciLibrary.performAndRelease(session, gciLibrary.falseOop(), 'not', resultOop => {
+                expectOopToBeTrue(resultOop);
+            });
+        });
+
+        it('returns the result of evaluating the callback', () => {
+            const expectedResult = 'callback result';
+
+            const result = gciLibrary.performAndRelease(session, gciLibrary.falseOop(), 'not', () => expectedResult);
+
+            expect(result).toBe(expectedResult);
+        });
+
+        it('releases the resulting oop after the callback returns', () => {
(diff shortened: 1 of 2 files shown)


client/src/__tests__/gciLibrary.test.ts
```diff
@@ -315,6 +321,98 @@ describe('GciLibrary', () => {
...
+        it ('throws when the selector cannot be resolved', () => {
+            expectToThrowGciLibraryError(
+                () => gciLibrary.perform(session, gciLibrary.falseOop(), 'foo'),
+                'a NameError occurred (error 2404), foo, There is no Symbol with the specified value'
+            )
+        })
+        
+    })
+    
+    describe('sending a message and releasing its result automatically', () => {
```
MatiasFernandez, line 341: 💭 This battery (releases after return, releases on throw, re-throws unchanged, returns the callback's result, still returns when release fails, still throws when both fail, doesn't run the callback on error) is nearly line-for-line identical to the `evaluating an expression and releasing its result automatically` block above — which makes sense, since both `executeAndRelease` and `performAndRelease` just delegate to `releaseAfterUse`.

Is it worth extracting a shared parameterized battery that takes a "produce a result oop" function, so the release contract is specified once and each public method just wires into it? I can see the counter-argument for documenting each public method's contract independently too — just flagging the duplication in case it's easy to fold together.

client/src/gciLibrary.ts
```diff
@@ -2549,6 +2615,14 @@ export class GciLibrary {
    * `GciTsFetchBytes` -- see that method's comment for why
    * `GciTsFetchUtf8Bytes` itself isn't used for the fetch either.
    *
+   * The `encodeAsUTF8` send happens as its own {@link performAndRelease}
+   * call on the already-evaluated result oop, rather than being appended to
+   * `code`'s source (e.g. `[ ${code} ] value encodeAsUTF8`). Appending it to
+   * the source is unsafe: a non-local return (`^`) inside `code` exits the
```
MatiasFernandez, line 2621: 💭 What do you think about adding a short mention about how "the extra round-trip" is an explicit decision/trade-off? I want to avoid agents in the future, assuming this function can be optimized, removing the extra round-trip

MatiasFernandez (approved):
Looks good, I left some very minor comments

client/src/__tests__/gciLibrary.test.ts
npapagna: Yes, I thought about that. I’ll probably address it in a separate PR. There are a lot of UTF-8 issues to fix, and I want to make sure everything keeps working, so I’m trying not to stray too far at the moment.

client/src/gciLibrary.ts
npapagna: Fixed!

PR #163, opened 2026-07-17
Clarify that the NetLDI login field accepts a name or a port number

MatiasFernandez:
## Summary
Clarifies in the login editor UI that the NetLDI field accepts either a NetLDI service name or a port number. An external user asked whether Jasper's login config can accept a NetLDI port number instead of a service name (for connecting to remote stones over SSH port-forwarding). It already does — the value is passed straight into the GCI network resource string (`#netldi:<value>`) as an opaque string — but the UI only hinted at a name-style example.

### Before

image

### After 

image

No parsing/validation change is needed — the value is already treated as an opaque string on login.

Addresses Gitlab#69.


srbaker (approved):
:shipit:

PR #169, opened 2026-07-17
Fix flaky GciLibrary selector-error tests

npapagna:
## Summary
- Two tests asserted that sending an unrecognized selector (`'foo'`) to a Boolean raises `NameError`. GemStone only raises `NameError` the first time a given symbol is sent anywhere in the stone; once any test interns it, later sends see `MessageNotUnderstood` instead. Under vitest's randomized test order, whichever test happened to intern `'foo'` first flipped the expected error for tests running after it, causing intermittent CI failures.
- Both tests now use `'new'`, a selector GemStone always has already interned, so they deterministically hit `MessageNotUnderstood` regardless of run order. Added comments explaining the reasoning so future selector choices in this file don't reintroduce the flake.

## Test plan
- [x] `npx vitest run src/__tests__/gciLibrary.test.ts -t "sending"` passes
- [x] Ran the full suite with `VITEST_SEED=1|2|3` to confirm no order-dependent flake

### DIFF
--- a/client/src/__tests__/gciLibrary.test.ts
+++ b/client/src/__tests__/gciLibrary.test.ts
@@ -350,10 +350,15 @@ describe('GciLibrary', () => {
       expectOopToBeTrue(result);
     });
 
-    it('throws when the selector cannot be resolved', () => {
+    it('throws when the receiver does not understand the selector', () => {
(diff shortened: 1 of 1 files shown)


ericwinger:
Looks good to me. :+1:

ericwinger (approved):

PR #170, opened 2026-07-17
Repro: perform: NameError vs MessageNotUnderstood

npapagna:
## Summary
- While tracking down the two flaky `GciLibrary` tests fixed in #169, we found the underlying GemStone behavior worth writing up separately: `perform:` raises `NameError` for a selector that's never been compiled as a Symbol in that session, but `MessageNotUnderstood` once it has, even if the thing that interned it had nothing to do with sending that message (a bare `#literal` or an `asSymbol` send is enough).
- Follows the same write-up format as the earlier non-ASCII compiler finding (`npapagna/non-ascii-source-cursor-finding-repro`): a `README.md` with the symptom/trigger/open-questions, a Vitest reproduction, and a standalone C program for easy hand-off. Also extends `client/src/__tests__/repro/**` into the `gci` vitest project (on-demand, not part of `npm test`), matching that branch's config change.
- We verified this is scoped to the session that did the compiling — it does **not** appear to reach other sessions, even across an explicit commit, which surprised us; see the "still not sure" section in the README.
- Checked against every locally-available GemStone release, 3.6.1 through 3.7.5 — reproduces identically across all of them (3.7.4.3 untested; a machine-local shared-memory issue blocked that one stone mid-sweep, unrelated to GemStone itself).

## Test plan
- [x] `npm run test:gci -- src/__tests__/repro/InternedSelectorRepro.test.ts` passes (5/5), repeatedly, across shuffled seeds
- [x] `./client/src/__tests__/repro/run-c-repro.sh [version]` builds and runs cleanly against 3.6.1, 3.6.2, 3.6.3, 3.6.4, 3.6.5, 3.6.6, 3.6.8, 3.7.2, and 3.7.5 — all cases pass on each
- [x] `npm run lint && npm run format:check && npm run compile && npm test` all pass

### DIFF
--- a/client/src/__tests__/repro/InternedSelectorRepro.c
+++ b/client/src/__tests__/repro/InternedSelectorRepro.c
@@ -1,0 +1,294 @@
+/*
+ * Standalone C reproduction of a GemStone finding. See README.md (same
+ * directory) for the full write-up -- what we found, why picking a
+ * made-up literal for a "selector cannot be resolved" test is fragile, and
+ * what we verified (and didn't) about how far the effect reaches.
+ *
+ * This file loads libgcits dynamically (dlopen/dlsym) instead of linking
+ * against GemStone's SDK headers, so it builds with nothing but a C compiler
+ * and the shared library itself. Signatures below are transcribed from
+ * gcits.hf / gci.ht (bundled in this repo under docs/3.7/) and cross-checked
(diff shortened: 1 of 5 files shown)


npapagna:
Closing this out — decided not to route the repro write-up through a PR after all. The branch (`repro/gci-perform-name-error-vs-message-not-understood`) is still up if it's useful to reference or link directly.

PR #173, opened 2026-07-20
Open Getting Started walkthrough on first GemStone view reveal

jgfoster:
The walkthrough's first step teaches how to connect to a stone, but it previously auto-opened only after a successful connect — so the "how to connect" guidance arrived after the user had already worked it out.

Move the trigger earlier: open it the first time the GemStone (Logins) view is revealed, via a TreeView.onDidChangeVisibility listener. Revealing the view proves intent and happens before connecting, so the connect step lands when it's useful. The once-per-machine gate (globalState key gemstone.hasSeenGettingStarted) is unchanged; the open logic is extracted into an idempotent, exported maybeOpenGettingStarted() so repeated visibility events can't double-open.

Update the resetGettingStarted message and resources/walkthrough/reopen.md copy to describe the new trigger, and add unit tests for the helper.

Fix https://code.gemtalksystems.com/GemStone/grail/-/work_items/49.

### DIFF
--- a/client/src/__mocks__/vscode.ts
+++ b/client/src/__mocks__/vscode.ts
@@ -230,7 +230,10 @@ export const window = {
   showErrorMessage: vi.fn(),
   showInformationMessage: vi.fn(),
   showWarningMessage: vi.fn(),
-  createTreeView: vi.fn(() => ({ dispose: () => {} })),
+  createTreeView: vi.fn(() => ({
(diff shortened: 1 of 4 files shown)


MatiasFernandez (approved):
I tested this and it works. But it may me think about the actual goal of this page and I think we'll need to add some extra changes to make it really useful for new users. I'll share that feedback directly with you, but in terms of what was described in the ticket, this PR achieves that.

PR #176, opened 2026-07-20
Fix Display It's non-ASCII printString results

npapagna:
## Summary

- Typing a non-ASCII literal into Execute It/Display It/Debug It (e.g. `UserGlobals at: #'A—B' put: 42`) stored it **corrupted** in GemStone: the old source-encoding resolution (`resolveOopClassString`, using GemStone's `String` class) treated each byte of the literal's UTF-8 encoding as its own raw character, so the em dash's 3-byte sequence became 3 separate garbage characters in the actual stored object.
- This was invisible from Display It itself: rendering the result went through an equally old raw-byte fetch (`GciTsPerformFetchBytes`) that shipped those same corrupted bytes back out essentially unencoded, which happened to UTF-8-decode into the original em dash on the client — two independent bugs canceling out by coincidence. `main` renders `#'A—B'` correctly in Display It today; this isn't a visible regression fix.
- The corruption only surfaces once you read that same data through anything that fetches correctly, e.g. this repo's already-UTF8-fixed dictionary/globals browser queries — there it shows as three garbage characters instead of the intended em dash.
- Fixes both halves together: `resolveUtf8ClassOopUsing` now resolves the true `Utf8` class for compiling source (so the literal is stored correctly), and `fetchResultString` now fetches via `executeAndFetchString` (UTF-8-safe) instead of the raw byte fetch. Display It's own rendering is unchanged; the underlying data is now genuinely correct instead of accidentally-appearing-correct.
- Depends on #175: `resolveUtf8ClassOopUsing` now lets a resolution failure propagate instead of swallowing it into a `bigint | undefined`, which is only safe because that PR made sure command failures actually surface instead of vanishing.

## Test plan

- [ ] Display It on `UserGlobals at: #'A—B' put: 42.` then `UserGlobals keys` — still renders the em dash correctly (no visible change vs. before)
- [ ] Open the Globals Browser on the same global — confirm it now *also* shows the em dash correctly (on unpatched `main`, this would show 3 garbage characters instead, diverging from what Display It showed)
- [ ] Display It on plain ASCII (`3 + 4`, `'hello'`, `nil`) — no regression
- [ ] Display It on a large result (e.g. `(1 to: 50000) asArray`) — completes without hanging (note: the old 64KB cap is gone; a large result may now take longer / show more than before, that's expected)
- [ ] `npm run lint && npm run format:check && npm run compile && npm test` all pass

### DIFF
--- a/client/src/__tests__/codeExecutor.test.ts
+++ b/client/src/__tests__/codeExecutor.test.ts
@@ -52,7 +52,7 @@ const OOP_NIL = 0x14n;
 
 function makeGci(overrides: Record<string, unknown> = {}) {
   return {
-    GciTsResolveSymbol: vi.fn(() => ({ result: 100n, err: { number: 0 } })),
+    utf8ClassOop: vi.fn(() => 100n),
     GciTsNbExecute: vi.fn((): Record<string, unknown> => ({
       success: true,
       err: { number: 0, message: '' },
@@ -66,11 +66,15 @@ function makeGci(overrides: Record<string, unknown> = {}) {
     })),
     GciTsPerformFetchBytes: vi.fn(() => ({ data: '42', err: { number: 0 } })),
     GciTsExecuteFetchBytes: vi.fn(() => ({ data: '', err: { number: 0 } })),
+    executeAndFetchString: vi.fn(() =>
+      expect.unreachable(
+        'executeAndFetchString mock not configured for this test -- call ' +
+          '(gci.executeAndFetchString as Mock).mockReturnValue(...) before triggering Display It.',
+      ),
+    ),
     GciTsClearStack: vi.fn(),
     GciTsObjExists: vi.fn(() => false),
     GciTsFetchClass: vi.fn(() => ({ result: 0n, err: { number: 0 } })),
-
-    GciTsExecute: vi.fn(() => ({ result: 0n, err: { number: 0 } })),
     ...overrides,
   };
 }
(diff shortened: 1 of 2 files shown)


client/src/codeExecutor.ts
```diff
@@ -160,8 +159,7 @@ export class CodeExecutor {
       return;
     }
 
-    const oopClassString = this.resolveOopClassString(session);
-    if (oopClassString === undefined) return;
+    const oopClassString = this.resolveUtf8ClassOopUsing(session);
```
MatiasFernandez, line 162: 💭 Now that this resolves the `Utf8` class oop (not `String`), the `oopClassString` name reads as a little misleading — it looks like "the oop of class String", or even "a string". Since what it actually holds is the source class passed to `GciTsNbExecute`, would `sourceClassOop` (or `utf8ClassOop`) be clearer? Same variable over in `executeAndInspect` (line 742).

client/src/__tests__/codeExecutor.test.ts
```diff
@@ -66,11 +66,15 @@ function makeGci(overrides: Record<string, unknown> = {}) {
     })),
     GciTsPerformFetchBytes: vi.fn(() => ({ data: '42', err: { number: 0 } })),
     GciTsExecuteFetchBytes: vi.fn(() => ({ data: '', err: { number: 0 } })),
+    executeAndFetchString: vi.fn(() =>
```
MatiasFernandez, line 69: Nice — having the default `executeAndFetchString` mock fail loudly with a pointer to how to configure it is much better than a silent default that could let a mis-set-up test pass by accident. ❤️

MatiasFernandez (approved):
Looks good. I left a minor question about variable naming

jgfoster (approved):

PR #177, opened 2026-07-20
Add STON syntax highlighting for .ston files

srbaker:
A TextMate grammar (`source.ston`) and language configuration for STON (Smalltalk Object Notation), registered in the extension manifest, so `.ston` files get highlighting, bracket matching, and folding.

The grammar follows **GemStone's STON dialect**, as defined by its `STONWriter` (in the RowanV3 STON package) rather than the original STON paper — the two differ:

- **Strings and quoted symbols use backslash escapes** (`\'`, `\"`, `\\`, `\b \f \n \r \t`, `\uXXXX`), not Smalltalk-style doubled quotes. So an embedded quote (`'Rowan\'s project'`) is highlighted as one string instead of terminating early.
- **Bare symbols may contain `-` `_` `.` `/`** alongside letters and digits (`#Rowan-Core`, `#rowan/src`), matching `STONWriter>>isSimpleSymbolChar:`.

Tested with vscode-textmate: registration integrity (the grammar/config files resolve and the scopeName matches) plus tokenization of every construct — class tags, symbol keys and symbols (including hyphens and slashes), strings and their backslash escapes, true/false/nil, character literals, object references, plain and radix numbers, and key/value punctuation.

### DIFF
--- a/client/src/__tests__/stonGrammar.test.ts
+++ b/client/src/__tests__/stonGrammar.test.ts
@@ -1,0 +1,133 @@
+import { describe, it, expect, beforeAll } from 'vitest';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as oniguruma from 'vscode-oniguruma';
+import { Registry, parseRawGrammar, INITIAL, type IGrammar } from 'vscode-textmate';
+
+const repoRoot = path.resolve(__dirname, '..', '..', '..');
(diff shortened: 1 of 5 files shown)


MatiasFernandez (approved):
I don't have context on the STON format but you said this is based on the `STONWriter` so I delegate to those tests you performed to verify this is working.

PR #185, opened 2026-07-21
Streamline onboarding: OS setup on Start Stone, and a setup-first walkthrough

jgfoster:
Remove the "Configure OS" sidebar view on macOS and Linux and fold its shared-memory (and Linux RemoveIPC) checks into a Start Stone preflight: if a prerequisite is missing, Jasper explains it will open a terminal for the sudo password and offers Continue/Cancel; cancelling, or a setup that leaves shared memory under 1 GB, cancels the start. The view is kept on Windows/WSL, where it hosts several ongoing settings. Shared-memory and RemoveIPC checks are extracted into reusable exports so the view and the preflight share one implementation.

Rework the "Get Started with GemStone" walkthrough so it begins with getting a stone (a new Quick Setup step that downloads a version, creates a database, starts the stone, and adds a login) before the connect step, instead of dropping a new user at "connect" with nothing to connect to. Fix the "Add a Login" button in both the walkthrough and the Logins view welcome: it pointed at gemstone.login (connects a selected login and throws when invoked from a static button) and now uses gemstone.addLogin.


package.json
```diff
@@ -490,12 +490,23 @@
       {
         "id": "gemstoneGettingStarted",
         "title": "Get Started with GemStone",
-        "description": "Connect to a stone and explore Smalltalk expressions in Jasper.",
+        "description": "Download GemStone, create a database, connect, and explore Smalltalk expressions in Jasper.",
```
MatiasFernandez, line 508: **Heads-up on how this walkthrough actually gets shown — not a blocker, but worth knowing.**

The "Get Started" walkthrough auto-opens via a listener on the `gemstoneLogins` view's `onDidChangeVisibility` (in `extension.ts`), guarded by a once-per-machine `globalState` flag. So today it depends on two things: (1) the user opening/revealing the GemStone sidebar, and (2) the Logins & Sessions section being **expanded by default**.

That second dependency is implicit and fragile. VS Code only fires `visible = true` for a section that's actually expanded — a section declared `"visibility": "collapsed"` never fires it (confirmed: our collapsed MCP Server section doesn't). `gemstoneLogins` works purely because it has no `visibility` set. If a future dev decides Logins should start collapsed, the walkthrough auto-open **breaks silently**: no error, and the one-word `package.json` change is nowhere near the onboarding code in `extension.ts`, so nobody would connect the two.

A guard test doesn't really save us either — it would just assert "Logins isn't collapsed", so whoever intentionally collapses it will update/delete the failing test to get CI green and *still* never realize they broke onboarding. The test protects the config value, not the feature.

The common alternative is to open the walkthrough on `onStartupFinished`, globalState-guarded, with no dependency on any view's visibility. The tradeoff: it pops as soon as the extension activates (basically right after install), before the user interacts with GemStone — a bit more invasive than waiting for the user to show intent. But it's robust: no view-layout change can silently disable it.

This is fine to ship as-is, but I'd recommend moving to the `onStartupFinished` approach — the silent-break failure mode isn't worth the "wait for intent" nicety.

MatiasFernandez, line 508: **Related naming misalignment (pre-existing, but relevant while we're in the onboarding area).** Not on this exact line — the offending string is the `gemstone.resetGettingStarted` command title further down in this file (`package.json`, `"title": "Reset Getting Started (Show Walkthrough on Next Connect)"`).

That title says **"on Next Connect"**, but the walkthrough doesn't key off connecting at all — as the thread above notes, it auto-opens from the `gemstoneLogins` view's `onDidChangeVisibility`. The command's *own* info message even says "the next time you open the GemStone view", and connecting is just one *step* of the walkthrough (completion event `onCommand:gemstone.login`). So the title is the odd one out.

Suggest aligning it with the actual trigger, e.g. `"Reset Getting Started (Show Walkthrough on Next GemStone View)"`. It predates this PR, so fixing it here is optional — flagging since this PR is what makes the onboarding flow prominent.

MatiasFernandez (commented):
James, here's a bit of feedback, and I also shared more visual feedback directly with you.

package.json
jgfoster: Updated.

jgfoster: Update.

jgfoster:
Also updated based on the video in Slack.

MatiasFernandez (approved):
Looks good

PR #188, opened 2026-07-21
Sync ESLint ignores with .gitignore via @eslint/compat

MatiasFernandez:
## Summary
- Replace the hand-maintained ESLint `ignores` array with `eslint-config-flat-gitignore`'s `recursive: true` mode, which walks the repo and picks up every `.gitignore` automatically — so lint ignores can't drift from what's actually gitignored.
- Concrete motivation: a missed `.vscode-test/` entry previously choked the ESLint parser on a downloaded VS Code test binary. Recursive discovery avoids that class of miss structurally, including catching `docs/enhancedInspectorSupport/.gitignore`, which a hand-maintained list would be easy to overlook.
- Keeps `**/*.d.ts` and `resources/**` as explicit lint-only ignores, since they're tracked files with no gitignore equivalent.

## Test plan
- [x] `npm run lint` passes
- [x] `npm run format:check` passes
- [x] `npm run compile` passes
- [x] Verified `.gemstone/` and `.vscode-test/` are correctly ignored by lint
- [x] Verified `.gitignore` negation (e.g. `!.claude/hooks/`) is still honored, so those files remain linted

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -3,19 +3,17 @@ import tseslint from 'typescript-eslint';
 import globals from 'globals';
 import eslintComments from '@eslint-community/eslint-plugin-eslint-comments';
 import eslintConfigPrettier from 'eslint-config-prettier';
+import gitignore from 'eslint-config-flat-gitignore';
 
 export default tseslint.config(
+  // Keep lint ignores in sync with every `.gitignore` in the repo, instead of
+  // a hand-maintained duplicate list that drifts (e.g. missed `.vscode-test/`
+  // choking the parser on a downloaded test binary).
+  gitignore({ recursive: true }),
   {
-    ignores: [
-      '**/out/**',
-      '**/node_modules/**',
-      '**/*.d.ts',
-      '**/*.vsix',
-      'resources/**',
-      '.gemstone/**',
-      'client/tmp/**',
-      'acceptance/test-results/**',
-    ],
+    // Tracked files that are intentionally excluded from lint, not from git —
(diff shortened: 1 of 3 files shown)


srbaker (approved):
😍

eslint.config.mjs
```diff
@@ -3,19 +3,19 @@ import tseslint from 'typescript-eslint';
...
 import eslintConfigPrettier from 'eslint-config-prettier';
+import { includeIgnoreFile } from '@eslint/compat';
+import path from 'node:path';
 
 export default tseslint.config(
+  // Keep lint ignores in sync with what's actually gitignored, instead of a
+  // hand-maintained duplicate list that drifts (e.g. missed `.vscode-test/`
+  // choking the parser on a downloaded test binary).
+  includeIgnoreFile(path.resolve(import.meta.dirname, '.gitignore')),
+  includeIgnoreFile(path.resolve(import.meta.dirname, 'client/.gitignore')),
```
npapagna: It looks like sometimes each project acts like a separate repo. (client has its own .gitignore) and other times we have configs at the root level (like this one).

I know this predates this PR, but I'm wondering if we could do something like "use all nested .gitignore files in this repo" so we don't have to maintain this list manually.

npapagna (approved):
Thanks @MatiasFernandez!

Left a comment with a possible improvement, otherwise this looks fine!

eslint.config.mjs
MatiasFernandez: that makes sense, I actually found a way better and more elegant solution that solves this too

PR #189, opened 2026-07-21
Migrate dictionary-kind queries off deprecated bind() executor

npapagna:
## Summary
- Removes two genuinely dead exports (`getPoolDictionaryNames`, `getSuperclassDictName`) instead of migrating them — zero production callers anywhere in `client/`, `server/`, or `mcp-server/`.
- Migrates the rest of the dictionary-kind group (`getDictionaryNames`, `getGlobalsForDictionary`, `getDictionaryEntries`, `getDictionaryClassFileOutOrder`, `addDictionary`, `removeDictionary`, `moveDictionaryUp`, `moveDictionaryDown`) from the deprecated `bind()`/`executeFetchString` executor to a new `defaultQueryExecutorUsing()`, backed by `GciLibrary.executeAndFetchString`.
- `executeFetchString`'s raw `GciTsExecuteFetchBytes` call mis-decodes non-ASCII results and truncates large ones; `executeAndFetchString` encodes the result as UTF-8 in Smalltalk before paging it out, fixing both.
- First group of a larger migration effort (`bind()` is used across ~60 call sites in `browserQueries.ts` alone); grouped by "kind" so each PR's manual verification stays small.

## Manual test plan
1. Right-click in the Dictionaries column and choose **Add Dictionary…**. Enter `Café`. Confirm the confirmation message shows the name correctly and it appears in the Dictionaries column.
2. Open a Workspace and Execute It on:
   ```
   (System myUserProfile symbolList objectNamed: #'Café') at: #'Café1' put: 'Café2'.
   ```
3. Click the `Café` dictionary in the Dictionaries column. Confirm the Globals panel shows `Café1` and its value `Café2`, both displayed correctly.
4. Right-click `Café` and choose **Move Up**. Confirm it moves one position earlier in the list.
5. Right-click `Café` and choose **Move Down**. Confirm it returns to its original position.
6. Right-click `Café` and choose **Remove Dictionary…**. Confirm the confirmation message shows the name correctly and it disappears from the list.

Footnote: if step 2 throws a compiler error, that's a known pre-existing GemStone issue unrelated to this change.


client/src/browserQueries.ts
```diff
@@ -316,19 +314,52 @@ export function sessionNeedsCommit(session: ActiveSession): boolean | undefined
...
+ *
+ * executeAndFetchString explicitly encodes the evaluated result as UTF-8 in
+ * Smalltalk before paging it out, so results decode correctly regardless of
+ * their original encoding and are not capped at a single fixed-size buffer.
+ */
+function defaultQueryExecutorUsing(session: ActiveSession): QueryExecutor {
+  return (label, code) => {
+    logQuery(session.id, label, code);
+
+    const { result: inProgress } = session.gci.GciTsCallInProgress(session.handle);
```
MatiasFernandez, line 340: 💭 This busy-check block (`GciTsCallInProgress` → throw `BrowserQueryError`) is now identical across `executeFetchString`, `executeFetchStringNb`, `executeFetchStringWithLimit`, and here — four copies, and it grew by one with this addition. Does it make sense to extract a small `ensureNotBusy(session)` (or a shared `guardedQuery` wrapper) so the precondition lives in one place? Not blocking.

client/src/browserQueries.ts
```diff
@@ -316,19 +314,52 @@ export function sessionNeedsCommit(session: ActiveSession): boolean | undefined
...
+      const msg = 'Session is busy with another operation. Please wait or use a different session.';
+      logError(session.id, msg);
+      throw new BrowserQueryError(msg);
+    }
+
+    try {
+      const data = session.gci.executeAndFetchString(session.handle, code);
+      logResult(session.id, data);
+      return data;
+    } catch (e) {
```
MatiasFernandez, line 351: 💭 The happy path through this executor is now covered by the migrated query tests (they read from the `executeAndFetchString` mock), but the two failure branches here — the session-busy throw above and the `GciLibraryError → BrowserQueryError` translation in this catch — don't seem exercised by any test. The one busy-path test we have (`sessionNeedsCommit`) still goes through the old `executeFetchString`. Worth adding a case against a migrated query that mocks `GciTsCallInProgress` to return `1`, and one where `executeAndFetchString` throws, so the wrapping doesn't silently regress as the rest of the migration lands?

srbaker (approved):

MatiasFernandez (approved):
Looks good, I left a few comments. Nice call on deleting the two dead exports outright rather than migrating them, and the `@deprecated` note on bind() pointing at the replacement is a good touch for the follow-up PRs.

client/src/browserQueries.ts
npapagna: That's actually on my TODO list: moving that check to GciLIbrary and giving it good coverage.
Since this PR and the following ones will require some manual checking, I'm trying to keep them lean wrt refactors and/or improvements.

PR #191, opened 2026-07-21
Server-side refactoring engine (grail #62): rename refactorings R1–R5 + editor Refactor… menu

ericwinger:
Adds a **server-side Refactoring Browser engine** to GemStone and the Jasper
client UI that drives it — the first five rename refactorings plus the shared
infrastructure the rest of the catalog (grail #62) will build on. Everything runs
across **GemStone 3.6.2 → 3.7.5+** and **never commits**: a refactoring stages a
non-committing, previewable change set; the client shows a before/after; apply
recompiles in the stone and leaves the transaction for the user to commit.

## Refactorings
- **R1 — Rename instance variable** (the first slice; drove all shared infra)
- **R2 — Rename method / selector** (unary/binary/keyword; keyword-part reorder;
  scope: class / hierarchy / dictionary / whole-system)
- **R3 — Rename class** (new class version, methods copied forward, descendants
  re-parented, external refs rewritten) + a **class-definition history viewer**
  with restore
- **R4 — Rename class variable** (value-preserving reflective rename; no version bump)
- **R5 — Rename temporary / argument** (method-local; AST scope + shadow aware)

## Engine + shared infrastructure
- `GsRefactoring` dictionary: `GsRefactoringEnvironment` (all-dictionary reflective
  queries), `GsRefactoringChangeSet`/`GsRefactoringChange` (non-committing,
  individually-selectable, before/after), and one `Gs…Refactoring` class per item.
- Minimal-diff AST rewrites over a **vendored, Rowan-free RowanV3 AST-Core**
  (RB parser/nodes/rewriter), plus `compat/` backports so it loads on 3.6.2.
- Paginated preview (SessionTemps token, byte-bounded pages) + server-side apply
  in one round trip; pure-ASCII JSON so the non-blocking GCI fetch never sees a
  Unicode-promoted result.
- A shared loader (`GsRefactoringLoader`) + runbook + post-load completeness gate;
  the engine ships as an **optional install** (bundled with the Enhanced Inspector),
  offered on connect and installable via one command — nothing is added to a stone
  until the user opts in.

## Client
- Rename **pencils** on the Explorer's ivar / class-var / method / class rows, each
  with the appropriate editor (keyword-part rows for methods, scope picks, name
  inputs), a paginated preview panel, and server-side apply.
- **Editor "Refactor…" menu**: in a method source editor, VS Code's native
  Refactor… menu now hosts Rename Temporary/Argument, Instance Variable, Class
  Variable (on the identifier at the cursor) and Rename Method (target follows the
  cursor — a sent selector renames that selector via the LSP, the header renames the
  edited method). "Offer all, decline politely": a rename that doesn't fit the word
  declines with a toast pointing at the right one. Shared plumbing in
  `renameAtCursorShared.ts`; the pencils and the editor commands run the same flows.

## Testing
- Engine **SUnit: 133 tests, green on 3.6.2 and 3.7.5** (both-boundary rule).
- Client vitest across the query/model/panel/command layers + guarded GCI
  integration tests (skip-with-reason when the engine isn't loaded). Full suite
  green: client ~3680, server 322, mcp 95. tsc / eslint / prettier clean.

## Notes for review
- **`resources/refactoring/*.gs` are generated** from `gs-src/refactoring/` by
  `build/build-refactoring.sh` — review the `.class.st` sources, not the payloads.
  The large line count is these payloads + the vendored AST.
- **No auto-commit anywhere** — asserted in SUnit and enforced in the client apply.
- Guarded GCI tests only *exercise* against an engine-loaded stone; CI stones are
  engine-less, so they skip green (wiring a dedicated engine-loaded CI leg is a
  Stage-6 / hardening item).
- Known won't-fix: a 3.6.2-only GCI compiler-state quirk after many compiles in one
  engine-loaded session (documented; doesn't affect CI or 3.7.5).

## Not in this PR
- Later catalog items (Method / Variable-structure / Optional families, Rename
  category/package) and Stage-6 hardening (full version-matrix CI, apply-time
  warning surfacing, acceptance specs).


npapagna (approved):
Thanks @ericwinger  ! I'm glad the list was useful. 🙂

From my perspective, I'm happy merging it as is. I don't see any blockers.

The only thing I found myself wondering about is how we want to set expectations once this reaches users. Since we already know there are a few rough edges that we'll keep iterating on, would it make sense to mark the refactorings as `(Experimental)` in the relevant context menus, window titles, or similar UI? That way users know they're trying an evolving feature while we continue polishing it.

npapagna (commented):
🤖 The following findings were identified by Claude via automated code review (`/code-review`), not manually by me. Flagging here for visibility.

1. **`client/src/renameInstVarPanelHtml.ts:38`**: The instance-variable rename preview lets the user uncheck the `classDefinitionEdit` change while leaving method-recompile changes checked, unlike the class/class-variable rename panels, which disable their structural checkboxes. If a user renames an ivar used by several methods, unchecks only the "class definition" card, and applies, the class is never recompiled with the new ivar name while the method bodies are recompiled referencing it, causing every selected method recompile to fail (or bind to a stray Undeclared global).

2. **`client/src/renameAtCursorShared.ts:44`**: `ensureRbSupport` returns `true` unconditionally after invoking the install command, without re-checking whether the refactoring engine actually became available. If the install fails (wrong password, missing payload) or the user declines the post-install "Refresh this session?" prompt, the caller still proceeds to send Smalltalk referencing refactoring classes that don't exist in that session yet, surfacing a cryptic compile/DNU error instead of a clean refusal. The three near-identical gates inlined in `gemstoneExplorer.ts` correctly re-check `session.rbSupportAvailable` after the install command returns.

3. **`client/src/renameMethodPanelHtml.ts:58`**: An implementor's `methodRename` change can be individually deselected while sender `methodRecompile` changes (and other implementors) for the same selector stay selected. Example: renaming `foo` to `bar` across a hierarchy where a subclass overrides `foo`. If the user unchecks only the subclass's rename row and applies, all sender call sites get rewritten to `bar`, but the subclass still only implements `foo`. Instances of that subclass then dispatch to an inherited `bar` (or DNU), silently changing behavior. The class-rename engine avoids this by marking `classRename`/`classReparent` changes as always-applied/disabled for exactly this reason.

4. **`client/src/gemstoneExplorer.ts:1321`**: The "refactoring engine not installed, offer to install, re-check" gate is duplicated verbatim in `renameInstVarNamed` (line 1321) and `renameMethodNamed` (line 1495) instead of calling the file's own private `ensureRbSupport` helper (line 1617), which three other call sites already use. Not a runtime bug on its own, but a maintenance hazard: a future fix to the gating logic (e.g. finding #2 above) has to be applied in up to four separate places, and it's easy to miss one, which is exactly the inconsistency already present between this file's copies and `renameAtCursorShared.ts`'s broken one.

MatiasFernandez (approved):
I'm good with merging as-is and continue working on stuff we discussed as follow-ups

PR #195, opened 2026-07-21
Guard against stale node_modules (dependency-freshness check)

jgfoster:
## Why

Two recent "Cannot find package '…'" failures (`vscode-oniguruma`, then `eslint-config-flat-gitignore`) had the same root cause: a merge/pull/branch-switch changed `package-lock.json` to add a dependency, but `npm install` wasn't re-run — so the package was never installed. The error surfaced cryptically from ESLint or vitest instead of pointing at the fix (`npm install`).

## What

A small dependency-freshness guard, wired into existing lifecycle hooks:

- **`postinstall`** (`scripts/write-install-stamp.mjs`) records a SHA-256 of `package-lock.json` in `node_modules/.install-stamp`.
- **`pretest`** (`scripts/check-deps.mjs`) re-hashes the lockfile and compares. A mismatch or missing stamp ⇒ `node_modules` is stale ⇒ **exit 1** with *"Dependencies are out of sync with package-lock.json … run `npm install`."* This runs in the pre-push hook (which runs `npm test`), so it catches drift before a push.
- It also **warns (never fails)** when Node doesn't match the pinned `.nvmrc` major, nudging `nvm use`. Deliberately not a hard gate: `engines` is broader than the pinned dev version and CI also tests the supported floor, so a strict "must be .nvmrc" gate would break those runs.

CI-safe: `npm ci` runs `postinstall`, so the stamp is always fresh there.

The lockfile's new `"hasInstallScript": true` is just npm recording that the package now has a `postinstall` script.

## Tests

`client/src/__tests__/depsGuard.test.ts` drives the scripts as a subprocess against a throwaway fixture repo (`DEPS_GUARD_ROOT` points the guard at it): in-sync passes, a drifted or absent stamp fails, and a Node-major mismatch warns without failing.


jgfoster:
I have no idea if this is worthwhile or done right, but when I told Claude that I'd like to avoid the recent errors, this is what it suggested.

MatiasFernandez:
I'll read this in detail tomorrow. But at first sight it looks weird. It's standard practice/requirement for a node project that you run npm install after pulling new changes. Or run it when you get failures. It's a step that one should normally run when switching branches and such.

PR #210, opened 2026-07-22
Migrate executeFetchString onto executeAndFetchString

npapagna:
## Summary
- Deduplicate `defaultQueryExecutorUsing` across `pythonQueries.ts`/`sunitQueries.ts` by exporting the one in `browserQueries.ts` and importing it instead.
- Migrate `executeFetchString` itself onto `GciLibrary.executeAndFetchString`, the last query-layer function still calling `GciTsExecuteFetchBytes` directly, so it decodes results correctly regardless of encoding and isn't capped at a fixed 256 KB buffer, matching the Python/SUnit executor migrations before it.
- Update the two test files whose mocks still stubbed `GciTsExecuteFetchBytes`, and correct comments left describing mechanics that moved or no longer apply.

## Test plan
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run compile`
- [x] `npm test` (server, client, mcp-server — 3738 tests, live stone)

### DIFF
--- a/client/src/browserQueries.ts
+++ b/client/src/browserQueries.ts
@@ -317,7 +317,7 @@ export function sessionNeedsCommit(session: ActiveSession): boolean | undefined 
  * Smalltalk before paging it out, so results decode correctly regardless of
  * their original encoding and are not capped at a single fixed-size buffer.
  */
-function defaultQueryExecutorUsing(session: ActiveSession): QueryExecutor {
+export function defaultQueryExecutorUsing(session: ActiveSession): QueryExecutor {
   return (label, code) => {
     logQuery(session.id, label, code);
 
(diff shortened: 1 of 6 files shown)


client/src/browserQueries.ts
```diff
@@ -309,35 +284,10 @@ export function sessionNeedsCommit(session: ActiveSession): boolean | undefined
...
-      return data;
-    } catch (e) {
-      const msg = e instanceof Error ? e.message : String(e);
-      logError(session.id, msg);
-      throw new BrowserQueryError(msg);
-    }
-  };
+// Binds a session to the QueryExecutor shape that shared queries expect,
+// backed by executeFetchString.
+export function defaultQueryExecutorUsing(session: ActiveSession): QueryExecutor {
```
MatiasFernandez: 💭 Now that `defaultQueryExecutorUsing` is exported and imported by both `pythonQueries.ts` and `sunitQueries.ts` (on top of its many call sites here), would you consider keeping it as a `/** */` JSDoc rather than the line comment? It carried a full JSDoc before, and the codebase leans on JSDoc for its reused/public surface (e.g. `gciLibrary`). A short one would help a casual reader — and an agent — orient on it as the shared entry point; the mechanism detail can stay on `executeFetchString` where it now lives.

npapagna: Fixed!

MatiasFernandez (approved):

PR #217, opened 2026-07-23
Start a stopped database on login failure, then connect

srbaker:
## Why

Logging in to a database that is not running just fails with an unhelpful error. You then have to leave the login, find the database, start it by hand, and try again.

## What changed

- **Offer to start a stopped database on login failure, then connect.** When a login fails because its database is stopped, offer to start it and connect automatically. The prompt offers Now / Never, and Never is remembered so you are not asked again.
- **Report signal kills honestly.** A GemStone process killed by a signal now says so (e.g. "killed by signal SIGKILL") instead of the misleading "exit code null".
- **Keep the Admin output panel shut** while auto-starting during a login -- no unsolicited panel popping open.
- **Recover a database when only one of its processes is wedged.** Anything that is *down* is offered first, so a wedged stone no longer leaves a stopped NetLDI stopped. `not-responding` is reported only when both are up, where starting cannot help.
- **Report a recovery failure instead of letting it escape the login command.** `maybeStartDatabaseAndRetry` documents itself as reporting the outcome on every path, but only guarded the failures it expected -- remembering the preference and reading the database list were bare, so an unwritable `settings.json` rejected out of the flow. With this branch having replaced the command`s `catch` with a bare `finally`, that surfaced as "command failed", naming neither the login nor the reason. The contract is now true at the source, and the command keeps a net as well.
- **Scope the auto-start preference so no workspace can override it.** The prompt writes `ConfigurationTarget.Global`, but the setting declared no scope and so took VS Code`s default of `window` -- which a workspace outranks. Anyone with such an override would click Never, have it written where it stays outranked, and be asked again with no sign of why.

## Verification

- Full CI matrix green on this head: GemStone 3.6.2 / 3.6.2 (Node 22.15.1) / 3.6.8 / 3.7.2 / 3.7.4.3 / 3.7.5, plus lint, format, CodeQL and `vsce` packaging.
- Locally: `compile`, `lint`, `format:check` clean; 4509 client unit tests pass. The remaining local failures are the integration suites that need a live stone, which CI covers above.
- Each commit ships its own tests (autoStartDatabase, autoStartDecision, autoStartPrompt, databaseForLogin, loginTreeProvider, processManager).


client/src/versionsMatch.ts
```diff
@@ -0,0 +1,21 @@
...
+ *  as matching when the shorter one is a dotted-component prefix of the longer
+ *  (e.g. "3.7.4" matches "3.7.4.3"). This keeps genuinely different installs —
+ *  "3.6.2" vs "3.7.5" — distinct, which is what lets the Databases panel tie a
+ *  running stone to the version that actually started it.
+ *
+ *  Lives in its own module (rather than processManager, which owns the only
+ *  other callers) so pure consumers — see databaseForLogin.ts — can use it
+ *  without pulling in vscode, fs, and the WSL bridge.
+ */
+export function versionsMatch(a: string, b: string): boolean {
```
MatiasFernandez: 💭 Why was this moved to a separate file if it was already exported?

client/src/autoStartPrompt.ts
```diff
@@ -0,0 +1,96 @@
...
+    mode: 'never',
+    label: 'Never',
+    detail: 'Never start a database; report the login failure as-is.',
+  },
+];
+
+/**
+ * Command Palette entry point for the preference, so someone who chose "Always"
+ * at the prompt can change their mind without hunting through Settings — and so
+ * "Never", which the prompt has no button for, is reachable at all.
```
MatiasFernandez: This comment says "'Never', which the prompt has no button for" — but `confirmStartDatabase` above passes `NEVER` as one of the four buttons to `showInformationMessage` (Yes/No/Always/Never). Should this comment be updated, or is there a subtlety (e.g. a platform where the 4th button gets dropped/collapsed) that I'm missing?

client/src/autoStartDecision.ts
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Decide what the recovery flow should do about a database's process state.
+ *
+ * A login needs both the stone and the NetLDI, so either being down is
+ * actionable. An unresponsive process is called out separately because
+ * starting it would just fail — that case wants the stale-lock tooling, not a
+ * second `startstone`.
+ */
+export function classifyStartNeed(state: DatabaseProcessState): StartNeed {
```
MatiasFernandez, line 72: 💭 If the stone is down but the netldi is merely wedged (running, not responding) — or vice versa — this reports `not-responding` and the whole recovery flow bails out, even though the down side could still be started independently (the code later starts stone/netldi independently of each other). Should a wedged process on one side still allow offering to start the other, or is bailing out entirely intentional here (e.g. to avoid a partial, confusing recovery)?

client/src/versionsMatch.ts
srbaker: Agreed — it did not need to move. `versionsMatch` is back in `processManager.ts` (relocated to the top of the file so `autoStartDecision.ts` and `databaseForLogin.ts` can import it without a forward reference), and `client/src/versionsMatch.ts` is gone. The `Exported for testing` note on its doc comment went away too, since it now has real callers outside the module.

client/src/autoStartPrompt.ts
srbaker: No subtlety — the comment was just wrong, left over from an earlier version that had only three buttons. The prompt does pass all four (Yes/No/Always/Never), and the stale sentence is gone; the comment on `configureAutoStartDatabase` now just explains why the Command Palette entry exists, which is so anyone who picked Always or Never can change it without hunting through Settings.

client/src/autoStartDecision.ts
srbaker: Good catch, and no — bailing out was not intentional. Fixed in 4a59e73 ("Recover a database when only one of its processes is wedged"): `classifyStartNeed` now tests for anything *down* first and returns `can-start` for just that side, so a wedged stone no longer leaves a stopped NetLDI stopped. `not-responding` is reported only when both processes are up, where starting genuinely cannot help and the stale-lock tooling is what is wanted. The reasoning is written into the doc comment, and `autoStartDecision.test.ts` covers each of the four combinations.

client/src/processManager.ts
```diff
@@ -475,12 +490,28 @@ export class ProcessManager {
...
+        // A signal-killed process reports a null exit code, so the plain
+        // "exit code ${code}" wording degrades to the meaningless "exit code
+        // null" — and it is usually killed before writing a word of output, so
+        // the message is all the user gets. macOS does this under memory
+        // pressure (jetsam), which is easy to mistake for a broken database.
+        if (signal) {
+          reject(
+            new Error(
+              `${label} was killed by the operating system (${signal}) before it could report ` +
+                `anything. This is usually memory pressure rather than a problem with the ` +
```
MatiasFernandez: 💭 This message commits to a cause for *any* signal, but `signal` can be more than a jetsam SIGKILL — a SIGSEGV from a `startnetldi` that crashes on a broken install, or a plain SIGTERM from someone running `kill`. In those cases "free some memory and try again" points the user in the wrong direction, and the real cause never gets named.

Should the memory-pressure advice be scoped to SIGKILL, with other signals reported plainly as "killed by signal X"? Worth noting the test asserts `/killed by the operating system|memory|log/i`, so it would pass either way.

client/src/databaseForLogin.ts
```diff
@@ -0,0 +1,35 @@
...
+import { GemStoneDatabase } from './sysadminTypes';
+import { versionsMatch } from './processManager';
+
+/**
+ * True when a login targets a stone on this machine. Only a local stone can be
+ * started by Jasper — starting a remote one would mean running `startstone`
+ * over there, which the extension has no way to do.
+ */
+export function isLocalLogin(login: Pick<GemStoneLogin, 'gem_host'>): boolean {
+  return login.gem_host === 'localhost' || login.gem_host === '127.0.0.1';
```
MatiasFernandez: 💭 This exact predicate already exists inline in `extension.ts` for Serve Seaside (`if (host !== 'localhost' && host !== '127.0.0.1')`). Now that there's an exported `isLocalLogin`, is it worth pointing that one at it too, so "what counts as local" lives in one place?

Related: should `::1` count as local? A login on the IPv6 loopback is the same machine, but it would fall out of `findDatabaseForLogin` as remote and quietly lose the auto-start offer — with no error explaining why, since we just re-show the original login failure. Cheaper to decide that once here than in two places later.

client/src/loginTreeProvider.ts
```diff
@@ -68,10 +73,25 @@ export class LoginTreeProvider implements vscode.TreeDataProvider<LoginTreeNode>
     sessionManager?.onDidChangeSelection(() => this.refresh());
   }
 
+  // Logins with a connect attempt in flight, by position in getLogins(). Held
+  // here rather than derived because nothing else knows about an attempt that
+  // has not produced a session yet.
+  private connecting = new Set<number>();
```
MatiasFernandez: 💭 This holds state keyed by the login's position, which is a different thing from the index use already in the file — those are derived fresh on every render, this one outlives a render.

Is it possible for a login to be deleted or reordered while a connect is in flight? Starting the stone and the NetLDI takes seconds, and the tree stays interactive throughout. If it is, the `setConnecting(item.index, false)` in the command's `finally` clears whatever now sits at that index, and the row that was actually connecting keeps spinning with no way to stop it.

Would keying by the login's identity (`loginKey`/`sameLogin` in `loginTypes`) sidestep that, or is the window too small to be worth it?

client/src/extension.ts
```diff
@@ -1502,7 +1513,20 @@ export function activate(context: vscode.ExtensionContext) {
           sysadminStorage.getGemstonePath(login.version) ?? path.dirname(path.dirname(gciPath));
         process.env.GEMSTONE = gsInstallPath;
 
-        let session;
+        // Spin the login's own row and a status bar item for the whole attempt.
+        // A connect is no longer just a login: it may start the stone and its
+        // NetLDI first, which takes seconds, so the progress notification alone
+        // is not enough to explain why nothing is happening.
+        const resolvedGciPath = gciPath;
```
MatiasFernandez, line 1560: nit: `resolvedGciPath` sits under the comment about the spinner, so I first read it as part of that change. If I follow it right, it exists because `gciPath` is a `let` and TS widens it back to `string | undefined` inside the new async closure — worth a short comment saying so? Otherwise the next reader is fairly likely to "clean it up" back to `gciPath` and get a type error they can't place.

MatiasFernandez (approved):
I left some comments worth checking

PR #222, opened 2026-07-23
Add Extract Method and Inline Method refactorings

ericwinger:
## Summary

Adds the next two items in the RB refactoring catalog — **Extract Method (M1)** and **Inline Method (M2)** — the first refactorings that create and remove methods (all prior ones only renamed). Both run entirely server-side through the shared refactoring engine, apply as one atomic unit, and **never commit** the GemStone transaction.

Built on top of the refactoring engine that landed in `main` via #191.

## Extract Method (M1)

- Select statements (or an expression) in a method editor → **Extract Method** (own editor context-menu item, not the Refactor… submenu).
- Prompts for the new selector; args are auto-named from the extracted variables (arity validated). Naming it the same as the source method, or a hard-invalid shape (e.g. a bare `^` return), is declined with a reason.
- Custom paginated preview panel: the new method + the rewritten original are locked-on core changes; any structurally-similar fragments in the class/hierarchy are offered as **opt-in** (unchecked) rows for safe void extractions.
- New `#methodAdd` engine change kind. Extracted method is formatted; the new method is revealed after apply.

## Inline Method (M2)

- Place the caret on a `self`/`super` message send → **Inline Method** (own editor context-menu item).
- Scope: `self`/`super` receivers, single-expression targets. When the inlined call was the target's **last sender**, the now-unused target is offered for removal as an **opt-in** (unchecked) row.
- New `#methodRemove` engine change kind; call-site parenthesization handled for non-atomic arguments.

## Also included

- General **Remove Dictionary** Explorer context menu (removes a dictionary from the current session's symbol list; non-committing).
- `rowanLoad`: isolate git subprocesses from the ambient `GIT_*` env, so git operations run against the intended repo even inside git hooks.

## Testing

- **Engine SUnit** verified on **both** 3.6.2 and 3.7.5 (non-committing, via topaz): extract 31/31, inline 39/39, full engine suite green on both boundaries.
- **Client**: full unit coverage of every function — preview models, query builders, panel HTML, jsdom DOM behavior, command flows, and webview host wiring — plus guarded GCI integration tests (skip on an engine-less stone).
- Full suite green: client 3826 pass / 26 skip, server 322, mcp 95; compile + lint + format clean.


client/src/refactoring/inlineMethodPanel.ts
```diff
@@ -0,0 +1,129 @@
+/**
+ * The paginated inline-method (M2) preview panel. Shows the required core change
+ * (the rewritten caller) plus, when the inlined call was the target's last sender,
+ * a deselectable removal of the now-unused target. Fetches further pages on demand
+ * and applies server-side reporting only the DESELECTED ids (so the core recompile,
+ * whose checkbox is disabled, always applies; the removal applies unless unticked).
```
MatiasFernandez: 💭 This says the removal "applies unless unticked", but everywhere else it reads as opt-in: `renderCard` in inlineMethodPanelHtml renders the removal checkbox *unchecked*, the summary tells the user to "tick the last row to also remove…", the Apply button starts at `coreCount` (1), and the shared view JS reports unchecked rows as deselected — so with the row unchecked by default the `methodRemove` id lands in `deselectedIds` and the target is *kept*. The keep-by-default behavior is the safe one; it's this line that's out of step. Worth fixing, because a future reader who trusts it and "corrects" the HTML to render the row checked would flip a method *deletion* to on-by-default. Reword to match the opt-in reality?

client/src/refactoring/extractMethodCommand.ts
```diff
@@ -0,0 +1,237 @@
...
+  refuse,
+  reloadMethodEditor,
+  saveIfDirty,
+} from './renameAtCursorShared';
+
+/** 1-based count of code points before `position` in `document`. The engine indexes
+ *  the stored source by CHARACTER (a GemStone Character is a full code point) while
+ *  VS Code offsets count UTF-16 code units — they differ once a non-BMP character
+ *  appears before the position, so count code points (`Array.from`). */
+function codePointsBefore(document: vscode.TextDocument, position: vscode.Position): number {
```
MatiasFernandez: 💭 `codePointsBefore` (and its doc comment) is duplicated verbatim in inlineMethodCommand.ts. Both files already import from `renameAtCursorShared` — would it be worth hoisting it there, so the UTF-16-vs-code-point subtlety lives in one place?

client/src/refactoring/extractMethodCommand.ts
```diff
@@ -0,0 +1,237 @@
...
+  if (entered === undefined) {
+    focusEditor();
+    return;
+  }
+  const newSelector = entered.trim();
+
+  // For a safe void extraction, find structurally-similar sites in the hierarchy and
+  // present them in the preview as opt-in (unchecked) rows — so there is no separate
+  // up-front dialog; the choice lives in the panel, off by default.
+  const replaceSimilar = analysis.safeVoidShape;
```
MatiasFernandez, line 126: 💭 `replaceSimilar` is turned on for *every* safe-void extraction, and on the server `stageSimilarSitesInto:` then walks the whole hierarchy (`targetClass allSuperclasses` + `environment descendantsOf:`) and parses every selector of every one of those classes, synchronously, inside `startExtractMethodPreview`. For a method on a widely-subclassed class that's a lot of parsing on the preview path even when the user never means to touch the duplicate rows (they come up unchecked). Is the scan bounded somewhere I'm missing, or could it be deferred until the user actually asks to replace similar sites?

client/src/refactoring/extractMethodPanel.ts
```diff
@@ -0,0 +1,127 @@
...
+  loadPage: (offset: number) => Promise<PreviewPage>;
+  /** Apply server-side, skipping `deselectedIds` (duplicates only); no commit. */
+  apply: (deselectedIds: string[]) => Promise<ApplyResult>;
+  /** Drop the preview session (called exactly once when the panel closes). */
+  cleanup: () => void;
+}
+
+/** Show the paginated preview; resolve with the apply result, or undefined if the
+ *  user cancelled/closed it. */
+export function showExtractMethodPanel(
```
MatiasFernandez, line 34: 💭 `showExtractMethodPanel` and `showInlineMethodPanel` are nearly identical (the fetchOne / loadMore / loadAll / apply / cancel message loop and the `settled`/`finish` guard), and so are the two `*PanelHtml` files (the whole `<style>` block, `escapeHtml`, `renderDiff`) and `jsonEscape:` / `hex2:` / `jsonQuote:` / `pageJsonFrom:` / `applyDeselected:` across the two engine classes. I realize this deliberately mirrors the rename-panel convention, so it may be an accepted trade-off — but since M1 and M2 landed together, is there a natural shared base for at least the panel message-loop and the engine JSON helpers, or do you find the parallel copies clearer kept apart?

client/src/gemstoneExplorer.ts
```diff
@@ -2324,6 +2326,34 @@ class ExplorerController {
     }
   }
 
+  async removeDictionary(node: DictItem): Promise<void> {
```
MatiasFernandez, line 2329: 💭 `removeDictionary` mutates the session's symbol list, has its own error branch, and calls `reset()` afterwards, yet I don't see a focused test for it — the other explorer behaviors each get one (explorerCategories, explorerDoubleClick, …). Worth an explorerRemoveDictionary test covering the confirm-declined, error, and success-then-reset paths?

client/src/refactoring/inlineMethodPanel.ts
ericwinger: Stale comment fixed

client/src/refactoring/extractMethodCommand.ts
ericwinger: pushed to common function renameAtCursorShared

client/src/refactoring/extractMethodCommand.ts
ericwinger: I'm going to make a work item to opt out of the scan before it's done.  Default on to preserve existing behavior.

client/src/refactoring/extractMethodPanel.ts
ericwinger: I'll write this up, too, or put it in the memory to defer but finish. The rename convention may be worth the extra cruft.

client/src/gemstoneExplorer.ts
ericwinger: Test added and pushed

MatiasFernandez (approved):
LGTM

PR #223, opened 2026-07-24
Run integration tests in CI with and without the server plugin installed

MatiasFernandez:
## Summary

Jasper's client-installable Server Plugin installs two features — Refactoring and Enhanced Inspector — but CI has only ever provisioned a bare stone and never filed either one in. Every feature test gates on a live presence probe and skips, so the suite stays green without ever exercising that behavior on any matrix version.

This PR wires up a two-pass health-check job:
1. `npm test` — absent world; plugin features' tests skip
2. `node client/bin/install-server-plugin.mjs` — installs the plugin and verifies the version→feature contract, exiting non-zero if it's violated.
3. `npm test` — present world; the existing live-probe gates now pass, so present-world test bodies actually run and assert.

## Scope

This is an initial, minimal PR to get the mechanism working end-to-end. It intentionally:
- **Duplicates code**: the install script reimplements ~15 lines of SystemUser-elevation login logic that already exists in `refactoringInstallCommand.ts` / `enhancedInspectorCommand.ts`, rather than extracting a shared helper.
- **Does not touch existing tests**: the 5 refactoring integration test files keep their inline `rbEnginePresent()` + `ctx.skip` gating as-is. They'll pick up the present-world pass for free (the live probe just returns true), but none of them are migrated to take fuller advantage of the new two-pass mechanism yet (e.g. asserting absent-world degradation behavior).

A follow-up PR will consolidate the duplicated install/login helpers, add world-aware gating helpers (`requireFeature`/`requireFeatureAbsent`), and migrate/adapt the existing tests to use them.

## Test plan
- [x] Confirm the matrix (including the 3.7.5 job) fits the CI timeout budget with the doubled test run; bump `timeout-minutes` if not
- [x] Triage any first-time failures surfaced by present-world tests running for the first time in CI

### DIFF
--- a/.github/workflows/health-check.yml
+++ b/.github/workflows/health-check.yml
@@ -154,5 +154,9 @@ jobs:
         env:
           GEMSTONE_VERSION: ${{ matrix.gemstone-version }}
         run: npm run test:server:start -- "$GEMSTONE_VERSION"
-      - name: Run tests
+      - name: Run tests (without server plugin)
+        run: npm test
+      - name: Install server plugin
+        run: node client/bin/install-server-plugin.mjs
+      - name: Run tests (with server plugin)
         run: npm test
--- a/client/bin/install-server-plugin.mjs
+++ b/client/bin/install-server-plugin.mjs
@@ -1,0 +1,229 @@
+#!/usr/bin/env node
+//
+// Provisions a test stone with the Jasper Server Plugin (Refactoring +
+// Enhanced Inspector, version-gated) so CI can run the feature-test suite
+// a second time in the "present" world, not just the bare-stone "absent"
+// world. Run after `npm run test:server:start` (which writes
+// `client/.env.test`) and after `npm run compile` (this script loads the
+// compiled client output under `client/out/`, not the TS source).
+//
+// Logs in as `VITE_GEMSTONE_USER` (from .env.test), elevates to a transient
+// SystemUser session on the same connection — install requires write access
+// to kernel classes — and files in each feature. Enhanced Inspector is
+// skipped on stones below its minimum version, exactly like the client's own
+// auto-install path.
+//
+// Finally it re-verifies the version→feature contract
+// (`isRefactoringSupportInstalled` / `isEnhancedInspectorInstalled` vs.
+// `supportsEnhancedInspector`) and exits non-zero on any mismatch. This is
+// the regression guard the two-pass CI workflow relies on: a silently-broken
+// install must fail here, loudly, rather than surface downstream as an
+// indistinguishable "feature not installed" skip in the second test pass.
+//
+//   node client/bin/install-server-plugin.mjs
+
+import path from 'node:path';
+import Module, { createRequire } from 'node:module';
+import { fileURLToPath } from 'node:url';
+import dotenv from 'dotenv';
+
+const require = createRequire(import.meta.url);
+
+const repoRoot = path.resolve(fileURLToPath(new URL('.', import.meta.url)), '..', '..');
+const clientOutDir = path.join(repoRoot, 'client', 'out');
+
+// The compiled client output is written for the extension host: some modules
+// (transitively, gciLog.ts) `require('vscode')` at load time to get an output
+// channel for incidental logging we don't care about here. There's no real
+// `vscode` module outside the extension host, so stub it in for the one thing
+// these modules use it for, rather than dragging the whole shared-helpers
+// question into this PR's scope.
+const originalModuleLoad = Module._load;
+Module._load = function (request, parent, isMain) {
+  if (request === 'vscode') {
+    return {
+      window: {
+        createOutputChannel: () => ({ appendLine: () => {}, show: () => {}, dispose: () => {} }),
+      },
+    };
+  }
+  return originalModuleLoad.call(this, request, parent, isMain);
+};
+
+const { GciLibrary } = require(path.join(clientOutDir, 'gciLibrary.js'));
+const { installRefactoringSupport, isRefactoringSupportInstalled } = require(
+  path.join(clientOutDir, 'refactoring', 'refactoringInstall.js'),
+);
+const {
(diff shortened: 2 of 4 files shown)


.github/workflows/health-check.yml
```diff
@@ -154,5 +154,9 @@ jobs:
         env:
           GEMSTONE_VERSION: ${{ matrix.gemstone-version }}
         run: npm run test:server:start -- "$GEMSTONE_VERSION"
-      - name: Run tests
+      - name: Run tests (without server plugin)
+        run: npm test
+      - name: Install server plugin
+        run: node client/bin/install-server-plugin.mjs
+      - name: Run tests (with server plugin)
         run: npm test
```
ericwinger, line 162: Claude: This is now the second full `npm test` in this job, and it's the heavier one — it files in `engine-tests.gs` and runs the in-stone SUnit suites that the first (bare) pass skips. But the job's `timeout-minutes` is still `5`. Two full test runs plus the install step against a 5-minute cap is tight; I think this needs bumping.

Eric: Timing is sooooo machine dependent. 5 minutes seems like plenty of time on my machine - which is beefy. But the CI machine is smaller if I understand correctly.

CPU: AMD Ryzen 7 3800X — 8 cores / 16 threads, up to 3.9 GHz
RAM: 125 GB
OS: Ubuntu 22.04.5 LTS (Linux 5.15, x86_64)
Node: v24.15.0 

Here is an example run on my machine - RH 

Image

client/bin/install-server-plugin.mjs
```diff
@@ -0,0 +1,229 @@
...
+Module._load = function (request, parent, isMain) {
+  if (request === 'vscode') {
+    return {
+      window: {
+        createOutputChannel: () => ({ appendLine: () => {}, show: () => {}, dispose: () => {} }),
+      },
+    };
+  }
+  return originalModuleLoad.call(this, request, parent, isMain);
+};
```
ericwinger, line 51: Claude: This stubs `vscode` with only `window.createOutputChannel`. It works because the modules pulled in here happen to touch nothing else — but any future transitive import that reads another `vscode` API at load time will throw right here, far from the file that added it. Worth a comment noting the stub is deliberately minimal.

client/bin/install-server-plugin.mjs
```diff
@@ -0,0 +1,229 @@
...
+const {
+  installEnhancedInspectorSupport,
+  isEnhancedInspectorInstalled,
+  supportsEnhancedInspector,
+} = require(path.join(clientOutDir, 'enhancedInspectorInstall.js'));
+
+// GemStone's default SystemUser password on a fresh stone. Not written as
+// `...PASSWORD`/`password = '...'` — see the same note in
+// refactoringInstallCommand.ts about Open VSX's secret-scan false hit.
+const DEFAULT_SYSTEMUSER_PW = 'swordfish';
```
ericwinger, line 66: Claude:  Hardcoded fresh-stone default. Fine for CI, but it's an unstated assumption — if the base user were already `SystemUser`, or the stone's `SystemUser` password differed, the install fails here. OK to keep as CI-only, just flagging the coupling.

Eric:  I can't imagine a development stone not having the default password so I'm not sure it's a real concern.

client/bin/install-server-plugin.mjs
```diff
@@ -0,0 +1,229 @@
...
+      `${name} is not set. Run \`npm run test:server:start\` first (it writes client/.env.test).`,
+    );
+  }
+  return value;
+}
+
+// `VITE_GEMSTONE_STONE_NRS` / `VITE_GEMSTONE_GEM_NRS` encode the connection's
+// host/stone/netldi, which the SystemUser elevation below needs to rebuild
+// its own NRS strings (mirrors gciTestConfig.ts's NETLDI_NAME parse).
+function parseLoginFromNrs(stoneNrs, gemNrs) {
```
ericwinger, line 90: Claude: `parseLoginFromNrs` and `loginAsSystemUser` are hand-copied from `refactoringInstallCommand.ts` / `gciTestConfig.ts` (deliberately, to avoid importing `vscode` outside the extension host). The tradeoff is a second copy of the NRS string formats — if those ever change, this script drifts silently. Maybe a shared, vscode-free helper someday.

Eric: I wonder if NRS strings should be parsed and built in one place without the import of vscode to avoid future Claude changing one and not the other?

ericwinger (approved):
Some Claude comments. Some Eric comments. Hopefully useful. Nothing is a show stopper so I'm approving. 

I'm glad the CI tests are run fully twice - once with RB/EI & once without.  That is good coverage.

.github/workflows/health-check.yml
MatiasFernandez: We are still well behind the 5 min mark. So I think we are ok for now. There are other test tweaks I have in my sleeve, so in the future this could be reduced further

client/bin/install-server-plugin.mjs
MatiasFernandez: I'm working on a follow-up PR to improve things like this

client/bin/install-server-plugin.mjs
MatiasFernandez: not a concern for now, we are in full control of the populated stone in CI

client/bin/install-server-plugin.mjs
MatiasFernandez: Yup, my follow-up PR is to clean up code duplication. This one focuses on something working to get the CI changes in quickly

client/bin/install-server-plugin.mjs
jgfoster: I believe that at some point it will no longer be possible to log in as SystemUser with a password. I think we're headed toward a situation where you must be on the stone machine, logged in as an OS user that is the same one that owns the stone, and then there is no password required. So this problem is one that should go away eventually.

PR #225, opened 2026-07-24
Migrate debugQueries.ts to executeAndFetchString

npapagna:
## Summary
- `getMethodUriInfo`, `getReceiverClassChain`, `getBrowseTarget`, `getDoesNotUnderstandInfo`, `fetchFullPrintString`, `fetchStackDump`, and `fetchFrameVariables` each repeated the same boilerplate (resolve the `Utf8` class, then call `GciTsExecuteFetchBytes` with the same fixed arguments) seven times over.
- They now share one local `executeAndFetchString` helper that calls the GCI library's own `executeAndFetchString`, logs any failure to the session's output channel, and re-throws so each caller's existing `try`/`catch` degrades exactly as before.
- Two things carry over unchanged, on purpose (reviewed via `/code-review` and discussed one finding at a time): no busy-session guard (matches every other function in this file), and no fetch size cap (the old 64KB/256KB/8MB limits are gone for good, since the underlying paging has no total ceiling either).
- Fixed a real regression along the way: `fetchFullPrintString` now surfaces the actual failure message instead of a fixed generic error string.

## Test plan
- [x] `npx tsc -p client/tsconfig.json --noEmit`
- [x] `npm run lint` (eslint, clean)
- [x] `npm run format:check` (prettier, clean)
- [x] `npm run test:client` (215 test files / 3827 tests passing)
- [x] Manually verified against a live GemStone 3.7.5 session: left-click opens editable `gemstone://` source, right-click → Browse resolves inherited/class-side methods correctly (including `SmallInteger>>#/`), right-click → Implement in... resolves the receiver's class chain correctly.

### DIFF
--- a/client/src/debugQueries.ts
+++ b/client/src/debugQueries.ts
@@ -69,6 +69,16 @@ function intToOop(session: ActiveSession, n: number): bigint {
   return result;
 }
 
+function executeAndFetchString(session: ActiveSession, code: string): string {
+  try {
+    return session.gci.executeAndFetchString(session.handle, code);
+  } catch (e) {
+    const msg = e instanceof Error ? e.message : String(e);
+    logError(session.id, msg);
+    throw e;
+  }
+}
+
 // ── Frame info ──────────────────────────────────────────
 
 export interface FrameInfo {
@@ -204,13 +214,6 @@ export function getMethodUriInfo(
   methodOop: bigint,
 ): MethodUriInfo | undefined {
   try {
-    const { result: classUtf8, err: resErr } = session.gci.GciTsResolveSymbol(
-      session.handle,
-      'Utf8',
-      OOP_NIL,
-    );
-    if (resErr.number !== 0) return undefined;
-
     const code = `| method class baseClass dictName category |
 method := Object _objectForOop: ${methodOop}.
 class := method inClass.
@@ -225,16 +228,7 @@ dictName, String tab,
   category, String tab,
   method selector asString`;
 
-    const { data, err } = session.gci.GciTsExecuteFetchBytes(
-      session.handle,
-      code,
-      -1,
-      classUtf8,
-      OOP_ILLEGAL,
-      OOP_NIL,
-      64 * 1024,
-    );
-    if (err.number !== 0) return undefined;
+    const data = executeAndFetchString(session, code);
 
     const parts = data.split('\t');
     if (parts.length < 5) return undefined;
@@ -292,13 +286,6 @@ export function getReceiverClassChain(
   selector: string,
 ): ClassHomeInfo[] {
   try {
-    const { result: classUtf8, err: resErr } = session.gci.GciTsResolveSymbol(
-      session.handle,
-      'Utf8',
-      OOP_NIL,
-    );
-    if (resErr.number !== 0) return [];
-
     // selector is a method selector (no quotes), but guard the quote anyway.
     const sel = selector.replace(/'/g, "''");
     const code = `| rcvr meta cls sel rows nm dn impl |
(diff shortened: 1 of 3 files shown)


client/src/debugQueries.ts
```diff
@@ -641,38 +588,21 @@ export function fetchPrintString(
...
 }
 
-const MAX_FULL_PRINT = 256 * 1024;
-
 /**
  * Returns the full output of printOn: for an object, bypassing GemStone's
  * internal printString size cap.  printString uses a LimitedWriteStream
  * internally; calling printOn: directly with a plain WriteStream has no limit.
  */
 export function fetchFullPrintString(session: ActiveSession, oop: bigint): string {
```
ericwinger, line 596: Claude: fetchFullPrintString is now unbounded. The old MAX_FULL_PRINT = 256 KB cap that wrapped this fetch was deleted, so it now pages the full string with no ceiling on a synchronous GCI call. A pathologically large object materializes fully (chunk buffers + Buffer.concat ~= 2x memory) and can freeze the extension host. Defensible since the function's point is the full string, but please confirm it's an intended trade-off.

client/src/debugQueries.ts
```diff
@@ -69,6 +69,16 @@ function intToOop(session: ActiveSession, n: number): bigint {
   return result;
 }
 
+function executeAndFetchString(session: ActiveSession, code: string): string {
+  try {
+    return session.gci.executeAndFetchString(session.handle, code);
+  } catch (e) {
+    const msg = e instanceof Error ? e.message : String(e);
+    logError(session.id, msg);
```
ericwinger, line 77: Claude: (also affects fetchStackDump:736, fetchFrameVariables:832)

Lost log context. The shared helper logs only the bare message. fetchStackDump and fetchFrameVariables previously logged with a "fetchStackDump: ..." / "fetchFrameVariables: ..." prefix, so the output channel no longer says which query failed. Consider passing a label arg into the helper.

Eric: It's essentially saying the function that produced the error was lost IIUC.

ericwinger, line 77: Claude: Silent errors are now logged. Via the helper, these four functions now log GCI errors they used to swallow. Generally an improvement; just noting the added output-channel noise.

Eric: Non-swallowed errors seem like a good thing to me.

ericwinger (approved):
Looks good.  The unbounded function might be worth investigating where it is sent to make sure the client doesn't get hung on some huge data hunk.

client/src/debugQueries.ts
npapagna: It sounds like the real concern is the extension host hanging silently with no feedback to the user, which is a worse outcome than a truncated string.

The trade-off is intentional. The old cap was a pragmatic safety net, but it meant the result could be incomplete with no indication of that. Unbounded is what `fetchFullPrintString` actually promises.

The right place to handle pathological sizes is in Smalltalk, where we know there's more content and can give the UI what it needs to surface that. Worth a follow-up issue — , tagging @jgfoster since we're actively shaping the roadmap.

client/src/debugQueries.ts
npapagna: It might seem like the function name is lost. It's still there, in the stack trace of the re-thrown error.

npapagna: Agreed! Better to surface them than swallow them.

client/src/debugQueries.ts
jgfoster: Yes, please create a follow-up issue for this. Feel free to attach it as you think best, or leave it and we'll assign its place later.

npapagna: Created an issue to document this: https://github.com/GemTalk/Jasper/issues/333

PR #226, opened 2026-07-24
Migrate refactoring GCI tests to CI, gated by server-plugin feature scenario

MatiasFernandez:
## Summary
- Extract shared install helpers (`client/src/serverPlugin/installHelpers.ts`, `systemUserAuth.ts`) out of the CI provisioning script and the install-command drivers, so login/probe/abort logic lives in one vscode-free place.
- Add `client/src/serverPlugin/pluginFeatures.ts` as the single source of truth for which server-plugin features exist, their version applicability, payload location, and probe/install behavior. It's a keyed record (`pluginFeatures.refactoring`) with `PluginFeatureId` / `PluginFeatureRef` derived from it via `as const satisfies`, and a `PLUGIN_FEATURES` array derived for install/verify loops. Every consumer — the CI script, install commands, and optional-support offer — is re-pointed at it and references feature **objects** directly, so a typo or an ad-hoc look-alike is a compile error rather than a runtime failure.
- Add a live `probe(session)` check to `ServerSupportFeature` (distinct from the cached `isMissing` latch) and `requireServerPluginFeature` / `requireServerPluginFeatureAbsent` test helpers that skip integration tests when the connected stone isn't in the applicable scenario. The gates take a `PluginFeatureRef` object (no id lookup, no runtime throw) and decide applicability live against the session — version support plus a presence probe — so no phase/env flag is needed and each CI pass (bare vs. plugin-installed stone) runs exactly its relevant subset.
- Migrate the `__tests__/gci/refactoring` class-variable and rename-class-variable smoke/e2e tests, and the inspector-routing smoke test, into the default integration/unit suites (now gated by the above helpers) instead of being excluded from normal CI runs.
- Document the two-pass CI model and when to reach for the `requireServerPluginFeature` / `...Absent` gates in a `.claude/rules/client/tests.md` client-test rules section for agents, pointing at the `pluginFeatures` registry and helper JSDoc as the source of truth.

## Test plan
- [x] `npm test` against a bare test stone
- [x] `npm test` against a stone with the server plugin installed


client/bin/install-server-plugin.mjs
```diff
@@ -99,38 +92,6 @@ function parseLoginFromNrs(stoneNrs, gemNrs) {
   return { gem_host: stoneMatch[1], stone: stoneMatch[2], netldi: netldiMatch[1] };
```
ericwinger, line 92: Claude: #223's review flagged both `loginAsSystemUser` and `parseLoginFromNrs` as hand-copied. This PR consolidated `loginAsSystemUser` into serverPlugin/installHelpers.ts (good), but `parseLoginFromNrs` still lives only here and overlaps the NRS-parsing in __tests__/gci/gciTestConfig.ts. Not blocking — it's CI-script-local and small — just flagging that the de-duplication follow-up is partial, not complete.

Eric:  Seems ticky-tack to me but I erred on the side of too much information for this comment.

ericwinger (dismissed):
Claude: Skip-count visibility (future thought, out of scope here): the two CI passes already yield equal *total* test counts by construction, but nothing surfaces the passed/skipped split. Surfacing it would be a change to the health-check workflow, which this PR doesn't touch — noting it only as a follow-up idea, not a finding against #226.

Eric - Could we talk about the following?  I'm not sure how I would write a client unit test that required RB support in the future such that it gets picked up by CI correctly. I think the default should be that a unit test should *not* expect RB/EI to be there, of course. Maybe I'm missing something. 

More comments from Claude on this: 

**Something to think about (not blocking):** how a future contributor tells CI to skip these tests when the refactoring engine isn't present.

There's no CI-side signal. The two `npm test` passes in the health-check workflow (added in #223 — unchanged here) run the identical test set; the skip is decided per-test, at runtime, by one line at the top of the `it()`:

    requireServerPluginFeature('refactoring', ctx, session());

That helper does a live probe against the connected stone (`checkRefactoringSupportAvailable` → resolves `GsRenameInstanceVariableRefactoring`); if the class isn't there, it calls `ctx.skip(...)`. So the same file skips on a bare stone and runs after `node client/bin/install-server-plugin.mjs` has filed the engine in — the only difference between the two passes is whether that install ran in between. No env flag, no phase variable.

So a future dev adds an RB-gated test by copying that one line. Nothing enforces or advertises it — discoverability is "read an existing test or the helper's doc comment." Worth a short note in the test area (and/or a `PluginFeatureId` union type so a mistyped id fails at compile time rather than throwing at runtime).

client/bin/install-server-plugin.mjs
MatiasFernandez: I have a follow up to deal with gci tests. I can add this to that work

ericwinger (approved):

PR #227, opened 2026-07-24
Extract Temporary / Inline Temporary refactorings

ericwinger:
Adds two method-local RB refactorings (catalog M3 + M4), a mirror-image pair built on
the R5 (rename-temporary) model: each stages a single `#methodRecompile`, applies
server-side (recompile, **never commits**), shows the standard preview panel, and needs
no class-definition edit or cross-method scan.

### Extract Temporary
Select an expression in a method editor → introduce `name := <expr>.` and replace the
occurrence(s) with the temporary; opt-in "replace all N" scoped to the declaring
sequence. A whole-statement selection becomes `name := expr` (no degenerate bare read).
A bare variable and an assignment expression are extractable; declines only what would
not compile (bare `super`/`thisContext`, an assignment target, a declaration).

### Inline Temporary
Put the cursor on a once-assigned temporary → replace each read with the assigned
expression (parenthesised only where precedence requires) and remove the declaration +
assignment. Declines multi-assignment / read-before-assign / never-read /
non-atomic-read-twice / argument / instance-variable / class-variable / global, each
with a classifying reason.

### Engine
`GsExtractTemporaryRefactoring` + `GsInlineTemporaryRefactoring` (+ GS SUnit: Extract 32,
Inline 24) — green on **both 3.6.2 and 3.7.5**.

### Client
Commands / preview / panels / queries under `client/src/refactoring/`, hosted in the
native **Refactor…** menu. As part of this, **all of M1–M4 moved into the Refactor…
menu** via `RefactorCodeActionProvider` and the four top-level editor-context entries
were removed (the menu had grown too large). 100+ client unit tests + guarded GCI
integration tests.

### Also included
- **Shared fix (grail #62 review):** the source-editor engine gate `ensureRbSupport`
  returned `true` unconditionally after the install prompt; it now takes the live
  session and re-checks `rbSupportAvailable`, so a failed/declined install cleanly
  refuses instead of emitting a cryptic DNU. All RB command call sites updated.

### Verification
Engine SUnit green on 3.6.2 + 3.7.5; full `npm test` green (client 3942 / server 322 /
mcp 92); lint / format / compile clean. Current with `main`.


client/src/refactoring/extractTemporaryPanelHtml.ts
```diff
@@ -0,0 +1,210 @@
...
+  if (oos.decline) {
+    lines.push(`⚠ ${escapeHtml(oos.decline)}`);
+  }
+  if (oos.collision) {
+    lines.push(
+      `⚠ ${escapeHtml(oos.collision)} — applying will fail unless you choose another name.`,
+    );
+  }
+  if (occurrenceCount > 1) {
+    lines.push(`Replacing all ${occurrenceCount} occurrences of the expression in this scope.`);
```
MatiasFernandez: 🚫 This banner reads "Replacing all N occurrences…" whenever `occurrenceCount > 1`, but `occurrenceCount` is the total number of occurrences in scope from the pre-flight — not what the preview actually rewrites. When the user picks "Replace only the selected occurrence" (`replaceAll: false`) on an expression that appears, say, 3 times, `start.occurrenceCount` is still 3, so the panel tells them "Replacing all 3 occurrences…" while the diff below rewrites only one — directly contradicting the choice they just made in the quick-pick.

The engine confirms this: `occurrenceCount` counts identical occurrences in the declaring scope regardless of `replaceAll`. Should the banner reflect `replaceAll` (thread it through to the panel, or have the start-envelope report the effective replaced count) so it only says "all N" when all N are really being replaced?

client/src/refactoring/extractTemporaryCommand.ts
```diff
@@ -0,0 +1,211 @@
...
+  ensureRbSupport,
+  refuse,
+  reloadMethodEditor,
+  saveIfDirty,
+} from './renameAtCursorShared';
+
+/** 1-based count of code points before `position` (the engine indexes stored source
+ *  by code point; VS Code offsets count UTF-16 units — they differ past a non-BMP
+ *  character). */
+function codePointsBefore(document: vscode.TextDocument, position: vscode.Position): number {
```
MatiasFernandez: 💭 This local `codePointsBefore` looks identical to the one already exported from `renameAtCursorShared.ts` (which this file already imports `resolveMethodEditor` / `ensureRbSupport` / `saveIfDirty` from). Could we drop the local copy and import the shared one, so the code-point offset convention stays defined in a single place for the whole rename/refactor family?

client/src/refactoring/__tests__/refactoringExtractTemporary.integration.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    const result = parseApplyResult(await applyExtractTemporary(asyncExec, token));
+    expect(result.applied).toBe(1);
+    expect(result.failed).toEqual([]);
+
+    const rewritten = exec(
+      `(${BASE} compiledMethodAt: #doStuff environmentId: 0 otherwise: nil) sourceString`,
+    );
+    // A temporary named `tmp` was introduced and assigned.
+    expect(rewritten).toContain('tmp');
```
MatiasFernandez: 💭 The fixture has `self hash` twice and this applies with `replaceAll: false`, so exactly one occurrence should be rewritten — but `toContain('tmp')` / `toContain(':=')` would still pass if the extraction were wrong (both occurrences replaced, or the temp assigned something other than the selection). Worth pinning the shape a bit more — e.g. asserting `tmp := self hash` is present and that `self hash` still appears once — so the test actually proves the replace-only-selected semantics rather than just that a `tmp` assignment showed up somewhere?

client/src/refactoring/extractTemporaryPanelHtml.ts
ericwinger: This is a legitimate bug.  I'll fix this.

client/src/refactoring/extractTemporaryCommand.ts
ericwinger: Deduped.

client/src/refactoring/__tests__/refactoringExtractTemporary.integration.test.ts
ericwinger: Should be fixed. If the refactoring was incorrrect and both were replaced, the test would still pass, IIUC.

MatiasFernandez (approved):
LGTM

PR #315, opened 2026-07-27
Reorganize the Enhanced Inspector under client/src/enhancedInspector/

MatiasFernandez:
Applies the `refactoring/` layout from 950fa5b to the Enhanced Inspector — the next-largest feature still spread across the flat `client/src`. Sources, webview script, feature-owned queries and tests now sit in one folder, moved with `git mv` so history follows. No behavior change.

### Commits

1. **Promote the refactoring webview loader to a shared helper** — `refactoring/webviewAssets.ts` → `client/src/webviewAssets.ts` as `readWebviewScript(fileName, subdir?)`, with `readRefactoringWebviewScript` left behind as a one-line wrapper so the move itself touched no call sites.
2. **Reorganize the Enhanced Inspector** — 6 modules + `enhancedInspectorColumns.js` + `queries/getEnhancedInspectorViewSpecs.ts` + 15 tests into `client/src/enhancedInspector/`, imports repointed, `.vscodeignore` whitelist updated, and the panel now reads its webview script through `readWebviewScript` (a `client/src`-relative `fs.readFileSync` no longer resolves one level deeper).
3. **Scan `client/src` recursively for runtime-read webview assets** — the packaging guard listed `client/src` non-recursively and matched only the literal `readFileSync(path.join(__dirname, '..', 'src', '<name>.js'))` form, so it had already gone blind to the five `refactoring/*View.js` scripts and would have missed this one. It now walks recursively and matches the helper forms too.
4. **Document the conventions** and **trim hardcoded file references** out of `client/CLAUDE.md`. The webview-script bullet pins the hazard where it actually is: a script inside a feature folder can't be resolved by either naive path form, which is why `readWebviewScript` exists. A bare `fs.readFileSync(path.join(__dirname, '..', 'src', <name>))` is still correct for the top-level scripts, and those call sites are left alone.
5. **Cover both webview script layouts in `readWebviewScript` tests** — the helper's two-branch lookup is now load-bearing for eleven scripts across two features, and the bundled-layout fallback (the branch whose failure mode is "the packaged extension never activates") had only ever been checked by hand. Mocked `fs`, `existsSync` driven both ways, assertions on the path handed to `readFileSync`. In the source layout both candidates normalize to the same absolute path, so the tests derive the expected string from their own `'..', 'src'` literals; pointing the fallback at `../out/` turns them red.
6. **Call the shared webview script loader directly from refactoring** — deletes the `readRefactoringWebviewScript` wrapper from step 1 in favor of `readWebviewScript(fileName, 'refactoring')` at its ten call sites. The wrapper existed only to supply a constant argument, and the packaging guard had to hardcode that same constant in a third regex to follow it — a guard that keeps passing while verifying a path production code no longer uses is worse than no guard. One regex and one module fewer; the discovered-asset list is unchanged.

`inspectRouter.ts` (the seam between the classic and enhanced inspectors), `resources/enhancedInspector/`, and `enhancedInspectorPrintString.test.ts` (enhanced-inspector by name, debugger-query by substance) stay where they are.

### Verification

- `npm run lint`, `npm run format:check`, `npm run compile` clean; full `npm test` green against a live 3.6.2 stone (client: 218 files, 3850 passed / 29 skipped).
- `npm run package` → the `.vsix` ships `client/src/enhancedInspector/enhancedInspectorColumns.js` and all five `refactoring/*View.js` scripts.
- The packaging guard was confirmed to still bite after the regex removal: dropping one `!client/src/refactoring/*View.js` line from `.vscodeignore` fails that asset's case, and the set of discovered assets is byte-identical to what the pre-change scan found.
- Moves show as renames under `git diff -M`, and `git log --follow` walks past the move.

Not covered by CI, needs a human at a GUI: **Inspect It** rendering the miller-column panel (proves the webview script loads from the packaged path), the perf output channel and status-bar commands, and one **Install Server Support** run. Worth doing from an installed `.vsix` rather than the dev host — only the packaged layout exercises the fallback branch in production, since `.vscodeignore` ships `client/out/extension.js` alone.


ericwinger (approved):
Claude's major finding was that one goal of moving the EI to it's own dictionary has not been met yet. But I think that's coming so no need to copy all its verbose output here. 

Nothing else really worth mentioning. Looks good.

PR #316, opened 2026-07-27
Change method signature (add / remove / reorder parameters)

ericwinger:
## Overview

Adds the catalog refactoring — **change a method's signature by adding, removing, or reordering parameters** — across a selector's implementors and senders, previewed before applying and applied server-side **without committing** (the user commits). It generalizes (rename method, which is arity-preserving) so the argument count can change.

## Is "remove parameter" a refactoring?

Yes — Remove-Parameter is behavior-preserving **only when the parameter is unused in the method body** (the exact mirror of Add-Parameter; both ship in the classic Refactoring Browser and are folded into Fowler's "Change Function Declaration"). M5 implements all three operations, with **remove gated by an unused-parameter precondition**: removing a parameter that is referenced in any in-scope implementor is declined, naming the class. Add is behavior-preserving because the new parameter is unused in the body and every call site is given a caller default; reorder is always safe.

## Engine — `GsChangeSignatureRefactoring`

Generalizes R2's `(newParts, permutation)` model: a `0` in the permutation marks a brand-new parameter (add), and an old argument index left out is removed. Reuses R2's scope model (`#class` / `#hierarchy` / `#dictionary` / `#wholeSystem` + out-of-scope counts), `methodRename`/`methodRecompile` staging, paginated preview, and server-side apply-except-deselected (no commit). Implementor headers rewrite via the vendored AST's `renameSelector:andArguments:` (which already rebuilds the signature region when arity changes); sends rewrite minimal-diff. Preconditions surfaced to the client: `decline` (remove-of-used, duplicate/temp-colliding arg name) and `collision` (new selector already implemented), both blocking Apply.

## Client — two entry points, one preview flow

- **Explorer** — "Change Signature…" on a method row (`gemstone.explorer.changeSignature`).
- **Method source pane** — the native **Refactor…** menu (`gemstone.changeMethodSignature`), targeting the edited method's own signature (no text selection needed).

Both open a signature editor (per-parameter rows: reorder ▲▼, **Add parameter** with a default-value input, **Remove**), preview as usual, apply server-side without committing, then reload the edited method and refresh the Explorer (the selector changed).

## Tests

- **Engine GS SUnit** — `GsChangeSignatureRefactoringTest`, 29 tests; full engine suite **288 run / 0 failed / 0 errors on both GemStone 3.6.2 and 3.7.5**.
- **Client** — 84 unit/jsdom/host-wiring tests across the preview model, editor, panel, command (both entry points), and query builders.
- **GCI integration** (`refactoringChangeSignature.integration.test.ts`) — gated via `requireServerPluginFeature(pluginFeatures.refactoring, …)` (runs in the plugin-installed CI pass, skips with a reason on a bare stone). Covers the engine SUnit suite in-stone plus add / reorder / remove-unused round trips and the decline / collision / deselection paths. Exercised green (9/9) against an engine-loaded 3.7.5 stone.

Full `npm run lint && format:check && compile && npm test` green (client / server / mcp).

## Notes

- Branches off the M3/M4 work (PR #227), now merged to main; the diff here is M5-only.
- No GemStone commit is ever performed by the refactoring; recompiles are staged and applied to the session only.


gs-src/refactoring/engine/GsChangeSignatureRefactoring.class.st
```diff
@@ -0,0 +1,676 @@
...
+				implTree := RBParser parseMethod: m sourceString onError: [:msg :p | nil].
+				implTree notNil ifTrue: [
+					implArgs := implTree arguments.
+					removed do: [:k |
+						(k <= implArgs size and: [implTree body references: (implArgs at: k) name])
+							ifTrue: [^declineString := 'parameter ', (implArgs at: k) name,
+								' is used in ', base name asString, ' and cannot be removed']]]]]
+			on: Error do: [:e | nil]]].
+	"5. collision with a distinct existing method on the defining class"
+	(newSelector ~= oldSelector and: [(defTarget compiledMethodAt: newSelector otherwise: nil) notNil])
```
MatiasFernandez: 🚫 The collision check only looks at `defTarget` — the *defining* class — but in `#hierarchy` / `#wholeSystem` scope we stage a `methodRename` for **every** in-scope implementor.

Concretely: `GsCSSub` implements both `movePointX:y:` and `moveY:x:`, and the user changes `movePointX:y:` → `moveY:x:` in hierarchy scope. `newSelectorCollision` is nil (the *base* class has no `moveY:x:`), so Apply is allowed; `applyChange:` then compiles the renamed source over `GsCSSub>>moveY:x:` and removes `movePointX:y:`. The subclass's original `moveY:x:` is gone from the session — and the preview card just shows a `GsCSSub>>movePointX:y:` rename, with nothing hinting that another method was overwritten.

Should the check run over every in-scope implementor class rather than only the defining one? `testCollisionWhenNewSelectorAlreadyExists` covers only the defining-class case, so a subclass fixture would make a good companion test.

gs-src/refactoring/engine/GsChangeSignatureRefactoring.class.st
```diff
@@ -0,0 +1,676 @@
...
+	parts := aMessageNode selectorParts.
+	args := aMessageNode arguments.
+	argTexts := (1 to: self newArity) collect: [:i |
+		(permutation at: i) > 0
+			ifTrue: [src copyFrom: (args at: (permutation at: i)) start to: (args at: (permutation at: i)) stop]
+			ifFalse: [(defaults at: i) asString]].
+	newArgNodes := (1 to: self newArity) collect: [:i |
+		(permutation at: i) > 0
+			ifTrue: [args at: (permutation at: i)]
+			ifFalse: [RBParser parseExpression: (defaults at: i) asString]].
```
MatiasFernandez, line 532: 💭 `RBParser parseExpression:` runs inside `buildChangeSet`'s per-method `on: Error do:`, so a default that doesn't parse (the user clears the `default:` input, or types something like `1 +`) never reaches `declineString` — the sender is counted as *skipped* and the implementor rename still applies.

The end state is the worst of both: the new selector is compiled and the old one removed, while every sender is left calling the removed selector — a DNU the next time that code runs. The skipped banner is the only signal, and it reads like a minor warning rather than "your call sites are now broken".

Should an empty or unparseable default for a `permutation = 0` position be a hard `declineString` in `computeAnalysis`, next to the duplicate-name check, so Apply is blocked instead of half-applying?

gs-src/refactoring/engine/GsChangeSignatureRefactoring.class.st
```diff
@@ -0,0 +1,676 @@
...
+	regionStart := parts first start.
+	regionStop := args isEmpty ifTrue: [parts last stop] ifFalse: [args last stop].
+	region := WriteStream on: String new.
+	newParts keysAndValuesDo: [:i :p |
+		i = 1 ifFalse: [region space].
```
MatiasFernandez, line 537: 💭 The region is rebuilt with a single space between every part and argument, so a send that was written across lines:

```smalltalk
^self
    movePointX: someLongExpression
    y: anotherLongExpression
```

comes back as one long line even for a pure reorder. The class comment says the rewrite "preserves comments, string literals, and formatting by mutating the AST and registering RBStringReplacements so … splices minimally" — which holds for the implementor header but not for sends.

Is the collapse an accepted trade-off here (I can see why rebuilding the whole region is the simple way to handle arity changes), or worth keeping the original inter-part whitespace? Either way, every fixture sender is single-line (`^self movePointX: 1 y: 2`), so a multi-line sender test would pin whichever behaviour we want.

client/src/refactoring/changeSignaturePreview.ts
```diff
@@ -0,0 +1,400 @@
...
+  };
+}
+
+/** A human label for a preview row: "Foo>>bar:baz:" or "Foo class>>bar:baz:". */
+export function methodChangeLabel(change: MethodSignatureChange): string {
+  const side = change.isMeta ? ' class' : '';
+  return `${change.className}${side}>>${change.selector ?? '?'}`;
+}
+
+// --- selector shape helpers ---------------------------------------------------
```
MatiasFernandez: 💭 `isKeywordSelector`, `isBinarySelector`, `selectorParts`, `selectorArgCount`, and `buildSelector` look byte-identical to the ones in `renameMethodPreview.ts` in this same directory (M5 already reuses R2's scope type and its panel view script, so the precedent for sharing is there).

Should these be imported from R2's module, or lifted into a small shared `selectorShape.ts` that both refactorings use? My next comment is a concrete example of a fix that now has to be made in two places.

client/src/refactoring/changeSignaturePreview.ts
```diff
@@ -0,0 +1,400 @@
...
+
+/** True if the selector is a keyword selector (contains a colon). */
+export function isKeywordSelector(selector: string): boolean {
+  return selector.includes(':');
+}
+
+/** True if the selector is a binary selector (all chars are binary characters). */
+export function isBinarySelector(selector: string): boolean {
+  return /^[-+*/~<>=&|@%,?!]+$/.test(selector);
```
MatiasFernandez: 💭 This character class is missing `\`, which is a valid GemStone binary selector character — the server's own `SELECTOR_CHARS` set in `server/src/lexer/lexer.ts` includes it.

So `isBinarySelector('\\')` is false, which makes `selectorArgCount('\\')` answer 0 and `validateSignatureParts` reject changing a signature *to* `\\` with "A unary selector must be an identifier; a binary selector one or more binary characters." Should we add `\` (and ideally source the set from the lexer's, so the two can't drift)?

client/src/refactoring/changeSignatureEditorView.js
```diff
@@ -0,0 +1,207 @@
...
+      const anyKeyword = p.some(function (s) {
+        return s.charAt(s.length - 1) === ':';
+      });
+      if (
+        anyKeyword &&
+        !p.every(function (s) {
+          return /^[A-Za-z_][A-Za-z0-9_]*:$/.test(s);
+        })
+      ) {
+        return 'Each keyword part must be an identifier ending in a colon.';
```
MatiasFernandez: 💭 The webview's `validate()` covers the all-keyword case, but not the two extra rules the host applies in `validateSignatureParts`: a single non-keyword part must be a valid identifier or binary operator, and more than one part *must* be keyword parts.

So if a user strips the colons off `at:put:`, OK stays enabled, the panel closes and disposes, and only then does the host pop "A selector with more than one part must use keyword parts (each ending in a colon)." — with the whole edit (reordering, added rows, defaults) lost, since there's no way back into the editor.

Should the webview mirror those two checks so it's caught while the rows are still on screen? The comment above says the live validation "mirrors changeSignaturePreview closely enough", but the gap is exactly the case where losing the edit hurts most.

client/src/refactoring/changeSignatureCommand.ts
```diff
@@ -0,0 +1,256 @@
...
+    void vscode.window.showErrorMessage(`Change signature: ${partsErr}`);
+    return false;
+  }
+  const dupe = duplicateArgName(edit.newArgNames);
+  if (dupe) {
+    void vscode.window.showErrorMessage(`Change signature: duplicate argument name '${dupe}'.`);
+    return false;
+  }
+  const newSelector = buildSelector(edit.newParts);
+  if (isNoOpChange(edit.newParts, edit.permutation, oldSelector)) return false; // nothing to do
```
MatiasFernandez: 💭 This returns silently, so a user who opens the editor, moves rows around and ends up back at the original signature just sees the panel close with nothing happening — no message, no preview.

Every other bail-out here says something (`total === 0` shows "nothing to change", collision and decline both `refuse(…)`), and `changeSignatureCommand.test.ts`'s own header calls out the "always tell the user why nothing happened" contract. Worth a `showInformationMessage` along the lines of "The signature is unchanged; nothing to do."?

gs-src/refactoring/engine/GsChangeSignatureRefactoring.class.st
ericwinger: This is a good catch. Don't want to lose data.

gs-src/refactoring/engine/GsChangeSignatureRefactoring.class.st
ericwinger: This is the only suggestion I'm not going to do.  I'm ok with the collapse as the user will see it in the diffs.  They can go back and format the method if they'd like. The following (preserve multiline) looks a little weird to me so it seems like 1/2 dozen of one 6 of the other type of problem. I'll go with the simpler code. 

^self
    movePointX: 1
    y: 1

MatiasFernandez (approved):
LGTM

PR #325, opened 2026-07-28
Replace GciTsResolveSymbol with resolveSymbol everywhere

npapagna:
## Summary

- `GciTsResolveSymbol` (the raw GCI binding for resolving a class/global
  by name) has a known memory leak the GemStone team is still
  investigating. Every call site that used it directly now goes
  through `resolveSymbol` (or `utf8ClassOop` for the cached `Utf8`
  lookup) instead, which resolves the same name via `createString` +
  `GciTsResolveSymbolObj` + `releaseObject` (three round trips, no
  leak). The raw binding and its koffi handle are deleted outright
  since nothing calls it anymore.
- Code review of this change turned up one real regression, which is
  fixed here: `wrapWithEnhancedInspectorPerfProxy` bound wrapped
  `GciLibrary` methods to the raw proxy target instead of the `get`
  trap's `receiver`, so `resolveSymbol`'s own internal round trips ran
  unproxied and silently dropped out of the Enhanced Inspector perf
  counter. Binding to `receiver` lets those nested calls re-enter the
  trap and get tracked individually.
- Also folded in along the way: `GciTsContinueWithAsync` was missing
  from the perf tracker's round-trip list despite driving the same
  wire call as the already-tracked `GciTsContinueWith`; two dead
  `expect(classOop).not.toBe(OOP_ILLEGAL)` assertions in
  `gciCompile.test.ts` were removed since `resolveSymbol` now throws
  instead of returning a sentinel value; and an em dash was removed
  from `resolveSymbol`'s JSDoc.

## Test plan

- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run compile`
- [x] `npm test` (240 client files / 4059 tests, 12 server files / 322
      tests, 3 mcp-server files / 92 tests, all passing)
- [x] `npm run test:gci -- --testNamePattern GciTsCompileMethod` (the
      file with the removed dead assertions, against a live stone)

### DIFF
--- a/client/src/__tests__/debugQueries.test.ts
+++ b/client/src/__tests__/debugQueries.test.ts
@@ -627,10 +627,9 @@ describe('debugQueries', () => {
         GciTsOopToI64: vi.fn(() => ({ value: 5n, err: { ...noErr } })),
         GciTsNewString: vi.fn(() => ({ result: EXPR_STRING, err: { ...noErr } })),
         GciTsNewSymbol: vi.fn(() => ({ result: AMOUNT_SYMBOL, err: { ...noErr } })),
-        GciTsResolveSymbol: vi.fn((_h: unknown, name: string) => ({
-          result: name === 'SymbolDictionary' ? 0xd1n : name === 'SymbolList' ? 0xd2n : 0xd3n,
-          err: { ...noErr },
-        })),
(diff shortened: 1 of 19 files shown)


MatiasFernandez (approved):
LGTM

PR #326, opened 2026-07-28
Migrate GCI query smoke tests into CI integration suite [Part 1]

MatiasFernandez:
## Summary

First of a multi-PR project to get all GCI tests running in CI across every supported server version. This part migrates the simplest tests; more phases will follow.

- Moves `queryPython`, `querySelectors`, `querySunit`, and `rowanExportFixpoint` smoke tests out of the excluded `gci` vitest project and into the default project, using `useIntegrationTest` instead of the ad hoc `queryHarness` login.
- Inlines the small `selectorExists` helper into `querySelectors` and has `querySunit` install its probe fixture per-test so the harness's auto-abort cleans it up, letting `queryHarness.ts` retire.

## Test plan

- [x] `npm test` against a live stone (including the migrated integration tests)
- [x] CI run across supported server versions


client/src/queries/python.ts
```diff
@@ -106,20 +106,26 @@ function buildPythonQuery(grailExpression: string, pythonSource: string): string
   // The hint is a literal ASCII string, but we still pipe it through
   // `encodeAsUTF8` at the unified return below so every result has the same
   // transfer-protocol class.
+  //
+  // Every literal in this source (not just the hint) must stay ASCII: an
```
ericwinger: The em-dash -> ASCII fix looks incomplete. The same 3.6.2 `ComStrmSetCursor` hazard still lives in generated Smalltalk this PR's tests don't exercise:
- `executeCode.ts:18-19` (`wrapExecuteCode`) - em-dashes inside `on: ... do: [ ... ]` blocks (the execute-code / doit path).
- `commitTransaction.ts:7` - em-dash in a string literal inside `ifFalse: [ ... ]`.
Reasonable to defer to a follow-up (Part 1 fixed only what its migrated tests surfaced), but flagging so it's tracked. A lint/grep gate for non-ASCII inside generated-Smalltalk template literals would prevent recurrence.

client/src/queries/__tests__/querySunit.integration.test.ts
```diff
@@ -183,41 +162,34 @@ describe('SUnit queries (live GCI)', () => {
...
-        expect(() => runFailingTests(s.exec)).toThrow(/too many to run|Narrow the run/);
-      } else {
-        const results = runFailingTests(s.exec);
-
-        expect(Array.isArray(results)).toBe(true);
-      }
-      // On a small image this runs SUnit across every test class; on a freshly provisioned
-      // (cold) stone that legitimately takes ~30s, past vitest's 5s default — give it room.
-    }, 120000);
+    // The blocking-call guard (MAX_RUN_CLASSES) is NOT covered here. Asserting
```
ericwinger, line 165: Good call parking this - the branch-on-image-size plus an un-interruptible blocking GCI call can't run safely in CI. Noting it removes the only end-to-end exercise of the MAX_RUN_CLASSES guard from CI; the discovery-fragment (round-2/round-5) regressions stay covered directly here. No change requested.

ericwinger (dismissed):
Claude:  Static, read-only review of the smoke → CI migration (from the PR ref; suites not run here). Overall a clean, well-documented migration that preserves and expands coverage
  into CI; verified no dangling refs to the deleted `queryHarness`/old `probeFixture`. Two substantive notes are inline below; a nit to confirm `useIntegrationTest` aborts
  per-test (querySunit now relies on that for probe cleanup).

client/src/queries/python.ts
MatiasFernandez: yes, that's a separate work @npapagna has been investigating. We can handle that separately

ericwinger (approved):

PR #327, opened 2026-07-28
Consolidate test-connection resolution and session helpers

MatiasFernandez:
## Summary
- Share a single `resolveTestConnection(env)` implementation (composed NRS first, atomic/shell vars as fallback) across `gciTestConfig.ts`, `testActiveSession.ts`, and the server-plugin install script, removing duplicated NRS-parsing logic
- Add a `testActiveSession.ts` factory for building a real `ActiveSession` in integration tests, replacing ad-hoc `as unknown as ActiveSession` casts across the refactoring integration test suite
- Consolidate `refreshWorkingSessionAfterInstall` into `systemUserAuth`, deleting the stale `systemUserSession.ts` duplicate
- Add `stoneNrsFor`/`parseStoneNrs`/`netldiNameFromGemNrs` to `loginTypes`, and emit `VITE_GEMSTONE_VERSION` (the one value not encoded in any NRS) from `gs-create-test-env-file.sh`
- `resolveTestConnection` now fails fast when the Stone NRS doesn't parse (e.g. a hand-edited `.env.test.local`), instead of silently falling back to a bogus `gem_host`/`stone` that could make a downstream SystemUser login fail confusingly or, worse, silently succeed against the wrong stone
- **Migrate the install script off `client/out/`.** Sharing the resolver initially meant `require()`-ing compiled output — an edge invisible to `tsc`, ESLint, and `git grep`, and one where a stale build produces *wrong behavior* rather than an error. It now imports `client/src` directly:
  - `client/bin/installServerPluginMain.ts` holds the real logic, type-checked via a new `client/tsconfig.bin.json` (`noEmit`, so nothing lands in `client/out/`) wired into `npm run compile` as `compile:bin`
  - `client/bin/install-server-plugin.mjs` shrinks to a launcher whose only job is installing the `vscode` stub before the imports it must precede — static imports evaluate before any module-body code, so the patch can't sit in the same file
  - `tsx` runs the `.ts` entry point: Node 24's native type-stripping can't, since `client/src` uses extensionless imports
  - new `test:server:install-plugin` script, which CI now calls instead of `node client/bin/install-server-plugin.mjs`

## Note on the deleted `testConnectionImports.test.ts`
That source-text scan of import lines existed only because the compiled-output edge was
unobservable by any tool — nothing in `npm test` ever loaded `client/out/`. With the script
importing `client/src` directly, the imports are type-checked and `git grep`-visible, so the
scan is redundant rather than merely under-matching.

Worth recording: its header claimed a regression would surface as `Cannot find module 'vscode'`,
which was never true — the launcher's stub silently satisfies both a `vscode` and a
`sessionManager` value import, so the scan's two main targets were exactly what the stub
absorbed.

The stub itself survives, because two modules reachable from the install path
(`gciLog.ts`, `nbRunner.ts`) import `vscode` at the top level purely for lazily-used APIs.
Retiring it means inverting those two dependencies, which is separable architectural work
and is tracked as a follow-up rather than bundled here.

## Test plan
- [x] `npm test` against a live test stone
- [ ] `npm run test:server:install-plugin` against a live test stone — installs and verifies the plugin
- [ ] `rm -rf client/out && npm run test:server:install-plugin` — confirms the stale-build coupling is gone (fails on `main`)
- [ ] `npm test` again afterwards, exercising the `requireServerPluginFeature` gates in the "plugin present" world


client/src/__tests__/testConnectionImports.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  it.each(files.map(({ name }) => [name] as const))('finds %s to scan', (name) => {
+    expect(files.some((f) => f.name === name)).toBe(true);
+  });
+
+  it.each(files.map(({ name, contents }) => [name, contents] as const))(
+    'has no non-type import of vscode, vitest, or a value import of sessionManager in %s',
+    (_name, contents) => {
+      const importLines = contents
+        .split(/\r?\n/)
+        .filter((line) => /^\s*import\b/.test(line) && !/^\s*\/\//.test(line));
```
ericwinger: Claude: This guard only scans lines that begin with `import`, so it catches today's single-line imports but would miss a multi-line *value* import of sessionManager, e.g.

    import {
      SessionManager,
    } from '../sessionManager';

— the `} from '../sessionManager'` line never starts with `import`, so it's filtered out before the offender check runs. Since this scan is billed as the only thing standing between a `vscode` regression and green CI, consider matching against the whole file text (or normalizing multi-line imports to one line) rather than filtering line-by-line. Non-blocking.

ericwinger: Eric - This seems reasonable just based on what is claimed.

ericwinger (dismissed):
Clean consolidation, and it closes the dedup follow-up I flagged on #226: the hand-copied parseLoginFromNrs in install-server-plugin.mjs is gone, and NRS parsing now goes through the shared parseStoneNrs / netldiNameFromGemNrs in loginTypes.ts — used by the CI provisioning script, the on-demand GCI config, and the acceptance helper alike. The integration tests swapping `as unknown as ActiveSession` for a real testActiveSession login is a nice correctness upgrade on top. CI is green across all six GemStone versions. LGTM — one minor, non-blocking note below.

client/src/__tests__/testConnectionImports.test.ts
MatiasFernandez: I'll review this. Please don't merge it yet

client/src/__tests__/testConnection.ts
```diff
@@ -0,0 +1,154 @@
+// Resolves the connection details the on-demand GCI test suite
+// (`gci/gciTestConfig.ts`), the refactoring integration tests (via
+// `testActiveSession.ts`), and `client/bin/installServerPluginMain.ts` all
+// need to reach a live stone.
+//
+// Deliberately imports only `../loginTypes` (verify with a grep on this file
```
ericwinger, line 6: LOW / optional. The vscode-free discipline this module and `testActiveSession.ts` rely on (so the `tsx` CI provisioning path never pulls `vscode`) is now enforced only by this "verify with a grep" comment plus the CI plugin-install step actually loading these modules. The automated guard I commented on earlier (`testConnectionImports.test.ts`) was removed rather than fixed, so there's no test that fails if someone flips one of the `import type` lines in `testActiveSession.ts` to a value import. Since `main` never had such a guard (these files are new here), this is "no automated guard added," not a regression — but note the CI step's `vscode` stub only implements `window.createOutputChannel`, so a stray value-import could either fail loudly OR (if it only touches that one API at load) slip through silently. If you want belt-and-suspenders, a corrected guard that scans the whole file text (not line-by-line, closing the multi-line-import gap I raised) would restore it. Non-blocking.

client/src/serverPlugin/__tests__/systemUserAuth.test.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+const DONE_MESSAGE = 'Refactoring engine installed.';
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  mocks.sessionNeedsCommit.mockReturnValue(false);
+});
+
+describe('refreshWorkingSessionAfterInstall', () => {
```
ericwinger, line 42: LOW / optional. This suite covers `refreshWorkingSessionAfterInstall` thoroughly (silent-refresh, both prompt wordings with exact text, Later/Refresh, abort-throws). The module's other export, `obtainSystemUserSession`, still has no unit coverage for its branches (default-accepted / default-rejected→prompt / non-interactive miss / user-cancel / login-fail). Not a regression (it was untested before too), but now that a test file for this module exists, the non-GCI branches (interactive=false miss, cancel) are cheap to add. Non-blocking.

ericwinger (approved):
I removed one of the comments because it was only informational  (and not all that helpful).  I'm good with it!

client/src/__tests__/testConnection.ts
MatiasFernandez: Fixing this issue proved to be harder than expected, so I didn't get to clean up everything in this PR. I have an action item to see how I can fix the VS Code stub for good, and I'm going to do it in a follow-up

client/src/serverPlugin/__tests__/systemUserAuth.test.ts
MatiasFernandez: I'll add this on a follow-up

PR #334, opened 2026-07-28
Migrate GCI binding tests for production-reachable wrappers into CI [Part 2]

MatiasFernandez:
Phase 2 of moving the deep GCI binding tests into CI. Part 1 (#326) has landed, so this sits directly on `main`.

The `gci` vitest project is excluded from `npm test`, so its tests only ever ran on whatever stone a developer had locally, never across the release matrix. This moves the ones worth running everywhere into the default project as `*.integration.test.ts`. Four commits, each self-contained; their messages carry the reasoning.

## An unfinished step in a longer reorganization

`client/src/gciLibrary/` is new, and is meant to progressively absorb the layer that wraps the GCI C library. Today it holds tests and nothing else — and only the raw-binding tests, on purpose. `gciLibrary.ts` itself has not moved, which does mean that module's tests are split across two locations for now (`__tests__/gciLibrary.test.ts` still holds the unit tests). That was raised in review and is being handled as separate follow-up work. `client/src/gciLibrary/README.md` is the signpost; the shape will evolve PR by PR, so please don't read the current arrangement as the target layout.

## Scope: only wrappers production actually calls

A wrapper moved only if something outside the tests calls it, directly or through an ergonomic `gciLibrary.ts` method that itself has a caller. That keeps the migrated set free of per-symbol availability gates: several `GciTs*` functions were added after 3.6.2 and are bound through `optionalFunc`, so calling one on the matrix floor throws `"<name> is not available in this GCI library"` — a CI failure with nothing actually wrong. **Everything migrated here runs on every release in the matrix, with no skips.**

Migrated — 9 files, 85 tests:

| new file | from |
|---|---|
| `gciLibrary/__tests__/gciVersion.integration.test.ts` | `gci/gciVersion.test.ts` |
| `gciLibrary/__tests__/gciUtf8.integration.test.ts` | `gci/gciUtf8.test.ts` |
| `gciLibrary/__tests__/gciHostUtils.integration.test.ts` | `gci/gciHostUtils.test.ts` |
| `gciLibrary/__tests__/gciExecute.integration.test.ts` | `gci/gciExecute.test.ts` |
| `gciLibrary/__tests__/gciNumeric.integration.test.ts` | `gci/gciNumeric.test.ts` |
| `gciLibrary/__tests__/gciFetchStore.integration.test.ts` | `gci/gciFetchStore.test.ts` (used describes) |
| `gciLibrary/__tests__/gciObjects.integration.test.ts` | `gci/gciObjects.test.ts` (used describes) |
| `gciLibrary/__tests__/gciCompile.integration.test.ts` | `gci/gciCompile.test.ts` (`GciTsCompileMethod` only) |
| `__tests__/browserQueries.integration.test.ts` | `gci/gciBrowser.test.ts` |

`browserQueries` is the one that doesn't land under `gciLibrary/` — it covers query builders, not raw bindings, so it goes beside the unit tests that assert the Smalltalk they build.

The three mixed files are **trimmed rather than deleted**: describes covering wrappers with no caller at all stay behind in `gci/` for the dead-wrapper cleanup to decide on. `gci/gciMisc.test.ts` is untouched — every describe in it is dead.

Where the 85 comes from, since it isn't 85 moved tests: the nine source files held 96 `it` blocks, 14 of which stay behind in the three trimmed files, leaving 82 in scope. One was dropped and one was split in two, and a new `resolveSymbol` describe adds three — 85.

## Beyond the move

- Each file drops its hand-rolled login/logout for `useIntegrationTest`, so mutations ride the harness's always-abort transaction instead of the files' own cleanup; `gciCompile`'s shared `GciTestClass` fixture becomes a per-test definition.
- Well-known OOPs come from `gciConstants` instead of being redeclared per file, the `console.log` diagnostics are gone, and test names state the guarantee rather than the example.
- Assertions that couldn't fail, or didn't check what their name claimed, were tightened: `gciVersion`'s filename comparison is an `it.skipIf` rather than a conditional assertion that could pass having asserted nothing; the two `GciTsNewUtf8String` tests now assert the class they produce (`Utf8` vs `Unicode16`) instead of asserting the same thing twice; `browserQueries` asserts the exact `Compiled: <Class> >> <selector>` string, and its fixture and dictionary-index helpers fail where they break instead of downstream.
- One assertion-free test dropped: `gciObjects`' `returns the class of OOP_NIL (UndefinedObject)`, which fetched the class and then asserted nothing about it. `reversed` gave way to `asUppercase` in a `GciTsPerform` and a `GciTsPerformFetchBytes` test — it isn't a String selector on 3.6.2, which the `gci` project never caught because it isn't in CI.
- **New coverage, not a move:** a three-test `resolveSymbol` describe in `gciObjects.integration.test.ts`. The ergonomic `resolveSymbol` on `gciLibrary.ts` had no tests anywhere before this — `main`'s `gciObjects.test.ts` covered only the raw `GciTsResolveSymbolObj` wrapper it is built on, and that describe stays behind in `gci/`. Worth having: every class-OOP lookup in the extension goes through it.
- `browserQueries`' synthesized `session()` now documents why it deliberately carries no `login`. `gemNrsFor` reads `gem_host` and `netldi`, and the harness only exposes the pre-assembled `VITE_GEMSTONE_GEM_NRS`, not its parts — so supplying a `login` here would quietly build an NRS full of `undefined` rather than fail. Queries that need one (`forkGemRunning`) can't be tested through that factory.

## Not migrated

`gci/gciClassSyncUnicode.test.ts` covers production code (the sync transport) but needs a stone-version gate nothing else in this set needs: 3.6.x cannot compile non-ASCII source at all, and a non-ASCII round-trip is precisely what it guards. It stays in `gci/` until that gate lands.

### DIFF
--- a/client/src/__tests__/browserQueries.integration.test.ts
+++ b/client/src/__tests__/browserQueries.integration.test.ts
@@ -5,6 +5,7 @@ import { GciLibrary } from '../gciLibrary';
 import * as queries from '../browserQueries';
 import type { ActiveSession } from '../sessionManager';
 import { useIntegrationTest } from './useIntegrationTest';
+import { testActiveSession } from './testActiveSession';
 
 /**
  * The System Browser's queries against a real stone. browserQueries.test.ts
@@ -24,13 +25,14 @@ describe('browser queries (integration)', () => {
   });
 
   /**
-   * The live session the queries run against. It deliberately carries no
-   * `login`: the harness only exposes the pre-assembled gem NRS, not the
-   * `gem_host` and `netldi` that `gemNrsFor` reads, so a `login` here would
-   * quietly build an NRS full of `undefined`. Queries that need one —
-   * `forkGemRunning`, for instance — can't be tested through this factory.
+   * The live session the queries run against, assembled from the same test
+   * environment the harness logs in with (`testActiveSession` reads it via
+   * `resolveTestConnection`), so it carries a real `login`. Every query here
+   * is ungated and needs only a running stone, but building the full session
+   * keeps `login`-dependent queries (e.g. `forkGemRunning`) reachable from
+   * this file.
    */
-  const session = (): ActiveSession => ({ id: 1, gci, handle }) as unknown as ActiveSession;
+  const session = (): ActiveSession => testActiveSession(gci, handle);
 
   /** The one-based index `getClassNames` wants for the named dictionary. */
   const dictionaryIndexOf = (name: string): number => {
--- a/client/src/__tests__/browserQueries.integration.test.ts
+++ b/client/src/__tests__/browserQueries.integration.test.ts
@@ -7,11 +7,11 @@ vi.mock('vscode', () => ({
   },
 }));
 
-import { GciLibrary } from '../../gciLibrary';
-import { GCI_LIBRARY_PATH, STONE_NRS, GEM_NRS, GS_USER, GS_PASSWORD } from './gciTestConfig';
-import { ActiveSession } from '../../sessionManager';
-import { GemStoneLogin } from '../../loginTypes';
-import * as queries from '../../browserQueries';
+import { GciLibrary } from '../gciLibrary';
+import { GCI_LIBRARY_PATH, STONE_NRS, GEM_NRS, GS_USER, GS_PASSWORD } from './gci/gciTestConfig';
+import { ActiveSession } from '../sessionManager';
+import { GemStoneLogin } from '../loginTypes';
+import * as queries from '../browserQueries';
 
 describe('Browser Queries (integration)', () => {
   let gci: GciLibrary;
--- a/client/src/__tests__/browserQueries.integration.test.ts
+++ b/client/src/__tests__/browserQueries.integration.test.ts
@@ -1,132 +1,126 @@
-import { describe, it, expect, beforeAll, afterAll, vi } from 'vitest';
-
-// Mock vscode since browserQueries → gciLog → vscode
-vi.mock('vscode', () => ({
-  window: {
-    createOutputChannel: () => ({ appendLine: () => {} }),
-  },
-}));
+import { describe, it, expect, vi } from 'vitest';
+vi.mock('vscode', () => import('../__mocks__/vscode'));
 
 import { GciLibrary } from '../gciLibrary';
-import { GCI_LIBRARY_PATH, STONE_NRS, GEM_NRS, GS_USER, GS_PASSWORD } from './gci/gciTestConfig';
-import { ActiveSession } from '../sessionManager';
-import { GemStoneLogin } from '../loginTypes';
 import * as queries from '../browserQueries';
+import type { ActiveSession } from '../sessionManager';
+import { useIntegrationTest } from './useIntegrationTest';
 
-describe('Browser Queries (integration)', () => {
+/**
+ * The System Browser's queries against a real stone. browserQueries.test.ts
+ * asserts the Smalltalk each one builds; these assert what the stone actually
+ * answers, which is what catches a selector that no longer exists or a
+ * result shape the parser mis-reads.
+ *
+ * Ungated: every query here needs only a running stone.
+ */
+describe('browser queries (integration)', () => {
   let gci: GciLibrary;
-  let session: ActiveSession;
+  let handle: unknown;
 
-  beforeAll(() => {
-    gci = new GciLibrary(GCI_LIBRARY_PATH);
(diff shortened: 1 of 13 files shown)


client/src/gciLibrary/__tests__/gciHostUtils.integration.test.ts
```diff
@@ -0,0 +1,73 @@
...
+import { useIntegrationTest } from '../../__tests__/useIntegrationTest';
+
+/**
+ * The GCI library's host-side utilities: allocation, clock, sleep and
+ * timestamp formatting. None of them take a session, but they are real native
+ * calls, so they run against the loaded library under the harness.
+ *
+ * GciHostCallDebuggerMsg is deliberately not covered — it blocks for 60
+ * seconds waiting for a C debugger to attach. That its binding loads at all is
+ * covered by gciOptionalFunctions.test.ts.
```
npapagna: 🤖 Claude finding

 The new doc comment says `GciHostCallDebuggerMsg`'s binding "loads at all is covered by gciOptionalFunctions.test.ts". That file calls `vi.mock('koffi', …)` and returns a stub for every signature not in its `OPTIONAL_FUNCTIONS` list, so it never opens the real `libgcits` and cannot verify any symbol resolves in it. The real coverage is the one the deleted comment named correctly: gciLibrary.ts:377 binds it via this.lib.func (non-optional), so any suite constructing a real `GciLibrary` fails if it's absent. A maintainer trusting this comment could make that binding optional and believe a mocked test still guards it.

client/src/__tests__/browserQueries.integration.test.ts
```diff
@@ -0,0 +1,170 @@
...
+    handle = testContext.session;
+  });
+
+  /**
+   * The live session the queries run against. It deliberately carries no
+   * `login`: the harness only exposes the pre-assembled gem NRS, not the
+   * `gem_host` and `netldi` that `gemNrsFor` reads, so a `login` here would
+   * quietly build an NRS full of `undefined`. Queries that need one —
+   * `forkGemRunning`, for instance — can't be tested through this factory.
+   */
```
npapagna, line 34: 🤖 Claude finding

The session factory's justification is wrong. It claims "the harness only exposes the pre-assembled gem NRS, not the gem_host and netldi that gemNrsFor reads", but client/src/__tests__/testActiveSession.ts builds a full ActiveSession from resolveTestConnection(), which parses netldi out of the Gem NRS and gem_host/stone out of the Stone NRS (testConnection.ts:100-129) and adds a real stoneVersion. Two consequences: forkGemRunning is declared untestable when it isn't, and because stoneVersion is undefined, the documented gating helper requireServerPluginFeature(feature, ctx, session()) — which version-gates on stoneVersion — silently takes the wrong branch if anyone adds a gated test to this file.

client/src/gciLibrary/__tests__/gciObjects.integration.test.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  describe('GciTsFetchClass', () => {
+    it('returns the class of an object', () => {
+      const { result: oop } = gci.GciTsNewString(session, 'class test');
+
+      expect(gci.GciTsFetchClass(session, oop).result).toBe(OOP_CLASS_STRING);
+    });
+  });
+
+  describe('resolveSymbol', () => {
```
npapagna: 🤖 Claude finding

This block that duplicates gciLibrary.test.ts:521-533 and covers an ergonomic-layer method, which the new client/src/gciLibrary/README.md explicitly says has not moved  so one method's  contract is now split across two files. And gci/gciCompile.test.ts / gciObjects.test.ts / gciFetchStore.test.ts keep filenames that no longer describe their trimmed contents.

npapagna (approved):
Gave it a pass, LGTM! Left some comments.

client/src/__tests__/browserQueries.integration.test.ts
MatiasFernandez: Fixed

client/src/gciLibrary/__tests__/gciHostUtils.integration.test.ts
MatiasFernandez: Fixed

client/src/gciLibrary/__tests__/gciObjects.integration.test.ts
MatiasFernandez: Fixed the duplicate test, but kept the original names for the test files within the gci folder. The whole goal of this project is to remove those files because they are being migrated to a new place, so I'll avoid renaming them in this PR; they are being removed soon.

PR #341, opened 2026-07-29
Fix #340 findings #1/#3/#4: disable structural rename checkboxes, share the RB engine-availability gate

ericwinger:
Resolves the remaining code-review findings from #340 (the automated `/code-review` pass on the #191 review).

## Findings addressed
Each item restates the problem #340 flagged, then the fix.

- **#1 — instance-variable rename let the class-definition change be deselected independently.** #340 flagged that the ivar-rename preview allowed unchecking the `classDefinitionEdit` card while keeping the method-recompile cards, so applying would recompile the method bodies against the new ivar name *without ever recompiling the class definition* — every recompile then fails, or binds to a stray Undeclared global. **Fix:** the `classDefinitionEdit` card now renders `checked disabled` (structural/required) so it can't be deselected (a disabled box is never reported as deselected to the engine).
- **#3 — method rename let an individual implementor's rename be deselected while senders recompiled.** #340 flagged that unchecking one implementor's `methodRename` while its sender recompiles stayed selected rewrote every call site to the new selector but left that implementor under the *old* selector — silently changing dispatch (instances fall through to an inherited method, or DNU). **Fix:** an implementor's `methodRename` card now renders `checked disabled`; sender `methodRecompile` rows stay individually deselectable.
- **#4 — install-gate logic was duplicated instead of using the shared helper.** #340 flagged that the "engine not installed → offer to install → re-check" block was copy-pasted into `renameInstVarNamed` and `renameMethodNamed` rather than calling the shared `ensureRbSupport` helper — a maintenance hazard, since a future gating fix (e.g. finding #2) would have to be repeated in every copy. **Fix:** both flows now route through `ensureRbSupport`.

The disable approach mirrors the class-rename panel, which already marks `classRename` / `classReparent` always-applied/disabled for exactly this reason. Finding **#2** (`ensureRbSupport` returning `true` without confirming the engine actually loaded) was already fixed in `40077c0d`.

## Tests
+8 unit tests: instVar/method panel HTML (structural row is `checked disabled`, siblings stay selectable) and a new `explorerRenameGate` controller test (the ivar/method flows gate through the shared helper).

## Verification
- `npm run lint` / `format:check` / `compile` clean.
- Full `npm test` green in **both** RB-engine states — engine **not installed** (engine-gated tests skip-guard) and engine **installed** (they exercise green): client 4159→4211 passing, server 322, mcp 92.

Refs #340.


npapagna (commented):
🤖 This review was generated by Claude (via the `/code-review` skill), then discussed and confirmed one finding at a time with @npapagna  before posting. Non-blocking, flagging for awareness or a follow-up.

## 1. Reuse the `isStructuralChange` pattern instead of inlining the `kind` check

`renameClassPreview.ts` already exports a named, testable predicate for this:

```ts
// renameClassPreview.ts:204
export function isStructuralChange(change: ClassRenameChange): boolean { ... }

// renameClassPanelHtml.ts:67
const structural = isStructuralChange(change);
```

This PR's two new call sites reinvent it inline instead:

```ts
// renameInstVarPanelHtml.ts:39
const structural = change.kind === 'classDefinitionEdit';

// renameMethodPanelHtml.ts:62
const structural = change.kind === 'methodRename';
```

Functionally identical today, but it breaks the established convention: the class-rename panel's "what counts as structural" is a named, exported predicate living next to its type definition; these two are anonymous one-liners buried in the rendering function. Consider adding an equivalent `isStructuralChange` export to `renameInstVarPreview.ts` and `renameMethodPreview.ts`.

## 2. Test names in `explorerRenameGate.test.ts` mirror internal method names

```ts
it('renameInstVarNamed aborts (no rename) when the engine is absent and the install offer is dismissed', ...)
it('renameInstVarNamed proceeds past the gate when the engine is present', ...)
it('renameMethodNamed aborts (no rename) when the engine is absent and the install offer is dismissed', ...)
```

This repo's own `.claude/rules/tests.md` calls this out directly as the anti-pattern to avoid: test names should describe the scenario in plain language, not mirror the internal method/field name being called. If `renameInstVarNamed`/`renameMethodNamed` are ever renamed or split, these titles rot even though the underlying guarantee (the shared install gate fires and blocks/passes appropriately) hasn't changed. Something like "renaming an instance variable aborts when the refactoring engine is absent and the install offer is declined" would hold up better.

## 3. `isRename` and `structural` computed separately in `renameMethodPanelHtml.ts`

```ts
const isRename =
  change.kind === 'methodRename' &&
  !!change.newSelector &&
  change.newSelector !== change.selector;
...
const structural = change.kind === 'methodRename';
```

Both gate on `change.kind === 'methodRename'`, but as two independent booleans 14 lines apart rather than one derived from the other. Not an active bug today since every current `methodRename` change satisfies both, but a future new `kind` value or an edge case could make a card show the plain label while still forced-disabled (or vice versa), with no test tying the two together.

## 4. Dead guard in `gemstoneExplorer.ts:1495` (informational only, not a regression)

```ts
if (!selectedIds || selectedIds.length === 0) return false;
```

Since the `classDefinitionEdit` checkbox is now always checked+disabled, `selectedIds` can never be empty via the panel, so this specific branch is unreachable. Worth noting this isn't a new capability: a user could already leave the class-def checkbox checked and uncheck every method row before this PR, applying a class-def-only rename with zero method recompiles — that path is unchanged, and it mirrors the existing class-rename panel's behavior. Just flagging the guard as a no-op now, not something to fix.

npapagna (dismissed):
Thanks a lot @ericwinger!                                                          
                                                                                                                                                                 
I ran it by hand today and found a few things worth flagging before this goes wider. 

Method rename worked well. But renaming an instance variable on its own, without touching the methods that use it, ended up deleting those methods. The lightbulb also stopped showing up at one point (had to use the context menu instead), and a method on a freshly-created class didn't show up in the rename preview at all. Recordings attached for both.

https://github.com/user-attachments/assets/73550ef7-7f8d-4b9b-8609-336f097c65af

https://github.com/user-attachments/assets/42fb44cb-1f86-40f6-a446-1042a85a1ba6                                   

I checked our test coverage too: we have a live-stone integration test for the preview step of instance-variable rename, but nothing for the apply step, where the actual recompile happens. That's the gap that let the delete-on-rename bug through.

Since this one deletes real methods, would it make sense to add apply-path integration tests and another manual pass through these scenarios before wider review?

Losing someone's work is worth double-checking for ahead of time.                                                                                      
                                                                                                                                                                 Thanks again!

client/src/refactoring/renameInstVarPanelHtml.ts
```diff
@@ -31,11 +31,17 @@ function renderCard(change: RenameChange): string {
         ? `<span class="badge">${escapeHtml(change.category)}</span>`
         : '';
   const diff = renderDiff(lineDiff(change.oldSource, change.newSource));
+  // A structural change is checked and DISABLED so it can't be deselected; see
+  // isStructuralChange for why. (Mirrors the class-rename panel.)
+  const structural = isStructuralChange(change);
```
ericwinger, line 36: **#1** — done. `renameInstVarPreview.ts` now exports `isStructuralChange(change)` next to the `RenameChange` type, mirroring `renameClassPreview.ts:204`.

Placement mattered more than deduplication: the five-line "why it can't be deselected" rationale was buried in an HTML generator and now sits on the predicate. Two unit tests on the predicate; existing panel tests unchanged.

client/src/refactoring/renameMethodPanelHtml.ts
```diff
@@ -39,23 +40,29 @@ function renderCard(change: MethodRenameChange): string {
...
-  // A methodRename REMOVES the old-selector method and ADDS the new-selector one;
-  // show both in the collapsed header. A pure argument reorder (same selector,
-  // recompiled in place) and a sender render as a plain label.
-  const isRename =
-    change.kind === 'methodRename' &&
-    !!change.newSelector &&
-    change.newSelector !== change.selector;
+  // A structural change is checked and DISABLED so it can't be deselected; see
+  // isStructuralChange for why. (Mirrors the class-rename panel.)
+  const structural = isStructuralChange(change);
```
ericwinger, line 45: **#1** — same fix: `renameMethodPreview.ts` exports `isStructuralChange(change) => kind === 'methodRename'`. All three modules use the same name rather than prefixed variants; each panel imports only its own preview module, so no collision. Happy to switch to the `changeLabel` / `methodChangeLabel` / `classChangeLabel` precedent if you'd rather.

**#3** — `structural` is computed first now and `isRename` derives from it, so they can't diverge.

Straight about the accompanying test: it feeds a `methodRename` with a null `newSelector` and checks the card shows a plain label while staying `checked disabled`. That holds under the old code too, so it pins the behaviour rather than proving the change. Exercising a real disagreement would need a `kind` value that doesn't exist yet — the protection is the derivation itself.

(#3 is here because the `const isRename =` line you quoted was outside the diff at the time.)

client/src/__tests__/explorerRenameGate.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  return new ExplorerController(sessionManager);
+}
+
+const infoMsg = vi.mocked(vscode.window.showInformationMessage);
+const inputBox = vi.mocked(vscode.window.showInputBox);
+
+beforeEach(() => vi.clearAllMocks());
+
+describe('the RB engine-availability gate on the rename flows (issue #340 #4)', () => {
+  it('cancels an instance-variable rename before prompting for a new name when the engine is missing and its install is declined', async () => {
```
ericwinger, line 32: **#2** — agreed, and the rule you cited was right. All three renamed, plus the `describe`, which named the helper too (now `the RB engine-availability gate on the rename flows`).

Swept the panel tests this PR added, as you suggested: one more hit in `renameInstVarPanel.test.ts` (`classDefinitionEdit` → `class-definition edit`). No assertions changed.

client/src/gemstoneExplorer.ts
```diff
@@ -1438,21 +1460,9 @@ export class ExplorerController {
...
-          "isn't loaded in this stone yet.",
-        LOAD,
-      );
-      if (choice !== LOAD) return false;
-      await vscode.commands.executeCommand('gemstone.installServerSupport');
-      if (!this.session()?.rbSupportAvailable) return false;
-    }
+    // The engine ships as an optional, separately-installed payload; gate through
+    // the shared helper so the install-then-re-check logic lives in one place.
+    if (!(await this.ensureRbSupport('Renaming an instance variable'))) return false;
```
ericwinger, line 1608: **#4** — keeping the guard deliberately, not as dead code.

It's unreachable today only because the panel can't produce an empty selection. That's a UI invariant, not a controller one — if the checkbox rendering changes, or another caller drives this flow, the controller shouldn't already depend on the HTML having been correct.

Agreed on your broader point: nothing here was a new capability. A class-def-only rename with zero method recompiles was reachable before this PR and still is, matching the class-rename panel.

gs-src/refactoring/engine/GsRenameInstanceVariableRefactoring.class.st
```diff
@@ -237,3 +237,262 @@ GsRenameInstanceVariableRefactoring >> stageMethodRecompilesInto: aChangeSet [
 						oldSource: oldSrc
 						newSource: newSrc]]]]
 ]
+
+{ #category : 'applying' }
+GsRenameInstanceVariableRefactoring class >> applyForToken: token deselected: deselectedIds [
```
ericwinger, line 242: **"renaming an instance variable … ended up deleting those methods"** — confirmed, worse than the video showed, and pre-existing in `main` rather than introduced by this PR.

Renaming an ivar reshapes the class, and in GemStone a reshape creates a new class version whose method dictionary starts **empty**. Apply lived in the client: it executed the staged class-definition string, then recompiled only the methods in the change set — which contains exactly the methods that *access* the variable. Everything else was destroyed, including class-side methods, which can never appear in a change set at all.

It fired **even with every box checked**. Unchecking wasn't the trigger, just how you noticed. Reproduced: a class with `bump` (uses the ivar), `plain` (doesn't) and a class-side `make` staged `[methodRecompile:#bump, classDefinitionEdit]`; after apply both selector sets were empty.

`main` already documented the rule this broke — `GsRenameClassRefactoring>>copyMethodsFrom:to:`: *"A new class version starts with an empty method dictionary, so this is what carries the behaviour forward."* Rename-class does that, rename-classVar avoids versioning entirely, rename-method never reshapes. Ivar rename was the only one that reshaped and didn't carry methods forward.

Apply now runs here, in one transaction, so the whole method dictionary can come along.

gs-src/refactoring/engine/GsRenameInstanceVariableRefactoring.class.st
```diff
@@ -237,3 +237,262 @@ GsRenameInstanceVariableRefactoring >> stageMethodRecompilesInto: aChangeSet [
...
+		classInstVars: (old class instVarNames collect: [:e | e asString])
+		poolDictionaries: old sharedPools
+		inDictionary: (self dictObjectFor: old)
+		inClassHistory: old classHistory
+		description: ([old commentForFileout] on: Error do: [:e | ''])
+		options: #()
+]
+
+{ #category : 'private - applying' }
+GsRenameInstanceVariableRefactoring >> copyMethodsFrom: old to: new deselected: deselected into: failures [
```
ericwinger, line 388: The carry-forward itself, both sides. Accessing methods compile from their staged rewritten source; every other method compiles from its existing source unchanged, keeping its category. Subclasses are re-versioned top-down and get the same treatment, since reshaping a superclass re-versions the whole subtree.

A method disappears **only** if you deselected it. Since that deletes it rather than leaving it alone, the client now shows a modal naming the casualties and requiring confirmation — deleting a method can be exactly what you want, but not as a surprise.

client/src/refactoring/queries/previewRenameInstVar.ts
```diff
@@ -23,6 +34,31 @@ cls isNil ifTrue: [^ 'Class not found: ${escapeString(className)}'].
 (GsRenameInstanceVariableRefactoring
   class: cls
   renameInstVar: '${escapeString(oldName)}'
-  to: '${escapeString(newName)}') previewJsonString`;
+  to: '${escapeString(newName)}') startPreviewToken: '${escapeString(token)}'`;
```
ericwinger, line 37: Why the preview is now token-addressed: apply **has** to run server-side. Only the engine can copy the whole method dictionary onto the new class version — any client that replays the staged changes destroys whatever the change set didn't mention. The token is what lets the client ask the engine to apply instead of doing it itself.

`applyRename` and its `compileClassDefinition` call are gone.

gs-src/refactoring/tests/GsRenameInstanceVariableRefactoringTest.class.st
```diff
@@ -241,3 +249,103 @@ GsRenameInstanceVariableRefactoringTest >> testStagesOneChangePerAffectedMethodP
...
+	ref := GsRenameInstanceVariableRefactoring
+		class: self baseFixture
+		renameInstVar: 'count'
+		to: aName.
+	ref changeSet.
+	^ref applyDeselected: idArray
+]
+
+{ #category : 'tests - apply' }
+GsRenameInstanceVariableRefactoringTest >> testApplyKeepsEveryMethodOnBothSidesOfTheDefiningClass [
```
ericwinger, line 277: **"nothing for the apply step, where the actual recompile happens"** — exactly right, and that's what let it through. Five SUnit apply tests now: method survival on both sides of the defining class and of the subclass, source rewriting, category preservation, and deselecting exactly one method.

The fixture gained methods no change set mentions — `getOther`, `subLabel`, and class-side `makeOne` / `describe` — because those are precisely what the bug destroyed and what nothing was asserting. Green on 3.6.2 and 3.7.5 (16/16 here, 366/366 across the engine suite on each).

client/src/refactoring/__tests__/refactoring.integration.test.ts
```diff
@@ -146,8 +183,73 @@ describe('rename instance variable (integration)', () => {
...
+  });
+
+  // ---- apply path -------------------------------------------------------------
+  // The preview being right says nothing about what the stone looks like after
+  // Apply. Renaming reshapes the class, so every class in the subtree is
+  // re-versioned onto an EMPTY method dictionary; only what the engine copies
+  // forward survives. `label`, `makeOne` and `subLabel` appear in no change set,
+  // which is exactly why they need asserting here.
+
+  it('keeps every method of the defining class, both sides, after applying', (ctx) => {
```
ericwinger, line 199: The same guarantees through the real client path, replacing preview-only coverage — 4 tests, green on both 3.6.2 and 3.7.5.

Worth noting the two layers aren't redundant: the SUnit tests caught a helper returning a change object instead of its source, and these caught a 3.6.2 Unicode string-comparison trap (error 2718) that cannot occur in-image. Neither would have found the other's bug.

**Your third report — the method on a freshly-created class missing from the preview — was real and is fixed.** The method had landed on a superseded class version; every rename under the old apply left stale versions behind, so the preview could resolve a different version than the one the method was compiled into. Doing the re-versioning once, in the engine, removed the cause. Verified by hand.

.claude/rules/refactoring-tests.md
```diff
@@ -0,0 +1,45 @@
...
+  state of the stone, not the change set that was staged.
+- **[GCI on-demand]** in `client/src/__tests__/gci/` — drive apply through the client path end to end.
+- Both boundaries: 3.6.2 **and** 3.7.5.
+
+## What an apply test must assert
+
+"The right changes were staged" (preview) and "the stone ended up right" (apply) are different
+questions. At minimum:
+
+1. **Nothing was silently lost.** Compare the full selector set before and after, **both sides**
```
ericwinger, line 24: Generalising the lesson so the next refactoring can't repeat it: a path-scoped rule requiring an apply-path test, not just a preview one, and spelling out what such a test must assert.

The first item is the one that would have caught this: compare the full selector set before and after, on **both** sides, naming survivors *including methods the refactoring has no reason to touch* — no change set mentions those, so nothing else catches their loss.

Also did another manual pass, 22 scenarios including both your recordings.

ericwinger (commented):
All four findings from your first review are addressed, and all three from the manual pass.
  The instance-variable rename data loss turned out to be pre-existing in main and broader than
  the video showed — it destroyed unrelated and class-side methods too, even with every box
  checked. Apply now runs in the engine and carries the whole method dictionary onto the new
  class version. Apply-path tests added at both layers, green on 3.6.2 and 3.7.5, plus a
  standing rule so the next refactoring can't repeat it. Your freshly-created-method report was
  real and is fixed. Details inline.

  The lightbulb item is the one thing not covered here — If my previous comment didn't address it, let's file a new issue. Already a lot in here and hard to track it all. 

  I added the last paragraph given your decision to split it out — otherwise a reader counts three reports and finds two
  answered.

  Everything else is unchanged. Pending review 4823509279, 10 comments, all anchored on the pushed head.

client/src/refactoring/renameInstVarPanelHtml.ts
jgfoster: Looks good.

jgfoster (approved):
It appears to me that the issues have been resolved.

PR #343, opened 2026-07-29
V2/V3/V5 instance-variable structure refactorings (push up/down ivar, temp→ivar)

ericwinger:
## Summary

Adds the **V2/V3/V5 instance-variable structure refactorings** to the server-side RB engine + the Explorer/editor UI, on top of the existing rename/extract/move/push families (grail #62).

- **V2 — Push Up Instance Variable** (subclass → superclass)
- **V3 — Push Down Instance Variable** (superclass → every immediate subclass)
- **V5 — Convert Temporary to Instance Variable** (method temp → ivar)

All three are one parametrized engine (`GsInstVarStructureRefactoring`). Because GemStone has no `addInstVarName:`/`removeInstVarName:`, changing a class's ivar list means creating a new class **version** — so the apply reuses R3's `newVersionOf:` machinery: for each affected class top-down, create a new version with the computed own-ivar list, copy methods forward, and re-parent the subtree. All-or-nothing; **never commits** unless an opt-in option is chosen.

## What it adds

- **Accessor move (V2/V3):** pushing an ivar also carries its *simple* `^ivar` getter / `ivar := arg` setter (strict AST detection — nothing with extra logic is moved; never overwrites an existing target method). Shown as move-out/move-in rows in the preview.
- **Instance & history options** (class-rename parity, opt-in, default OFF, the only committing path): **Migrate existing instances** and **Remove old versions from class history**.
- **UI:** inline ▲/▼ push arrows on ivar rows (▼ hidden when the class has no subclasses); the moved ivar is selected after a push; the method-editor Refactor… menu drives V5. Decline messages name the offending method(s).

## Testing

- **Engine SUnit: 33/33 on BOTH 3.6.2 and 3.7.5** (topaz, non-committing).
- **Client unit:** query builders, preview parsing (committed/migratedFailures), panel option flow, the shared-panel-script option collection, and a full `instVarStructureCommand.test.ts`. Compile / lint / format clean.
- **GCI integration + on-demand e2e** (gated on the engine): push up/down, convert-temp, accessor-move, and the committing migrate path (validated live).

Includes fixes from two parallel code reviews: a class-side (`meta:true`) convert-temp corruption (now declined), a swallowed apply-`error` in the client (now surfaced), and a multi-class migrate bug (`migrateInstancesTo:` needs a clean transaction — now commits per class).

## Notes

- The long local PR-gate matrix (`test-ci-local.sh`) was **not** run for this PR — relying on the CI health check. Engine SUnit both boundaries + client unit/compile/lint/format were run locally.
- `GsRenameClassRefactoring` (R3, already in main) shares the multi-class migrate + prune-with-instances weaknesses fixed here for V235; tracked as a separate follow-up.


client/src/refactoring/instVarStructureCommand.ts
```diff
@@ -0,0 +1,220 @@
+/**
+ * The instance-variable structure refactorings (V2 push up, V3 push down, V5 convert
+ * temporary to instance variable). V2/V3 are driven from the Explorer's instance-variable
+ * row; V5 from a method source editor (cursor on the temporary). All three share one
+ * server-side engine and one preview flow: run a pre-flight (viable? why not?), preview the
+ * change set (class-definition edits + descendant reparents [+ a method recompile for V5]),
+ * apply it server-side (new class versions, no commit), then let the caller refresh.
+ *
+ * Nothing is ever committed; the user commits explicitly. Existing instances keep their
+ * prior class version and are NOT migrated (standard GemStone class evolution).
```
MatiasFernandez: 🚫 This header says nothing is ever committed and that existing instances are never migrated, but the same PR adds the `migrateInstances` / `removeOldFromHistory` apply options — and both commit (the panel even labels them `(commits)`). A future reader, or an agent, trusting this doc would assume anything going through `runInstVarStructure` can't touch the database. Could you reword it so the default path is described as the no-commit one, and the two opt-in options are called out as committing?

gs-src/refactoring/engine/GsInstVarStructureRefactoring.class.st
```diff
@@ -0,0 +1,872 @@
...
+	"Record the SIMPLE accessors of varName on srcClass to move alongside the ivar: remove
+	 each from srcClass and add it to the target class(es). Never overwrites an existing method
+	 on a target -- if a target already implements the selector, that target is skipped. On
+	 push-up (the ivar stays reachable via inheritance) an accessor is left in place unless the
+	 superclass can receive it; on push-down (the ivar leaves srcClass) the accessor must be
+	 removed from srcClass regardless, and added to each subclass that doesn't already define it."
+	(self simpleAccessorsOf: srcClass forIvar: varName) do: [:a |
+		| sel src cat targets |
+		sel := a at: 1. src := a at: 2. cat := a at: 3.
+		targets := targetClasses reject: [:t | t includesSelector: sel].
```
MatiasFernandez: 💭 `includesSelector:` only looks at the target's own method dictionary, so the "Never overwrites an existing method on a target" guarantee in the comment above doesn't cover an *inherited* implementation.

Concrete push-up case: `GrandParent>>pushable` computes something, `Parent` doesn't implement it, `Child` has `pushable ^pushable`. Pushing `pushable` up from `Child` with `moveAccessors` on installs `pushable ^pushable` onto `Parent`, which then silently shadows `GrandParent`'s version for `Parent` and its whole subtree — the accessor didn't overwrite anything in `Parent`'s dictionary, but it did change behaviour.

Should the guard consult the inherited chain (`canUnderstand:` / `whichClassIncludesSelector:`) instead? Or, if shadowing is deliberate here, is it worth narrowing the comment so the next reader knows the check is own-dictionary-only?

gs-src/refactoring/engine/GsInstVarStructureRefactoring.class.st
```diff
@@ -0,0 +1,872 @@
...
+		(all at: i) jsonOn: ws.
+		i := i + 1].
+	ws nextPut: $].
+	^'{"changes":', ws contents,
+	  ',"nextOffset":', i printString,
+	  ',"done":', (i > all size) printString, '}'
+]
+
+{ #category : 'applying' }
+GsInstVarStructureRefactoring >> applyDeselected: deselectedIds [
```
MatiasFernandez, line 641: 💭 The comment calls this ALL-OR-NOTHING, but the loop keeps going after a failure — so a mid-way error leaves the session holding a half-reversioned hierarchy: some classes on new versions parented correctly, the rest still on the old chain. Nothing commits, so it's recoverable, but only by aborting — and neither this doc nor the client says so. `runInstVarStructure` just shows `Change failed: <class>: <error>` and returns false.

`GsRenameClassRefactoring>>applyDeselected:` spells the contract out ("left uncommitted for the user to inspect and abort"). Should this one carry the same note, and should the client's failure message tell the user their transaction now needs an abort?

client/src/refactoring/queries/previewInstVarStructure.ts
```diff
@@ -0,0 +1,124 @@
...
+  maxBytes: number,
+  dict?: number | string,
+  extra?: ConvertTempArgs,
+  moveAccessors = false,
+): Promise<string> {
+  const accessorStmt =
+    moveAccessors && (op === 'pushUp' || op === 'pushDown') ? 'ref moveAccessors: true.\n' : '';
+  const code = `| cls ref |
+cls := ${classLookupExpr(className, dict)}.
+cls isNil ifTrue: [^ 'Class not found: ${escapeString(className)}'].
```
MatiasFernandez: 💭 `analyzeInstVarStructure` answers a proper `{"decline":"Class not found: …"}` envelope, but this one answers a bare `'Class not found: X'` — which `parseStartPreview` hands straight to `JSON.parse`, so the user sees `Preview failed: Unexpected token 'C'…` rather than the actual reason.

It should be unreachable today because the pre-flight declines first, but a class removed between the two sends, or a future caller that skips the pre-flight, would land here. Worth answering the same decline shape so the message survives?

client/src/refactoring/queries/previewInstVarStructure.ts
```diff
@@ -0,0 +1,124 @@
...
+export function startInstVarStructurePreview(
+  execute: AsyncQueryExecutor,
+  op: IvarStructureOp,
+  className: string,
+  varName: string,
+  token: string,
+  maxBytes: number,
+  dict?: number | string,
+  extra?: ConvertTempArgs,
+  moveAccessors = false,
```
MatiasFernandez, line 81: 💭 Nine positional parameters with three optional ones at the tail makes the call sites hard to read — the GCI e2e test ends up with `…, PREVIEW_PAGE_BYTES, undefined, undefined, true` and the integration helper threads the same shape through. Does it make sense to group the tail into one object (`{ dict, extra, moveAccessors }`)? It'd also let the command tests name what they're asserting rather than indexing into the call (see my other comment on `instVarStructureCommand.test.ts`).

client/src/refactoring/__tests__/instVarStructureCommand.test.ts
```diff
@@ -0,0 +1,211 @@
...
+  it('always opts into moving simple accessors on a push', async () => {
+    vi.mocked(queries.analyzeInstVarStructure).mockResolvedValue(analysis());
+    vi.mocked(queries.startInstVarStructurePreview).mockResolvedValue(startEnvelope());
+    vi.mocked(showInstVarStructurePanel).mockResolvedValue(applyResult());
+
+    await pushUp();
+
+    // moveAccessors is the 7th arg of analyze and the 9th of start.
+    expect(vi.mocked(queries.analyzeInstVarStructure).mock.calls[0][6]).toBe(true);
+    expect(vi.mocked(queries.startInstVarStructurePreview).mock.calls[0][8]).toBe(true);
```
MatiasFernandez: 💭 These index-based assertions are only correct as long as nobody inserts a parameter — add one to `analyzeInstVarStructure` ahead of `moveAccessors` and `calls[0][6]` silently starts asserting on the wrong argument while staying green, which is the opposite of what this test is here to pin.

Could we assert on the full argument list instead (a `toHaveBeenCalledWith(...)` naming every arg), so a signature change fails loudly rather than quietly changing the test's meaning? An options object on the query builders would make this read even better.

client/src/refactoring/queries/previewInstVarStructure.ts
```diff
@@ -0,0 +1,124 @@
...
+  offset: number,
+  maxBytes: number,
+): Promise<string> {
+  const code =
+    `${ENGINE} pageForToken: '${escapeString(token)}' ` + `from: ${offset} maxBytes: ${maxBytes}`;
+  return execute(`pageIvarPreview(${token} @ ${offset})`, code);
+}
+
+/** Apply a started preview server-side (create new class versions, copy methods forward,
+ *  re-parent the subtree), WITHOUT committing. The refactoring is all-or-nothing, so the
```
MatiasFernandez: The "WITHOUT committing" here isn't accurate any more now that `migrateInstances` / `removeOldFromHistory` are parameters — both commit. Same wording as the module doc in `instVarStructureCommand.ts`.

client/src/refactoring/instVarStructurePreview.ts
```diff
@@ -0,0 +1,216 @@
...
+  | 'classDefinitionEdit'
+  | 'classReparent'
+  | 'methodRecompile'
+  // V2/V3 "move accessors" opt-in: a simple accessor removed from its source class and added
+  // to the target class(es) alongside the ivar declaration.
+  | 'methodRemove'
+  | 'methodAdd';
+
+/** One staged change: a class-definition edit, a descendant reparent, or (V5) a method
+ *  recompile. `oldSource`/`newSource` are the before/after for the diff. */
```
MatiasFernandez, line 28: nit: This (and the module comment above it) enumerates the kinds as class-definition edit / descendant reparent / V5 method recompile, but `IvarChangeKind` also carries `methodRemove` and `methodAdd` for the accessor move. Worth listing them so the doc matches the type?

client/src/refactoring/instVarStructureCommand.ts
ericwinger: Good catch — reworded in `3c25fa61`: the default path is now described as the no-commit one, and the `migrateInstances` / `removeOldFromHistory` opt-ins are explicitly called out as committing (they do write to the database).

gs-src/refactoring/engine/GsInstVarStructureRefactoring.class.st
ericwinger: Fixed in `3c25fa61`: on push-up the guard now rejects a target when it *or any ancestor* already implements the selector — via a new `class:orAncestorImplements:` helper (a plain `superclass` + `includesSelector:` walk, so no version-specific primitive) — so a moved accessor can't silently shadow an inherited method. Push-down keeps the own-dictionary check. Added `testPushUpLeavesAccessorWhenSuperclassInheritsIt`.

gs-src/refactoring/engine/GsInstVarStructureRefactoring.class.st
ericwinger: Addressed in `3c25fa61`: the `applyDeselected:` comment now spells out that a mid-way failure leaves an uncommitted, half-reversioned hierarchy the caller must **abort** (mirroring `GsRenameClassRefactoring>>applyDeselected:`), and the client's failure toast now tells the user to abort the transaction.

client/src/refactoring/queries/previewInstVarStructure.ts
ericwinger: Fixed in `3c25fa61`: the start builder now answers the same `{"decline":…}` envelope on class-not-found, and `parseStartPreview` surfaces a token-less decline through the banner instead of feeding a bare string to `JSON.parse`. Added tests. Agreed it's unreachable in the normal flow (the pre-flight declines first), so this is purely defensive.

client/src/refactoring/queries/previewInstVarStructure.ts
ericwinger: Leaving the positional signatures as-is for now to keep this PR focused — it's a readability nit. I did take the related point on the test below by switching it to `toHaveBeenCalledWith(...)`. Happy to do the options-object refactor as a follow-up.

client/src/refactoring/__tests__/instVarStructureCommand.test.ts
ericwinger: Fixed in `3c25fa61`: the test now asserts the full argument list with `toHaveBeenCalledWith(...)` instead of `calls[0][6]`, so inserting or reordering a parameter fails loudly rather than silently asserting on the wrong argument.

client/src/refactoring/queries/previewInstVarStructure.ts
ericwinger: Fixed in `3c25fa61`: reworded — the default path doesn't commit; passing `migrateInstances` / `removeOldFromHistory` makes the engine commit. Same wording fix as the module header above.

client/src/refactoring/instVarStructurePreview.ts
ericwinger: Fixed in `3c25fa61`: the module doc now lists `methodRemove` / `methodAdd` (the accessor move) alongside the other change kinds.

client/src/refactoring/instVarStructureCommand.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  // The engine can report a whole-apply error (e.g. the preview token expired between preview
+  // and apply) with an empty `failed` list — surface it instead of a hollow "applied 0" success.
+  if (result.error) {
+    void vscode.window.showErrorMessage(`${heading} failed: ${result.error}`);
+    return false;
+  }
+
+  if (result.failed.length > 0) {
+    // A structure change is all-or-nothing, but the engine applies top-down and stops reporting
```
MatiasFernandez: nit: Thanks for adding the abort warning — that's exactly what I was after. One wording thing: "stops reporting at the first failure" doesn't match the engine. `applyDeselected:` keeps looping past a failed change and collects *every* failure into `failed:` — it's this client that only surfaces `failed[0]`. As written, the next reader will assume `failed` has at most one entry.

Maybe something like "the engine applies top-down and collects every failure; we surface the first"?

client/src/refactoring/instVarStructureCommand.ts
```diff
@@ -0,0 +1,228 @@
...
+          options.removeOldFromHistory,
+        ),
+      ),
+    cleanup: safeClear,
+  });
+  if (!result) return false;
+
+  // The engine can report a whole-apply error (e.g. the preview token expired between preview
+  // and apply) with an empty `failed` list — surface it instead of a hollow "applied 0" success.
+  if (result.error) {
```
MatiasFernandez, line 147: 💭 This is the H2 path, and surfacing the error instead of the hollow "applied 0" toast is right. But it's now the one failure path that *doesn't* carry the abort advice you added just below at line 155.

For the expected case (expired preview token) that's fine — nothing applied, nothing to abort. But `result.error` is a whole-apply error, so it also covers a raise partway through the change loop, which leaves the same half-reversioned uncommitted hierarchy — and here the user gets no hint that their transaction needs an abort.

Is the error envelope distinguishable (token-expiry vs. mid-apply raise)? If not, would it be safer to append the same "abort the transaction" sentence here too, given the cost of a missing warning is worse than a redundant one?

MatiasFernandez (approved):
I left non-blocking comments but take a look at them just in case they are important to be addressed

client/src/refactoring/instVarStructureCommand.ts
ericwinger: Fixed in e77c3496 — reworded to almost exactly your suggestion:

> A structure change is all-or-nothing, but the engine applies top-down and collects every failure (the earlier changes stay applied, uncommitted); we surface the first.

Thanks for catching that `failed` can hold more than one entry.

client/src/refactoring/instVarStructureCommand.ts
ericwinger: Good question to force — I traced it through the engine, and the envelope is distinguishable *by construction*, so this branch is correct as-is (no abort advice needed).

The top-level `error` field is set in exactly one place: the `ifNil:` (token-not-found) branch of `GsInstVarStructureRefactoring>>applyForToken:deselected:migrateInstances:removeOldFromHistory:`:

```smalltalk
^(SessionTemps current at: token asSymbol ifAbsent: [nil])
    ifNil: [`{"applied":0,"committed":false,"migratedFailures":0,"failed":[],"error":"preview session expired"}`]
    ifNotNil: [:ref | ... ref applyDeselected: deselectedIds]
```

`applyDeselected:` (the token-*found* path) never emits a top-level `error` at all — its envelope is `{applied, committed, migratedFailures, failed:[...]}`. A raise *during* apply is caught per-change (`on: Error do:`) and each is collected into `failed:`; commit/migrate/prune failures are caught the same way. So a mid-apply raise surfaces via `result.failed`, which routes to the branch just below that already carries the abort warning.

Net: `result.error` truthy ⟺ token expired ⟺ `applied:0`, nothing changed ⟺ nothing to abort. Appending the abort sentence here would be a false alarm on a plain expired-token. I pushed e77c3496 tightening the comment to state exactly this (token-expiry is the *only* whole-apply error, and the missing abort advice is deliberate) so the next reader doesn't re-derive the concern.

PR #344, opened 2026-07-29
Migrate the ungated GCI binding tests into CI [Part 3]

MatiasFernandez:
Phase 3 of moving the deep GCI suite out of the excluded `gci` vitest project and into the default project, where CI actually runs it across the whole GemStone matrix.

## What moves

Traversal, export set and the session-free special-OOP calls become their own `*.integration.test.ts` files under `client/src/gciLibrary/__tests__/`. The leftover describes in `gciObjects` (`FetchUnicode`, the `IsKindOf*`/`IsSubclassOf*` family, `ResolveSymbolObj`) and `gciCompile` (`ClassRemoveAllMethods`, `ProtectMethods`) fold into the integration files migrated in Part 2.

Fifty tests, none of them gated: every wrapper they exercise is bound via the required `func`, so it resolves on the 3.6.2 matrix floor. Verified green there.

## What stays behind, and why

Exactly the set that needs the per-symbol availability gate (every wrapper `optionalFunc`-bound, so it throws on 3.6.2): all of `gciSessionUtils`, all of `gciMisc`, the four `gciFetchStore` leftovers, and the `GciI32ToOop` describe — the only part of `gciSpecials` that survives in `gci/`. `RemoveOopsFromNsc` is bound unconditionally but its describe can't run without `AddOopsToNsc`, which isn't.

## Worth a look

- **The dead-wrapper list was wrong about `gciExportSet`.** `debugQueries` pins and unpins objects with `GciTsSaveObjs`/`GciTsReleaseObjs` for variable-revert, and calls `GciTsOopIsSpecial`. Only `GciTsGetFreeOops` has no caller. They migrate here anyway — used or not, none of them need a gate.
- **`StoreTravDoTravRefs`** performed `reversed`, the same selector 3.6.2 lacks that tripped Part 2. It sends `asUppercase` now.
- **`MoreTraversal`** asserted "either a second call was needed or it wasn't". A 4000-byte String in the 2048-byte minimum buffer always reports an incomplete traversal, so it asserts that instead.
- **`ProtectMethods`** is session state the harness's abort can't roll back. The test asserts the DataCurator refusal and switches protection back off if a stone ever lets it through.

## Verification

`npm run lint`, `npm run format:check` and `npm run compile` are clean; the five migrated files pass against a live 3.6.2 stone. The 11 failing `refactoring/*.integration.test.ts` files in a full `npm test` are pre-existing on this branch's base — that local stone has no refactoring plugin — and fail identically on a pristine tree.


client/src/gciLibrary/__tests__/gciCompile.integration.test.ts
```diff
@@ -91,4 +92,46 @@ describe('GCI method compilation (integration)', () => {
...
+      // provisioned by `npm run test:server:start` — and only SystemUser may
+      // switch method protection on.
+      const { success, err } = gci.GciTsProtectMethods(session, true);
+      // Protection is session state the harness's abort can't roll back, so
+      // undo it if this stone did let us through.
+      if (success) gci.GciTsProtectMethods(session, false);
+
+      // Asserting refusal assumes no release in the matrix provisions its test
+      // user as SystemUser; revisit this if one ever does.
+      expect(success).toBe(false);
```
ericwinger: This asserts the *harness user's privileges* rather than the wrapper's behaviour, and that becomes riskier now the file runs in the default project.

The comment covers the CI matrix, but this now runs on every developer's `npm test` against whatever stone their environment points at. `.env.test.local` is a supported way to target a custom stone, and if that stone's user happens to be SystemUser, the call succeeds, protection is toggled back off, and the test then fails on `expect(success).toBe(false)` -- a red that says "binding broken" when the truth is "different user".

Deriving the expectation from the session instead would make it robust wherever it runs, e.g.

```ts
const isSystemUser =
  execute(`System myUserProfile userId`) === 'SystemUser'; // or however you'd rather probe it
const { success, err } = gci.GciTsProtectMethods(session, true);
if (success) gci.GciTsProtectMethods(session, false);
if (isSystemUser) {
  expect(success).toBe(true);
} else {
  expect(success).toBe(false);
  expect(err.number).not.toBe(0);
}
```

That also turns the "revisit this if a release ever provisions SystemUser" note into something the test handles by itself.

Separately and more minor: the `if (success)` restore is not in a `finally`, so if an assertion between the call and the restore ever threw, protection would stay on for the session. It happens to be safe as written (the restore precedes both asserts), but a `try`/`finally` would keep it safe under future edits.

client/src/gciLibrary/__tests__/gciTraversal.integration.test.ts
```diff
@@ -0,0 +1,149 @@
...
+      const { result: oop } = gci.GciTsNewString(session, 'X'.repeat(4000));
+      const MIN_TRAV_BUFF_SIZE = 2048;
+
+      const first = gci.GciTsFetchTraversal(session, [oop], 1, 0, OOP_NIL, MIN_TRAV_BUFF_SIZE);
+
+      expect(first.err.number).toBe(0);
+      expect(first.status).toBe(TRAVERSAL_INCOMPLETE);
+      const rest = gci.GciTsMoreTraversal(session, MIN_TRAV_BUFF_SIZE);
+      expect(rest.err.number).toBe(0);
+      expect(usedBytes(rest.travBuf)).toBeGreaterThan(0);
```
ericwinger: Good change on the first assertion -- pinning `TRAVERSAL_INCOMPLETE` is deterministic now (4000 bytes cannot fit a 2048-byte buffer), which is a real improvement on "either a second call was needed or it wasn't".

The second assertion is still weak, though: `usedBytes > 0` proves *more bytes arrived*, not that they are *the right bytes*. A wrapper that returned a non-empty but wrong or duplicated buffer would pass.

Since `parseTravBuffer` and `bodyText` are already to hand, you could assert the property that actually matters -- that the two buffers reassemble into the original String:

```ts
const firstBody = GciLibrary.parseTravBuffer(first.travBuf).map(bodyText).join('');
const restBody = GciLibrary.parseTravBuffer(rest.travBuf).map(bodyText).join('');
expect(firstBody + restBody).toBe('X'.repeat(4000));
```

If continuation reports don't reassemble that cleanly, then even `expect(restBody).toBe('X'.repeat(restBody.length))` plus a length check would be stronger than the current byte-count assertion. Worth confirming against a live stone before taking it -- I haven't run this variant.

ericwinger (approved):
LGTM

client/src/gciLibrary/__tests__/gciCompile.integration.test.ts
MatiasFernandez: Good find. I fixed this

client/src/gciLibrary/__tests__/gciTraversal.integration.test.ts
MatiasFernandez: Fixed

PR #346, opened 2026-07-30
Add Remove/Add Instance Variable refactoring (catalog V1)

ericwinger:
## Summary

Adds the **Add / Remove Instance Variable** server-side refactoring (catalog item
V1, part of #252) to the GemStone Explorer, on the shared RB refactoring engine.

GemStone has no `addInstVarName:`/`removeInstVarName:`, so changing a class's own
instance-variable list means creating a **new class version**: the engine builds a
new version in the same class history with the edited ivar list, copies methods
forward, and re-parents descendants top-down (the same `newVersionOf:` machinery the
class-rename refactoring uses). Nothing commits unless you opt in.

## What it does

**Engine** (`GsInstVarRefactoring`, `#add` / `#remove`):
- Versions the affected subtree top-down; each class is a `classDefinitionEdit`
  (own-ivar list changes) or a `classReparent` (recompiled only to re-point at the
  new parent version).
- **Preserves class compile options** across the new version (earlier engines passed
  `options: #()` and dropped them).
- **Predicts methods that will not recompile** after a remove (they reference the
  gone variable) from bytecode-level instance-variable access, surfaces them in the
  preview, and reports them as dropped after apply.
- **Opt-in, explicitly-committing** migrate-instances and delete-history steps; the
  structural change itself never commits.
- Declines are surfaced, not applied (invalid name, duplicate/inherited name,
  class-variable shadow, non-own ivar on remove, unknown operation).

**UX** (GemStone Explorer):
- `+` on the instance variable-side node (and class right-click) to add.
- `-` inline on an instance-variable row to remove.
- Preview panel: the staged class changes, a prominent "will not recompile" warning,
  and migrate-instances / delete-history checkboxes clearly flagged as committing
  (with a modal confirm).
- After an add, the new instance-variable row is selected in the Explorer.

## Scope note

An earlier revision of this branch also included **V4 (move instance variable to
another class)**. It was intentionally **dropped**: V2/V3 (push up / push down,
#343) cover the common cases, and moving an ivar to an arbitrary class isn't worth
the extra UI. This PR is V1 only.

## Testing

**112 new tests** across every layer:
- **27** GemStone SUnit (in-stone engine correctness)
- **75** client unit tests (command orchestration, Explorer handlers, query
  builders/wrappers, preview parsing, panel host + webview)
- **5** gated GCI integration tests (files in the payload, runs the SUnit suite
  in-stone, and a real add/remove round trip)
- **5** on-demand GCI e2e tests (incl. the committing migrate / delete-history paths)

Full local CI-mirror matrix on **3.6.2 and 3.7.5**, with the RB engine gates both
**off** (refactoring tests skip-guard) and **on** (they exercise):
- client + server + mcp: **green on both versions in both engine states**.
- In-stone RB SUnit (`GsInstVarRefactoringTest`): **27/27 on both boundaries**.

The only remaining reds are the pre-existing managed-stone `#encodeAsUTF8`
base-image gap in the on-demand `test:gci` suite (identical to the existing
`gciMoveMethod.e2e`; only green on a full 3.7.5 image / CI) and one unrelated
`querySunitRunLimit` smoke-test timeout, neither introduced here.


client/src/refactoring/instVarRefactorPreview.ts
```diff
@@ -0,0 +1,219 @@
...
+ * only to re-point at the freshly created parent version). The whole change is
+ * ALL-OR-NOTHING: the class-shape edits and the descendant reparents must move together,
+ * so every row is a required (checked + disabled) row and the deselected set is always
+ * empty.
+ *
+ * Two things this preview surfaces beyond the rest of the family:
+ *   - `willNotRecompile` — the methods that reference the removed variable and so will
+ *     fail to compile onto the new class version (they are dropped, and reported again in
+ *     the apply result's `dropped`);
+ *   - the acted-on class's `currentOptions` + the `optionVocabulary`, driving the
```
MatiasFernandez: 🚫 This says `currentOptions` + `optionVocabulary` drive "the editable class-options group", but the panel deliberately doesn't render one — `instVarRefactorPanelHtml.ts` says options are "intentionally not surfaced for editing here" and `instVarRefactorPanelView.js` always posts `options: null`. The engine repeats the same claim (`GsInstVarRefactoring` class comment: "the acted-on class's options are additionally EDITABLE from the preview panel", plus `actedOnClassName`'s "whose options are user-editable in the panel" and `outOfScopeJsonString`'s "drive the editable class-options group").

Could we align the docs with what actually ships? Right now a reader (or an agent) following these comments would go looking for a UI that isn't there. Related: `currentOptions`/`optionVocabulary` are parsed into `InstVarOutOfScope` and never read by anything — worth a note that they're carried for a future options editor, so nobody deletes them as dead payload (or drops them if that's the intent).

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
```diff
@@ -0,0 +1,656 @@
...
+	decline := nil.
+	newIvarLists := Dictionary new.
+	willNotRecompile := OrderedCollection new.
+	operation == #add ifTrue: [^self analyzeAdd].
+	operation == #remove ifTrue: [^self analyzeRemove].
+	decline := 'Unknown operation: ', operation printString
+]
+
+{ #category : 'private - analysis' }
+GsInstVarRefactoring >> analyzeAdd [
```
MatiasFernandez, line 204: 💭 `analyzeRemove` predicts the methods that will not recompile, but `analyzeAdd` doesn't compute anything — and `testAddHasNoWillNotRecompile` pins that at 0.

Is an add really always safe? The new variable becomes visible to the whole subtree, so a descendant method that already has a temporary/argument/block-arg named `tally` (or a block temp) would now collide with an instance variable when it's recompiled onto the new version. If GemStone rejects that, the method is silently dropped: the preview promised "0 will not recompile" and the user only finds out from the post-apply `dropped` list.

Worth checking on a stone whether a shadowing temp actually fails to compile? If it does, we could reuse the same accessor-scan machinery to surface it in the preview.

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
```diff
@@ -0,0 +1,656 @@
...
+			inject: '' into: [:acc :s | acc isEmpty ifTrue: [s] ifFalse: [acc, ',', s]]),
+	  '],"committed":', (committed ifTrue: ['true'] ifFalse: ['false']), '}'
+]
+
+{ #category : 'applying' }
+GsInstVarRefactoring >> commitStructuralThenMigrate: aBool deleteHistory: dBool on: failures [
+	"With the structural change applied in the transaction, commit it (migrateInstancesTo: needs a
+	 clean transaction), then -- if asked -- migrate instances of each old version to its new one
+	 and commit, and delete every prior version from each versioned class's history and commit."
+	[System commitTransaction.
```
MatiasFernandez, line 486: 💭 `System commitTransaction` commits the whole session transaction, not just the structural change we staged. If the developer has other uncommitted work in the same Jasper session (a method edited in an editor, a class variable added earlier, an aborted-but-not-really experiment), ticking "Migrate existing instances" quietly commits all of it too.

The panel's modal does say "This will COMMIT the transaction", which is honest, but should we go further and tell them what else is riding along — e.g. check `System needsCommit` before staging and warn "you have other uncommitted changes; they will be committed too"? That's the kind of surprise that's hard to undo.

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
```diff
@@ -0,0 +1,656 @@
...
+	| applied failures |
+	self ensureAnalysis.
+	"nil optsArray keeps the acted-on class's current options; an Array (even empty) sets them."
+	editedOptions := optsArray isNil ifTrue: [nil] ifFalse: [optsArray collect: [:e | e asString]].
+	oldToNew := IdentityDictionary new.
+	dropped := OrderedCollection new.
+	committed := false.
+	failures := OrderedCollection new.
+	applied := 0.
+	self changeSet changes do: [:change |
```
MatiasFernandez, line 460: 💭 The change set is documented as ALL-OR-NOTHING, but this loop keeps going after a failure: if versioning `GsIVSub` raises, `GsIVLeaf` is still versioned (and re-parented onto whatever `oldToNew` holds for its superclass, which is now the *old* sub). We answer `applied: n` plus the failures and leave that half-versioned subtree sitting in the transaction; the client shows "Failed: …", returns `undefined`, and doesn't even refresh the Explorer — so the user is looking at a stale tree over a partially reshaped hierarchy.

Since nothing has committed at that point, would it make sense to `System abortTransaction` (or at least stop at the first failure and say "nothing was applied, the transaction was rolled back")? Right now recovering means knowing to abort by hand.

client/src/refactoring/instVarRefactorCommand.ts
```diff
@@ -0,0 +1,145 @@
...
+      parsePage(await queries.pageInstVarPreview(session, token, off, PREVIEW_PAGE_BYTES)),
+    apply: async (options, migrate, deleteHistory) =>
+      parseApplyResult(
+        await queries.applyInstVar(session, token, [], options, migrate, deleteHistory),
+      ),
+    cleanup: safeClear,
+  });
+  if (!result) return undefined;
+
+  if (result.failed.length > 0) {
```
MatiasFernandez, line 131: 💭 We check `result.failed`, but never `result.error` — and the engine's expired-token path answers `{"applied":0,"failed":[],"dropped":[],"committed":false,"error":"preview session expired"}`. That parses cleanly, `failed` is empty, so we fall through to `showInformationMessage('Add bar to Foo.')` and the caller refreshes/reveals as if it worked, when nothing happened at all.

I know `moveMethodCommand` has the same shape, so this isn't new — but here the token can expire while the user sits on the preview deciding about the committing checkboxes, which makes it more reachable. Should we surface `result.error` (and/or `applied === 0`) as a failure before the success toast?

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
```diff
@@ -0,0 +1,656 @@
...
+]
+
+{ #category : 'private' }
+GsInstVarRefactoring >> isValidIvarName: aString [
+	"A syntactically valid Smalltalk instance-variable name: a letter or underscore followed by
+	 letters, digits, or underscores."
+	| first |
+	aString isEmpty ifTrue: [^false].
+	first := aString at: 1.
+	(first isLetter or: [first == $_]) ifFalse: [^false].
```
MatiasFernandez, line 178: 💭 This accepts an uppercase first letter, so `Tally` passes as an instance-variable name (the Explorer's `validateInput` regex `/^[A-Za-z_][A-Za-z0-9_]*$/` agrees). The class comment above notes GemStone won't compile a reference to an undeclared *lowercase* identifier — the flip side being that a capitalized identifier in a method body reads as a global, so an ivar named `Tally` is at best confusing and at worst silently shadowed by / shadowing a global of that name.

Should we require a lowercase (or `_`) first character here? Also, the client regex and this predicate are now two copies of the same rule that can drift — since the engine already declines with a good message, is the client-side validator worth keeping as more than a fast-fail?

client/src/refactoring/instVarRefactorPanelView.js
```diff
@@ -0,0 +1,165 @@
...
+      const commits =
+        (migrateBox && migrateBox.checked) || (deleteHistoryBox && deleteHistoryBox.checked);
+      applyBtn.textContent = commits ? 'Apply & Commit' : 'Apply';
+      applyBtn.classList.toggle('commits', !!commits);
+    };
+    if (migrateBox) migrateBox.addEventListener('change', syncCommitHint);
+    if (deleteHistoryBox) deleteHistoryBox.addEventListener('change', syncCommitHint);
+
+    if (applyBtn) {
+      applyBtn.addEventListener('click', function () {
```
MatiasFernandez, line 116: 💭 Apply isn't disabled once clicked, and the host has no `applying` guard (unlike the `loading` guard for pagination), so a double-click posts two `apply` messages and `handlers.apply` runs `applyForToken:` twice against the same token. The second run re-applies an already-applied change set — a second round of class versions on top of the first, and with a committing option ticked, a second commit.

Should we `applyBtn.disabled = true` here (and re-enable on `busyDone`), or guard in `instVarRefactorPanel.ts` the way `loading` does?

client/src/__tests__/gci/gciInstVar.e2e.test.ts
```diff
@@ -0,0 +1,282 @@
...
+        await applyInstVar(asyncExec, 'gci-iv-remove', [], null, false, false),
+      );
+      expect(result.failed).toEqual([]);
+      expect(result.dropped.map((m) => m.selector)).toEqual(
+        expect.arrayContaining(['combine', 'doubleCount']),
+      );
+
+      expect(hasIvar(BASE, 'count')).toBe(false);
+      expect(includesSelector(BASE, 'combine')).toBe(false);
+      expect(includesSelector(BASE, 'getOther')).toBe(false); // never existed
```
MatiasFernandez, line 171: 💭 This assertion can't fail — `getOther` is never compiled by `defineFixture` (it exists in the SUnit fixture, not this one), so `false` is guaranteed no matter what the refactoring does. The interesting property is the opposite one: a method that does *not* touch the removed variable must survive the new version.

Could we add `getOther\n\t^ other` to `defineFixture` and then assert `includesSelector(BASE, 'getOther')` is `true` (and that it's absent from `result.dropped`)? That turns the vacuous line into the one that proves copy-forward is selective.

gs-src/refactoring/tests/GsInstVarRefactoringTest.class.st
```diff
@@ -0,0 +1,460 @@
...
+		instVarNames: #()
+		classVars: #()
+		classInstVars: #()
+		poolDictionaries: #()
+		inDictionary: UserGlobals.
+	self compile: 'combine ^count + other' in: base.
+	self compile: 'getOther ^other' in: base.
+	self compile: 'doubleCount ^count * 2' in: sub.
+	self compile: 'readSubOwn ^subOwn' in: sub.
+	self compile: 'leafCount ^count negated' in: leaf
```
MatiasFernandez, line 114: 💭 The fixture is instance-side only: no class-side methods and no class variables anywhere in the hierarchy. That leaves two things `makeNewVersionOf:` / `copyMethodsFrom:` do untested — the meta-side copy (`old class selectors do: …`, including the `dstCls thisClass name, ' class'` label used when a class method is dropped) and the `classVars:`/`classInstVars:` preserved onto the new version.

Given class options got their own regression test because an earlier engine silently dropped them, should the fixture grow a class method and a class variable so the same class of bug can't slip through here?

client/src/refactoring/instVarRefactorPanelHtml.ts
```diff
@@ -0,0 +1,255 @@
...
+
+function renderWillNotRecompile(oos: InstVarOutOfScope): string {
+  const broken = oos.willNotRecompile;
+  if (broken.length === 0) return '';
+  const items = broken
+    .map((m) => `<li><code>${escapeHtml(m.className)}&gt;&gt;${escapeHtml(m.selector)}</code></li>`)
+    .join('');
+  const n = broken.length;
+  return `<div class="warn-box">
+    <div class="warn-head">⚠ ${n} method${n === 1 ? '' : 's'} will NOT recompile (they reference the ${n === 1 ? 'variable' : 'variable'} being removed/moved) and will be dropped:</div>
```
MatiasFernandez, line 87: nit: `${n === 1 ? 'variable' : 'variable'}` picks the same word on both branches — leftover from a plural that isn't needed (it's always one variable), so the ternary can just go. In the same string, "being removed/moved" mentions a move that this PR intentionally dropped (V4); same for the `title` example on line 104 ("Move count to Bar") and the header of `instVarRefactorPanelView.js` ("add / remove / move"). Worth trimming so the docs match the shipped catalog item?

client/src/__tests__/gci/gciInstVar.e2e.test.ts
```diff
@@ -0,0 +1,282 @@
+import { describe, it, expect, beforeAll, afterAll, vi } from 'vitest';
```
MatiasFernandez, line 1: 💭 I'm working on migrating `gci` tests to the `useIntegrationTest` approach so tests like this run in CI. Did you ask the agent to migrate this test to the other approach? If it won't be possible, I would like to understand why so I can fix it in the integration suite and enable more tests to be written that way.

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
ericwinger: Deferring this out of #346 to keep the PR focused on the V1 add/remove engine. Tracked as an item in #360; resolving here.

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
ericwinger: Deferring this out of #346 to keep the PR focused on the V1 add/remove engine. Tracked as an item in #360; resolving here.

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
ericwinger: Deferring this out of #346 to keep the PR focused on the V1 add/remove engine. Tracked as an item in #360; resolving here.

client/src/refactoring/instVarRefactorCommand.ts
ericwinger: Deferring this out of #346 to keep the PR focused on the V1 add/remove engine. Tracked as an item in #360; resolving here.

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
ericwinger: Deferring this out of #346 to keep the PR focused on the V1 add/remove engine. Tracked as an item in #360; resolving here.

client/src/refactoring/instVarRefactorPanelView.js
ericwinger: Deferring this out of #346 to keep the PR focused on the V1 add/remove engine. Tracked as an item in #360; resolving here.

client/src/__tests__/gci/gciInstVar.e2e.test.ts
ericwinger: Deferring this out of #346 to keep the PR focused on the V1 add/remove engine. Tracked as an item in #360; resolving here.

gs-src/refactoring/tests/GsInstVarRefactoringTest.class.st
ericwinger: Deferring this out of #346 to keep the PR focused on the V1 add/remove engine. Tracked as an item in #360; resolving here.

client/src/refactoring/instVarRefactorPanelHtml.ts
ericwinger: Deferring this out of #346 to keep the PR focused on the V1 add/remove engine. Tracked as an item in #360; resolving here.

client/src/__tests__/gci/gciInstVar.e2e.test.ts
ericwinger: Deferring this out of #346 to keep the PR focused on the V1 add/remove engine. Tracked as an item in #360; resolving here.

client/src/refactoring/instVarRefactorPreview.ts
ericwinger: Fixed in 380003ed. Went with docs + trim:

- **Docs aligned with what ships** — the preview header, the `GsInstVarRefactoring` class comment, `actedOnClassName`, and `previewDefinitionFor:` no longer claim an editable class-options group. They now say class options are *preserved* onto the new version but *not surfaced for editing* in the panel.
- **Dropped the dead payload** — removed `currentOptions`/`optionVocabulary` from `InstVarOutOfScope` (type + parse) and the engine's `outOfScopeJsonString`, plus the now-orphaned `optionVocabulary` / `jsonStringArray:` (engine) and `asStringArray` (client) helpers. Regenerated `resources/refactoring/*.gs`.

The real, tested options-*preservation* path (`optionsForApply:`/`editedOptions`) is untouched. The remaining `editedOptions` apply-time plumbing is now also unreachable from the UI; I left it and documented it as future-options-editor plumbing rather than widening this PR's scope.

The other, non-blocking review comments are tracked in #360.

MatiasFernandez (approved):

PR #348, opened 2026-07-30
Make MCP socket tests portable to Windows

npapagna:
## What

Makes the MCP socket tests runnable on Windows, and closes two gaps found while
porting them.

The socket tests each rolled their own tmp-path helper and assumed a POSIX
filesystem, so the Windows CI run could never go green on them.
`temporarySocketPath()` replaces those ad hoc helpers with one shared,
truly-unique path (pid + `process.hrtime`, plus a named-pipe form on Windows)
that survives Vitest's per-file module resets, unlike a plain in-module counter.

While porting the ownership tests, a mislabeled `describe` block turned up: four
unrelated behaviors sat under the title *"fails to start if the parent directory
is a file"*, a scenario nothing actually tested. The block is renamed and the
real regression test for that `ENOTDIR` path now exists, inside the POSIX-only
*"real socket file behavior"* block, since it builds a filesystem path directly
and a Windows named pipe fails with a generic `EACCES` instead.

Also folds two pre-existing, differently-spelled "skip on Windows" test helpers
(`unixOnly`, `itUnlessWin32`) into the same `platformGates.ts` module this work
introduces, and switches `mcpSocketServer.ts` to the shared `isWindows()`
instead of re-deriving `process.platform === 'win32'`.

## Testing

`npm run lint`, `npm run format:check`, `npm run compile`, and `npm test` pass
locally on macOS. The Windows Health Check CI run is the one that matters here.

### DIFF
--- a/client/src/__tests__/mcpOwnerSidecar.test.ts
+++ b/client/src/__tests__/mcpOwnerSidecar.test.ts
@@ -1,6 +1,5 @@
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import * as fs from 'fs';
-import * as os from 'os';
 import * as path from 'path';
 import {
   McpOwnerInfo,
@@ -9,10 +8,7 @@ import {
   readOwnerSidecar,
   writeOwnerSidecar,
 } from '../mcpOwnerSidecar';
-
-function makeTempSidecarPath(): string {
-  return path.join(fs.mkdtempSync(path.join(os.tmpdir(), 'mcp-owner-sidecar-')), 'mcp.owner.json');
-}
+import { safelyRemovePath, temporarySidecarPath, withTemporaryFolderDo } from './support/file';
 
 function sampleInfo(overrides: Partial<McpOwnerInfo> = {}): McpOwnerInfo {
   return {
@@ -27,14 +23,10 @@ function sampleInfo(overrides: Partial<McpOwnerInfo> = {}): McpOwnerInfo {
 describe('writeOwnerSidecar / readOwnerSidecar', () => {
   let sidecarPath: string;
   beforeEach(() => {
-    sidecarPath = makeTempSidecarPath();
+    sidecarPath = temporarySidecarPath();
   });
   afterEach(() => {
-    try {
-      fs.rmSync(path.dirname(sidecarPath), { recursive: true, force: true });
-    } catch {
-      /* ignore */
-    }
+    safelyRemovePath(sidecarPath);
   });
 
   it('round-trips an info record through disk', () => {
@@ -45,9 +37,13 @@ describe('writeOwnerSidecar / readOwnerSidecar', () => {
   });
 
   it('creates the parent directory if it does not exist', () => {
-    const nested = path.join(path.dirname(sidecarPath), 'sub', 'mcp.owner.json');
-    writeOwnerSidecar(sampleInfo(), nested);
-    expect(fs.existsSync(nested)).toBe(true);
+    withTemporaryFolderDo((temporaryFolder) => {
+      const nested = path.join(temporaryFolder, 'sub', 'mcp.owner.json');
+
+      writeOwnerSidecar(sampleInfo(), nested);
+
+      expect(fs.existsSync(nested)).toBe(true);
+    });
   });
 
   it('returns undefined when the sidecar does not exist', () => {
@@ -103,14 +99,10 @@ describe('writeOwnerSidecar / readOwnerSidecar', () => {
 describe('deleteOwnerSidecar', () => {
   let sidecarPath: string;
   beforeEach(() => {
-    sidecarPath = makeTempSidecarPath();
+    sidecarPath = temporarySidecarPath();
   });
   afterEach(() => {
-    try {
-      fs.rmSync(path.dirname(sidecarPath), { recursive: true, force: true });
-    } catch {
-      /* ignore */
-    }
+    safelyRemovePath(sidecarPath);
   });
 
   it('removes the sidecar when the recorded pid matches', () => {
--- a/client/src/__tests__/mcpServerTreeProvider.test.ts
+++ b/client/src/__tests__/mcpServerTreeProvider.test.ts
@@ -2,16 +2,10 @@ import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 
 vi.mock('vscode', () => import('../__mocks__/vscode.js'));
 
-import * as fs from 'fs';
-import * as os from 'os';
-import * as path from 'path';
 import { ActiveSession } from '../sessionManager';
 import { writeOwnerSidecar } from '../mcpOwnerSidecar';
 import { resolveOwnership, renderOwnership } from '../mcpServerTreeProvider';
-
-function makeTempSidecarPath(): string {
-  return path.join(fs.mkdtempSync(path.join(os.tmpdir(), 'mcp-tree-')), 'mcp.owner.json');
-}
+import { safelyRemovePath, temporarySidecarPath } from './support/file';
 
 function fakeSession(id = 7): ActiveSession {
   return {
@@ -36,14 +30,10 @@ function fakeSession(id = 7): ActiveSession {
 describe('resolveOwnership', () => {
   let sidecarPath: string;
   beforeEach(() => {
-    sidecarPath = makeTempSidecarPath();
+    sidecarPath = temporarySidecarPath();
   });
   afterEach(() => {
-    try {
-      fs.rmSync(path.dirname(sidecarPath), { recursive: true, force: true });
-    } catch {
-      /* ignore */
-    }
+    safelyRemovePath(sidecarPath);
   });
 
   it('returns "this" with the active session when this window is owner', () => {
--- a/client/src/__tests__/mcpSocketOwnership.test.ts
+++ b/client/src/__tests__/mcpSocketOwnership.test.ts
@@ -1,8 +1,8 @@
 /**
- * Owner/passive lifecycle tests for {@link McpSocketServer}. Uses real
- * Unix sockets in `os.tmpdir()`, so it intentionally does NOT mock `fs`.
- * The mocked-fs tests for the Claude Desktop config writer live in
- * mcpSocketServer.test.ts.
+ * Owner/passive lifecycle tests for {@link McpSocketServer}. Uses a real
+ * socket (Unix domain socket, or named pipe on Windows) in `os.tmpdir()`,
+ * so it intentionally does NOT mock `fs`. The mocked-fs tests for the
+ * Claude Desktop config writer live in mcpSocketServer.test.ts.
  */
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 
@@ -14,20 +14,17 @@ import * as os from 'os';
 import * as path from 'path';
 import * as crypto from 'crypto';
 import { McpSocketServer } from '../mcpSocketServer';
-
-const tmpSocketPath = () =>
-  path.join(os.tmpdir(), `jasper-mcp-test-${crypto.randomBytes(6).toString('hex')}.sock`);
+import { temporarySidecarPath, withTemporaryFileDo } from './support/file';
+import { temporarySocketPath } from './support/socket';
+import { onSupportedPosixDescribe } from './platformGates';
 
 // Each test gets its own sidecar path so parallel runs (and a real Jasper on
 // the same machine) don't collide on the default `EXTENSION_FOLDER/mcp.owner.json`.
-const tmpSidecarPath = () =>
-  path.join(os.tmpdir(), `jasper-mcp-owner-${crypto.randomBytes(6).toString('hex')}.json`);
-
 const newServer = (socketPath: string) =>
   new McpSocketServer({
     getSession: () => undefined,
     socketPath,
-    sidecarPath: tmpSidecarPath(),
+    sidecarPath: temporarySidecarPath(),
   });
 
 describe('McpSocketServer ownership', () => {
@@ -43,8 +40,8 @@ describe('McpSocketServer ownership', () => {
     }
   });
 
-  it('the first server to start becomes the owner and creates the socket file', async () => {
-    const socketPath = tmpSocketPath();
+  it('the first server to start becomes the owner', async () => {
+    const socketPath = temporarySocketPath();
     const server = newServer(socketPath);
     servers.push(server);
 
@@ -52,11 +49,10 @@ describe('McpSocketServer ownership', () => {
 
     expect(isOwner).toBe(true);
     expect(server.isOwner).toBe(true);
-    expect(fs.existsSync(socketPath)).toBe(true);
   });
 
   it('a second server on the same path is passive (EADDRINUSE)', async () => {
-    const socketPath = tmpSocketPath();
+    const socketPath = temporarySocketPath();
 
     const first = newServer(socketPath);
     servers.push(first);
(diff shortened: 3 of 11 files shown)


client/src/mcpSocketServer.ts
```diff
@@ -48,6 +49,29 @@ export function defaultSocketPath(): string {
   return extensionPathFrom('mcp.sock');
 }
 
+/**
+ * A socket / named-pipe path, unique per call. Unlike {@link defaultSocketPath},
+ * never collides across parallel test runs or with a real Jasper window's
+ * shared socket, so tests can freely start and dispose many
+ * {@link McpSocketServer} instances without stepping on each other.
+ *
+ * Uniqueness is guaranteed, not just overwhelmingly likely: `process.pid`
```
MatiasFernandez: 💭 Is uniqueness really *guaranteed* here? `pid` is unique among **live** processes, but pids get recycled, and `process.hrtime.bigint()` is documented as relative to an arbitrary past time rather than wall-clock — so a later run can in principle produce a name an earlier crashed run already left behind as a `.sock` in `os.tmpdir()`. In practice `start()` reclaims a stale socket file, so nothing breaks today; my concern is that the doc promises a stronger contract than the code delivers, and the helpers this replaces got that guarantee for free from `crypto.randomBytes(6)`.

What do you think about keeping a random component alongside the pid (`jasper-mcp-temp-${process.pid}-${crypto.randomBytes(6).toString('hex')}`), which is unique across runs *and* across module resets? Or, if pid + hrtime is deliberate, wording the doc as "unique within a run" so a future reader doesn't lean on more than it gives?

client/src/mcpSocketServer.ts
```diff
@@ -48,6 +49,29 @@ export function defaultSocketPath(): string {
...
+ * distinguishes concurrent processes (no two running processes on the same
+ * machine share one), and `process.hrtime.bigint()` distinguishes multiple
+ * calls within one process. Unlike a module-level counter, it's tied to the
+ * OS process rather than the module's state, so it keeps counting across a
+ * test runner's per-file module resets instead of restarting at zero.
+ */
+export function temporarySocketPath(): string {
```
MatiasFernandez: 💭 This is now exported production API (it ships in the extension bundle) whose only callers are tests — the doc says as much: "so tests can freely start and dispose many instances". Since this PR introduces `__tests__/platformGates.ts` as a home for shared test helpers, would this fit better there? It only needs `isWindows()`, which a test helper can import just as easily.

If it stays here as public API, is it worth two small tests pinning the contract the doc promises — two calls never return the same path, and the Windows branch returns a `\\.\pipe\` name? Right now the strongest claim in the file has nothing asserting it.

client/src/mcpSocketServer.ts
```diff
@@ -12,6 +12,7 @@ import { registerMcpTools } from './mcpTools';
 import { appendSysadmin } from './sysadminChannel';
 import { defaultSidecarPath, deleteOwnerSidecar, writeOwnerSidecar } from './mcpOwnerSidecar';
 import { extensionPathFrom } from './extensionPath';
+import { isWindows } from './wslBridge';
```
MatiasFernandez, line 15: The PR description says this "switches `mcpSocketServer.ts` to the shared `isWindows()` instead of re-deriving `process.platform === 'win32'`", but only the new `temporarySocketPath()` uses it — `defaultSocketPath()`, `start()`, `dispose()`, and the macOS/Windows branch further down still re-derive it directly (lines 46, 146, 256, 317).

💭 Since the import is already here, should we finish the switch so the file asks the platform question exactly one way? Otherwise a reader has to check whether the two forms are meant to differ.

client/src/__tests__/platformGates.ts
```diff
@@ -0,0 +1,14 @@
+import { describe, it } from 'vitest';
+
+const isPosix = process.platform === 'linux' || process.platform === 'darwin';
```
MatiasFernandez: 💭 This narrows the semantics of the two helpers it replaces: `skipIf(process.platform === 'win32')` ran everywhere except Windows, while `runIf(linux || darwin)` now *skips* on anything that isn't Linux/macOS (freebsd, aix, sunos…). Not a practical problem for the current CI matrix, but the failure mode is a silent skip rather than a visible failure, which is the harder one to notice.

Given the production code branches on `!isWindows()` (from `wslBridge`), should the gate be defined as `!isWindows()` too, so the gate and the code it's gating agree on what "not Windows" means? That also gives us one definition of the platform check instead of a third one.

client/src/__tests__/mcpSocketOwnership.test.ts
```diff
@@ -238,23 +204,81 @@ describe('McpSocketServer ownership', () => {
     expect(written.workspacePath).toBe('/tmp/first');
   });
 
-  it('creates the parent directory if missing', async () => {
-    const dir = path.join(os.tmpdir(), `jasper-mcp-dir-${crypto.randomBytes(6).toString('hex')}`);
-    const socketPath = path.join(dir, 'mcp.sock');
-    expect(fs.existsSync(dir)).toBe(false);
+  onPosixDescribe('real socket file behavior', async () => {
```
MatiasFernandez: nit: the callback is `async` but nothing inside the suite body is awaited — the `it`s carry their own async. Drop the `async` to keep it a plain suite factory?

client/src/__tests__/mcpSocketOwnership.test.ts
```diff
@@ -238,23 +204,81 @@ describe('McpSocketServer ownership', () => {
...
+        `jasper-mcp-notadir-${crypto.randomBytes(6).toString('hex')}`,
+      );
+      fs.writeFileSync(notADir, '');
+      const socketPath = path.join(notADir, 'mcp.sock');
+      const server = newServer(socketPath);
+      servers.push(server);
+
+      await expect(server.start()).rejects.toThrow(/ENOTDIR/);
+
+      fs.unlinkSync(notADir);
```
MatiasFernandez: 💭 If the `rejects.toThrow(/ENOTDIR/)` assertion fails, this `unlinkSync` never runs and we leak a file in `os.tmpdir()` — which then makes the *next* run of this test start from a dirty state (the path is random, so it won't actually collide, but the junk accumulates). Worth a `try`/`finally`, or tracking the path in a `let` and unlinking in an `afterEach` alongside the server cleanup that's already there?

client/src/__tests__/mcpSocketServerIntegration.test.ts
```diff
@@ -151,11 +151,7 @@ describe('McpSocketServer integration', () => {
     session = makeMockSession();
     server = new McpSocketServer({
       getSession: () => session,
-      // Randomize per test so parallel runs don't collide on the fixed socket path.
-      socketPath: path.join(
-        os.tmpdir(),
-        `jasper-mcp-test-${crypto.randomBytes(6).toString('hex')}.sock`,
-      ),
+      socketPath: temporarySocketPath(),
       sidecarPath: path.join(
```
MatiasFernandez: 💭 The socket path now comes from the shared helper, but the sidecar path is still hand-rolled here and again as `tmpSidecarPath()` in `mcpSocketOwnership.test.ts` — same intent, two copies, and this one keeps `crypto`/`os` imports alive just for it. Does it make sense to add a `temporarySidecarPath()` companion (wherever `temporarySocketPath()` ends up living) so both files get their unique paths the same way?

jgfoster (approved):

client/src/__tests__/platformGates.ts
npapagna: That's a fair point. I kept it this way because it reflects our current understanding of what's supported rather than everything that isn't Windows.

`process.platform` covers quite a few platforms (even `android` can show up), and I don't really know how sockets behave on the others. I was concerned that using `!isWindows()` here could imply we've intentionally validated all non-Windows platforms when we haven't. I'd rather explicitly lock the test to the platforms we know about than silently start running it somewhere we don't understand yet.

client/src/__tests__/mcpSocketOwnership.test.ts
npapagna: Fixed!

client/src/__tests__/mcpSocketOwnership.test.ts
npapagna: Fixed!

client/src/__tests__/mcpSocketServerIntegration.test.ts
npapagna: Fixed!

client/src/mcpSocketServer.ts
npapagna: Fixed! Moved it to `__tests__/support/socket.ts`

client/src/mcpSocketServer.ts
npapagna: Fixed!

client/src/mcpSocketServer.ts
npapagna: Fixed!

client/src/__tests__/mcpOwnerSidecar.test.ts
```diff
@@ -27,7 +23,7 @@ function sampleInfo(overrides: Partial<McpOwnerInfo> = {}): McpOwnerInfo {
 describe('writeOwnerSidecar / readOwnerSidecar', () => {
   let sidecarPath: string;
   beforeEach(() => {
-    sidecarPath = makeTempSidecarPath();
+    sidecarPath = temporarySidecarPath();
```
MatiasFernandez, line 26: 🚫 This swap changes what `path.dirname(sidecarPath)` means, and the `afterEach` right below (line 30) still recursively deletes it.

The old `makeTempSidecarPath()` was `path.join(fs.mkdtempSync(...), 'mcp.owner.json')`, so `dirname` was a **dedicated** temp directory and `rmSync(dirname, { recursive: true, force: true })` was safe. The new `temporarySidecarPath()` returns `path.join(os.tmpdir(), 'jasper-mcp-tempfile-<hex>.json')` — no intermediate directory — so `dirname` is now **`os.tmpdir()` itself**, and that hook is effectively `fs.rmSync(os.tmpdir(), { recursive: true, force: true })`.

Three hooks are in this state: lines 30 and 106 here, and `mcpServerTreeProvider.test.ts:38`. Concretely, on Linux CI that's `rm -rf /tmp` after each of those tests, and on macOS the same for `/var/folders/…/T`. Since vitest runs files in parallel workers, it will delete the sockets and sidecars other test files are actively using mid-run — and `force: true` plus the surrounding `try {} catch {}` means it fails silently, so the symptom shows up as unrelated flaky failures elsewhere rather than as an error here.

Since the file lives directly in `os.tmpdir()` now, could these hooks just `fs.rmSync(sidecarPath, { force: true })` (no `recursive`)? That matches what the helper actually creates.

Related: `it('creates the parent directory if it does not exist')` (line 44) builds `nested` as `path.join(path.dirname(sidecarPath), 'sub', 'mcp.owner.json')`, which is now a fixed `os.tmpdir()/sub` shared by every run instead of a per-test path — so that test loses its isolation too. Worth giving it its own unique directory?

client/src/__tests__/mcpServerTreeProvider.test.ts
```diff
@@ -36,7 +31,7 @@ function fakeSession(id = 7): ActiveSession {
 describe('resolveOwnership', () => {
   let sidecarPath: string;
   beforeEach(() => {
-    sidecarPath = makeTempSidecarPath();
+    sidecarPath = temporarySidecarPath();
```
MatiasFernandez, line 33: 🚫 Same issue as my comment on `mcpOwnerSidecar.test.ts:26` — the `afterEach` below now recursively removes `os.tmpdir()`, since `temporarySidecarPath()` no longer nests the file in its own directory.

client/src/__tests__/support/file.ts
```diff
@@ -0,0 +1,70 @@
...
+function temporaryFileName(): string {
+  return `jasper-mcp-tempfile-${crypto.randomBytes(6).toString('hex')}`;
+}
+
+/**
+ * Attempts to delete a file, logging an error instead of throwing if deletion fails.
+ *
+ * @param {string} file - The path to the file to delete.
+ */
+function safelyRemoveFile(file: string): void {
```
MatiasFernandez: 💭 The doc on `withTemporaryFileDo` promises the file is "deleted afterward regardless of success or failure", but this swallows the failure — so what's actually guaranteed is an *attempt*, and a leaked file surfaces only as a `console.error` line in the vitest output that nobody reads.

Is the swallow load-bearing for a case you hit? If the file is always created by `createTemporaryFile()` a line earlier, `unlinkSync` failing means something genuinely surprising happened (the consumer replaced the path with a directory, a permissions change), which seems worth failing on rather than logging. Or, if you'd rather keep the test green, soften the doc to say cleanup is best-effort so a future reader doesn't lean on the stronger promise?

Also, is `console.error` the right channel here? Everything else in the client routes through `appendSysadmin`, and in a test helper it just adds unattributable noise to the run output — there's no test name attached to it.

client/src/__tests__/support/socket.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/**
+ * A unique file path for a JSON "sidecar" file, generated via {@link temporaryFilePath}.
+ * Typically used to store metadata or state alongside a {@link temporarySocketPath},
+ * such as connection info for a test's {@link McpSocketServer} instance.
+ *
+ * @returns An absolute path to a non-existent `.json` file in the OS's temp directory.
+ */
+export function temporarySidecarPath() {
```
MatiasFernandez: 💭 This one is about sidecars, not sockets, and two of its three callers have no socket in sight: `mcpOwnerSidecar.test.ts` tests the sidecar reader/writer, and `mcpServerTreeProvider.test.ts` tests ownership rendering in the tree. Both now reach into `support/socket` for it, which reads oddly at the import site.

Would `support/sidecar.ts` (or just putting it next to `temporaryFilePath` in `support/file.ts`, since that's all it delegates to) be a better home? The doc would get more accurate as a side effect too — "metadata or state alongside a `temporarySocketPath()`" describes only the `mcpSocketOwnership` usage.

client/src/__tests__/support/socket.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * `process.pid` keeps the name human-readable for debugging leftover
+ * `.sock` files, but pids get recycled and don't by themselves guarantee
+ * uniqueness across runs. The `crypto.randomBytes(6)` suffix is what
+ * actually carries the guarantee: 2^48 possible values makes collision
+ * probability negligible, independent of process lifetime, reboots, or a
+ * test runner's per-file module resets.
+ **/
+export function temporarySocketPath(): string {
+  const socketName = `jasper-mcp-temp-${process.pid}-${crypto.randomBytes(6).toString('hex')}`;
```
MatiasFernandez: nit: `jasper-mcp-temp-` (sockets) and `jasper-mcp-tempfile-` (from `temporaryFileName()`) are near-identical prefixes, and `file.ts` says its prefix is there "for easy identification". In practice `ls /tmp | grep jasper-mcp-temp` matches both, so the one thing the prefix is for doesn't quite work. Worth making this one `jasper-mcp-socket-`?

client/src/__tests__/platformGates.ts
MatiasFernandez: That's a fair point, and I'd rather have the explicit allow-list than silently start running these somewhere we haven't validated — thanks for spelling out the reasoning.

💭 Given that's the intent, should the names say so? `isPosix` / `onPosix*` claim a property of the platform (freebsd and aix *are* POSIX, and would be excluded here), whereas what you're actually encoding is "the platforms we know socket behavior on". Something like `isSupportedPosixPlatform` with `onSupportedPosixIt/Describe` — or keeping `onPosix*` but having the doc say "Linux/macOS only; other POSIX platforms are excluded because we haven't validated socket behavior there" — would put your rationale where the next reader finds it instead of in this thread.

MatiasFernandez (approved):

PR #349, opened 2026-07-30
Make npm 11 the single lockfile authority across the CI matrix

MatiasFernandez:
## Problem

`package-lock.json` was last regenerated under **npm 10**, but [.nvmrc](.nvmrc) pins Node `24.15.0`, whose bundled npm is **11**. The two majors disagree about the dependency tree: npm 10 records nested optional-peer duplicates (`jsdom`'s and `html-encoding-sniffer`'s `@noble/hashes@2.2.0`) that npm 11 prunes as redundant. Both write `lockfileVersion: 3`, so the format version gives no warning.

The disagreement is **one-directional**, which is why it went unnoticed:

| lockfile generated by | `npm ci` on npm 11 | `npm ci` on npm 10 |
| --- | --- | --- |
| npm 10 (previously committed) | ✅ installs | ✅ installs |
| npm 11 | ✅ installs | ❌ `EUSAGE` |

npm 10's lockfile is a *superset*, so npm 11 tolerates the extra entries. npm 11's is a *subset*: npm 10 walks the tree, can't find `@noble/hashes` nested under `jsdom`, and fails outright:

```
npm error `npm ci` can only install packages when your package.json and
npm error package-lock.json or npm-shrinkwrap.json are in sync.
npm error Missing: @noble/hashes@2.2.0 from lock file
```

So anyone bumping a dependency on the pinned Node 24 silently breaks the `health-check` matrix's Node `22.15.1` floor job, which gets npm 10 from `setup-node`.

## Change

- Regenerate `package-lock.json` under npm 11 (58 deletions, all metadata — no version or integrity changes).
- Add an `npm install --global npm@11` step to the `health-check` job, gated with `if: startsWith(matrix.node-version, '22.')`, so the floor leg uses the same npm major that owns the lockfile and the dev legs skip it entirely.

The two must land together: the lockfile alone breaks the floor job, and the pin alone does nothing.

## Verification

- **The Node 22.15.1 floor job passes** (`GemStone S/64 3.6.2 (Node 22.15.1)`), along with every other matrix leg.
- npm 11 emits a **byte-identical** lockfile on Node `22.15.1` and `24.15.0` (0-line diff) — the churn tracks the npm *major*, not the Node version, so pinning npm makes the lockfile stable across both legs.
- The new lockfile installs cleanly via `npm ci` under Node `22.15.1` + npm 11.
- `npm run lint`, `npm run format:check`, `npm run compile` pass; `actionlint` and `zizmor` both clean locally at the versions CI pins.

## Trade-offs

**The install trips zizmor's `adhoc-packages` audit** (low severity), suppressed inline with a documented reason, following the convention already used in [roadmap.yml](.github/workflows/roadmap.yml). This is a risk acceptance rather than a fix, bounded by the job running with `contents: read` and `persist-credentials: false`. The alternative that would satisfy the audit outright — corepack with a hash-pinned `packageManager` field — requires pinning an exact npm patch, reintroducing precisely the drift this change removes, and corepack's future in the Node distribution is uncertain. Happy to switch if the preference is never to suppress a supply-chain audit.

**The floor job's signal narrows slightly:** it now tests the Node 22 floor with a newer npm than a default Node 22 install provides. That's the intended trade — npm isn't the runtime under test, and the floor exists to prove the oldest *runtime* works, per [raising-the-version-floor.md](docs/how-to/raising-the-version-floor.md). The alternative (generating lockfiles on npm 10 forever) permanently contradicts `.nvmrc`.


npapagna (approved):
LGTM 🚀

PR #350, opened 2026-07-30
V4: move instance variable to a chosen hierarchy class via the ▲/▼ picker

ericwinger:
Generalizes the instance-variable **push up / push down** arrows (V2/V3) into a single
**"move to a chosen class in the hierarchy"** refactoring — the revived V4. Each arrow now
opens a picker of destination classes:

- **▲** lists the class's **ancestors** (single-pick) — move up to the immediate superclass
  *or any higher ancestor*.
- **▼** lists the class's **descendants** (multi-pick) — push down to a chosen subset, at any depth.

So the two arrows cover push-up (V2), push-down to a subset (V3), and move across the hierarchy
(V4) with one gesture. The move stays **hierarchy-constrained** (up = a proper ancestor, down =
proper descendants) — no move to an arbitrary unrelated class, so no instance data-loss case is
reachable.

### Engine (`GsInstVarStructureRefactoring`)
- New `#move` operation (`class:moveInstVar:toClasses:direction:`) reusing the existing
  reversion / preview / paginated-apply machinery. `#pushUp` / `#pushDown` kept as tested
  special cases.
- New **losing-class precondition**: on a partial push-down, any class left *without* the ivar
  (the source, or a descendant subtree not among the targets) must not still reference it — else
  it declines, naming the class + method.
- Accessor relocation now keys off `effectiveDirection` (works for `#move` both ways).
- All-or-nothing, never auto-commits (opt-in migrate/prune still commit, as before).

### Client
- New `getClassDescendantNames` query feeds the ▼ picker; the ▲ picker reuses `getClassHierarchy`
  (now dict-scoped so a shadowed class name offers the right lineage).
- New `moveInstVar` flow (op `move`, threads targets + direction, carries simple accessors); the
  Explorer ▲/▼ arrows route through `pickInstVarMoveTargets`.
- Command titles → "Move Instance Variable Up/Down…".

### Tests
- Engine SUnit: **48/48 instVarStructure** and **409/409 whole RB suite** on **both 3.6.2 and 3.7.5**
  (12 new `#move` cases incl. every decline path).
- Client unit + engine-gated integration (new descendants query, move query/command, down-partial
  accessor, two move integration cases).
- Full CI-mirror: both plugin worlds PASS on both versions.
- Manual F5: all 14 scenarios verified (up-to-grandparent, multi-select down, every decline shape,
  no-commit, instances survive).

Stacked on `eric/RBinstVarStructureChanges` (#343, V2/V3/V5) since that isn't in main yet.


gs-src/refactoring/engine/GsInstVarStructureRefactoring.class.st
```diff
@@ -345,6 +400,98 @@ GsInstVarStructureRefactoring >> analyzePushDown [
...
+			(self otherDescendant: def ofTop: sup ownsIvar: varName) ifNotNil: [:cls |
+				^decline := 'Cannot move ', varName, ' up: ', cls, ' also declares an instance variable of that name, which would collide once it is inherited.'].
+			topClass := sup]
+		ifFalse: [
+			dir == #down
+				ifTrue: [
+					targetClasses do: [:t |
+						(self isAncestor: def of: t) ifFalse: [
+							^decline := 'Cannot move ', varName, ' down: ', t name asString, ' is not a subclass of ', def name asString, '.']].
+					(self anyDescendantOf: def ownsIvar: varName) ifNotNil: [:cls |
```
MatiasFernandez, line 442: 🚫 On `#down` the only ownership precondition is `anyDescendantOf: def ownsIvar: varName`, which looks at what classes own the ivar *today* — nothing checks the chosen targets against each other. The ▼ picker lists every descendant at any depth as a multi-select, so a user can tick both `GsVSLeaf` and a subclass of `GsVSLeaf`. `newIvarLists` then puts `varName` on both, and the child ends up declaring an instance variable it also inherits from the freshly-edited parent — which GemStone rejects when it recompiles the definition, so the user gets a `"failed"` entry out of apply rather than a clean decline.

Should `analyzeMove` reject targets that are descendants of another chosen target (or just drop the redundant ones, since the parent already covers them)? A `toClasses: #('GsVSLeaf' 'GsVSGrand')` case would pin whichever behaviour you pick. Same for the `resources/refactoring/engine.gs` mirror.

gs-src/refactoring/engine/GsInstVarStructureRefactoring.class.st
```diff
@@ -111,6 +122,25 @@ GsInstVarStructureRefactoring class >> class: aClass pushDownInstVar: aName [
...
+	| ref env targets |
+	env := GsRefactoringEnvironment new.
+	ref := self new
+		setEnvironment: env
+		operation: #move
+		class: aClass
+		varName: aName
+		selector: nil
+		meta: false.
+	targets := classNames collect: [:n | env classNamed: n asString].
```
MatiasFernandez, line 140: 💭 `GsRefactoringEnvironment>>classNamed:` is first-match across the whole symbol list, so the destinations get resolved *unscoped* even though the rest of this flow is carefully dict-scoped — the client resolves the source class through `classLookupExpr(className, dict)`, and you just threaded `state.dictIndex` into both pickers precisely so a shadowed name offers the right lineage. The names then travel back as bare strings and lose that scoping.

So if `Foo` is bound in two dictionaries, the picker can show the dict-2 `Foo` while the engine binds the dict-1 one. The friendly outcome is a puzzling `is not a superclass of` decline; the unfriendly one is the ivar landing on a same-named class in another dictionary that happens to sit in the right relation. Should the move carry the dictionary alongside the names (or take resolved classes instead of names)?

Worth noting the `hierNeighbors` comment in `gemstoneExplorer.ts` still asserts "the refactoring itself is dict-scoped … and stays correct" — after this change that's only true of the source class.

gs-src/refactoring/engine/GsInstVarStructureRefactoring.class.st
```diff
@@ -345,6 +400,98 @@ GsInstVarStructureRefactoring >> analyzePushDown [
...
+							^decline := 'Cannot move ', varName, ' down: ', t name asString, ' is not a subclass of ', def name asString, '.']].
+					(self anyDescendantOf: def ownsIvar: varName) ifNotNil: [:cls |
+						^decline := 'Cannot move ', varName, ' down: ', cls, ' already declares an instance variable of that name.'].
+					topClass := def]
+				ifFalse: [
+					^decline := 'Cannot move ', varName, ': unknown direction ', dir printString]].
+	newIvarLists at: def name asString put: ((self ownInstVarsOf: def) reject: [:n | n = varName]).
+	targetClasses do: [:t |
+		newIvarLists at: t name asString put: ((self ownInstVarsOf: t) copyWith: varName)].
+	self moveAccessors ifTrue: [self planAccessorMovesFrom: def to: targetClasses].
```
MatiasFernandez, line 450: 💭 The losing-class precondition runs *after* `newIvarLists` is filled in and `planAccessorMovesFrom:` has recorded its removals/adds, so a declined move leaves the instance holding a half-built plan — `analyzePushDown` does the opposite, all checks first and the accessor planning last. Nothing escapes today because the decline tests show `changeSet` comes back empty, but that relies on the changeSet builder noticing `decline` rather than on the object never being in that state.

The check recomputes `simpleAccessorsOf:` itself, so as far as I can tell it doesn't depend on the planning having happened — could it move above these two lines so "declined ⇒ nothing recorded" holds by construction?

client/src/refactoring/instVarStructureCommand.ts
```diff
@@ -199,6 +203,38 @@ export async function pushInstVar(
   });
 }
 
+/** V4: move an instance variable to chosen destination class(es) in the hierarchy — the general
+ *  form the Explorer's ▲/▼ arrows drive after the user picks a destination. `direction` is `up`
+ *  (a single chosen ancestor) or `down` (one or more chosen descendants). Always carries the
+ *  ivar's simple getter/setter accessors along, like push up/down. */
+export async function moveInstVar(
```
MatiasFernandez, line 210: 💭 Now that the Explorer's ▲/▼ arrows route through `moveInstVar`, is `pushInstVar` (just above) still reachable from production code? The only references I can find at this head are in `instVarStructureCommand.test.ts`. Keeping `#pushUp`/`#pushDown` in the engine as tested special cases makes sense to me, but a client helper whose only callers are its own tests tends to quietly rot — should it go in this PR, or is there a caller coming that I'm not seeing?

client/src/gemstoneExplorer.ts
```diff
@@ -1511,6 +1506,67 @@ export class ExplorerController {
     }
   }
 
+  // Ask the user which hierarchy class(es) to move an ivar to. ▲ lists ancestors (immediate
+  // superclass first) as a single-select; ▼ lists every descendant (top-down) as a multi-select.
+  // Answers the chosen destination class names, or undefined when there is nowhere to move or the
+  // user cancels.
+  private async pickInstVarMoveTargets(
```
MatiasFernandez, line 1513: 💭 Should this get unit coverage? A few decisions here are load-bearing and nothing else pins them down: the `.reverse()` that puts the immediate superclass first (the ▲ list is actively misleading if that ever flips), the two "nothing to move to" information messages, and `canPickMany` being set on the ▼ path only. `explorerPushMethod.test.ts` looks like the closest precedent for driving an Explorer flow with a stubbed `showQuickPick`.

client/src/gemstoneExplorer.ts
```diff
@@ -1478,29 +1470,32 @@ export class ExplorerController {
...
-  // (V3), from the ivar row's context menu. The engine recompiles the affected class
-  // definitions (new versions), previews, and applies WITHOUT committing. After a
-  // successful apply the class shape changed, so re-cascade the class panes.
+  // Move an instance variable up the hierarchy (▲) or down into subclasses (▼), from the ivar
+  // row. Each arrow opens a picker of destination classes IN THE HIERARCHY — ancestors for ▲
+  // (pick one), descendants for ▼ (pick one or more) — so the two arrows cover push-up (V2),
+  // push-down to a chosen subset (V3), and move to any hierarchy class (V4). The engine
+  // recompiles the affected class definitions (new versions), previews, and applies WITHOUT
+  // committing. After a successful apply the class shape changed, so re-cascade the class panes.
   async pushInstVar(item: IvarItem, direction: 'up' | 'down'): Promise<void> {
```
MatiasFernandez, line 1479: 💭 The comment above now describes a move-with-a-picker, and you renamed the command titles to "Move Instance Variable Up/Down…", but the method is still `pushInstVar`. Should the method and command IDs be renamed?

gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
```diff
@@ -352,6 +355,196 @@ GsInstVarStructureRefactoringTest >> testPageForTokenOnAnExpiredSessionAnswersAn
...
+	 ancestor. Mid gets a simple getter for 'pushable'; move it up to Base."
+	| ref cs |
+	self compile: 'pushable ^ pushable' in: (self classNamed: 'GsVSMid').
+	ref := (GsInstVarStructureRefactoring
+		class: (self classNamed: 'GsVSMid') moveInstVar: 'pushable' toClasses: #('GsVSBase') direction: #up)
+		moveAccessors: true.
+	cs := ref changeSet.
+
+	self assert: ref decline isNil.
+	self assert: (cs changes anySatisfy: [:c | c kind = #methodAdd and: [c className = 'GsVSBase']]).
```
MatiasFernandez, line 517: 💭 These two assertions check only the change *kind* and the class, so the test would still pass if the method that travelled up were some unrelated selector rather than the accessor. `testMoveDownPartialMovesAccessorToSelectedSubclassOnly` a few tests down does pin `c selector = #pushable` — worth matching that here, given the whole point is that the *accessor* follows the ivar?

gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
```diff
@@ -352,6 +355,196 @@ GsInstVarStructureRefactoringTest >> testPageForTokenOnAnExpiredSessionAnswersAn
...
+		class: (self classNamed: 'GsVSLeaf') moveInstVar: 'leaf' toClasses: #('GsVSMid' 'GsVSBase') direction: #up.
+
+	self assert: ref decline notNil.
+	self assert: ref decline includesSubstring: 'more than one superclass'.
+	self assert: ref changeSet isEmpty
+]
+
+{ #category : 'tests - V4 move' }
+GsInstVarStructureRefactoringTest >> testMoveUpDeclinesOnSiblingCollision [
+	"'dup' is owned by both Leaf and Leaf2; pushing Leaf's up to Mid would collide with Leaf2's
```
MatiasFernandez, line 413: nit: this comment still says `Leaf2` after the sibling was renamed to `GsVSTwig` — same in `testMoveDownToASelectedSubset` below, and in both of their counterparts in the `resources/refactoring/engine-tests.gs` mirror. Worth updating so the prose matches the assertion right underneath it.

client/src/refactoring/instVarStructureCommand.ts
```diff
@@ -199,6 +203,38 @@ export async function pushInstVar(
...
+  targets: string[],
+  dict?: number | string,
+): Promise<boolean> {
+  const where =
+    direction === 'up'
+      ? `up to ${targets[0] ?? 'superclass'}`
+      : targets.length === 1
+        ? `down to ${targets[0]}`
+        : `down to ${targets.length} subclasses`;
+  const heading = `Move instance variable '${ivarName}' ${where}`;
```
MatiasFernandez, line 224: nit: `pushInstVar`'s heading named the class the ivar was leaving ("… up from GsVSMid"); this one drops it, and the multi-target case collapses to "down to 3 subclasses". Since `heading` is what the preview panel titles itself with and what the final "applied N change(s)" toast repeats, would something like `Move instance variable 'x' from Mid down to Leaf, Twig` read better — falling back to the count once there are more names than fit?

MatiasFernandez (approved):
I left some comments. Feel free to take a look at them on this PR on in a follow-up if you think they are useful

gs-src/refactoring/engine/GsInstVarStructureRefactoring.class.st
ericwinger: Tracked in #352 for follow-up. Resolving here.

gs-src/refactoring/engine/GsInstVarStructureRefactoring.class.st
ericwinger: Tracked in #352 for follow-up. Resolving here.

gs-src/refactoring/engine/GsInstVarStructureRefactoring.class.st
ericwinger: Tracked in #352 for follow-up. Resolving here.

client/src/refactoring/instVarStructureCommand.ts
ericwinger: Tracked in #352 for follow-up. Resolving here.

client/src/gemstoneExplorer.ts
ericwinger: Tracked in #352 for follow-up. Resolving here.

client/src/gemstoneExplorer.ts
ericwinger: Tracked in #352 for follow-up. Resolving here.

gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
ericwinger: Tracked in #352 for follow-up. Resolving here.

gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
ericwinger: Tracked in #352 for follow-up. Resolving here.

client/src/refactoring/instVarStructureCommand.ts
ericwinger: Tracked in #352 for follow-up. Resolving here.

PR #353, opened 2026-07-31
Issue #352: follow-ups from PR #350 review (V4 move instance variable)

ericwinger:
Addresses all nine review comments @MatiasFernandez left on PR #350, tracked in #352. Four focused commits, one per concern-area.

## Correctness (engine)
- **#1 — `#down` nested targets.** The ▼ picker lists descendants at any depth, so a user could pick a class *and* one of its own descendants; the ivar was then declared on the child while it also inherited it from the freshly-edited parent — a double declaration GemStone rejects, surfacing as a `"failed"` apply. `analyzeMove` now declines that up front (e.g. *"Cannot move x down: Leaf is already a subclass of Mid, which was also chosen; drop the redundant target."*).
- **#2 — dictionary scoping of destinations.** Destination names were resolved unscoped (global first-match), so a name shadowed across dictionaries could bind the wrong class. They're now resolved **within the source class's own lineage** (superclasses for `#up`, descendants for `#down`) via a new `ancestorsOf:` helper, falling back to a global lookup only so analysis can still report the friendlier "is not a superclass/subclass" decline.
- **#3 — precondition ordering.** The losing-class check ran *after* the plan was half-built; it's hoisted above `newIvarLists`/`planAccessorMovesFrom:` so a decline records nothing by construction (matching `analyzePushDown`).

## Client rename & hygiene
- **#5** — renamed to match the "Move Instance Variable Up/Down…" command titles: Explorer method `pushInstVar`→`moveInstVar`, command IDs `explorer.pushUp/pushDownInstVar`→`explorer.moveUp/moveDownInstVar`, and the "Push … failed" error strings. (Engine `#pushUp`/`#pushDown` ops and their `pushUpInstVar:`/`pushDownInstVar:` selectors are kept as tested special cases.)
- **#4** — deleted the now-dead `pushInstVar` client helper; its op-agnostic flow-contract tests were re-pointed to the live `moveInstVar` path so no coverage was lost.
- **#9** — the move heading now names the source class and lists destinations, falling back to a count past three: *"Move instance variable 'x' from Mid down to Leaf, Twig"* / *"… down to 4 subclasses"*.
- **#2 (comment)** — corrected the stale `hierNeighbors` "dict-scoped … stays correct" comment to describe the lineage-scoped binding.

## Tests
- **#6** — new `explorerMoveInstVarPicker.test.ts` unit-covers the ▲/▼ picker: the immediate-superclass-first `.reverse()`, both "nowhere to move" messages, and `canPickMany` on the ▼ path only.
- **#7** — the move-up accessor test now pins `c selector = #pushable`, so an unrelated selector travelling up can no longer pass.
- **#8** — stale `Leaf2` comments updated to `GsVSTwig`.

## Verification
- RB engine SUnit **49/49 on both 3.6.2 and 3.7.5**.
- Full CI-mirror matrix (both plugin worlds × both boundaries) + on-demand `test:gci`: green for this branch. The only reds are pre-existing/environmental — the documented `encodeAsUTF8` base-image gap in the gci e2e suite and a load-induced ChangeSignature timeout (its SUnit re-run 33/33) — none introduced here.

Closes #352.

### DIFF
--- a/gs-src/refactoring/engine/GsInstVarStructureRefactoring.class.st
+++ b/gs-src/refactoring/engine/GsInstVarStructureRefactoring.class.st
@@ -126,10 +126,15 @@ GsInstVarStructureRefactoring class >> class: aClass pushDownInstVar: aName [
 GsInstVarStructureRefactoring class >> class: aClass moveInstVar: aName toClasses: classNames direction: aSymbol [
 	"V4 (generalised push up/down): move aClass's own instance variable aName to the named
 	 destination class(es). aSymbol is #up (classNames is a single ancestor of aClass) or #down
-	 (classNames are descendants of aClass, one or many). Names are resolved against the
-	 environment; an unresolved name declines in analysis."
-	| ref env targets |
+	 (classNames are descendants of aClass, one or many). Each name is resolved WITHIN aClass's own
+	 lineage first -- its superclasses for #up, its descendants for #down -- so a name shadowed
+	 across dictionaries binds the class that is genuinely in aClass's hierarchy rather than an
+	 unrelated global first match. A name nowhere in the lineage falls back to a global lookup so
+	 analysis can still report the friendlier 'is not a superclass/subclass' decline; an unresolved
+	 name yields nil and declines as a destination that could not be found."
+	| ref env dir lineage targets |
 	env := GsRefactoringEnvironment new.
+	dir := aSymbol asSymbol.
 	ref := self new
 		setEnvironment: env
 		operation: #move
@@ -137,8 +142,13 @@ GsInstVarStructureRefactoring class >> class: aClass moveInstVar: aName toClasse
 		varName: aName
 		selector: nil
 		meta: false.
-	targets := classNames collect: [:n | env classNamed: n asString].
-	^ref setMoveTargets: targets direction: aSymbol asSymbol
+	lineage := dir == #up
+		ifTrue: [ref ancestorsOf: aClass]
+		ifFalse: [env descendantsOf: aClass].
+	targets := classNames collect: [:n |
+		(lineage detect: [:c | c name asString = n asString] ifNone: [nil])
+			ifNil: [env classNamed: n asString]].
+	^ref setMoveTargets: targets direction: dir
 ]
 
 { #category : 'instance creation' }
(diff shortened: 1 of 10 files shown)


gs-src/refactoring/engine/GsInstVarStructureRefactoring.class.st
```diff
@@ -126,19 +126,29 @@ GsInstVarStructureRefactoring class >> class: aClass pushDownInstVar: aName [
...
 		class: aClass
 		varName: aName
 		selector: nil
 		meta: false.
-	targets := classNames collect: [:n | env classNamed: n asString].
-	^ref setMoveTargets: targets direction: aSymbol asSymbol
+	lineage := dir == #up
+		ifTrue: [ref ancestorsOf: aClass]
+		ifFalse: [env descendantsOf: aClass].
+	targets := classNames collect: [:n |
```
MatiasFernandez, line 148: 💭 This lineage-scoped resolution is the fix for #2 (a name shadowed across dictionaries binding an unrelated global), but I don't see a test that pins it — all the fixtures install into a single `UserGlobals`, so `lineage detect:` and the old `env classNamed:` global-first path resolve to the same class. A revert to the unscoped lookup would still pass 49/49, leaving the subtle correctness fix unprotected.

Is it feasible in this harness to set up a same-named class in a second dictionary and assert the move binds the lineage class (or declines with the friendly "is not a superclass/subclass") rather than the shadow? If the fixture cost is too high I'm fine leaving it — but the case that motivated the change is exactly the one nothing currently exercises.

MatiasFernandez (approved):

gs-src/refactoring/engine/GsInstVarStructureRefactoring.class.st
ericwinger: Good catch — you're right that the single-`UserGlobals` fixtures don't distinguish the lineage path from the old global-first lookup, so nothing currently fails on a revert. Rather than expand this PR, I've filed #356 to add a self-contained cross-dictionary test (shadow class in a second dictionary ahead of `UserGlobals`, assert the move binds the real lineage class / friendly decline, torn down via `ensure:`). Tracking it there so the fix gets pinned as a focused follow-up. Thanks!

PR #354, opened 2026-07-31
Bump vscode-languageclient to 10 to unblock vsce package

MatiasFernandez:
Fixes #351

## Summary
- PR #288 patched the `brace-expansion` DoS advisory (GHSA-mh99-v99m-4gvg) with a scoped npm override forcing `vscode-languageclient`'s `minimatch` to `^10.2.5`. That pulled in the patched `brace-expansion` 5.0.8, but v9 declares `minimatch: ^5.1.0`, so npm marked the resolution invalid — `vsce`'s `npm list --production --parseable --depth=99999` check then fails with `ELSPROBLEMS`, aborting packaging and blocking the 1.8.10 release.
- Upgrades `vscode-languageclient` to `10.1.0`, which declares `minimatch: ^10.2.5` itself, so the override is unnecessary and the tree resolves cleanly. `brace-expansion` stays at the patched 5.0.8.
- v10 is exports-only (dropped the root `node.js`/`node.d.ts` shims v9 shipped). Since `client/tsconfig.json` implies `node10` module resolution (which ignores `exports`), added a typecheck-only `paths` mapping for `vscode-languageclient/node` until the repo moves to `moduleResolution: node16`/`bundler`.
- Server stays on `vscode-languageserver` 9 (LSP 3.18 is additive over 3.17; `server.ts` only advertises pre-3.17 capabilities).

## Test plan
- [x] `npm list --production --parseable --depth=99999` exits 0
- [x] `vsce ls` exits 0, lists all files including koffi's prebuilds

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ All notable changes to the **GemStone Smalltalk** extension will be documented i
 
 ### Security
 
-- **Patched the `brace-expansion` denial-of-service advisory (GHSA-mh99-v99m-4gvg).** A scoped npm override forces `vscode-languageclient`'s `minimatch` to `^10.2.5` so the only `brace-expansion` in the tree is the patched 5.0.8, avoiding a `vscode-languageclient` major bump that doesn't compile under the repo's module resolution. ([#288](https://github.com/GemTalk/Jasper/pull/288))
+- **Patched the `brace-expansion` denial-of-service advisory (GHSA-mh99-v99m-4gvg) by upgrading `vscode-languageclient` to 10.1.0.** Version 10 is an exports-only package, so `client/tsconfig.json` carries a `paths` mapping for `vscode-languageclient/node` until the repo moves off `node10` module resolution. ([#288](https://github.com/GemTalk/Jasper/pull/288), [#354](https://github.com/GemTalk/Jasper/pull/354))
 
 ## [1.8.9] - 2026-07-22
 
--- a/client/package.json
+++ b/client/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "selfsigned": "^5.5.0",
-    "vscode-languageclient": "^9.0.0"
+    "vscode-languageclient": "^10.1.0"
   },
   "devDependencies": {
     "@types/jsdom": "^28.0.3",
--- a/client/tsconfig.json
+++ b/client/tsconfig.json
@@ -3,7 +3,17 @@
   "compilerOptions": {
     "outDir": "./out",
     "rootDir": "./src",
-    "allowJs": true
+    "allowJs": true,
+    // vscode-languageclient@10 is an exports-only package: it dropped the root-level `node.js` /
+    // `node.d.ts` shims v9 shipped and declares no `main`/`types`, exposing `./node` solely through
+    // the `exports` map. tsconfig.base.json uses `"module": "commonjs"` with no `moduleResolution`,
+    // which implies `node10` — and node10 resolution ignores `exports`, so tsc cannot find
+    // `vscode-languageclient/node`. esbuild (which builds the shipped bundle) does honour
+    // `exports`, so this only ever needed to be a hint for the type checker.
+    // Delete this mapping once the project moves to `moduleResolution: node16`/`bundler`.
+    "paths": {
+      "vscode-languageclient/node": ["../node_modules/vscode-languageclient/lib/node/main"]
+    }
   },
   "include": ["src/**/*.ts", "src/**/*.js"]
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,7 +50,7 @@
       "version": "0.1.0",
       "dependencies": {
         "selfsigned": "^5.5.0",
-        "vscode-languageclient": "^9.0.0"
+        "vscode-languageclient": "^10.1.0"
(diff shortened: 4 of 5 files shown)


client/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "selfsigned": "^5.5.0",
-    "vscode-languageclient": "^9.0.0"
+    "vscode-languageclient": "^10.1.0"
```
ericwinger, line 7: Checked the two things that would have made this risky, and both are fine:

- **`engines.vscode`** — v10 requires `^1.91.0`; the extension declares `^1.101.0`, so no install-time or marketplace conflict.
- **The security property is strengthened, not just preserved** — this lockfile resolves to exactly one `brace-expansion` (5.0.8, patched) and one `minimatch` (10.2.5), with no stale duplicates anywhere in the tree. The override produced an npm-*invalid* resolution; this produces a clean one, which is exactly why `npm list … --depth=99999` stops failing.

One thing worth recording somewhere: the client now pulls `vscode-languageserver-protocol` **3.18.2** while the server stays on `vscode-languageserver` 9 (**3.17.5**). Your reasoning holds — I verified `server/src/server.ts:61-62` advertises only `textDocumentSync: Full` plus a `workspace.configuration` check, all pre-3.17, and 3.18 is additive. The risk isn't today, it's the next person who bumps one side not knowing the other is deliberately pinned. A one-line note in `server/package.json` would cover it. Non-blocking.

client/tsconfig.json
```diff
@@ -3,7 +3,17 @@
...
     "rootDir": "./src",
-    "allowJs": true
+    "allowJs": true,
+    // vscode-languageclient@10 is an exports-only package: it dropped the root-level `node.js` /
+    // `node.d.ts` shims v9 shipped and declares no `main`/`types`, exposing `./node` solely through
+    // the `exports` map. tsconfig.base.json uses `"module": "commonjs"` with no `moduleResolution`,
+    // which implies `node10` — and node10 resolution ignores `exports`, so tsc cannot find
+    // `vscode-languageclient/node`. esbuild (which builds the shipped bundle) does honour
+    // `exports`, so this only ever needed to be a hint for the type checker.
+    // Delete this mapping once the project moves to `moduleResolution: node16`/`bundler`.
```
ericwinger, line 13: This removal instruction may be optimistic. Upstream's manifest for 10.1.0:

```
'./node': { node: './lib/node/main.js', types: './lib/node/main.d.ts' }
```

`exports` conditions are order-sensitive, and `node` precedes `types` here. Under `node16` TS matches the `node` condition too, so it can land on `main.js` instead of `main.d.ts` — usual symptom being an implicit-`any` import rather than clean resolution. The mapping may still be needed after the migration, or need a different workaround.

Suggest softening to something like: *"Re-check this mapping when the project moves to `moduleResolution: node16`/`bundler` — upstream orders `node` before `types` in its exports map, so it may still be required."* No code change.

For what it's worth I confirmed the rest of the mapping is exactly right: target resolves to upstream's real `./node` entry, `paths` without `baseUrl` is valid on TS ≥4.1 and resolves relative to this file, only `extension.ts:10` and `extension.test.ts:5` touch the package (nothing imports the root `.`, so one entry suffices), and no other tsconfig compiles that chain — `tsconfig.bin.json` includes only `bin/**/*.ts` and never reaches `extension.ts`.

CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ All notable changes to the **GemStone Smalltalk** extension will be documented i
 
 ### Security
 
-- **Patched the `brace-expansion` denial-of-service advisory (GHSA-mh99-v99m-4gvg).** A scoped npm override forces `vscode-languageclient`'s `minimatch` to `^10.2.5` so the only `brace-expansion` in the tree is the patched 5.0.8, avoiding a `vscode-languageclient` major bump that doesn't compile under the repo's module resolution. ([#288](https://github.com/GemTalk/Jasper/pull/288))
+- **Patched the `brace-expansion` denial-of-service advisory (GHSA-mh99-v99m-4gvg) by upgrading `vscode-languageclient` to 10.1.0.** Version 10 is an exports-only package, so `client/tsconfig.json` carries a `paths` mapping for `vscode-languageclient/node` until the repo moves off `node10` module resolution.
```
ericwinger: This entry now carries no PR backlink. Rewriting in place is right — both this and #288's entry are still `[Unreleased]`, so no shipped history is being rewritten — but every sibling bullet ends with a `([#NNN])` link and this one ends with prose.

Suggest `([#354](https://github.com/GemTalk/Jasper/pull/354))`, and possibly keeping #288 alongside it: someone tracing GHSA-mh99-v99m-4gvg will find the override commit and wonder why it disappeared. e.g. `([#288](https://github.com/GemTalk/Jasper/pull/288), [#354](https://github.com/GemTalk/Jasper/pull/354))`.

ericwinger (approved):
Claude only found low priority items.  I think its good to commit.

client/tsconfig.json
MatiasFernandez: it's ok for now, I have another PR that handles this temporary patch

client/package.json
MatiasFernandez: I'll track this as a follow-up

PR #355, opened 2026-07-31
Move TypeScript to node16 module resolution

MatiasFernandez:
## Summary
- `tsconfig.base.json` now sets `"module": "node16"` instead of `commonjs`, so `tsc`'s resolution algorithm can read `package.json` `exports` maps the way Node/esbuild/vitest already do.
- Deletes the `paths` shim in `client/tsconfig.json` that worked around `vscode-languageclient` v10's exports-only layout, and the node10 pin in `acceptance/tsconfig.json`.
- Fixes a real runtime regression this surfaces: `readWebviewScript` (`client/src/webviewAssets.ts`) used to check the `__dirname`-adjacent path first, which under `node16` could pick up a stale, `exports`-stamped compiled copy of a webview script from `client/out` during F5 debugging and throw in the webview DOM. It now only ever reads from the sibling `src/` tree.
- Adds `.js` to the 104 `vi.mock('vscode', () => import(...))` specifiers that needed it, and drops a redundant dynamic `fs` import in `extension.ts`.
- Second commit enables `isolatedModules` as a separate, easily-bisectable follow-up (zero new errors).

## Test plan
- [x] `npm run compile` — all four projects (client, client-bin, server, mcp-server), 0 errors
- [x] `npm run bundle` — esbuild output rewritten cleanly
- [x] `npm test` — all three workspaces green (4287+322+92 tests passing)
- [x] `npm run lint && npm run format:check` — clean
- [x] Manual F5 check: connect to a stone, open a webview-script-backed panel (Debugger view / a refactoring panel), confirm no `ReferenceError: exports is not defined` in the Dev Tools console — this repo's own CLAUDE.md flags F5 as a human-only verification step

### DIFF
--- a/client/src/webviewAssets.ts
+++ b/client/src/webviewAssets.ts
@@ -6,17 +6,17 @@ import * as path from 'path';
 // workspace conventions). The scripts live alongside their owning panels, so a
 // feature folder passes its own directory name as `subdir`.
 //
-// The lookup has to work in two layouts:
+// The lookup has to work in three layouts:
 //   • Bundled runtime — `extension.js` sits in `client/out/`, so `__dirname` is
 //     `client/out` and the scripts are at `../src/<subdir>/`.
 //   • From source (unit tests, ts-node) — `__dirname` is `client/src` and the
-//     scripts sit in `<subdir>/` beside this module.
-// We try the source-adjacent path first, then fall back to the bundled path (and
-// let that path surface in the error if neither exists).
+//     scripts sit in `<subdir>/` beside this module; `../src/<subdir>/` collapses
+//     back to the same place.
+//   • tsc's per-file `client/out` emit (the F5 debug build) — `allowJs` compiles
+//     these scripts into `client/out/<subdir>/` too, but that copy is stamped as
+//     a CJS module (`exports` reference) and throws in the webview DOM, which has
+//     no module system. We must not pick it up, so we always resolve against
+//     `../src/` first and never fall back to the `__dirname`-adjacent path.
 export function readWebviewScript(fileName: string, subdir?: string): string {
-  const beside = path.join(__dirname, subdir ?? '', fileName);
-  if (fs.existsSync(beside)) {
-    return fs.readFileSync(beside, 'utf8');
-  }
   return fs.readFileSync(path.join(__dirname, '..', 'src', subdir ?? '', fileName), 'utf8');
 }
--- a/client/tsconfig.json
+++ b/client/tsconfig.json
@@ -3,17 +3,7 @@
   "compilerOptions": {
     "outDir": "./out",
     "rootDir": "./src",
-    "allowJs": true,
-    // vscode-languageclient@10 is an exports-only package: it dropped the root-level `node.js` /
-    // `node.d.ts` shims v9 shipped and declares no `main`/`types`, exposing `./node` solely through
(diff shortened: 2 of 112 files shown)


client/src/webviewAssets.ts
```diff
@@ -6,17 +6,17 @@ import * as path from 'path';
...
+//     these scripts into `client/out/<subdir>/` too, but that copy is stamped as
+//     a CJS module (`exports` reference) and throws in the webview DOM, which has
+//     no module system. We must not pick it up, so we always resolve against
+//     `../src/` first and never fall back to the `__dirname`-adjacent path.
 export function readWebviewScript(fileName: string, subdir?: string): string {
-  const beside = path.join(__dirname, subdir ?? '', fileName);
-  if (fs.existsSync(beside)) {
-    return fs.readFileSync(beside, 'utf8');
-  }
   return fs.readFileSync(path.join(__dirname, '..', 'src', subdir ?? '', fileName), 'utf8');
```
ericwinger, line 21: **The real runtime fix — correct, but the one thing automation can't reach.** Reading only `../src/…` is right: the `__dirname`-adjacent copy in `out/` is the `allowJs` compile output, which node16 stamps as CJS (`exports`) and throws in the webview DOM. Packaging is covered (`.vscodeignore` whitelist L63–73 + the scan test), and the unit test pins this path.

But the tests mock `fs` and only assert the resolved *path* — they can't exercise the actual webview-DOM load that regressed. Since that failure (`ReferenceError: exports is not defined`) is exactly what this PR fixes and nothing automated covers it, worth doing the F5 check the PR lists (open a Debugger view / refactoring panel, confirm a clean console) before merge.

client/tsconfig.json
```diff
@@ -3,17 +3,7 @@
...
-    // which implies `node10` — and node10 resolution ignores `exports`, so tsc cannot find
-    // `vscode-languageclient/node`. esbuild (which builds the shipped bundle) does honour
-    // `exports`, so this only ever needed to be a hint for the type checker.
-    // Delete this mapping once the project moves to `moduleResolution: node16`/`bundler`.
-    "paths": {
-      "vscode-languageclient/node": ["../node_modules/vscode-languageclient/lib/node/main"]
-    }
+    "allowJs": true
   },
   "include": ["src/**/*.ts", "src/**/*.js"]
```
ericwinger: LOW / optional: with `allowJs` + `"src/**/*.js"` in `include`, tsc still emits these webview scripts into `out/<subdir>/` as the CJS-stamped copies this PR now deliberately avoids reading — they're still generated, just dead. If you'd rather remove the footgun at the source than route around it, dropping `src/**/*.js` from `include` (or excluding the `*View.js` scripts) would stop tsc producing the broken copies at all. Not blocking.

ericwinger (approved):
Lots of files but claude didn't find anything super scary. 
Probably should at least consider doing the F5 in a follow-up. 
Some of my other PR's will need .js on their vscode mocks also. 
Hope the documentation that an agent might read is up to date.

client/tsconfig.json
MatiasFernandez: Fixed!

client/src/webviewAssets.ts
MatiasFernandez: I already tested this but forgot to mark the test plan checkbox

PR #357, opened 2026-07-31
Enable vitest/no-conditional-expect

MatiasFernandez:
## Summary
- Enables `vitest/no-conditional-expect` (scoped down from a too-broad draft that flipped on all five deferred `vitest.configs.recommended` rules at once); the other four (`expect-expect`, `no-standalone-expect`, `no-mocks-import`, `no-disabled-tests`) stay deferred for a later PR in this series. Disabled for `client/src/__tests__/gci/**`, which is mid-migration to a different test approach.
- Fixes the 43 real violations this exposed, grouped by shape:
  - Collapse discriminated-union narrowing guards (`expect(x.kind).toBe(...); if (x.kind === ...) {...}`) into a single `toMatchObject` call in `parser.test.ts` and `mcpServerTreeProvider.test.ts`.
  - Add a `captureError()` test helper so a thrown error's properties can be asserted outside the `try/catch` in `executeFetchStringWithLimit.test.ts`.
  - Close several vacuous-guard gaps (`if (x) expect(...)` with no `else`) in `debugQueries`, `keybindings`, `enhancedInspectorGetViewSpecs`, and `querySunit.integration` tests.
  - `stepping.integration.test.ts` had a real silent-pass gap: nothing asserted `ranToCompletion` ever became true. Making that assertion real exposed that the 3-iteration step cap was never enough on this (interpreted-only, Darwin/ARM) stone — raised to 25 with an explanation.
  - Split `queryPython.integration.test.ts`'s Grail present/absent branching into two separately gated tests via new `requireGrail`/`requireGrailAbsent` helpers, mirroring the existing `requireServerPluginFeature` gate pattern.

## Test plan
- [x] `npm run lint` — zero `vitest/no-conditional-expect` violations outside `client/src/__tests__/gci/**`
- [x] `npm run format:check`
- [x] `npm run compile`
- [x] `npm test` against a live stone — 4202 passed, 63 skipped (Grail-gated tests skip since this stone doesn't have Grail installed), 0 failed

### DIFF
--- a/client/src/__tests__/debugQueries.test.ts
+++ b/client/src/__tests__/debugQueries.test.ts
@@ -222,9 +222,12 @@ describe('debugQueries', () => {
       debug.getObjectClassName(session, RECEIVER_OOP);
 
       const performCalls = (session.gci.GciTsPerform as ReturnType<typeof vi.fn>).mock.calls;
-      for (const call of performCalls) {
-        const selectorStr = call[3] as string | null;
-        if (selectorStr) expect(selectorStr).not.toContain(' ');
+      const selectorStrs = performCalls
+        .map((call) => call[3] as string | null)
+        .filter((s): s is string => s !== null);
+      expect(selectorStrs.length).toBeGreaterThan(0);
+      for (const selector of selectorStrs) {
+        expect(selector).not.toContain(' ');
       }
 
       const fetchBytesCalls = (session.gci.GciTsPerformFetchBytes as ReturnType<typeof vi.fn>).mock
@@ -503,9 +506,12 @@ describe('debugQueries', () => {
       debug.getInstVarNames(session, RECEIVER_OOP);
 
       const performCalls = (session.gci.GciTsPerform as ReturnType<typeof vi.fn>).mock.calls;
-      for (const call of performCalls) {
-        const selectorStr = call[3] as string | null;
-        if (selectorStr) expect(selectorStr).not.toContain(' ');
+      const selectorStrs = performCalls
+        .map((call) => call[3] as string | null)
+        .filter((s): s is string => s !== null);
+      expect(selectorStrs.length).toBeGreaterThan(0);
+      for (const selector of selectorStrs) {
+        expect(selector).not.toContain(' ');
       }
 
       const fetchBytesCalls = (session.gci.GciTsPerformFetchBytes as ReturnType<typeof vi.fn>).mock
--- a/client/src/__tests__/debugQueries.test.ts
+++ b/client/src/__tests__/debugQueries.test.ts
@@ -110,6 +110,24 @@ function receiverClassName(oop: bigint): string {
   return 'Object';
 }
 
+function expectNoMultiWordSelectors(session: ActiveSession): void {
+  const performCalls = (session.gci.GciTsPerform as ReturnType<typeof vi.fn>).mock.calls;
+  const selectorStrs = performCalls
+    .map((call) => call[3] as string | null)
+    .filter((s): s is string => s !== null);
+  expect(selectorStrs.length).toBeGreaterThan(0);
+  for (const selector of selectorStrs) {
+    expect(selector).not.toContain(' ');
+  }
+
+  const fetchBytesCalls = (session.gci.GciTsPerformFetchBytes as ReturnType<typeof vi.fn>).mock
+    .calls;
+  for (const call of fetchBytesCalls) {
+    const selector = call[2] as string;
+    expect(selector).not.toContain(' ');
+  }
+}
+
 describe('debugQueries', () => {
   describe('getStepPoint', () => {
     const GS_PROCESS = 9000n;
@@ -221,21 +239,7 @@ describe('debugQueries', () => {
       const session = createMockSession();
       debug.getObjectClassName(session, RECEIVER_OOP);
 
-      const performCalls = (session.gci.GciTsPerform as ReturnType<typeof vi.fn>).mock.calls;
-      const selectorStrs = performCalls
-        .map((call) => call[3] as string | null)
-        .filter((s): s is string => s !== null);
-      expect(selectorStrs.length).toBeGreaterThan(0);
-      for (const selector of selectorStrs) {
-        expect(selector).not.toContain(' ');
-      }
-
-      const fetchBytesCalls = (session.gci.GciTsPerformFetchBytes as ReturnType<typeof vi.fn>).mock
-        .calls;
-      for (const call of fetchBytesCalls) {
-        const selector = call[2] as string;
-        expect(selector).not.toContain(' ');
-      }
+      expectNoMultiWordSelectors(session);
     });
   });
 
@@ -505,21 +509,7 @@ describe('debugQueries', () => {
       const session = createMockSession();
       debug.getInstVarNames(session, RECEIVER_OOP);
 
-      const performCalls = (session.gci.GciTsPerform as ReturnType<typeof vi.fn>).mock.calls;
-      const selectorStrs = performCalls
-        .map((call) => call[3] as string | null)
-        .filter((s): s is string => s !== null);
-      expect(selectorStrs.length).toBeGreaterThan(0);
-      for (const selector of selectorStrs) {
-        expect(selector).not.toContain(' ');
-      }
-
-      const fetchBytesCalls = (session.gci.GciTsPerformFetchBytes as ReturnType<typeof vi.fn>).mock
-        .calls;
-      for (const call of fetchBytesCalls) {
-        const selector = call[2] as string;
-        expect(selector).not.toContain(' ');
-      }
+      expectNoMultiWordSelectors(session);
     });
   });
 
--- a/client/src/__tests__/executeFetchStringWithLimit.test.ts
+++ b/client/src/__tests__/executeFetchStringWithLimit.test.ts
@@ -45,7 +45,7 @@ describe('executeFetchStringWithLimit', () => {
     expect(session.gci.GciTsExecuteFetchBytes).not.toHaveBeenCalled();
   });
 
-  it('throws a BrowserQueryError carrying the GCI error number', () => {
+  it('throws a BrowserQueryError with the underlying GCI error message', () => {
     const session = makeSession({
       GciTsExecuteFetchBytes: vi.fn(() => ({
         bytesReturned: -1,
@@ -53,12 +53,8 @@ describe('executeFetchStringWithLimit', () => {
         err: { number: 2010, message: 'boom' },
       })),
     });
-    try {
-      executeFetchStringWithLimit(session, 'l', 'C', 1024);
-      expect.unreachable('should have thrown');
-    } catch (e) {
-      expect(e).toBeInstanceOf(BrowserQueryError);
-      expect((e as BrowserQueryError).gciErrorNumber).toBe(2010);
-    }
+
+    expect(() => executeFetchStringWithLimit(session, 'l', 'C', 1024)).toThrow(BrowserQueryError);
+    expect(() => executeFetchStringWithLimit(session, 'l', 'C', 1024)).toThrow('boom');
   });
 });
--- a/client/src/__tests__/mcpServerTreeProvider.test.ts
+++ b/client/src/__tests__/mcpServerTreeProvider.test.ts
@@ -4,7 +4,7 @@ vi.mock('vscode', () => import('../__mocks__/vscode.js'));
 
 import { ActiveSession } from '../sessionManager';
 import { writeOwnerSidecar } from '../mcpOwnerSidecar';
-import { resolveOwnership, renderOwnership } from '../mcpServerTreeProvider';
+import { McpOwnership, resolveOwnership, renderOwnership } from '../mcpServerTreeProvider';
 import { safelyRemovePath, temporarySidecarPath } from './support/file';
 
 function fakeSession(id = 7): ActiveSession {
@@ -45,12 +45,12 @@ describe('resolveOwnership', () => {
       getSession: () => session,
       sidecarPath,
     });
-    expect(ownership.kind).toBe('this');
-    if (ownership.kind === 'this') {
-      expect(ownership.selectedSession).toBe(session);
-      expect(ownership.socketPath).toBe('/tmp/sock');
-      expect(ownership.httpsUrl).toBe('https://127.0.0.1:27101/sse');
-    }
+    expect(ownership).toMatchObject({
+      kind: 'this',
+      socketPath: '/tmp/sock',
+      httpsUrl: 'https://127.0.0.1:27101/sse',
+    });
+    expect((ownership as Extract<McpOwnership, { kind: 'this' }>).selectedSession).toBe(session);
   });
 
   it('returns "other" with the sidecar info when another live window owns the socket', () => {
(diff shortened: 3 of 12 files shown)


client/src/__tests__/executeFetchStringWithLimit.test.ts
```diff
@@ -53,12 +54,8 @@ describe('executeFetchStringWithLimit', () => {
...
-    try {
-      executeFetchStringWithLimit(session, 'l', 'C', 1024);
-      expect.unreachable('should have thrown');
-    } catch (e) {
-      expect(e).toBeInstanceOf(BrowserQueryError);
-      expect((e as BrowserQueryError).gciErrorNumber).toBe(2010);
-    }
+    const error = captureError(() => executeFetchStringWithLimit(session, 'l', 'C', 1024));
+    expect(error).toBeInstanceOf(BrowserQueryError);
+    expect((error as BrowserQueryError).gciErrorNumber).toBe(2010);
```
npapagna: It looks like `gciErrorNumber` only ever earns its keep inside our tests. What's making us check it separately here, if the code's already riding along in the message? How would `toThrowInstanceOf` fall short if we leaned on that instead?

client/src/queries/__tests__/support/grailAvailability.ts
```diff
@@ -0,0 +1,36 @@
+import type { TestContext } from 'vitest';
```
npapagna, line 1: It seems like `require` in the filename might be promising something narrower than what's actually in there. What would it take to just call this `grail.ts`?

client/src/queries/__tests__/queryPython.integration.test.ts
```diff
@@ -30,15 +31,9 @@ describe('python queries (integration)', () => {
   describe('eval_python', () => {
     // Simplest possible success case — any encoding bug in the success
     // path would surface as the wrong digits or visible NUL bytes.
-    it('evaluates a basic Python expression', () => {
+    it('evaluates a basic Python expression', (ctx) => {
+      requireGrail(ctx, exec);
```
npapagna, line 35: I really like this!
It sounds like this pattern leans on the dev knowing something extra going in, like whether it has to be the first call in the test. What happens if someone gets that order wrong?
How would something like `onGrailIt('test name', () => { ... })` change that, the same pattern we're planning for posix tests?
What's stopping us from keeping `requireGrail` private, so `onGrailIt` is the only door in?
And when there's more than one `onGrailIt` in the same test, how do we want that handled, is that where `onGrailDescribe` comes in?

client/src/queries/python.ts
```diff
@@ -59,6 +59,16 @@ scopes ifNotNil: [scopes removeKey: '${escScope}' ifAbsent: []].
   return execute(code);
 }
 
+// Probe for Grail's presence: mirrors the same dispatcher lookup the queries
+// above do internally, so tests can gate on the same condition that decides
+// whether evalPython/compilePython run the real path or emit GRAIL_HINT.
+export function isGrailInstalled(execute: QueryExecutor): boolean {
```
npapagna, line 65: Found by Claude (AI code review): `isGrailInstalled` compares this GciTsPerform-fetched printString with strict `===`, but every other `printString === 'true'` check in this codebase (`browserQueries.ts:357,381,397`) trims first. If the fetch ever returns a trailing newline or space, this always evaluates to false, which would make `requireGrail` skip every Grail-present test unconditionally and let the Grail-absent test run and fail on a Grail-installed stone. Worth trimming here too.

client/src/queries/__tests__/querySunit.integration.test.ts
```diff
@@ -206,10 +206,10 @@ describe('SUnit queries (integration)', () => {
         SUNIT_PROBE_TEST_CLASS,
         SUNIT_PROBE_FAILING_SELECTOR,
       );
-      if (details.stackReport !== undefined) {
-        expect(details.stackReport.length).toBeGreaterThan(0);
-        expect(details.stackReport).not.toContain('\0');
-      }
+      const { stackReport } = details;
```
npapagna: Found by Claude (AI code review): `stackReportShapeIsValid = stackReport === undefined || (...)` is true whenever `stackReport` is undefined, so this still passes with zero real assertion on any stone where the toggle isn't honored. This reproduces the same vacuous-guard shape the PR fixed elsewhere (debugQueries, keybindings, enhancedInspectorGetViewSpecs), just moved into a boolean instead of an `if`.

client/src/__tests__/mcpServerTreeProvider.test.ts
```diff
@@ -55,12 +55,12 @@ describe('resolveOwnership', () => {
...
     });
-    expect(ownership.kind).toBe('this');
-    if (ownership.kind === 'this') {
-      expect(ownership.selectedSession).toBe(session);
-      expect(ownership.socketPath).toBe('/tmp/sock');
-      expect(ownership.httpsUrl).toBe('https://127.0.0.1:27101/sse');
-    }
+    expect(ownership).toMatchObject({
+      kind: 'this',
+      selectedSession: session,
```
npapagna: Found by Claude (AI code review): folding `selectedSession: session` into `toMatchObject` swaps the old `toBe(session)` reference-identity check for structural equality. `resolveOwnership`'s whole point here is that the same session object passes through unchanged; a future regression that copies or reconstructs the session would still satisfy `toMatchObject` and pass silently. Might be worth keeping a separate `expect(ownership.selectedSession).toBe(session)` alongside a typed `toMatchObject` for the rest.

eslint.config.mjs
```diff
@@ -143,6 +144,13 @@ export default tseslint.config(
       'vitest/valid-title': 'error',
     },
   },
+  {
+    // client/src/__tests__/gci/** is mid-migration to a different test
```
npapagna, line 148: Found by Claude (AI code review): this turns `vitest/no-conditional-expect` off for the entire `client/src/__tests__/gci/**` glob rather than a scoped, single-line disable with a reason at each violating site. That directory alone has roughly 34 conditional-expect sites across 16 files, more than the other four rules still deferred for later PRs. Worth a linked issue or tracking note so this carve-out doesn't linger indefinitely.

client/src/queries/__tests__/queryPython.integration.test.ts
```diff
@@ -69,20 +64,26 @@ describe('python queries (integration)', () => {
     // Source is supplied verbatim, including single quotes. The escape
     // path is shared with every other query in the file — if it's broken,
     // every test on this branch breaks.
-    it('escapes single quotes in Python source', () => {
+    it('escapes single quotes in Python source', (ctx) => {
```
npapagna, line 67: Found by Claude (AI code review): a few of the restructured tests here (this one, lines 73-77, lines 84-91, plus enhancedInspectorGetViewSpecs.test.ts:59-65 and querySunit.integration.test.ts:203-212) run setup/exercise/assert back-to-back with no blank line separating them. Minor, but worth a pass for scanability.

client/src/queries/__tests__/support/grailAvailability.ts
```diff
@@ -0,0 +1,36 @@
...
+ * gated tests: inapplicable tests report as skipped rather than failing.
+ *
+ * Both gates decide live against the connected session via `isGrailInstalled`, so no
+ * phase or environment flag is involved.
+ */
+
+/**
+ * Present-world gate: call at the top of a test that only applies when Grail IS
+ * installed, so the test exercises Grail-dependent behavior.
+ */
```
npapagna, line 19: Found by Claude (AI code review): `requireGrail`/`requireGrailAbsent` mirror requireServerPluginFeature.ts's present/absent gating shape closely (near-identical doc comments) but as an independent, parallel mechanism rather than an extension of it. Might be worth a shared "live-probe present/absent gate" primitive both could sit on top of, so the next optional feature doesn't need a third copy.

client/src/queries/__tests__/support/grailAvailability.ts
```diff
@@ -0,0 +1,36 @@
...
+export function requireGrail(ctx: TestContext, execute: QueryExecutor): void {
+  ctx.skip(
```
npapagna, line 21: Found by Claude (AI code review): each of the 5 Grail-gated tests in queryPython.integration.test.ts calls requireGrail/requireGrailAbsent independently, and each call re-probes Grail's presence via a live GCI round-trip through isGrailInstalled. Since presence can't change mid-suite, caching the probe once (e.g. in a beforeAll) would save 4 extra round-trips per run.

client/src/__tests__/debugQueries.test.ts
```diff
@@ -222,9 +222,12 @@ describe('debugQueries', () => {
       debug.getObjectClassName(session, RECEIVER_OOP);
 
       const performCalls = (session.gci.GciTsPerform as ReturnType<typeof vi.fn>).mock.calls;
```
npapagna: Found by Claude (AI code review): this 7-line selector-filter-and-assert block is pasted verbatim at line 508 too (for getInstVarNames). Might be worth factoring into a small shared helper (e.g. expectNoMultiWordSelectors(performCalls)) so a future tweak to the check doesn't need to land in two places.

client/src/__tests__/stepping.integration.test.ts
```diff
@@ -71,21 +71,30 @@ cls compileMethod: 'outer | x | x := self inner. ^ 42' dictionaries: System myUs
...
+    const stepErrorNumbers: number[] = [];
 
     try {
-      for (let i = 0; i < 3 && !ranToCompletion; i++) {
+      // On stones without native code (e.g. Darwin/ARM — see the class comment),
+      // gciStepOverFromLevel: advances by bytecode-level step points rather than
+      // by source statement, so completing this two-statement method takes well
+      // more than a couple of steps; 25 leaves comfortable headroom above the
+      // ~16 observed in practice.
+      for (let i = 0; i < 25 && !ranToCompletion; i++) {
```
npapagna: Found by Claude (AI code review): the cap bump from 3 to 25 (with "~16 observed in practice" as the rationale) is a bigger fixed number tied to today's bytecode shape, not a mechanism that adapts. A different compiler version, CI runner, or slightly bigger test method could push past 25 again, needing the same fix once more.

server/src/parser/__tests__/parser.test.ts
```diff
@@ -269,151 +267,139 @@ describe('Parser', () => {
       const expr = findMessages('foo 3 @env2:squared');
       expect(expr.messages).toHaveLength(1);
       const msg = expr.messages[0];
-      expect(msg.kind).toBe('UnaryMessage');
-      if (msg.kind === 'UnaryMessage') {
-        expect(msg.selector).toBe('squared');
-        expect(msg.envSpecifier).toBe('@env2:');
-      }
+      expect(msg).toMatchObject({
```
npapagna: Found by Claude (AI code review): collapsing the discriminated-union if (x.kind === 'Y') {...} guards into toMatchObject({kind: 'Y', ...}) trades TypeScript's compile-time narrowed-field checking for a runtime-only structural match across roughly 13 sites here. No coverage is lost today, but a renamed or removed AST field will now surface as a runtime match failure on the next test run instead of a build error.

npapagna (approved):
This is great @MatiasFernandez!

Left a couple of comments

client/src/queries/python.ts
MatiasFernandez: I think the risk of `notNil printString` returning something other than `'true'` is small, and the impact of that issue is visible; tests will fail. For this function, I copied the approach the production code uses to identify whether Grail is present in the stone, which I think could be updated/improved. But given Grail is still under development and the final mechanism to detect Grail installed is probably not yet defined, I'm good leaving this as-is for now.

eslint.config.mjs
MatiasFernandez: I'm actively migrating those tests out of that folder. So those files are going to be gone soon

client/src/__tests__/executeFetchStringWithLimit.test.ts
MatiasFernandez: This is just transforming the approach we use to run the asserts; removing `gciErrorNumber` would require changes in many other unrelated test files. If we want to deprecate `gciErrorNumber`, I think we should do that in another PR. I'm not sure what we gain/lost if we do so tho. Do you think we'll never need to programmatically access the code in the raised error?

client/src/queries/__tests__/support/grailAvailability.ts
MatiasFernandez: Fixed

client/src/__tests__/executeFetchStringWithLimit.test.ts
npapagna: We have no evidence we need that in production ATM.
The reason I suggested to replace this single instance of that (not every usage of `gciErrorNumber`) is because  we needed to introduce another support helper to capture the error.

client/src/__tests__/debugQueries.test.ts
MatiasFernandez: Fixed

client/src/__tests__/mcpServerTreeProvider.test.ts
MatiasFernandez: Fixed

client/src/queries/__tests__/support/grailAvailability.ts
MatiasFernandez: I'm still discovering the gating needs for the tests I'm migrating from the "gci" folder into "integration" tests, so I'll defer the refactor here so I don't generalize too often. But I agree there's some room for improvement here

client/src/__tests__/executeFetchStringWithLimit.test.ts
MatiasFernandez: Fixed

client/src/__tests__/stepping.integration.test.ts
MatiasFernandez: I agree this is not great, but the real fix is big, and we can't just rely on Vitest timeouts here because this is sync code that can't be interrupted by Vitest's async timeout mechanism. 

I applied a change to make it more obvious that this should not fail if there's a compiler change, but still keeping a guard to avoid an endless iteration.

I also simplified the code used to test this to reduce the number of steps needed.

client/src/queries/__tests__/queryPython.integration.test.ts
MatiasFernandez: In parallel to this work, I'm working on other conditional test gating for the tests I'm migrating from the legacy "gci" folder. I'm still discovering the mechanisms that are helpful for this gating, so I want to continue working on that a bit more before cleaning up this case, because I'll need to solve similar problems on that parallel project and even tolerate more variations. So, using that project to test the alternatives is better than starting the refactoring in this PR

client/src/queries/__tests__/queryPython.integration.test.ts
MatiasFernandez: Fixed

client/src/queries/__tests__/support/grailAvailability.ts
MatiasFernandez: Thanks for flagging this — I'm going to leave it as is for now. The current cost is small (a handful of extra GCI round trips in an already-live-stone integration suite), and caching the probe risks becoming confusing/stale later (e.g. if a test starts using logout/relogin against a different session, or gating gets restructured). I'm actively working on other test-gating changes in this same area right now, so I'll fold in probe-caching as part of that broader pass if it still makes sense then rather than as an isolated change here.

client/src/queries/__tests__/querySunit.integration.test.ts
MatiasFernandez: I verified that all stone versions in CI include a stackReport, so the conditionality here is actually not needed. I removed the check and folded the asserts into a previous test.

server/src/parser/__tests__/parser.test.ts
MatiasFernandez: Re-added typed literals to ensure we also get type errors

PR #358, opened 2026-07-31
CI gate: catch a semantically-invalid production dependency tree

MatiasFernandez:
## Summary
- Add a `package-check` CI job that runs `npm run package` (vsce's full packaging path), catching the class of failure that has broken `vsce package` on `main` twice (#351), and that `npm ci`/lint/compile/test never touch.
- Document the underlying rule using a Claude rule file: an `overrides` entry may only narrow within the declaring package's existing range.

## Failure screenshot

image

## Test plan
- [x] `npm run lint:github:workflows` — new job's `timeout-minutes` satisfies the enforced-timeout linter
- [x] `npm run lint && npm run format:check && npm run compile && npm test` — full suite passes
- [x] `package-check` appears in GitHub Actions and is wired into `ci-complete` as a required, non-skippable job

### DIFF
--- a/.github/workflows/health-check.yml
+++ b/.github/workflows/health-check.yml
@@ -210,6 +210,37 @@ jobs:
           path: test-results/skips-${{ matrix.gemstone-version }}-${{ matrix.node-version || 'nvmrc' }}-plugin.json
           if-no-files-found: ignore
 
+  package-check:
+    name: Verify vsce packaging
+    timeout-minutes: 5
+    needs: prepare
+    runs-on: ubuntu-latest
+    steps:
+      # zizmor(unpinned-uses): actions must be pinned to a commit SHA, not
+      # a mutable tag like @v6
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          # zizmor(artipacked): checkout must not persist the token past
+          # this job
+          persist-credentials: false
+      - name: Setup Node.js
+        # zizmor(unpinned-uses): actions must be pinned to a commit SHA,
+        # not a mutable tag like @v6
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'npm'
+      - name: Install dependencies
+        run: npm ci
+      # Runs vsce's full packaging path (esbuild bundle + the production
+      # dependency walk `npm list --production --parseable --depth=99999`).
+      # An `overrides` entry resolved outside its declaring package's range
+      # marks that edge `invalid`, `npm list` exits non-zero, and packaging
+      # aborts — invisible to npm ci / lint / compile / test, and it has
+      # broken a release twice (see issue #351).
+      - name: Verify vsce packaging
+        run: npm run package
+
   # Single stable check for branch protection / the merge queue to require.
   # The matrix job's own check names embed the matrix values, so requiring
   # `test` directly is brittle; this fans them in to one name instead.
@@ -219,7 +250,7 @@ jobs:
   ci-complete:
     name: ci-complete
     if: always()
-    needs: [prepare, lint-workflows, lint, health-check]
+    needs: [prepare, lint-workflows, lint, health-check, package-check]
     runs-on: ubuntu-22.04
     timeout-minutes: 1
     steps:
@@ -228,11 +259,13 @@ jobs:
           if [ "${{ needs.prepare.result }}" != "success" ] ||
             [ "${{ needs.lint-workflows.result }}" != "success" ] ||
             [ "${{ needs.lint.result }}" != "success" ] ||
-            [ "${{ needs.health-check.result }}" != "success" ]; then
+            [ "${{ needs.health-check.result }}" != "success" ] ||
+            [ "${{ needs.package-check.result }}" != "success" ]; then
             echo "prepare=${{ needs.prepare.result }}"
             echo "lint-workflows=${{ needs.lint-workflows.result }}"
             echo "lint=${{ needs.lint.result }}"
             echo "health-check=${{ needs.health-check.result }}"
+            echo "package-check=${{ needs.package-check.result }}"
             exit 1
           fi
 
--- a/.claude/rules/npm-overrides.md
+++ b/.claude/rules/npm-overrides.md
@@ -1,0 +1,8 @@
+---
+paths:
+  - "package.json"
+---
+
+# npm `overrides`
+
+An `overrides` entry may only narrow *within* the declaring package's existing range — an out-of-range override breaks `vsce package`. Read [docs/how-to/npm-overrides.md](../../docs/how-to/npm-overrides.md) before adding or editing one.
--- a/docs/how-to/npm-overrides.md
+++ b/docs/how-to/npm-overrides.md
@@ -1,0 +1,33 @@
+# Using `overrides` in root `package.json`
+
+## The rule
+
+An `overrides` entry may only **narrow** the range the declaring package already permits — it pins a specific version *within* that package's own dependency range, it cannot replace the range with something outside it.
+
+If the version you need to force falls outside the declaring package's existing range, bump the parent package instead of (or as well as) adding/adjusting the override.
+
+## Why
+
+An out-of-range override makes npm mark that edge of the dependency tree `invalid`. `vsce package` refuses to package the extension when `npm ls` reports invalid edges, so a bad override silently breaks packaging rather than failing loudly at the time it's added. This has blocked a release twice.
+
+## How to apply
+
+Before adding or editing an entry in the root `package.json`'s `"overrides"` block:
+
+1. Find the package that actually depends on the module you want to override, and check the range it declares for that dependency (e.g. via `npm ls <package>` or reading its own `package.json`).
+2. Confirm the version you want to pin falls inside that range.
+3. If it doesn't, bump the parent package to a version whose declared range includes the version you need — then add the override only if one is still necessary.
+4. Run `npm run package` (or at least `npm ls`) to confirm no edge comes back `invalid`.
+
+The `package-check` CI job (see `health-check.yml`) exists specifically to catch this class of breakage before it reaches a release.
+
+## Example of the mistake
+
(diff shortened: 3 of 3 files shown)


client/src/__tests__/dependencyTreeHealth.test.ts
```diff
@@ -98,4 +99,28 @@ describe('transitive dependency vulnerability floors', () => {
...
   );
+
+  it('resolves every production dependency inside its declaring package range', () => {
+    // `npm ls` defaults to --depth=0 and never visits the offending edge, so --all is
+    // required or this assertion is permanently green. --omit=dev is required too: the
+    // dev tree carries known-benign noise (platform-specific optional deps of
+    // @rolldown/binding-wasm32-wasi; a hoisted @noble/hashes for jsdom) that never ships.
+    let stdout: string;
+    try {
+      stdout = execFileSync('npm', ['ls', '--omit=dev', '--all', '--json'], {
```
ericwinger: **Windows portability.** `execFileSync('npm', …)` throws `ENOENT` on native Windows — the executable there is `npm.cmd`, and `execFile` doesn't do `PATHEXT` resolution the way a shell would. CI is `ubuntu-latest` and WSL is Linux, so this only bites a dev running the suite in a native Windows terminal, but Jasper does support that platform.

The sibling shell-outs are all in platform-gated files (`windowsClient`, `wslFs`, `wslBridgeNetwork`), so this is the first unconditional one.

```ts
const npm = process.platform === 'win32' ? 'npm.cmd' : 'npm';
```

**Minor, same block:** if `npm` can't be spawned at all, `err.stdout` is undefined → `stdout` becomes `''` → `JSON.parse('')` throws `Unexpected end of JSON input`, which hides the real cause. Worth failing with the spawn error instead when stdout is empty.

client/src/__tests__/dependencyTreeHealth.test.ts
```diff
@@ -98,4 +99,28 @@ describe('transitive dependency vulnerability floors', () => {
       }
     },
   );
+
+  it('resolves every production dependency inside its declaring package range', () => {
+    // `npm ls` defaults to --depth=0 and never visits the offending edge, so --all is
```
ericwinger: The `invalid:`-only filter is doing real work and shouldn't be tightened later — I verified that on a current tree `npm ls --omit=dev --all --json` still reports 4 problems even with `--omit=dev`:

```
extraneous: @emnapi/core@1.10.0
extraneous: @emnapi/runtime@1.10.0
extraneous: @emnapi/wasi-threads@1.2.1
extraneous: @tybys/wasm-util@0.10.2
```

So the `@rolldown/binding-wasm32-wasi` family leaks into the production walk regardless of `--omit=dev`. A well-meaning future tightening to `problems.length === 0` would go permanently red. Might be worth saying that in the comment explicitly — the current wording implies `--omit=dev` clears the noise, when it only clears *some* of it.

client/src/__tests__/dependencyTreeHealth.test.ts
```diff
@@ -1,6 +1,7 @@
 import { describe, it, expect } from 'vitest';
 import * as fs from 'fs';
 import * as path from 'path';
+import { execFileSync } from 'child_process';
```
ericwinger, line 4: The file is now `dependencyTreeHealth.test.ts`, but the `describe` (line 37) is still `'transitive dependency vulnerability floors'` and this header still opens with *"Regression guard for the transitive dependency vulnerabilities Dependabot flagged…"*. The third invariant isn't a vulnerability floor — it's a resolution-validity check, and the file name is the only place that now says so.

Suggest renaming the `describe` to match the file (e.g. `'production dependency tree health'`) and adding a sentence to the header covering the new invariant. Cheap, and it keeps the file self-describing for whoever adds the fourth test.

.github/workflows/health-check.yml
```diff
@@ -210,6 +210,37 @@ jobs:
...
+      - name: Install dependencies
+        run: npm ci
+      # Runs vsce's full packaging path (esbuild bundle + the production
+      # dependency walk `npm list --production --parseable --depth=99999`).
+      # An `overrides` entry resolved outside its declaring package's range
+      # marks that edge `invalid`, `npm list` exits non-zero, and packaging
+      # aborts — invisible to npm ci / lint / compile / test, and it has
+      # broken a release twice (see issue #351).
+      - name: Verify vsce packaging
+        run: npm run package
```
ericwinger, line 242: Good gate — this is the one path that lint/compile/test genuinely can't see, and wiring it into `ci-complete` as non-skippable is the right call.

Two thoughts, neither blocking:

1. `npm run package` builds a `.vsix` and then discards it. Since the failure mode you're catching is the dependency walk, `npx vsce ls` would exercise the same check faster — or, if you keep full `package`, consider `actions/upload-artifact` on the `.vsix` so release candidates fall out of CI for free.
2. `timeout-minutes: 5` covers `npm ci` + esbuild + the `--depth=99999` walk on a cold cache. Fine today; worth remembering it's the first thing to trip if the dependency tree grows.

CLAUDE.md
```diff
@@ -47,4 +47,6 @@ Before considering something done, run `npm run lint && npm run format:check &&
 
 Never add `eslint-disable` or change a rule's severity to silence a lint error — fix the underlying code. If a rule genuinely can't be satisfied, add a scoped disable (single line, not file-wide) with a comment explaining why, and flag it to the user before committing. `eslint-comments/require-description` enforces that every disable comment carries a `-- reason`, so an undocumented one fails `npm run lint` outright.
 
+An `overrides` entry may only narrow *within* the declaring package's existing range. If the patched version falls outside that range, bump the parent package instead — an out-of-range override makes npm mark the edge `invalid`, which breaks `vsce package`. This has blocked a release twice.
```
ericwinger: Right call putting this in the always-loaded file — it's a prerequisite whose violation has cost two releases, which is exactly the bar for root `CLAUDE.md`, and it's not something an agent would infer from browsing the code.

One suggestion: name the failure signal so it's greppable when someone hits it cold — `npm list`/`vsce` report it as `invalid: <pkg>@<version>` and exit `ELSPROBLEMS`. A reader who has just seen that string in a terminal will then match it to this rule.

.github/workflows/health-check.yml
```diff
@@ -219,7 +250,7 @@ jobs:
   ci-complete:
     name: ci-complete
     if: always()
-    needs: [prepare, lint-workflows, lint, health-check]
+    needs: [prepare, lint-workflows, lint, health-check, package-check]
```
ericwinger, line 253: **Merge-ordering / coordination — flagging for you to decide, not asking for a change here.**

Making `package-check` a non-skippable input to `ci-complete` is correct, and it means this PR is deliberately red until the #351 fix lands. So the sequencing matters more than usual: merging this first blocks the merge queue for everyone, not just this branch.

The part I'd like your call on: the description points at `mfernandez/use-node16-modules` (#355, *Move TypeScript to node16 module resolution*) as "the actual fix", but **#354** (*Bump vscode-languageclient to 10 to unblock vsce package*) is also open and also fixes #351 — by bumping the dependency to a version that declares the patched `minimatch` itself, so the override becomes unnecessary. I verified #354's lockfile resolves to a single `brace-expansion@5.0.8` and `minimatch@10.2.5` with no invalid edges, so it does clear this gate on its own.

Those two may well be complementary rather than competing — #354 even says its `paths` shim exists *"until the repo moves to `moduleResolution: node16`"*, which is precisely #355 — but this PR's description reads as though #355 is the only fix, so I wanted to make sure the three-way ordering is intentional:

- if **#354** merges first, this gate should go green without #355;
- if **#355** merges first, does the override still need removing separately?
- either way this branch rebases last.

Whichever you intend, worth updating the description so the next reader doesn't have to reconstruct it.

ericwinger: Eric: Claude asked and I didn't quite understand this so throwing back to your court instead of trying to decide.

client/src/__tests__/dependencyTreeHealth.test.ts
ericwinger: Eric:  We haven't done a lot with Windows but it's worth thinking about these issues eventually.

ericwinger (approved):
Everything is optional.  Claude wanted a couple things highlighted here.  I'll let you decide. 

Claude: 
  1. Don't merge before the #351 fix — this is the real gate. Not a defect; the PR is intentionally red, and because
  package-check is a non-skippable input to ci-complete, merging it first blocks the merge queue for everyone, not just
  this branch. That plus the unresolved #354-vs-#355 ordering is the only thing I'd actually hold on.

  2. The one finding I'd ask for before merge: Windows portability. execFileSync('npm', …) throws ENOENT on native
  Windows — it needs npm.cmd. This isn't cosmetic: it's an unconditional test, so npm test fails outright for anyone
  running the suite in a native Windows terminal, on a platform Jasper supports. Every other shell-out in the suite lives
  in a platform-gated file. One-line fix, so cheap to just do.

.github/workflows/health-check.yml
MatiasFernandez: re: `actions/upload-artifact`, that may be useful. I need to check what's the impact of doing that. I'll leave it as future work for now

.github/workflows/health-check.yml
MatiasFernandez: This was my mistake; the description was confusing, and it confused Claude. The fix was already merged, it's ok to merge this PR on top of that

client/src/__tests__/dependencyTreeHealth.test.ts
MatiasFernandez: fair point, and I actually realied this test was overlapping and useles once we have the CI step in. So I removed it. The new CI step already catches this issue

client/src/__tests__/dependencyTreeHealth.test.ts
MatiasFernandez: I reverted this change

client/src/__tests__/dependencyTreeHealth.test.ts
MatiasFernandez: I reverted this test, we don't need it

CLAUDE.md
MatiasFernandez: I tweaked the Claude rule and added a document explaining the issue better

PR #359, opened 2026-07-31
Issue #356: pin lineage-scoped move-destination resolution with a cross-dictionary test

ericwinger:
Follow-up to the review of #353, where @MatiasFernandez noted (approved, non-blocking) that the lineage-scoped move-destination resolution had no test protecting it: every `GsInstVarStructureRefactoringTest` fixture installs into a single `UserGlobals`, so the new lineage `detect:` path and the old unscoped `env classNamed:` global-first path resolved to the same class — a revert to the unscoped lookup would still pass 49/49.

Adds three [GS SUnit] tests. **Test-only; the engine is unchanged.**

## Tests
- **`testMoveUpBindsLineageClassNotASameNamedShadowInAnotherDictionary`** (the #356 core) — binds an unrelated `GsVSBase` in a `SymbolDictionary` inserted ahead of `UserGlobals` so a global first-match lookup returns that shadow, then moves `GsVSMid`'s own `'mid'` up to `'GsVSBase'` and asserts it still binds the **real** lineage `GsVSBase` (`decline` nil, edit targets the real class).
- **`testMoveDownBindsLineageClassNotASameNamedShadowInAnotherDictionary`** — the `#down` half, which resolves through `descendantsOf:` — a different code path than `#up`'s ancestor walk, so a revert of just the down branch needs its own shadow test.
- **`testMoveUpToANonImmediateAncestor`** — covers V4's move-up-to-**any**-ancestor (`GsVSLeaf`'s `'leaf'` up to its grandparent `GsVSBase`, skipping `GsVSMid`), which nothing previously exercised.

Each shadow test builds its second dictionary inside the test and removes it in an `ensure:` block, so it cannot leak into the other tests.

## Verification
- RB engine SUnit **52/52 on both 3.6.2 and 3.7.5** (49 prior + 3 new), and 52/52 in-stone via GCI on a plugin-installed stone.
- **Revert-proof (the point):** temporarily reverting the lineage resolution to the old unscoped lookup makes exactly the two shadow tests **fail** — instead of the old "still passes 49/49". The third test is V4-capability coverage and passes either way, by design.

Closes #356.

### DIFF
--- a/gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
+++ b/gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
@@ -535,6 +535,74 @@ GsInstVarStructureRefactoringTest >> testMoveUpMovesSimpleAccessorsToAncestor [
 ]
 
 { #category : 'tests - V4 move' }
+GsInstVarStructureRefactoringTest >> testMoveUpBindsLineageClassNotASameNamedShadowInAnotherDictionary [
+	"#2 dictionary scoping: a destination name resolves within the source's OWN lineage, not by
+	 unscoped global first-match. Bind an unrelated GsVSBase in a dictionary AHEAD of UserGlobals so a
+	 global lookup returns that shadow (which is not an ancestor of GsVSMid). Moving GsVSMid's own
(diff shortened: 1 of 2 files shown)


MatiasFernandez (approved):
LGTM

PR #361, opened 2026-07-31
RB hardening: JSON-helper dedup (C1) + move/push client dedup (C2) + RBScanner wide-source fix

ericwinger:
## Summary

Three focused hardening/cleanup items for the server-side RB refactoring engine (part of #252), plus permanent tests and extensive verification. **No user-facing behavior change.**

**C1 — dedup the engine's JSON helpers.** Every `Gs*Refactoring` serializer carried a byte-identical copy of `jsonQuote:`/`jsonEscape:`/`hex2:`. The one implementation now lives in a canonical `GsRefactoringJson`; the 14 copies (+ `GsRefactoringChange>>hex2:`) are one-line delegators. `engine.gs` −465 lines. The escaper emits pure ASCII, so the client's non-blocking GCI fetch is never handed a Unicode-promoted (wide) result.

**RBScanner wide-source fix.** Feature-detected `Character>>isSqueakSeparator` compat shim so parsing a method whose stored source is a wide `DoubleByteString` (non-ASCII in a comment / string / character literal) no longer DNUs inside `RBParser` on 3.6.2 — which had been silently breaking convert-temp and push-up/down accessor detection.

**C2 — dedup the move/push client GUI.** Shared `methodRelocation{Preview,PanelHtml,Panel}` modules; `moveMethod*`/`pushMethod*` reduced to thin wrappers. Public exports/signatures unchanged.

## Tests added
- **[GS SUnit]** `GsRefactoringJsonTest` (15) — escaper correctness + the pure-ASCII invariant. `GsRefactoringParseTest` (9) — wide-source parse (comment/string/char) + `isSqueakSeparator`.
- **[client]** `jsonHelperDedup.test.ts` (9) — structural guard that exactly one real copy of each helper exists and every other class delegates.
- **[client]** `methodRelocation.test.ts` (38) — C2 parsers / HTML / edges.

## Verification
- Engine RB SUnit **437 / 437** on **both 3.6.2 and 3.7.5** (non-committing file-in + abort).
- Full client suite green; `tsc` / lint / format clean.
- Aggressive one-shot checks (all pass): golden-master (C1 static + **63,754-point** dynamic diff vs pre-dedup `main`, 0 mismatches); cross-language **Node JSON oracle** (409 inputs, 0 ASCII/round-trip failures); RBScanner **real-corpus soak** (12,000 base methods + 58 compiled wide-source across scripts, 0 wide failures); **mutation test** (each permanent test confirmed to catch its regression); **C2 golden vs main** (0 mismatches across 8 parser/label fns).

## Not included
- **C3** (family-wide preview/panel/command scaffold) is deliberately deferred to a separate PR — a large sweep; C2 is the seed for it.

> Draft: the full PR-gate suite (both boundaries + on-demand GCI + RB SUnit) will be run before this is marked Ready for review.


gs-src/refactoring/tests/GsRefactoringParseTest.class.st
```diff
@@ -0,0 +1,105 @@
+"
+Regression tests for the RBScanner non-ASCII (wide-source) parse gap. The
+vendored RBScanner classifies characters with #isSqueakSeparator; GemStone's
+Character lacked it, so parsing any method whose STORED SOURCE is a wide
+DoubleByteString (a non-ASCII character in a comment, string or character
+literal) raised MessageNotUnderstood inside RBParser -- silently breaking
+convert-temp and push-up/down accessor detection. The compat shim
```
MatiasFernandez, line 7: 💭 These pin the shim itself (`isSqueakSeparator` answers instead of DNU-ing, `RBParser parseMethod:` survives wide source), which is the right foundation — but the thing that was actually broken for users sits one level up. Every caller wraps the parse in `[RBParser parseMethod: src] on: Error do: [:e | nil]` (e.g. `GsExtractTemporaryRefactoring.class.st:194`, `GsInlineMethodRefactoring.class.st:179`), so the DNU came back as `nil` and convert-temp hard-declined / push-up quietly skipped accessor detection.

That swallow is still there, so a future regression anywhere along the path would be just as silent — and these tests wouldn't see it, because they call `RBParser` directly rather than going through a refactoring.

Is it worth one test at the refactoring level: a convert-temp or push-up analysis over a method whose stored source carries an em-dash comment, asserting it does NOT decline? That's the test that would have failed before this PR, for exactly the reason the PR describes.

client/src/refactoring/methodRelocationPreview.ts
```diff
@@ -0,0 +1,264 @@
...
+/** The engine pre-flight: the resolved target class, a global decline (if any), the
+ *  count that will move, and a per-selector verdict. */
+export interface RelocationAnalysis<S extends BaseSelectorAnalysis> {
+  targetClass: string | null;
+  globalDecline: string | null;
+  movableCount: number;
+  selectors: S[];
+}
+
+export function asCount(v: unknown): number {
```
MatiasFernandez, line 85: nit: `asCount` is the one export in this module without a TSdoc, and arguably the one whose behaviour is least guessable from the name — it clamps non-numbers, negatives, `NaN` and `Infinity` all to `0` rather than throwing or returning `undefined`. One line would help, especially for an agent deciding whether to reuse it instead of re-rolling a `typeof x === 'number'` check.

MatiasFernandez (approved):

PR #368, opened 2026-08-03
#360: harden the V1 add/remove instance-variable refactoring (deferred #346 review items)

ericwinger:
Addresses the 10 non-blocking review items deferred from #346 and tracked in #360.

## What changed
- **Prevents the mid-apply failure the reviewer flagged (item 1):** `analyzeAdd` now detects when a subclass already declares the new name as its own instance variable and declines *before* the preview, naming the offending subclass — so the reshape can't strand a half-versioned subtree. The in-panel abort banner remains as a safety net for the unforeseen/racy failures analysis can't predict.
- **In-panel failure UX (item 1):** apply failures now show a prominent banner in the preview with a direct Abort action, instead of a toast plus a secondary modal.
- **Double-apply guard (item 2):** the Apply button disables on click and every preview panel has a host-side `applying` guard.
- **No false success (item 3):** an apply error or a zero-applied result surfaces as a failure, never the success toast.
- **Add shadowing prediction (item 4):** `#add` predicts a method temporary/argument that would shadow the new variable as will-not-recompile; block-level temps/args and class-side methods are correctly ignored.
- **Instance-variable name rule (item 5):** names must start with a lowercase letter or `_`; the engine declines otherwise and the client fast-fail validator is kept in step.
- **Committing-option warning (item 6):** the migrate / delete-history preview warns when the session holds other uncommitted work that a commit would sweep along.
- **Test hardening (items 7, 8):** the copy-forward assertion is no longer vacuous (asserts an untouched method survives and isn't dropped); the SUnit fixture gained class-side methods, a class variable, and a class-instance variable, with survival asserted.
- **Docs/nits (item 9):** removed a redundant ternary and stale "move" wording from the panel strings.
- **Test-infra decision (item 10):** documented why the two committing GCI e2e tests stay in the on-demand suite rather than migrating to `useIntegrationTest` (an abort can't undo a commit); the non-committing cases already run in the CI integration suite.

## Testing
- GS SUnit, on-demand GCI, and client unit tests added/updated for each change.
- Full suite run on both the **3.6.2** and **3.7.5** boundaries, with the server plugin absent and present.
- The only reds are pre-existing/flaky and unrelated to this branch: a change-signature in-stone SUnit hitting a 5s vitest timeout under matrix load (passes idle), the documented `querySunitRunLimit` timeout, and the known 3.6.2 on-demand-GCI API gaps.

Follow-up to #360.


client/sunit2.gs
```diff
@@ -0,0 +1,12 @@
+| p |
+p := '/tmp/issue360-worktree/resources/refactoring/engine-tests.gs'.
```
MatiasFernandez: 🚫 This looks like a leftover debugging scratchpad rather than something intended to ship — it hard-codes `/tmp/issue360-worktree/...`, so it only works on the machine it was written on, and it files in `engine-tests.gs` and runs a single test method by name. Can it be dropped from the branch?

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
```diff
@@ -448,21 +495,38 @@ GsInstVarRefactoring >> applyDeselected: deselectedIds options: optsArray migrat
...
-			failures add: (Array with: change id with: change className with: e messageText)]].
-	(failures isEmpty and: [aBool or: [dBool]]) ifTrue: [
-		self commitStructuralThenMigrate: aBool deleteHistory: dBool on: failures].
+			on: Error do: [:e |
+				failures add: (Array with: change id with: change className with: e messageText)].
+		index := index + 1].
+	failures isEmpty
+		ifTrue: [
+			(aBool or: [dBool]) ifTrue: [
+				self commitStructuralThenMigrate: aBool deleteHistory: dBool on: failures]]
```
MatiasFernandez: 🚫 `commitStructuralThenMigrate:on:` appends to `failures` (its `on: Error do:` adds a `'commit'` entry), but that happens *inside* the `ifTrue:` branch — so on that path `partiallyApplied` is never assigned and keeps the `false` it was initialised with at the top.

Concretely: structural apply succeeds for 3 classes, the user ticked "migrate existing instances", `System commitTransaction` succeeds (`committed := true`) and then `migrateInstancesTo:` raises. The result is `applied:3, failed:[commit...], committed:true, partiallyApplied:false`. Client-side, `describeApplyFailure` reads `result.partiallyApplied ?? result.applied > 0` → `false ?? …` → `false`, so the banner tells the user **"Nothing was applied."** when in fact three classes were versioned *and already committed*. That's the same class of false report item 3 was meant to close.

Should `partiallyApplied` be set from `applied > 0` (or from `committed`) regardless of which branch produced the failure — e.g. compute it once after the `ifTrue:`/`ifFalse:` rather than only in the `ifFalse:`?

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
```diff
@@ -448,21 +495,38 @@ GsInstVarRefactoring >> applyDeselected: deselectedIds options: optsArray migrat
...
+		ifFalse: [
+			"The engine does NOT abort: rolling the transaction back would discard any work the user
+			 had in flight before this refactoring started, which is not ours to throw away. Instead we
+			 report whether anything was staged before the failure; the client RECOMMENDS an abort when
+			 something was, and says what an abort would cost."
+			partiallyApplied := applied > 0].
```
MatiasFernandez: 💭 `applied` is only incremented *after* `applyChange:` returns, but `applyClassChange:` isn't atomic: `makeNewVersionOf:` creates the new version (a real staged mutation, and it records nothing in `oldToNew` yet) and only then runs `copyMethodsFrom: old to: new`. If the copy raises on the very first change, `applied` is still 0, so `partiallyApplied` is `false`, and the client reports "Nothing was applied." with the Abort button hidden — while a method-less new class version sits stranded in the transaction, which is exactly the state the Abort button exists for.

The new SUnit test provokes the failure at `old isNil ifTrue: [^self error: 'Class not found']`, i.e. the one point in `applyClassChange:` where nothing has been staged yet, so it wouldn't catch this. Is it worth flipping a flag as soon as a change is *attempted* past `makeNewVersionOf:`, so the conservative "you can abort" advice is the default rather than the exception?

client/src/refactoring/instVarRefactorPanel.ts
```diff
@@ -110,23 +125,57 @@ export function showInstVarRefactorPanel(
                 migrate ? 'migrate existing instances' : '',
                 deleteHistory ? 'delete prior class versions' : '',
               ].filter(Boolean);
+              // `System commitTransaction` commits the WHOLE session transaction, not just the
+              // change staged here, so anything else the user has uncommitted rides along. The
+              // engine reports `System needsCommit` as of the preview; say so explicitly.
+              const collateral = start.outOfScope.sessionHasUncommittedChanges
```
MatiasFernandez: 💭 `start.outOfScope.sessionHasUncommittedChanges` is a snapshot of `System needsCommit` from when the preview was *built*. The preview can sit open for a long time (it's paginated, and the user may go read code before ticking migrate), so by the time this confirmation renders the session may have picked up uncommitted work that this message then doesn't mention — and the user confirms a commit that sweeps it along without warning.

The reverse is fine (a stale `true` is just a harmless over-warning), so it's only the false-negative that bites. `sessionNeedsCommit(session)` already exists in `browserQueries.ts` — and the JSDoc you added on `abortSessionTransaction` even says callers "should probe `sessionNeedsCommit` first when they need to say how much is at stake." Should the panel re-probe it here (and in `describeApplyFailure`'s input) instead of using the preview-time value?

gs-src/refactoring/engine/GsRefactoringEnvironment.class.st
```diff
@@ -191,6 +191,68 @@ GsRefactoringEnvironment >> instanceMethodsAccessing: anInstVarName inClass: aCl
...
+	name := anInstVarName asString.
+	result := OrderedCollection new.
+	aClass selectors do: [:sel | | m src tree |
+		m := aClass compiledMethodAt: sel environmentId: 0 otherwise: nil.
+		m isNil ifFalse: [
+			src := m sourceString.
+			"Cheap pre-filter: a method can only DECLARE the name if the name appears in its source
+			 at all. This runs on every add-preview over the whole subtree, so it keeps the parse --
+			 by far the expensive part -- off the overwhelming majority of methods."
+			(src indexOfSubCollection: name) > 0 ifTrue: [
```
MatiasFernandez, line 239: 💭 How expensive does this get on a wide hierarchy? This runs on *every* add-preview, over `aClass` plus all of `allSubclasses`, and `indexOfSubCollection:` is a plain substring test — so for a short or common name (`x`, `id`, `key`, `at`) the pre-filter matches almost every method that mentions it in any form, and each of those gets an `RBParser parseMethod:`. Adding an ivar to a broadly-subclassed domain base class could mean parsing thousands of methods synchronously before the preview appears.

Have you timed it against a real hierarchy of any size? If it's material, a tighter pre-filter might pay for itself cheaply — the name can only be *declared* between a `|`…`|` pair or in the pattern line, so scanning only the source up to the first statement (or requiring the match not be immediately followed by a `:` / preceded by an alphanumeric) would drop most of the parses.

gs-src/refactoring/tests/GsInstVarRefactoringTest.class.st
```diff
@@ -200,9 +237,164 @@ GsInstVarRefactoringTest >> testAddDeclinesInheritedClassVariableShadow [
...
+
+{ #category : 'tests - class side' }
+GsInstVarRefactoringTest >> testClassSideMethodsStillRunOnTheNewVersion [
+	"Copied forward AND still functional -- the class variable and class-instance variable they
+	 read have to have come across with them."
+	| newBase |
+	self add: 'tally'.
+	newBase := UserGlobals at: #GsIVBase.
+	self assert: newBase describe equals: 'base'.
+	self assert: newBase tag isNil.
```
MatiasFernandez: 💭 The comment says the class variable and class-instance variable "have to have come across with them", but `isNil` is true whether or not they came across with any *content* — `setUp` never assigns `BaseTag` or `registry`, so these two assertions can only ever be `nil` and would still pass if the new version's slots were freshly created and empty.

Would it be worth seeding them in `setUp` (`BaseTag := 'tagged'`, and a class-side setter or `instVarAt:put:` for `registry`) and asserting the *values* survive the reshape? That's the assertion that would actually catch a `makeNewVersionOf:` that recreates the class-variable association instead of carrying the existing one over. `describe equals: 'base'` already carries its weight; these two don't yet.

client/src/refactoring/instVarRefactorPanelView.js
```diff
@@ -139,7 +198,14 @@
     const handleMessage = function (msg) {
       if (!msg) return;
       if (msg.command === 'appendChanges') appendChanges(msg.html, msg.done === true);
-      else if (msg.command === 'busyDone') setBusy(false);
+      else if (msg.command === 'busyDone') {
+        setBusy(false);
+        // The host declined/aborted the apply (or a page load settled) and the panel is
+        // staying open — let Apply be pressed again.
+        setApplying(false);
```
MatiasFernandez, line 211: 💭 `busyDone` clears `applying` unconditionally, but the host keeps its own `applying` latched forever once it has posted `applyFailed` (it's deliberately "cleared only on the paths that leave the panel open"). So after a failed apply the two latches can disagree: the pager is still visible when the preview wasn't fully loaded, and the host posts `busyDone` on a `loadMore` it decides is already done — at which point the view re-enables Apply while the ✖ banner is still on screen, and clicking it does nothing because the host drops the message.

Should `showApplyFailed` set a separate terminal flag that `busyDone` won't clear, so Apply stays visibly dead once the failure banner is up?

client/src/__tests__/explorerInstVar.test.ts
```diff
@@ -77,6 +77,9 @@ describe('ExplorerController add instance variable', () => {
     expect(validate('9bad')).toBeTruthy();
     expect(validate('has space')).toBeTruthy();
     expect(validate('  ')).toBeTruthy();
+    // Kept in step with the engine's isValidIvarName: an UPPERCASE first letter reads as a global
+    // and must be rejected here too, not just declined after a round trip (#360 item 5).
+    expect(validate('Tally')).toBeTruthy();
```
MatiasFernandez: Since the whole point of this case is keeping the *message* in step with the engine (which now names the rule), `toBeTruthy()` passes for any non-empty string — including the old generic "Not a valid instance-variable name." Would `expect(validate('Tally')).toContain('lowercase')` be better? It'd mirror `testAddDeclinesUppercaseFirstLetter`, which pins the same word on the engine side.

MatiasFernandez (changes requested):
I left some comments, but the most important ones are the ones with the 🚫 emoji at the beginning.

gs-src/refactoring/engine/GsRefactoringEnvironment.class.st
ericwinger: Good point — timing it needs a real wide hierarchy, so I'm deferring this from the PR and tracking it as #369. The plan there: benchmark first, then (if it's material) tighten the pre-filter to a whole-identifier match, with the existing add-shadowing tests as the guard against dropping a genuine declaration.

client/sunit2.gs
ericwinger: Fixed — removed this file from the branch (`git rm client/sunit2.gs`). It was a leftover debugging scratchpad and hard-coded a local path.

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
ericwinger: Fixed — `partiallyApplied` is now computed once after both branches, `partiallyApplied := failures notEmpty and: [applied > 0]`, so a commit/migrate-step failure reports it too (not only a structural failure). Also refined the client message: for a failure *after* the commit, `describeApplyFailure` now says the change was already committed and an abort can't undo it (canAbort:false), instead of the false "Nothing was applied."

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
ericwinger: Fixed — promoted `applied` to an instance variable and bump it inside `applyClassChange:` right after `makeNewVersionOf:` (before `copyMethodsFrom:`). So if anything raises past the staging point, `applied > 0` and the partial state is reported rather than "Nothing was applied." (Caught a nice bug doing this: `applied` had been a method temp, so the first reinstall failed to compile until it became an ivar.)

client/src/refactoring/instVarRefactorPanel.ts
ericwinger: Fixed — the panel now re-probes `sessionNeedsCommit(session)` **live** at commit-confirm and only falls back to the preview-time snapshot if the probe can't tell (undefined). New optional handler on the panel; the command supplies `() => queries.sessionNeedsCommit(session)`. A stale `true` merely over-warns; this closes the stale-`false` false-negative you flagged.

Eric - This staleness is good to clean up

gs-src/refactoring/tests/GsInstVarRefactoringTest.class.st
ericwinger: Fixed — `setUp` now seeds values (`BaseTag := 'tagged'`, `registry := 'reg'` via a `seedClassState` class method), and the meta-side tests assert the *values* survive the reshape (`tag = 'tagged'`, `registry = 'reg'`) rather than only `isNil` — so a `makeNewVersionOf:` that recreated empty slots would now fail the test.

client/src/refactoring/instVarRefactorPanelView.js
ericwinger: Fixed — `showApplyFailed` sets a terminal flag and `setApplying` keeps Apply disabled while it's set, so a later `busyDone` (e.g. from an already-done `loadMore`) can no longer re-enable Apply under the failure banner.

client/src/__tests__/explorerInstVar.test.ts
ericwinger: Fixed — changed to `expect(validate('Tally')).toContain('lowercase')`, mirroring the engine's `testAddDeclinesUppercaseFirstLetter`, so the two copies of the rule stay in step and a regression to a generic message would fail here.

client/sunit2.gs
ericwinger: Yup.  Claude is sloppy.

ericwinger (commented):
Changes made. And one deferred for future analysis when we're doing working on performance .

MatiasFernandez (approved):

PR #371, opened 2026-08-04
Retire the duplicated refactoring-engine gci e2e suites

MatiasFernandez:
Phase 4 of the GCI-test CI migration. It was planned as "move 10 refactoring-engine e2e tests
into the integration suite"; it turned out to be a **deduplication**, because that coverage
already landed.

## What was actually found

Every file in this group already had a sibling in `client/src/refactoring/__tests__/` driving
the same query builders and parsers over an equivalent fixture, already gated with
`requireServerPluginFeature`, already running in CI across the release matrix in both the bare
and plugin-installed passes — and asserting a **superset**. The `gci` project is excluded from
`npm test`, so none of the deleted copies ever ran in CI.

Also unaccounted for in the plan: `gciInstVar.e2e.test.ts` (8 tests), same shape, and where
the interesting residue lived.

14 of the 18 tests across the four files were already covered.

| gci file | tests | outcome |
|---|---|---|
| `gciMoveMethod.e2e` | 2 | deleted; one uncovered edge ported |
| `gciPushMethod.e2e` | 3 | deleted, nothing to port |
| `gciInstVar.e2e` | 8 | 3 duplicates dropped, 2 ported, reduced to its 3 committing tests |
| `gciInstVarStructure.e2e` | 5 | 4 duplicates dropped, reduced to its 1 committing test |

## What moved into the CI suite

Not a mechanical move — only the assertions the siblings genuinely lacked:

- **`refactoringMoveMethod.integration`** — a preview where *every* selected method is
  non-movable. The sibling only ever saw the super-sender alongside movable selectors.
- **`refactoringInstVar.integration`** — the shadowed-temporary prediction (a subclass method
  whose temp is shadowed by the new inherited ivar must be reported up front); the decline that
  names the offending subclass; the **selective copy-forward control** (a method that never
  referenced the removed variable must survive the new class version and must not be reported
  as dropped — per `.claude/rules/refactoring-tests.md`, the assertion nothing else catches);
  and the staged-change count on add.

## What stays in `gci/`, and why

Four tests that **commit**: migrate instances, delete class history, a mid-apply failure over a
committed fixture, and push-up-with-migration. `useIntegrationTest` aborts after every test and
an abort cannot undo a commit. All four are already self-cleaning, so the technical work is
small — the open question is whether the harness gets a sanctioned committing mode
(`withTransientSession` / a compensating-cleanup contract) or keeps "never commit" as an
invariant and these stay out of CI. **Same question as the `gciLogin` migration**, so it is
tracked as its own phase to be settled once for both.

`.claude/rules/refactoring-tests.md` is corrected here too: it pointed at the gci project as the
home for client-side apply coverage and at `gciInstVar.e2e.test.ts` as the shape to copy.

## Verification

- `npm run lint && npm run format:check && npm run compile && npm test` — green (4729 passed,
  10 skipped; all 10 pre-existing).
- Each touched file on **3.6.2** and **3.7.5** with the server plugin installed: 35 passed, 0 skipped.
- **Bare 3.7.5** stone: the engine tests skip *with a reason* rather than fail; the two
  probe-agreement tests still pass. Correct in both CI passes.

Draft: the four parked committing tests are a deliberate omission pending the harness decision
above.


client/src/__tests__/gci/gciInstVar.e2e.test.ts
```diff
@@ -12,57 +12,41 @@ import { GCI_LIBRARY_PATH, STONE_NRS, GEM_NRS, GS_USER, GS_PASSWORD } from './gc
...
- * transient (they roll back with `System abortTransaction` in a `finally`); the two committing
- * tests (migrate instances / delete history) instead remove the class they created — and commit
- * that removal — in their `finally`, since an abort cannot undo a commit. All Smalltalk is
- * ASCII-only for 3.6.x.
+ * `GsInstVarRefactoring`); the tests skip, with a reason, otherwise. Each test is self-cleaning
+ * — it removes the committed class (and any persisted instance) and commits that removal in
+ * `finally`, leaving no residue. All Smalltalk is ASCII-only for 3.6.x.
  */
-describe('add / remove instance variable (gci e2e)', () => {
+describe('instance-variable refactoring, committing paths (gci e2e)', () => {
```
ericwinger, line 39: Keeping these committing tests here (rather than dropping them) is the right call. Flagging for the "committing harness" follow-up phase: this suite is the **only** coverage of the migrate-instances / delete-history apply paths, and it never runs in CI. That's exactly where the #368 review found a real *silent data-loss* bug — `migrateInstancesTo:` reports failures via its return value, not a raise, so an apply loop that ignores the return reports success while stranding instances. Of everything staying CI-dark, this is the highest-risk path — worth prioritizing the phase that lets it run.

ericwinger: Eric: Yes, worth flagging. If CI doesn't see these migrate instances + remove history committing tests, then we've got a test gap. It's deferred to a later phase AFAIK but we should figure out how to get these tests into CI (or test those features in a different way that will run in CI)

ericwinger (approved):
LGTM with a note about some deferred tests being the only place some functionality is tested.

client/src/__tests__/gci/gciInstVar.e2e.test.ts
MatiasFernandez: Yes, that's tracked as another phase in the test migration project. I still need to take a look at what we how now to understand how to make them work in CI

PR #372, opened 2026-08-04
GemStone Explorer: pinning, method/comment/category editing, and UX fixes

ericwinger:
## Summary

A batch of GemStone Explorer UX enhancements and fixes, plus a couple of
non-Explorer changes. Everything here is client-side (the extension + tests);
no server/engine changes. Nothing commits implicitly — every stone-mutating
action follows the existing "edit, then you commit explicitly" model.

## Explorer — Enhancements

- **Method-source editor placement, redesigned around pinning.** All method /
  definition / comment source editors now live as tabs in **one** editor group:
  a single transient (unpinned) tab that single-clicks replace, plus explicit
  **Pin** tabs beside it (native pinned-tab glyph). Replaces the old
  "open to the side" column load-balancer (deleted `explorerColumnBalance`),
  which scattered editors across up to three columns and ping-ponged
  unpredictably. Pinning never steals the view, and pinning the transient
  promotes it. The side actions are relabeled **Pin Method / Pin Definition**
  (`$(pin)`). Mirrors VS Code's own preview-tab + pinned-tabs model.
- **Remove Method.** A 🗑 button on each method row (rightmost inline + a danger
  context-menu entry) removes a method after a modal confirmation, via the same
  `removeSelector:` path the System Browser uses.
- **Editable class comments** surfaced in the Explorer.
- **Add / rename method categories, and add methods into them** from the
  Methods pane.
- **Session Commit / Abort** buttons on the Explorer title bar, so the
  transaction can be committed or aborted without switching to the Sessions
  view (they act on the selected session; the existing "unsaved edits" / "work
  to lose" confirmations still apply, and they sit to the right of Refresh).
- **Active-editor tint:** the Explorer method row backing the currently-focused
  source editor is tinted, so the tree and the open editor stay visually linked.
- **Clearer variable groups.** Under a class, the two variable groups are now
  labeled **"instance variables" / "class variables"** (was "instance" /
  "class"). When a class defines any variables, **both** groups always show —
  the empty one grayed out and caret-less — for a consistent layout. New Instance
  Variable is available from the (empty) instance side; the class side is
  informational (adding a class variable isn't offered in the Explorer).

## Explorer — Fixes

- **No duplicate editors:** clicking (or pinning) a method that's already open
  reveals the existing tab instead of opening a second copy.
- **Methods pane no longer jumps/scrolls** when a method is clicked (a
  preview-swap + redundant reveal was knocking the clicked row out of view;
  now guarded by a self-opened-URI check + an already-selected check).
- **Explorer dictionaries refresh** after server support is installed.

## Non-Explorer changes

- **Refactor… menu grouping:** the editor's Refactor context menu is split into
  separated sections (extract / inline / rename / rewrite) via `CodeActionKind`
  sub-kinds, so related refactorings cluster instead of forming one flat list.

## Testing

- **Client unit + integration suite:** 296 files, 4785 passed, 0 failed
  (`vitest --project default`, against a live 3.7.5 test stone).
- **RB refactoring engine SUnit `[GS SUnit]`:** 493 run, 0 failures, 0 errors on
  **both** boundary versions (3.6.2 and 3.7.5) — staged session-locally and
  aborted, nothing committed.
- `npm run compile` (node16), ESLint, and Prettier all clean.
- **New e2e coverage follows the integration convention** (`useIntegrationTest`
  + `*.integration.test.ts`, not the on-demand `gci/` folder), so CI's default
  project runs it automatically in both plugin passes:
  `explorerRemoveMethod.integration.test.ts`,
  `explorerMethodCategories.integration.test.ts`.
- **Full PR-gate matrix** (`test-ci-local.sh 3.6.2 3.7.5` — each version: fresh
  stone → whole suite without then with the server plugin + RB engine → on-demand
  `test:gci`): **no regressions from this branch.** The client/server/mcp suites
  pass on both versions in both plugin worlds. The only reds are pre-existing,
  on tests this branch does not touch: the expected 3.6.x on-demand GCI set
  (already known-red on 3.6.x), and two 3.7.5 *timeouts* (the change-signature
  in-stone SUnit and the `querySunitRunLimit` guard both legitimately exceed
  vitest's 5s limit) — both reproduce on `main`.


client/src/gemstoneExplorer.ts
```diff
@@ -2566,14 +2739,35 @@ export class ExplorerController {
...
+    const className = this.state.className;
+    const selector = node.info.selector;
+    const confirmed = await vscode.window.showWarningMessage(
+      `Remove method #${selector} from ${node.isMeta ? `${className} class` : className}?`,
+      { modal: true },
+      'Remove',
+    );
+    if (confirmed !== 'Remove') return;
+
+    queries.deleteMethod(session, className, node.isMeta, selector, this.state.dictIndex);
```
MatiasFernandez: 💭 `deleteMethod` reports its outcome two ways, and we're ignoring both. It returns a status string (`'Deleted: …'`, but also `'Class not found: …'` / `'Selector not found: …'`), and the underlying `removeSelector:` can raise — e.g. on a class that isn't writable in this repository. Neither is handled here, and the registration below (`registerCommand('gemstone.explorer.removeMethod', … void ctl.removeMethod(node))`) has no `.catch` either, unlike its neighbour `renameMethodCategory`, which both try/catches the query and attaches a `.catch` that surfaces the error.

So the failing path is: user confirms the modal → the removal fails → `reloadCurrentClassMethods()` redraws the pane with the method still there → no message at all. The user reasonably concludes the click didn't register and tries again.

Worth mirroring what `renameMethodCategory` does just below — try/catch + `showErrorMessage`, and check the returned string for the non-`Deleted:` cases? Related: `createNewMethod` guards with `canClassBeWritten` before opening a template precisely so a kernel-class write doesn't silently no-op; should `removeMethod` do the same before prompting?

client/src/gemstoneExplorer.ts
```diff
@@ -83,45 +86,98 @@ const VIEW_METHODS = 'gemstoneExplorerMethods';
...
+
+// Close the single reusable source tab a prior single-click opened, but only if
+// the user isn't working in it: a dirty tab has unsaved edits and a pinned tab was
+// deliberately kept — both are cases where a preview tab would likewise have
+// promoted itself to permanent rather than being replaced.
+function closeDisposableTab(uriStr: string): void {
+  for (const group of vscode.window.tabGroups.all) {
+    for (const tab of group.tabs) {
+      if (tabInputUri(tab)?.toString() !== uriStr) continue;
+      if (tab.isDirty || tab.isPinned) return;
```
MatiasFernandez: 💭 `closeDisposableTab` walks *every* tab group and matches on the URI alone, so it can close a tab that isn't ours. Concrete scenario: the user opens `Array>>at:` in the System Browser (which lands in the System Browser's own group), then single-clicks `at:` in the Explorer and then `size`. The `closeDisposableTab(at:)` call for the outgoing transient scans from group 1 and can hit the System Browser's copy first, closing it.

That's exactly the cross-browser interference `SourceEditorPlacement` exists to prevent ("It never adopts a foreign browser's group"). Since we always know the column we opened into, should the scan be scoped to `sourceColumn` — or at least require `placement` to own the tab — instead of the whole window?

package.json
```diff
@@ -1644,10 +1720,20 @@
         }
       ],
       "view/title": [
+        {
+          "command": "gemstone.explorer.commit",
```
MatiasFernandez: 💭 Two things about the new Explorer toolbar Commit/Abort:

1. They aren't mentioned anywhere in the PR description — which does say "every stone-mutating action follows the existing 'edit, then you commit explicitly' model". A one-click Commit is still explicit, but it's a notable new capability that reviewers (and release notes) would want called out.
2. Placement: `commit` is `navigation@3`, `abort` `@4`, `refresh` `@5` — so Commit sits two icons left of the button people reach for constantly. `commitSession` only confirms when there are unsaved `.gs` edits, so a mis-click on an otherwise clean session commits the transaction with no prompt and no undo. Should Commit/Abort be at the far right of the group (after Refresh), or carry a confirmation of their own?

client/src/gemstoneExplorer.ts
```diff
@@ -268,23 +324,38 @@ class VarSideItem extends vscode.TreeItem {
...
     // Split by side so the inline "+" (Add Instance Variable) targets only the
     // instance side; the class side keeps the base token.
     this.contextValue = isMeta ? 'explorerVarSide.class' : 'explorerVarSide.instance';
     this.iconPath = new vscode.ThemeIcon('symbol-class');
-    this.tooltip = isMeta
-      ? `Class variables of ${className}`
-      : `Instance variables of ${className}`;
+    if (isEmpty) {
+      // Synthetic resourceUri → FileDecorationProvider grays the header label.
+      this.resourceUri = emptyVarSideUri(className, isMeta);
```
MatiasFernandez, line 353: 💭 The empty **class variables** header looks like it ends up being a dead row. It's `TreeItemCollapsibleState.None` (so no caret), it carries `contextValue: 'explorerVarSide.class'` — and in package.json the only variable-side menu contribution is `gemstone.explorer.addInstVar`, wired to `explorerVarSide.instance`. There's no add-class-variable command in the Explorer at all (only `renameClassVariable`, on `explorerClassVar` rows). So right-clicking the grayed "class variables" header gives nothing.

The PR description says both groups stay "right-clickable so *New Instance/Class Variable* works from it", which holds for the instance side but not the class side. Is the plan to add an `addClassVar` command here, or should the class-side header stay purely informational for now?

client/src/gemstoneExplorer.ts
```diff
@@ -83,45 +86,98 @@ const VIEW_METHODS = 'gemstoneExplorerMethods';
...
+//   - NAVIGATION (single click): the ONE transient (unpinned) tab, reused — a new
+//     click replaces it; pinned tabs are left alone. Already-open tab → just reveal.
+//   - PIN (the 📌 action): the method becomes a PINNED tab in that same group, added
+//     WITHOUT stealing the view (the tab you were reading stays showing).
+// `placement` scopes this to editors this Explorer opened, so it never invades the
+// System Browser's group (see sourceEditorPlacement.ts).
+// Exported for unit testing the placement rules.
+export async function openGemstoneDocument(
   doc: vscode.TextDocument,
   toSide: boolean,
```
MatiasFernandez: 💭 Now that the whole placement model is pinning rather than columns, `toSide` no longer describes what this flag does — the doc comment right above already calls the two paths NAVIGATION and PIN, and the user-facing titles became "Pin Method" / "Pin Definition" / "Pin Comment". Rename the parameter to `pin` (here and in `openMethod` / `openClassComment` / `openDefinitionFor`)? Given we don't have real users yet, we can also rename the command IDs

client/src/gemstoneExplorer.ts
```diff
@@ -559,9 +645,23 @@ export class ExplorerController {
...
+    before: Set<string>;
+  };
+  // URIs of editors we opened ourselves (method/definition clicks); syncToEditor
+  // ignores its own opens so a tree click doesn't bounce the selection. A Set (not
+  // a single value) because opens can overlap — clicking through methods faster
+  // than each preview settles — and each open must match its own later
+  // onDidChangeActiveTextEditor event; a scalar got overwritten by the next click,
+  // letting the earlier open's event slip past the guard and re-reveal (scroll) the
+  // Methods pane.
+  private readonly selfOpenedUris = new Set<string>();
```
MatiasFernandez, line 675: 💭 Nothing prunes `selfOpenedUris` when the expected `onDidChangeActiveTextEditor` never arrives. The event only fires when the active editor actually *changes*, so clicking the already-focused method in the tree adds a mark that's never consumed. Later, when the user manually focuses that method's tab from elsewhere, `syncToEditor` swallows that first genuine event and the tree doesn't follow.

Granted it's one stale entry per URI and the new `alreadySelected` check covers the common case — the comment below even calls that "the reliable guard". Given that, is the URI set still pulling its weight, or could it be dropped in favour of the selection check alone?

client/src/gemstoneFileSystemProvider.ts
```diff
@@ -116,6 +116,11 @@ export function parseUri(uri: vscode.Uri): ParsedUri {
   if (parts.length === 4 && parts[3] === 'comment') {
     return { kind: 'comment', sessionId, dictName: parts[1], className: parts[2], dictIndex };
   }
+  // 5-segment form `/dict/Class/comment/Class comment` — the trailing label makes
+  // the editor tab read as "Class comment" (see buildClassCommentUri).
+  if (parts.length === 5 && parts[3] === 'comment') {
```
MatiasFernandez: nit: this branch returns exactly the same object as the 4-segment one right above it — could they collapse into `if ((parts.length === 4 || parts.length === 5) && parts[3] === 'comment')`, keeping your comment about the trailing label segment?

MatiasFernandez (approved):
Approved and left some feedback my agent found

client/src/gemstoneExplorer.ts
ericwinger: Fixed in 3535cbe4. Added a `canClassBeWritten` pre-guard (mirroring `createNewMethod`), plus a try/catch around `deleteMethod`, a check of its returned status string, and a `.catch` on the command registration — a failed removal now surfaces an error instead of silently redrawing the pane with the method still there.

Also **manually verified both ways we discussed**: a removal that clears the writability pre-check but then fails now shows an error dialog rather than a silent no-op — both the `Selector not found:` status-string path (e.g. removing a method that was deleted out from under a stale tree) and the raised-error `catch` path.

client/src/gemstoneExplorer.ts
ericwinger: Fixed in 3535cbe4. `closeDisposableTab` now closes only within the column the transient was opened into — recorded as `SourceEditorPlacement.reusableColumn` from the returned editor. The System Browser builds the *identical* `gemstone://…?dict=N` URI, so the URI alone can't tell the two copies apart; the recorded column can. New cross-browser test, mutation-verified (red on the whole-window scan).

package.json
ericwinger: Fixed in 3535cbe4. Moved Commit/Abort to the right of Refresh (Refresh is now leftmost of the three), and documented the buttons in the PR description. Kept the existing confirmations rather than adding an always-on prompt: Abort already confirms when `sessionNeedsCommit`/there are unsaved edits, and an unconditional prompt would double-confirm those cases and nag on every clean commit.

client/src/gemstoneExplorer.ts
ericwinger: Resolved as informational. Adding an instance variable goes through the RB engine (`runInstVarRefactor`), and there's no class-variable engine op, so there's no cheap symmetric `addClassVar` — it'd be real engine scope. The class-side header stays informational, and I corrected the PR-description overclaim: New Instance Variable is offered from the instance side only; the class side is not actionable.

client/src/gemstoneExplorer.ts
ericwinger: Done in 3535cbe4. Renamed the parameter to `pin` across `openGemstoneDocument` / `openMethod` / `openClassDefinition` / `openClassComment` / `openDefinitionFor` / `openCommentFor`. Left the command IDs as-is for now to avoid churn across keybindings and menu `when`-clauses.

client/src/gemstoneExplorer.ts
ericwinger: Fixed in 3535cbe4 without dropping the set (it still guards the Methods-pane jump fix). `openMethod` now skips the `selfOpenedUris` mark when the method is already the active editor — the case where no `onDidChangeActiveTextEditor` fires — so the mark can't go stale and swallow a later genuine focus event. New test, mutation-verified.

client/src/gemstoneFileSystemProvider.ts
ericwinger: Fixed in 3535cbe4. Collapsed both the `definition` and `comment` 4-/5-segment branches into single conditions, keeping the trailing-label comments, and added the missing 5-segment definition parse test.

PR #373, opened 2026-08-04
Rule agents into the CI suite, not the gci project

MatiasFernandez:
Agent-facing context only — no production or test code changes.

While migrating the on-demand `gci` tests into the CI integration suite, the
rules files still read as though `gci/` were a normal place to add a test. An
agent following them would keep writing tests that nobody ever runs, and would
happily write the same case twice, once in each suite.

## What this adds

`.claude/rules/client/tests.md` gains a "Choosing where a stone-dependent test
lives" section that makes `useIntegrationTest` the documented default home and
states the rule plainly: a case that can be written that way MUST be written
that way, and only that way.

It lists the five constraints that genuinely rule the harness out — must
commit, needs exclusive or destructive stone access, drives its own session
lifecycle, makes one un-interruptible blocking call that can outlast the CI
budget, or needs `cwd` cleanup only the `gci` project registers.

Just as important, it lists the near-misses that are **not** reasons, because
each already has a mechanism that runs in CI: needing the server plugin is a
gate, a fixture a bare extent lacks is something the test creates and the
auto-abort rolls back, version-dependent behavior resolves off the session, and
`SystemUser` is a login option. It also bans asserting on whatever the live
stone happens to contain — a branch that asserts both arms can never fail.

`.claude/rules/client/gci.md` reframes the suite as legacy and being retired,
keeping the mechanics needed to actually run it, and points at the rule above
rather than restating the exceptions.

## Placement

The decision rule is client-scoped, so it lives in the path-scoped client rules
file rather than the repo-wide one, which would load stone and GCI detail for
every `server/` and `mcp-server/` test. 

Two incidental cleanups fell out of the no-duplication rule: the "deep GCI
binding suite is separate" aside is gone now that a whole section covers it, and
the koffi deep-equality note moved up so the new section's gating
cross-reference sits next to the gating section it refers to.

### DIFF
--- a/.claude/rules/client/tests.md
+++ b/.claude/rules/client/tests.md
@@ -42,9 +42,35 @@ Tests run in random order; the seed is printed at the top of the output. Reprodu
 
 ## Integration tests
 
-Tests using `useIntegrationTest` require a live GemStone instance so plain `npm test` needs a running stone. Run `npm run test:server:start` once to provision one; it writes connection details to `.env.test` (which the user may override with `.env.test.local`). CI runs these as a matrix over `client/.gemstone-integration-releases.json`. The deep GCI binding suite (`npm run test:gci`) is separate.
+Tests using `useIntegrationTest` require a live GemStone instance so plain `npm test` needs a running stone. Run `npm run test:server:start` once to provision one; it writes connection details to `.env.test` (which the user may override with `.env.test.local`). CI runs these as a matrix over `client/.gemstone-integration-releases.json`.
 
 GCI session/oop values are koffi `External` pointer wrappers with no enumerable properties, so vitest's deep equality (`toEqual`, and therefore `expect(spy).toHaveBeenCalledWith(someSession, ...)`) cannot tell two *different* sessions or oops apart — it treats any two of them as equal regardless of the underlying native pointer. To assert *which* session/oop a call received, pull the argument out of `spy.mock.calls` and compare with `toBe` (reference equality) instead.
+
+### Choosing where a stone-dependent test lives
+
+The **default home** is a `*.integration.test.ts` using `useIntegrationTest`, co-located in the `__tests__/` dir next to the code it exercises (raw `GciTsXxx` wrapper tests → `client/src/gciLibrary/__tests__/`). Those run in CI over the whole release matrix, in both passes.
+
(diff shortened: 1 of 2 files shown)


.claude/rules/client/tests.md
```diff
@@ -42,10 +42,35 @@ Tests run in random order; the seed is printed at the top of the output. Reprodu
...
+
+- it must **commit** — the harness aborts every test, and an abort cannot undo a commit;
+- it needs **exclusive access to the stone**, or destroys and restores the whole repository;
+- it drives a **session lifecycle** beyond `login` / `logout` / `withTransientSession` (e.g. many logins with varying flags), or installs process-wide state once for a whole file;
+- it makes a **single blocking GCI call** that can outlast the CI budget and that vitest cannot interrupt, so a hang takes out the whole job;
+- it writes into `process.cwd()` and its cleanup is only registered by the `gci` project.
+
+**These are not reasons** — each already has a mechanism that runs in CI:
+
+- *needs the server plugin, Rowan, or Grail in the image* → gate it (see below). CI runs the suite twice, so both worlds are covered.
```
ericwinger: Small accuracy gap that could misdirect an agent: this bullet groups "server plugin, **Rowan, or Grail** → gate it (see below)," and "see below" is the `requireServerPluginFeature` section. But `pluginFeatures` registers only `refactoring` and Enhanced Inspector — there is no `pluginFeatures.rowan`/`.grail`, and the "install server plugin" CI pass files those two, not Rowan/Grail. So "gate it → both worlds covered" holds for the **server plugin**, but an agent applying it to the Rowan/Grail sub-case will reach for a gate that doesn't exist.

If Rowan/Grail are in fact always present in the matrix stones, they're a non-reason for a *different* reason (always available, nothing to gate) — worth saying that, or splitting the server-plugin bullet from Rowan/Grail so the cited mechanism matches each.

ericwinger: I know Rowan/Grail won't yet be in the image so I don't know if it's worth tweaking this.

ericwinger (approved):
LGTM

.claude/rules/client/tests.md
MatiasFernandez: I tweaked the wording to avoid confusion

PR #376, opened 2026-08-04
Cache lint and format runs for faster repeats

npapagna:
## Summary

- Enable ESLint's and Prettier's built-in caches for `lint`, `lint:fix`, `format`, and `format:check`, so unchanged files are skipped on repeat runs instead of being fully re-checked every time.
- Cache files live under `node_modules/.cache/`, already covered by the existing `node_modules/` gitignore rule, so nothing new needs to be ignored.

## Time improvements (measured locally)

| Command | Before (always cold) | After (warm cache) |
|---|---|---|
| `npm run lint` | ~15s | ~1.4s |
| `npm run format:check` | ~5.6s | ~1s |

Verified cache correctness by introducing a real lint violation with a warm cache in place; it was still caught, since ESLint invalidates per changed file.

Note: this speeds up local, repeated runs. CI does a fresh `npm ci` checkout each time with nothing to warm the cache from, so this doesn't change CI timing.

## Test plan

- [x] `npm run lint` passes (cold and warm)
- [x] `npm run format:check` passes (cold and warm)
- [x] Confirmed a real lint error in a changed file is still caught with a warm cache


package.json
```diff
@@ -2347,10 +2347,10 @@
     "dev:fresh:keep-installs": "bash scripts/dev-fresh.sh --keep-installs",
     "serve:seaside": "bash scripts/serve-seaside.sh",
     "lint:github:workflows": "bash scripts/lint-workflow-timeouts.sh",
-    "lint": "eslint .",
-    "lint:fix": "eslint . --fix",
-    "format": "prettier --write \"**/*.{ts,mts,cts,js,mjs,cjs}\"",
-    "format:check": "prettier --check \"**/*.{ts,mts,cts,js,mjs,cjs}\"",
+    "lint": "eslint . --cache --cache-location node_modules/.cache/eslint/.eslintcache",
```
MatiasFernandez: 💭 Should we also pass `--cache-strategy content` here? ESLint's default strategy is `metadata` (file mtime + size), while Prettier's `--cache` hashes content, so the two halves of this PR end up with different safety properties. The `metadata` strategy misses an edit that keeps the file the same size *and* doesn't move the mtime forward — e.g. restoring files with `rsync -t` / `touch -r`, or a tool that preserves timestamps. Content hashing costs one hash per changed file (negligible next to running the rules) and makes both commands behave the same way. WDYT?

package.json
```diff
@@ -2347,10 +2347,10 @@
...
+    "lint:fix": "eslint . --fix --cache --cache-location node_modules/.cache/eslint/.eslintcache",
```
MatiasFernandez: 💭 Is it worth giving devs a documented way to bust the cache? Right now, if someone ever does hit a stale-cache result, the only recourse is knowing to `rm -rf node_modules/.cache` — which isn't discoverable from `package.json` or from CLAUDE.md's Commands section (both of which describe `npm run lint` as a full check) or any documentation.

The file already has a nice convention for this a few lines below: the `//hooks:install` sibling key documenting the script next to it. Something like a `//lint` note ("results are cached under `node_modules/.cache/`; delete it to force a full re-run") — or a small `lint:cache:clear` script — would fit that pattern, and would help agents too, since they'll otherwise read a fast, quiet `npm run lint` as a full-repo verification.

package.json
```diff
@@ -2347,10 +2347,10 @@
...
+    "format": "prettier --write --cache --cache-location node_modules/.cache/prettier/.prettier-cache \"**/*.{ts,mts,cts,js,mjs,cjs}\"",
```
MatiasFernandez: nit: `node_modules/.cache/prettier/.prettier-cache` is already Prettier's default `--cache-location` (checked in the pinned prettier 3.9.5), so `--cache-location` is a no-op on these two lines. Dropping it would shorten both `format` and `format:check` without changing behavior. The ESLint one is worth keeping explicit, since its default is `.eslintcache` at the repo root, which isn't gitignored.

package.json
```diff
@@ -2347,10 +2347,10 @@
...
+    "format:check": "prettier --check --cache --cache-location node_modules/.cache/prettier/.prettier-cache \"**/*.{ts,mts,cts,js,mjs,cjs}\"",
```
MatiasFernandez: 💭 On the CI note in the description: agreed that nothing warms the cache today, but would it be worth a follow-up to restore `node_modules/.cache/eslint` in the lint job via `actions/cache`? `npm ci` wipes `node_modules`, so it'd need to be restored *after* install, keyed on the lockfile + `eslint.config.*`. One caveat that ties back to my other comment: with the default `metadata` strategy this wouldn't work at all in CI, since each checkout writes fresh mtimes — it'd only pay off with `--cache-strategy content`. Not for this PR, just curious whether you looked at it.

MatiasFernandez (approved):

package.json
npapagna: Fixed! Both `lint` and `lint:fix` now use `--cache-strategy content`.

package.json
npapagna: Fixed! Added `lint:cache:clear` and `format:cache:clear` scripts, with `//lint` and `//format` comments pointing at them the same way `//hooks:install` documents its neighbor.

package.json
npapagna: Fixed! Dropped `--cache-location` from both Prettier scripts, confirmed `node_modules/.cache/prettier/.prettier-cache` is still where the cache lands with no flag. Kept it on the ESLint scripts since that default isn't gitignored.

package.json
npapagna: Looked into it. Found the job: `.github/workflows/health-check.yml` `lint` (checkout → `npm ci` → `format:check` → `lint`, 5 min timeout). Lockfile is `package-lock.json` at the root, ESLint config is `eslint.config.mjs` at the root, so a key of `hashFiles('package-lock.json', 'eslint.config.mjs')` would work.

Now that this PR uses `--cache-strategy content`, your caveat is resolved, content hashing doesn't care that every checkout gets fresh mtimes, so an `actions/cache` restore would actually pay off here (it wouldn't have under the old `metadata` strategy).

Still think it's a genuine follow-up rather than something to fold in here:
- Locally the warm/cold gap is ~14s, but that's against a job that also pays for checkout, Node setup, and `npm ci`, so the fraction of total job time saved is smaller than it looks, well under the 5 min timeout either way.
- It needs both a restore and a save step (`npm ci` wipes `node_modules`, so `actions/cache` alone doesn't persist across it), keyed so a config/dependency change invalidates it, a bit more moving parts than a one-liner.
- Prettier's cache could get the same treatment for symmetry, which widens the scope a little further.

Happy to open that as a separate PR if useful.

PR #379, opened 2026-08-05
V6/V7: Extract Superclass (GsExtractSuperclassRefactoring + #classAdd)

ericwinger:
## V6 Insert Superclass / V7 Extract Superclass

Adds the extract-superclass refactorings to the server-side RB engine (part of #252). One engine — `GsExtractSuperclassRefactoring` — drives both, with V6 as the degenerate no-sibling/no-member case of V7.

### Engine
- **`GsExtractSuperclassRefactoring`** + a new change kind **`#classAdd`** (a brand-new class, `inClassHistory: nil`) on `GsRefactoringChange`/`ChangeSet`.
- Apply generalises `GsInstVarStructureRefactoring`'s reversion machinery: create the new superclass, re-parent the extracted classes onto it (dropping hoisted ivars), reversion each descendant subtree **top-down** onto its freshly-versioned ancestor, hoist methods onto the new class and skip them on copy-forward.
- Member classification (`candidateMethods`/`candidateInstVars`): identical / divergent / partial / unhoistable (super-send, foreign ivar, foreign class/pool var). All-or-nothing, **never commits**; existing instances keep their prior version.

### Client
- Explorer class-row + hierarchy-node commands `gemstone.explorer.insertSuperclass` / `extractSuperclass`.
- **V6** = name only. **V7** = opt-in sibling QuickPick (none pre-picked) → name → member checklist (identical pre-checked; divergent/partial opt-in with a warning; unhoistable not offered) → paginated preview panel → apply → reveal.

### Tests — ~85 new, all in CI-run layers; **no new `gci/` tests**
- **30** engine GS SUnit tests (`GsExtractSuperclassRefactoringTest`), incl. a two-deep (`Dog→Puppy→Runt`) with-sibling reparent test.
- **49** client unit tests (command, panel, panel HTML, preview parsers, query builders, and the Explorer handlers).
- **6** GCI integration tests (`refactoringExtractSuperclass.integration.test.ts`) — client apply path, plugin-gated, run in CI over the whole release matrix.

**No tests were added to the on-demand `client/src/__tests__/gci/` project** — only integration tests. That project never runs in CI and is being retired, so the (redundant) `gciExtractSuperclass.e2e.test.ts` was removed; its coverage is a superset in the integration test above.

### Incorporates a code-review pass
F3 (super-substring false-pass), F4 (panel double-apply latch — the #360 guard the new panel had missed), F5 (flush dirty method editors before hoisting stored source), F6 (deep-subtree reparent test), plus LOWs (no false success on `applied:0`; `applyDeselected` surfaces a declined ref). The redundant `gciExtractSuperclass.e2e.test.ts` was removed (its coverage is a superset in the integration test; consistent with main retiring the `gci/` project).

**Design note:** hoisting *divergent* / *partial* members stays an opt-in (the client offers it with a warning, and it's verified working — the anchor's version wins and siblings inherit) rather than an engine decline. Called out for reviewer visibility.

### Testing status
Local gate green (lint / format:check / compile / `npm test`: server 322, mcp 92, client green — extract-superclass integration passing in-stone on 3.6.2). **Still owed before undrafting:** the full both-boundary gate (engine SUnit on 3.6.2 **and** 3.7.5) per the house pre-PR rule.


client/src/refactoring/extractSuperclassCommand.ts
```diff
@@ -0,0 +1,331 @@
...
+  return true;
+}
+
+async function promptNewClassName(anchor: string): Promise<string | undefined> {
+  const name = await vscode.window.showInputBox({
+    title: `New superclass for ${anchor}`,
+    prompt: 'Name of the new superclass to insert',
+    placeHolder: 'e.g. AbstractShape',
+    ignoreFocusOut: true,
+    validateInput: (v) => validateClassName(v) ?? undefined,
```
MatiasFernandez: 🚫 The name is only format-validated here, and the engine's collision precondition uses `environment classNamed:`, which returns nil for anything that is not a `Class`. So a name that is bound to a NON-class global — say `UserGlobals at: #Pet` holding a collection or a constant — passes both checks, and `applyClassAdd:`'s `subclass: newName … inDictionary:` then rebinds that key to the new class, silently clobbering the existing global.

We already guard exactly this for rename: `ExplorerController.validateRenameTarget` (whose comment this PR touches) calls `queries.globalNameInUse(session, newName)` on top of the format check. Should `promptNewClassName`'s `validateInput` do the same — and/or should `computeAnalysis` check the symbol list rather than `classNamed:` — so the collision is caught while the user is still typing instead of quietly overwriting a global?

gs-src/refactoring/engine/GsExtractSuperclassRefactoring.class.st
```diff
@@ -0,0 +1,807 @@
...
+	"Every hoisted ivar must be the anchor's OWN ivar and must not already be defined on S (a
+	 double declaration once the new class -- a child of S -- declares it)."
+	| own supAll |
+	own := self ownInstVarsOf: anchorClass.
+	supAll := self allInstVarsOf: sharedParent.
+	hoistInstVars do: [:v |
+		(own includes: v) ifFalse: [
+			^decline := 'Cannot hoist instance variable ', v, ': it is not declared in ', anchorClass name asString, '.'].
+		(supAll includes: v) ifTrue: [
+			^decline := 'Cannot hoist instance variable ', v, ': ', sharedParent name asString, ' already defines it.']]
```
MatiasFernandez: 💭 This validates a hoisted ivar against the anchor's own list and against S, but never against the DESCENDANTS of the extracted classes. Is it possible that a descendant already declares the same name?

With the suite's own fixture: if `GsESPuppy` owned `'name'` (instead of `'cuteness'`) and we hoist `'name'` from `GsESDog`, the new `GsESPet` declares `name`, `GsESDog`'s new version drops it, and then `applyClassChange:` reversions `GsESPuppy` with its own `name` under a superclass chain that now provides one — GemStone rejects the duplicate declaration. The error is caught by the `applyDeselected:` loop, so we end up in the half-applied state that comment on line 638 describes: the class created and some reparents done, and the user told to abort.

Since the class comment sells these as "global declines that empty the change set", would it be worth adding this to the precondition set (walk `environment descendantsOf:` for each extracted class and reject a hoisted ivar that any of them owns) so it declines in the preview instead of failing mid-apply? A test for it would also pin the current behaviour either way.

client/src/refactoring/extractSuperclassPreview.ts
```diff
@@ -0,0 +1,289 @@
...
+    changes: env.changes.map(parseChange),
+    nextOffset: asCount(env.nextOffset),
+    done: env.done === true,
+  };
+}
+
+function toMemberKind(v: unknown): MemberKind {
+  return v === 'identical' || v === 'divergent' || v === 'partial' || v === 'unhoistable'
+    ? v
+    : 'partial';
```
MatiasFernandez: 💭 The fallback for an unrecognised kind is `'partial'`, which is an OFFERABLE kind — `buildMemberPicks` only filters out `'unhoistable'`. So if the engine ever grows a new classification (or a payload is truncated/garbled), the member shows up in the checklist as opt-in-able rather than being withheld.

Would `'unhoistable'` be the safer default here? It fails closed: an unknown kind is simply not offered, and the worst case is a member the user has to hoist some other way rather than one we let them hoist without knowing whether it compiles.

gs-src/refactoring/tests/GsExtractSuperclassRefactoringTest.class.st
```diff
@@ -0,0 +1,561 @@
...
+	self deny: json includesSubstring: '"failed":[{'.
+	self assert: ((self classNamed: 'GsESPet') includesSelector: #eat).
+	self deny: ((self classNamed: 'GsESDog') includesSelector: #eat).
+	self deny: ((self classNamed: 'GsESCat') includesSelector: #eat).
+	self assert: ((self classNamed: 'GsESDog') canUnderstand: #eat).
+	self assert: ((self classNamed: 'GsESCat') canUnderstand: #eat).
+	"the non-hoisted methods stay put"
+	self assert: ((self classNamed: 'GsESDog') includesSelector: #describe).
+	self assert: ((self classNamed: 'GsESCat') includesSelector: #describe)
+]
```
MatiasFernandez, line 292: 💭 The suite covers hoisting an *identical* method well, but I don't see a test for hoisting a **divergent** one — which is exactly the design decision the PR description calls out for visibility (the anchor's version wins, siblings inherit it and lose their own).

That's the behaviour most likely to surprise someone, and the one a future change is most likely to break silently. Worth a test that hoists `#describe` with `siblings: #('GsESCat')` and asserts `GsESPet` defines it, `GsESCat` no longer defines its own, and `GsESCat` now answers `'a dog'`? Same argument for a `partial` member.

gs-src/refactoring/tests/GsExtractSuperclassRefactoringTest.class.st
```diff
@@ -0,0 +1,561 @@
...
+	self assert: (self changeOfKind: #classDefinitionEdit for: 'GsESDog' in: cs) notNil
+]
+
+{ #category : 'tests - V6 insert' }
+GsExtractSuperclassRefactoringTest >> testInsertEmptySuperclassApplies [
+	| json |
+	json := (GsExtractSuperclassRefactoring class: (self classNamed: 'GsESDog') insertSuperclassNamed: 'GsESPet' inDictionary: nil)
+		applyDeselected: #().
+
+	self deny: json includesSubstring: '"failed":[{'.
```
MatiasFernandez: 💭 `deny: json includesSubstring: '"failed":[{'` also passes when `json` is not the envelope we think it is at all (an error string, a truncated payload) — it only proves the literal `"failed":[{` is absent. `self assert: json includesSubstring: '"failed":[]'` asserts the same thing positively and can't pass on a malformed answer. Same pattern in the other six apply tests — worth flipping them all?

client/src/refactoring/queries/getSiblingClassNames.ts
```diff
@@ -0,0 +1,30 @@
...
+  className: string,
+  dict?: number | string,
+): string[] {
+  const code = `| cls sup out |
+cls := ${classLookupExpr(className, dict)}.
+cls isNil ifTrue: [^ ''].
+sup := cls superclass.
+sup isNil ifTrue: [^ ''].
+out := WriteStream on: String new.
+((ClassOrganizer new subclassesOf: sup) asSortedCollection: [:a :b | a name <= b name]) do: [:c |
```
MatiasFernandez, line 30: 💭 This resolves siblings via `ClassOrganizer new subclassesOf: sup` (consistent with `getClassDescendantNames`), but the engine's `resolveSibling:` resolves them with `sharedParent subclasses` and declines with "… is not a subclass of …" when it can't find a match.

Is there a case where the two disagree — an older class version, a subclass reachable through one lookup but not the other — where we'd offer a name in the picker that the engine then refuses after the user has already picked members and named the new class? If they're guaranteed equivalent, a one-line note here saying so would save the next reader the same trip.

gs-src/refactoring/tests/GsExtractSuperclassRefactoringTest.class.st
```diff
@@ -0,0 +1,561 @@
...
+{ #category : 'tests - apply' }
+GsExtractSuperclassRefactoringTest >> testApplyIsAllOrNothingIgnoringDeselection [
+	"Extract-superclass is all-or-nothing: a deselected id is ignored and the new class is still
+	 created."
+	(GsExtractSuperclassRefactoring class: (self classNamed: 'GsESDog') insertSuperclassNamed: 'GsESPet' inDictionary: nil)
+		startPreviewToken: 'esTok' maxBytes: 100000.
+	[GsExtractSuperclassRefactoring applyForToken: 'esTok' deselected: #('1' '2' '3')]
+		ensure: [GsExtractSuperclassRefactoring clearToken: 'esTok'].
+
+	self assert: (UserGlobals includesKey: #GsESPet)
```
MatiasFernandez: nit: This asserts the new class exists, but the deselected ids are arbitrary (`#('1' '2' '3')`) — so it doesn't quite show that a change whose id WAS deselected still applied. Adding `self assert: (self classNamed: 'GsESDog') superclass name asString equals: 'GsESPet'` would make it prove the reparent survived the deselection too, which is the all-or-nothing property the test name claims.

MatiasFernandez (approved):
Left some comments worth reviewing and handling as follow-ups as appropiate

client/src/refactoring/extractSuperclassCommand.ts
ericwinger: Fixed in `67804717`. `promptNewClassName` now validates through `validateNewSuperclassName`, which layers `globalNameInUse` on top of the format check — and `globalNameInUse` probes via `objectNamed:`, so it catches a **non-class** global too, not just a class. It runs as the input box's live validator (mirrors `ExplorerController.validateRenameTarget`), so the collision surfaces inline while the user is still typing. Tests added.

The engine-side precondition (`computeAnalysis` still uses `classNamed:`) is left as-is here and tracked as a defense-in-depth hardening item — the client is the only place a class name is typed.

gs-src/refactoring/engine/GsExtractSuperclassRefactoring.class.st
ericwinger: Fixed in `67804717`. `validateHoistInstVars` now also declines via a new `descendantDeclaring:`, which walks `environment descendantsOf:` each extracted class (skipping the extracted classes themselves) and declines **up front** when a descendant already owns the ivar — a global decline that empties the change set, not a part-way `applyClassChange:` failure. The reachable shape is a descendant of a *sibling* (GemStone already forbids a subclass redeclaring an ancestor's ivar). Positive + over-reach tests, green on both boundaries.

client/src/refactoring/extractSuperclassPreview.ts
ericwinger: Fixed in `67804717`. `toMemberKind` now fails **closed**: an unrecognised classification → `'unhoistable'` (withheld by `buildMemberPicks`) instead of `'partial'` (offerable). Test covers unknown / missing / non-string kinds.

gs-src/refactoring/tests/GsExtractSuperclassRefactoringTest.class.st
ericwinger: Added in `67804717`: `testExtractHoistsDivergentMethodAnchorVersionWins` — hoists a divergent `#describe` (`'a dog'` vs `'a cat'`), asserts it lands on the new superclass, both extracted classes lose their own copy, and both now answer the anchor's version (`new describe = 'a dog'`).

gs-src/refactoring/tests/GsExtractSuperclassRefactoringTest.class.st
ericwinger: Fixed in `67804717`. All seven `deny: json includesSubstring: '"failed":[{'` flipped to `assert: json includesSubstring: '"failed":[]'`. The positive form only passes on the real success envelope — an error string or a truncated payload can no longer slip through.

client/src/refactoring/queries/getSiblingClassNames.ts
ericwinger: Fixed in `67804717`. `getSiblingClassNames` now intersects the `ClassOrganizer subclassesOf:` result with `sup subclasses` — the exact collection `resolveSibling:` searches — so every offered name is guaranteed engine-resolvable (no more offering a name the engine would then decline after the user picked members and named the class). Documented in the query + a test asserting the intersection.

gs-src/refactoring/tests/GsExtractSuperclassRefactoringTest.class.st
ericwinger: Fixed in `67804717`. `testApplyIsAllOrNothingIgnoringDeselection` now also asserts `GsESDog superclass = GsESPet`, proving the reparent survived the (arbitrary) deselection — not just that the new class was created.

PR #380, opened 2026-08-05
Fix release doc: use npm version instead of hand-editing package-lock.json

MatiasFernandez:
## Summary

- During a supply-chain-hardening audit we traced a version mismatch to release commit [`cd0bff1`](https://github.com/GemTalk/Jasper/commit/cd0bff1cc79ffd65825061e7005800fc36f96f25) ("Release 1.6.1"): `package-lock.json`'s version bump was done by hand (a find-and-replace of `1.6.0` → `1.6.1`), and it wasn't scoped to just the two root fields — it also matched two unrelated dependencies pinned at `1.6.0`, `mime@1.6.0` and `sax@1.6.0`, silently rewriting their `version` field to `1.6.1` while their `resolved` URL and `integrity` hash still point at the real `1.6.0` tarball. Both entries are now internally inconsistent in the lockfile on `main` today.
- This isn't a one-off: the collision risk scales with the next release. `1.8.11` (a plausible next patch version) collides with `typed-rest-client@1.8.11`; a `2.0.0` release would match roughly 20 entries across the tree.
- `package-lock.json` should never be edited by hand for exactly this reason. npm already provides `npm version <X.Y.Z> --no-git-tag-version`, which bumps only the two root version fields (`version` and `packages."".version`) atomically and structurally can't reach nested dependency entries.
- Updates [CONTRIBUTING.md](CONTRIBUTING.md)'s release step accordingly, with an inline note on the collision risk and on why `--no-git-tag-version` is required (later steps already commit/tag).

## Test plan

- [x] Verified in a scratch copy that `npm version 1.8.11 --no-git-tag-version` updates only `package.json`'s `version` and `package-lock.json`'s two root fields, leaving the other 739 lockfile entries untouched.
- [x] Docs-only change; no build/test impact expected.
- [x] Separately: fix the drifted `mime`/`sax` entries in `package-lock.json` (not in scope of this PR).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,7 +125,7 @@ CI runs on **GitHub Actions**. The [Health Check workflow](.github/workflows/hea
 
 ## Publishing a release
 
-1. Update the version in `package.json` (and `package-lock.json` — both the top-level `version` and `packages."".version`) and promote the `[Unreleased]` section in `CHANGELOG.md` to a new dated `[X.Y.Z]` heading. Sweep `main` since the last release for merged PRs that didn't add their own changelog entries.
+1. `npm version <X.Y.Z> --no-git-tag-version` — bumps `package.json`'s `version` and `package-lock.json`'s two root fields (`version` and `packages."".version`) atomically. Don't hand-edit these or find-and-replace the version string across the lockfile: the version can collide with an unrelated dependency's own version elsewhere in `package-lock.json` (e.g. `1.8.11` matches `typed-rest-client@1.8.11`), corrupting that entry. `--no-git-tag-version` skips npm's own commit/tag, since steps 3-4 below handle that. Then promote the `[Unreleased]` section in `CHANGELOG.md` to a new dated `[X.Y.Z]` heading. Sweep `main` since the last release for merged PRs that didn't add their own changelog entries.
 2. `npm run compile && npm test`
 3. Commit the version + changelog changes (e.g. `Release X.Y.Z: <one-line summary>`).
 4. `git tag -a vX.Y.Z -m "Release X.Y.Z"` — annotated tag, on the release commit.


CONTRIBUTING.md
```diff
@@ -125,7 +125,7 @@ CI runs on **GitHub Actions**. The [Health Check workflow](.github/workflows/hea
 
 ## Publishing a release
 
-1. Update the version in `package.json` (and `package-lock.json` — both the top-level `version` and `packages."".version`) and promote the `[Unreleased]` section in `CHANGELOG.md` to a new dated `[X.Y.Z]` heading. Sweep `main` since the last release for merged PRs that didn't add their own changelog entries.
+1. `npm version <X.Y.Z> --no-git-tag-version` — bumps `package.json`'s `version` and `package-lock.json`'s two root fields (`version` and `packages."".version`) atomically. Don't hand-edit these or find-and-replace the version string across the lockfile: the version can collide with an unrelated dependency's own version elsewhere in `package-lock.json` (e.g. `1.8.11` matches `typed-rest-client@1.8.11`), corrupting that entry. `--no-git-tag-version` skips npm's own commit/tag, since steps 3-4 below handle that. Then promote the `[Unreleased]` section in `CHANGELOG.md` to a new dated `[X.Y.Z]` heading. Sweep `main` since the last release for merged PRs that didn't add their own changelog entries.
```
ericwinger, line 128: This is the right guidance, but it is **advisory only** — it prevents the mistake for someone who reads and follows step 1, and nothing detects it if they don't.

Worth knowing how invisible the current corruption is: `main` carries both broken entries today and its CI is fully green (the only non-success checks are skipped `Regenerate ROADMAP.md` runs). `npm ci` does not notice, because it installs from `resolved` and verifies `integrity` — both of which are still the genuine 1.6.0 values. So the drifted `version` field is inert until something reads it, which is exactly why it sat unnoticed from 1.6.1 through 1.8.10.

Since the failure mode is mechanical, a mechanical guard closes it for good. The invariant is simply "every entry's `version` matches the version in its own `resolved` URL", which is the check I ran to confirm your claim:

```js
// e.g. scripts/check-lockfile-versions.mjs, wired into the Lint job
import { readFileSync } from 'node:fs';
const { packages } = JSON.parse(readFileSync('package-lock.json', 'utf8'));
const bad = Object.entries(packages).flatMap(([path, meta]) => {
  const m = /-(\d+\.\d+\.\d+[^/]*)\.tgz$/.exec(meta.resolved ?? '');
  return m && meta.version && m[1] !== meta.version ? [`${path}: version=${meta.version} but resolved=${m[1]}`] : [];
});
if (bad.length) { console.error('Lockfile version drift:\n' + bad.join('\n')); process.exit(1); }
```

That would have failed the `Release 1.6.1` commit, and it fails today until `mime`/`sax` are repaired — which makes it a natural companion to that cleanup rather than a separate chore. Entirely your call whether it belongs in this PR or a follow-up; the doc change stands on its own either way.

Separately, a readability nit: step 1 now carries the command, two sentences of rationale, the `--no-git-tag-version` explanation, and the changelog instruction in a single numbered item. Splitting the rationale into a sub-bullet under the command would keep the numbered list scannable as a checklist, which is how it reads during an actual release.

ericwinger: Eric: It does seem that we should have more than an advisory note which can be skipped over by humans and ignored by Claude. A failed CI step would help prevent this. 

It might also be worth putting the specific instruction in the comment in its own line so it's easier for a human to see. 

Is there an issue to fix mime and sax so it doesn't get forgotten?

ericwinger (approved):
I'd suggest making this a harder check to avoid version comments and versions getting out of sync. Maybe a CI check. See comment. 

@jgfoster should review as well.

CONTRIBUTING.md
MatiasFernandez: I'm fixing the issue in my separate PR where I'm hardening the NPM supply chain tooling. As part of that change I'm introducing a new step in CI that will catch future issues like this one. So both things are already covered in my other work in progress.

jgfoster (approved):
Thanks for this. It looks like a good direction.

PR #381, opened 2026-08-05
Harden npm supply-chain posture

MatiasFernandez:
## Summary

Adds preventive supply-chain controls so a compromised dependency (the Shai-Hulud / SHA1-Hulud worm class) can't execute arbitrary code at install time — closes GemTalk/Jasper#246.

- **Raise the dev-toolchain npm floor to 11.16** (`.nvmrc` → 24.19.0, root `devEngines`) — the minimum version that supports `strict-allow-scripts`/`npm approve-scripts`, with the `prepare` CI job and `acceptance/Dockerfile` fixed to match.
- **Adopt npm's install-script allowlist** — `strict-allow-scripts=true` + a version-pinned `allowScripts` table in `package.json` makes an unreviewed install script fail-closed instead of silently running. Approves the tree's 5 current install-script deps, denies `fsevents`, and backs the control with a new `lint:supply-chain` CI check so `rm .npmrc` can't silently disable it.
- **Add a release-age cooldown and Dependabot config** — `min-release-age=7` guards local `npm install`/`update` (the only path that can still pull a freshly published malicious release, since the lockfile pins every version), mirrored by a 7-day Dependabot cooldown with an `install-scripts` group so `ESTRICTALLOWSCRIPTS` bumps land in one predictable PR per cycle.
- **Repair drifted `mime`/`sax` lockfile entries** — a prior release-bump find-and-replace left their `version` field disagreeing with the tarball their own `resolved` URL points to, which silently breaks advisory scanning and hard-blocks `npm audit signatures`.
- **Enforce lockfile integrity and registry signatures in CI** — `lockfile-lint` (host/HTTPS/integrity), `npm audit signatures`, and a new version-vs-`resolved` drift check that closes the exact gap that let `mime`/`sax` drift unnoticed for two months.
- Seeds a Diátaxis-organized `docs/` split (explanation/how-to/reference) for the new controls, with an index.

## Test plan

- [x] `npm ci` succeeds with no `EBADDEVENGINES` / `ESTRICTALLOWSCRIPTS`
- [x] `npm run package` (vsce's dependency walk, ships koffi's native binary) + `npm run test:gci` against a live stone
- [x] `.vsix` contents exclude `.npmrc`/`.nvmrc`
- [x] Negative tests: allowlist fail-closed on a removed `allowScripts` entry, `lint:supply-chain` catches a missing `.npmrc` and a drift lockfile entry, `lockfile-lint` catches a tampered `resolved` host
- [x] `npm audit signatures` passes over the full tree
- [x] CI: `prepare`, `lint`, and floor jobs green end to end


.github/workflows/health-check.yml
```diff
@@ -130,36 +146,69 @@ jobs:
...
-        # the same lockfile drift one level up.
-      - name: Pin npm to the lockfile's major version
+        # it any other way. Pinned to an exact version rather than the `11`
+        # dist-tag so this leg installs the same toolchain on every run and a
+        # fresh npm release can't change it underneath us — bump it
+        # deliberately, together with the same pin in acceptance/Dockerfile.
+      - name: Pin npm to an exact version that owns the lockfile
         if: startsWith(matrix.node-version, '22.')
-        run: npm install --global npm@11 # zizmor: ignore[adhoc-packages] -- npm 11 owns the lockfile and no Node 22 release bundles it
+        run: npm install --global npm@11.17.0 # zizmor: ignore[adhoc-packages] -- npm 11 owns the lockfile and no Node 22 release bundles it
```
npapagna, line 211: 🤖 Flagged by Claude (code review): `npm@11.17.0` is hardcoded identically here and in `acceptance/Dockerfile:37`, with nothing linking the two or checking they stay in sync. `docs/reference/supply-chain-controls.md` already calls this a known gap.

**Failure mode:** a future npm floor bump updates this workflow but the Dockerfile pin is forgotten, so the acceptance-test container silently runs a different npm than CI — on a branch specifically about tightening supply-chain guarantees, that's the kind of drift this work is meant to prevent.

**Possible fixes (increasing effort):**
1. Note in the "Known gaps" doc to grep for the version string in both places when bumping.
2. Add an assertion in `lint-supply-chain.mjs` (or a small script) that reads both files and fails if the pinned versions diverge.
3. Have the Dockerfile read the version from a single source (env var / `.npm-version` file) that the workflow also references, instead of duplicating the literal.

.github/workflows/health-check.yml
```diff
@@ -130,36 +146,69 @@ jobs:
...
-          # The floor include job sets node-version, which takes precedence.
-          node-version: ${{ matrix.node-version }}
           node-version-file: '.nvmrc'
           cache: 'npm'
+      # The floor leg's Node 22 bundles npm 10, which can't satisfy this
+      # repo's devEngines.packageManager floor (npm >=11.16.0, onFail:
+      # error) — so *any project-scoped* npm invocation on it fails with
+      # EBADDEVENGINES, including the innocuous `npm config get cache` that
+      # setup-node runs itself to locate the cache directory. That happens
+      # during setup, before the "Pin npm to the lockfile's major version"
```
npapagna: 🤖 Flagged by Claude (code review): this comment quotes the step name as "Pin npm to the lockfile's major version", but the actual step below (line 209) is now named `Pin npm to an exact version that owns the lockfile` — looks like a rename that didn't propagate to this reference.

**Failure mode:** low severity, but someone debugging the cache-suppression ordering greps for the quoted name and finds nothing, having to re-derive the connection by reading sequentially instead.

**Fix:** update the quoted name here to match the current step name.

scripts/lint-supply-chain.mjs
```diff
@@ -0,0 +1,209 @@
...
+  { key: 'allow-git', expected: 'none' },
+  { key: 'allow-remote', expected: 'none' },
+  { key: 'allow-scripts', expected: '', hint: ALLOW_SCRIPTS_HINT },
+  { key: 'min-release-age', expected: '7', floor: true },
+];
+
+const REGISTRY_PREFIX = 'https://registry.npmjs.org/';
+
+function getConfig(key) {
+  return execFileSync('npm', ['config', 'get', key], { encoding: 'utf8' }).trim();
```
npapagna: 🤖 Flagged by Claude (code review): `getConfig()` shells out to `npm config get <key>` once per entry in `CONFIG_ASSERTIONS` (5 keys), so `checkConfig()` pays for 5 full npm-CLI cold starts per lint run instead of one.

**Cost:** the test file's own comment (`client/src/__tests__/lintSupplyChain.test.ts:21`) already documents this: "Each case costs one `node` + five `npm config get` spawns, roughly half a second." With 22 test cases in that file, that's 100+ subprocess spawns and 10+ seconds of runtime, most of it avoidable process-startup overhead rather than actual work.

**Fix:** batch the reads into a single call, e.g. `npm config list --json` (or `npm config get key1 key2 ...`), parsed once and reused across all 5 assertions instead of spawning per key.

.github/workflows/health-check.yml
```diff
@@ -101,6 +111,12 @@ jobs:
         run: npm run format:check
       - name: Lint
         run: npm run lint
+      - name: Supply-chain config check
```
npapagna, line 110: 🤖 Flagged by Claude (code review): `lint:supply-chain` and `lint:lockfile` (lines 114-117) only depend on `package.json`, `package-lock.json`, and npm config, not on anything `format:check` or the full ESLint `lint` step touches, yet they run *after* both of those slower steps.

**Failure mode:** on a PR where one of the new checks fails (e.g. a dropped `.npmrc` key, or a lockfile-drifting dependency bump), the job still burns the full Prettier + ESLint runtime across all three workspaces before ever reporting a failure a sub-second check could have caught right after `npm ci`. Doesn't affect correctness, just delays feedback on exactly the class of problem this branch exists to catch quickly.

**Fix:** move `Supply-chain config check` and `Lockfile lint` up, right after `Install dependencies` and before `Format check` / `Lint`.

client/src/__tests__/lintSupplyChain.test.ts
```diff
@@ -0,0 +1,403 @@
...
+
+  // The incident this check was born from: an in-range `version` masking an out-of-range
+  // tarball. `npm ci` fetches `resolved` and only verifies `integrity`; `npm audit
+  // signatures` looks the package up by the declared `version`; lockfile-lint never reads
+  // `version` at all. Nothing else in the toolchain notices.
+  it('fails when version disagrees with the resolved tarball', () => {
+    const { status, stderr } = runLint({
+      lockfile: {
+        packages: {
+          'node_modules/mime': {
```
npapagna: 🤖 Flagged by Claude (code review): this test (and the similar ones at lines 259, 277, and 391) hand-writes the same three-key lockfile-entry object shape (`version`/`resolved`/`integrity`) that the existing `entry()` helper (line 94) already builds for the matching case, just with a deliberately mismatched `resolved` tarball version inlined by hand instead of extending the helper to express drift.

**Risk:** if the lockfile-entry shape ever gains a field `entry()` needs to account for, these four call sites have to be updated in lockstep with the helper. Miss one and that test silently keeps testing the old shape, a security-relevant test quietly stops verifying what it claims to, with no failure to flag it.

**Fix:** give `entry()` an optional drift parameter, e.g. `entry(name, version, resolvedVersion = version)`, computing `resolved` from `resolvedVersion` instead of `version`. Each drift test then becomes `entry('mime', '1.6.0', '2.0.0')` instead of the inline literal.

Side note: the line-391 case (`'reports every failing check in one run'`) is missing `integrity` entirely, unlike the other three, worth confirming that's intentional or just copy-paste drift.

client/src/__tests__/lintSupplyChain.test.ts
```diff
@@ -0,0 +1,403 @@
...
+});
+
+// Runs the script against a synthetic repo. The child's environment is scrubbed of every
+// npm_config_* variable, because `npm test` exports this repo's own npm config that way and
+// it would otherwise leak past the fixture's .npmrc and decide the config assertions for us.
+// User- and global-level npm config are likewise redirected at empty files (two distinct
+// paths — npm refuses to load one file at two levels), so a developer's ~/.npmrc can neither
+// rescue nor break a case.
+function runLint({ npmrc = COMPLIANT_NPMRC, packageJson = {}, lockfile = {}, env = {} }: Fixture) {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'lint-supply-chain-'));
```
npapagna: 🤖 Flagged by Claude (code review): `runLint()` hand-rolls tempdir creation (`fs.mkdtempSync`) and cleanup (a module-level `tempDirs` array drained in `afterEach`, lines 45-51), duplicating a helper that already exists in this same test-support directory: `withTemporaryFolderDo` (`client/src/__tests__/support/file.ts:33`), which does the identical mkdtemp + guaranteed-cleanup-via-finally, and is already used elsewhere (`mcpOwnerSidecar.test.ts:40`).

`client/src/__tests__/rowanLoad.test.ts` independently reimplements the same idiom a third time with `beforeEach`/`afterEach` (lines 16-22), and — worse — a *second*, unrelated tempdir created inside its `updateGitRepo` describe block (`setup()`, line 161) is never cleaned up at all, a real leak. See the review summary for details; not commenting inline there since that file isn't part of this diff.

**Why it matters here:** `runLint`'s body (`writeFileSync`/`spawnSync`) is fully synchronous, so it fits `withTemporaryFolderDo`'s synchronous consumer signature directly — wrapping it would drop the `tempDirs` array and its `afterEach` entirely (lines 45-51) and get crash-safe cleanup for free, instead of batching cleanup at the end of each test.

**Fix:**
```js
function runLint({ npmrc = COMPLIANT_NPMRC, packageJson = {}, lockfile = {}, env = {} }: Fixture) {
  let result;
  withTemporaryFolderDo((dir) => {
    // ...same body, minus tempDirs.push(dir) and the afterEach block above...
    result = spawnSync(...);
  });
  return { status: result.status, stdout: result.stdout, stderr: result.stderr };
}
```
(Env-scrubbing logic stays as-is; it's orthogonal to the tempdir lifecycle.)

npapagna (commented):
🤖 Claude's code review of this branch. Six findings below, all inline except one: `client/src/__tests__/rowanLoad.test.ts` isn't part of this diff, so a line comment can't be anchored there, but flagging it here since it's directly relevant to the last finding on `lintSupplyChain.test.ts` — `updateGitRepo`'s `setup()` helper (line 161) creates a second tempdir via `mkdtempSync` that is never cleaned up anywhere in the file (only the module-level `root` is removed, in the top-level `afterEach`). Every run of the two tests that call `setup()` (lines 191, 197) leaks a directory (seed repo, bare remote, clone, and any `bump()`-created pusher clones) into the OS temp folder permanently. This is exactly the class of bug `withTemporaryFolderDo` (`client/src/__tests__/support/file.ts:33`) already exists to prevent via its `finally`-block cleanup — worth a follow-up fix even though it's outside this PR's diff.

npapagna (approved):
Works great @MatiasFernandez!
Left a couple of comments.

.github/workflows/health-check.yml
MatiasFernandez: Fixed

scripts/lint-supply-chain.mjs
MatiasFernandez: Fixed — `getConfigs()` now does a single `npm config get k1 k2 ...` for all five keys and parses the `key=value` lines. Locally that's 0.38s → 0.08s for the config reads, and 25 script runs go from 10.5s to 2.5s; the test file's stale "five spawns" comment is updated too.

Went with multi-key `get` over `npm config list --json` because the JSON form changes the value types (`true`/`7`/`[]` instead of `'true'`/`'7'`/`''`), which every comparison and message here would have to re-stringify — including the literal `'null'` an unset `min-release-age` relies on. A missing key now throws `Unexpected npm config output` rather than silently reading as a failed assertion.

client/src/__tests__/lintSupplyChain.test.ts
MatiasFernandez: Fixed

client/src/__tests__/lintSupplyChain.test.ts
MatiasFernandez: Done — `runLint` now returns through `withTemporaryFolderDo`, so `tempDirs` and its `afterEach` are gone. The helper is generic over its return type now (`<T>(consumer: (p: string) => T): T`) so the `spawnSync` result can escape the callback.

The `rowanLoad.test.ts` leak was real, so it's fixed here too: `setup()`'s temp dir is now cleaned up in an `afterEach`. `withTemporaryFolderDo` doesn't fit there — the paths it returns are used after it returns.

.github/workflows/health-check.yml
MatiasFernandez: Fixed — `Supply-chain config check` and `Lockfile lint` now run right after `npm ci`, ahead of `Format check` and `Lint`. Left `npm audit signatures` last since it hits the registry.

.github/workflows/health-check.yml
MatiasFernandez: Fixed — folded into the commit that did the rename, since the reference was missed there.

PR #383, opened 2026-08-05
GemStone Explorer: method preview tabs, Methods-pane toggles, and instance-variable method filter (#259)

ericwinger:
## Summary
A batch of GemStone Explorer improvements, culminating in #259 item 2a (filter methods by instance-variable readers/writers). Client-side only, plus one read-only reflection query — no RB engine or server-plugin change.

### Method source editors — native preview tabs
- Single-click a method opens a VS Code **preview** tab (reused/replaced by the next click); **double-click** promotes it to a permanent tab; the 📌 action pins a tab. Replaces the hand-rolled transient-tab machinery.
- Fixed the senders/implementors **CodeLens pop-in** that shoved the source down on a method's first open (register the provider by scheme, and tag the doc `gemstone-smalltalk` before it enters the reused preview editor, avoiding the post-show language flip that re-queried CodeLens).

### Hierarchy pane
- Collapsed by default (more room for the other panes) and no longer force-opened when selecting a class/method or filtering by an ivar.

### Methods pane — leaner tree
- **Instance/class side** is now a title-bar toggle rather than a tree level.
- **Category grouping** is a title-bar toggle backed by `gemstone.explorer.groupMethodsByCategory` (lists methods flat when off). Filtering respects it.

### #259 Filtered searches
- **2a — filter methods by instance-variable access**: `reads:` / `writes:` / `accesses:<ivar>` filter tokens (`*` wildcard, composable with a selector prefix); right-click an instance-variable row → **Filter: Readers / Writers / References**; `r` / `w` / `rw` glyphs on matching rows; the filtered ivar highlighted in the opened source. Backed by a new `getMethodInstVarAccess` query (`GsNMethod>>instVarsRead` / `instVarsWritten`, scanning environments `0..maxEnv`).
- **Filter chip**: while a pane is filtered, its list leads with a distinct funnel-icon chip (pattern in grey + inline ✕), replacing the title-bar Clear button — rolled out to all four filterable panes (Dictionaries, Class Categories, Classes, Methods).
- Items 1 (filter classes by name) and 2b (quick method search) are the existing pane filters, now always-visible. **Item 2c (methods extended in packages) is split to #382** — deferred until Rowan package support is available.

### Housekeeping
- gitignore the GCI trace-log artifacts a failed `gci`-project run leaves behind.
- Give the flaky change-signature in-stone SUnit test a generous timeout (it was intermittently failing the pre-push gate under a busy shared stone).

## Testing
Client unit suite (including a live matrix integration test for the ivar query) + full lint + format + compile, all green. A `/review` pass over the diff found no HIGH issues; its one MED (ivar-access cache going stale after a method edit) and the actionable LOWs are fixed in this branch.

Addresses #259 (items 1, 2a, 2b); item 2c → #382.

### DIFF
--- a/client/src/__tests__/explorerViewRegistration.test.ts
+++ b/client/src/__tests__/explorerViewRegistration.test.ts
@@ -69,3 +69,27 @@ describe('GemStone Explorer Class Hierarchy pane', () => {
     expect(src).toContain(`createTreeView('${HIERARCHY}'`);
   });
 });
+
+// A command referenced from a menu (title bar, row context, palette) but never
+// declared in contributes.commands shows up only at runtime as "command not
+// found". Guard our own gemstone.* commands so a new menu entry can't drift out
+// of sync with its declaration.
+describe('command manifest consistency', () => {
+  const declared = new Set<string>(
+    (pkg.contributes.commands as Array<{ command: string }>).map((c) => c.command),
+  );
+  const menus = pkg.contributes.menus as Record<string, Array<{ command?: string }>>;
+  const menuCommands = new Set<string>();
+  for (const group of Object.values(menus)) {
+    for (const item of group) if (item.command) menuCommands.add(item.command);
+  }
+
+  it('declares every gemstone command that a menu references', () => {
+    const missing = [...menuCommands]
+      .filter((c) => c.startsWith('gemstone'))
+      .filter((c) => !declared.has(c))
+      .sort();
+
+    expect(missing).toEqual([]);
+  });
+});
--- a/client/src/__tests__/explorerViewRegistration.test.ts
+++ b/client/src/__tests__/explorerViewRegistration.test.ts
@@ -48,3 +48,24 @@ describe('GemStone Explorer views are registrable when created', () => {
     },
   );
 });
+
+// The Class Hierarchy pane starts collapsed so it doesn't crowd out the other
+// panes. VS Code only applies a view's declared `visibility` when it has no
+// stored layout for that view id, so the pane id must also stay in sync with
+// the id passed to createTreeView — a mismatch throws "No view is registered".
+const HIERARCHY = 'gemstoneExplorerClassHierarchy';
+
+describe('GemStone Explorer Class Hierarchy pane', () => {
+  it('starts collapsed to leave room for the other panes', () => {
+    const hierarchy = explorerViews.find((v) => v.id === HIERARCHY) as
+      { visibility?: string } | undefined;
+
+    expect(hierarchy?.visibility).toBe('collapsed');
+  });
+
+  it('registers the pane under the id the source creates it with', () => {
+    const src = fs.readFileSync(path.resolve(__dirname, '..', 'gemstoneExplorer.ts'), 'utf-8');
+
+    expect(src).toContain(`createTreeView('${HIERARCHY}'`);
+  });
+});
--- a/client/src/__tests__/explorerViewRegistration.test.ts
+++ b/client/src/__tests__/explorerViewRegistration.test.ts
@@ -93,3 +93,48 @@ describe('command manifest consistency', () => {
     expect(missing).toEqual([]);
   });
 });
+
+// The mirror of the command check above, for view ids. A menu `when` that gates
+// on a `view ==` id no view declares is silently, permanently false — the entry
+// just never appears, with no runtime error. That is exactly how the Insert /
+// Extract Superclass items vanished when the Hierarchy pane was renamed but two
+// of its menu clauses weren't. Guard every view reference so a renamed or
+// mistyped id fails a test instead of quietly dropping a menu item.
+describe('menu view-id consistency', () => {
+  const declaredViews = new Set<string>();
+  for (const views of Object.values(
+    pkg.contributes.views as Record<string, Array<{ id: string }>>,
+  )) {
+    for (const v of views) declaredViews.add(v.id);
+  }
+
+  const whens: string[] = [];
+  for (const group of Object.values(
+    pkg.contributes.menus as Record<string, Array<{ when?: string }>>,
+  )) {
+    for (const item of group) if (item.when) whens.push(item.when);
+  }
+
+  it('references only declared view ids in exact `view ==` menu clauses', () => {
+    const referenced = new Set<string>();
+    for (const when of whens) {
+      for (const m of when.matchAll(/view\s*==\s*([A-Za-z0-9_.]+)/g)) referenced.add(m[1]);
+    }
+
+    const missing = [...referenced].filter((id) => !declaredViews.has(id)).sort();
+
+    expect(missing).toEqual([]);
+  });
+
+  it('uses only `view =~` menu patterns that match at least one declared view', () => {
+    const dead: string[] = [];
+    for (const when of whens) {
+      for (const m of when.matchAll(/view\s*=~\s*\/([^/]+)\//g)) {
+        const re = new RegExp(m[1]);
+        if (![...declaredViews].some((id) => re.test(id))) dead.push(m[1]);
+      }
+    }
+
+    expect(dead).toEqual([]);
+  });
+});
--- a/client/src/gemstoneExplorer.ts
+++ b/client/src/gemstoneExplorer.ts
@@ -15,6 +15,13 @@ import {
   tabInputUri,
 } from './gemstoneFileSystemProvider';
 import { filterMatches } from './explorerFilter';
+import {
+  parseMethodFilter,
+  methodMatchesFilter as matchesMethodFilter,
+  ivarAccessMark as computeIvarAccessMark,
+  ivarIdentifierRanges,
+  MethodFilter,
+} from './explorerMethodFilter';
 import { DoubleClickDetector } from './explorerDoubleClick';
 import { categoryChildNodes, categoryParentPath, categoryMatches } from './explorerCategories';
 import { registerExplorerOpenEditors } from './explorerOpenEditors';
@@ -85,6 +92,14 @@ const VIEW_CLASSES = 'gemstoneExplorerClasses';
 const VIEW_METHODS = 'gemstoneExplorerMethods';
 // Panes that support the live filter (the Hierarchy pane doesn't).
 const EXPLORER_VIEWS = [VIEW_DICTS, VIEW_CATEGORIES, VIEW_CLASSES, VIEW_METHODS];
+
+// Highlights the filtered instance variable(s) in an opened method source while a
+// reads:/writes:/accesses: filter is active — theme-aware, styled like a search
+// match. One shared type for the extension's lifetime (disposed in activation).
+const ivarHighlightDecoration = vscode.window.createTextEditorDecorationType({
+  backgroundColor: new vscode.ThemeColor('editor.findMatchHighlightBackground'),
+  borderRadius: '2px',
+});
 
 // How the Explorer should open a gemstone:// source editor:
 //   - 'preview': a single-click NAVIGATION open — VS Code's own preview tab (the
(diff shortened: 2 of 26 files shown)


package.json
```diff
@@ -2064,22 +2131,22 @@
...
         {
           "command": "gemstone.explorer.renameClass",
-          "when": "view == gemstoneExplorerHierarchy && viewItem == explorerHierClass",
+          "when": "view == gemstoneExplorerClassHierarchy && viewItem == explorerHierClass",
           "group": "inline@3"
         },
         {
           "command": "gemstone.explorer.classHistory",
-          "when": "view == gemstoneExplorerHierarchy && viewItem == explorerHierClass",
+          "when": "view == gemstoneExplorerClassHierarchy && viewItem == explorerHierClass",
```
MatiasFernandez, line 2149: 🚫 The `gemstoneExplorerHierarchy` → `gemstoneExplorerClassHierarchy` rename misses two entries a few lines below: `gemstone.explorer.insertSuperclass` (line 2154) and `gemstone.explorer.extractSuperclass` (line 2159) still say `view == gemstoneExplorerHierarchy`.

No view has that id any more, so both `when` clauses are permanently false — "Insert Superclass" and "Extract Superclass" just disappear from the Hierarchy pane's context menu, and nothing anywhere reports an error to point at it.

client/src/gemstoneExplorer.ts
```diff
@@ -3300,22 +3525,10 @@ export class ExplorerController {
...
-        ],
-        { placeHolder: `New method on ${this.state.className}` },
-      );
-      if (!pick) return;
-      isMeta = pick.meta;
-    }
+    // Honor the last-touched side, else default to the side the Methods pane is
+    // currently showing (the instance/class title toggle) — a new method lands on
+    // the side you're looking at.
+    const isMeta = this.state.selectedIsMeta ?? this.showClassMethods;
```
MatiasFernandez, line 3544: 💭 Now that the side is a title-bar toggle, can `selectedIsMeta` go stale here? `setMethodSide` doesn't touch it, and `onDidChangeSelection` only fires when the selection actually changes. So: select an instance method (`selectedIsMeta = false`), flip the toggle to the class side, hit the title-bar "New Method" — the comment says "a new method lands on the side you're looking at", but `selectedIsMeta` wins and you get an instance method while the pane shows the class side.

Should `setMethodSide` clear `selectedIsMeta`, or should the pane's side win when the two disagree? And since this replaces the palette's Instance/Class QuickPick, is it worth a test pinning which side the new method lands on?

client/src/gemstoneExplorer.ts
```diff
@@ -1432,6 +1451,10 @@ export class ExplorerController {
   // in sync with the Classes pane.
   async revealHierarchySelf(): Promise<void> {
     if (this.hierChain.length === 0) return;
+    // Don't reveal when the Hierarchy pane is collapsed — reveal() would force
+    // VS Code to expand the section, defeating the collapsed-by-default layout
+    // and re-opening the pane every time the user selects a class.
+    if (!this.views?.hierarchy.visible) return;
```
MatiasFernandez, line 1459: 💭 `TreeView.visible` is false not only when the section is collapsed, but whenever the whole Explorer view container isn't showing (the user is in Source Control, the sidebar is hidden, …).

So if someone navigates classes while the Explorer isn't on screen — including via the editor-focus → `syncToEditor` path — the reveal is skipped, and nothing re-reveals it later: when they open the pane again it sits on whatever class was selected the last time it happened to be visible. Worth re-revealing on `onDidChangeVisibility` (or remembering a pending reveal) so the pane catches up when it reappears?

client/src/gemstoneExplorer.ts
```diff
@@ -2678,20 +2701,224 @@ export class ExplorerController {
...
+  // The instance-variable names to highlight in a method-source editor: the ivars
+  // the method actually reads/writes that match an active reads:/writes:/accesses:
+  // token. Empty unless a gemstone:// method source and an ivar filter line up.
+  ivarsToHighlight(uri: vscode.Uri): string[] {
+    if (uri.scheme !== 'gemstone') return [];
+    const raw = this.filters.get(VIEW_METHODS);
+    if (raw === undefined) return [];
+    const filter = this.parseFilter(raw);
+    if (filter.ivar.length === 0) return [];
+    // parts: ['', dictName, className, side, category, selector...]
```
MatiasFernandez: 💭 Should this reuse `parseUri` from `gemstoneFileSystemProvider` rather than splitting the path by hand? Beyond the duplication, `parseUri` already covers two cases this misses: it recognizes the 6-segment `…/new-method` URI (here it parses as a method whose selector is `new-method`), and it strips the `sel (base)` / `sel (session override)` labels the override-diff editors carry — so highlighting would work in those editors too, where today the lookup can never match an access entry.

(I see line 3250 hand-rolls the same parse, so this is a pre-existing pattern — but that's arguably an argument for converging on the helper.)

client/src/gemstoneExplorer.ts
```diff
@@ -3888,56 +4118,55 @@ class MethodProvider extends RefreshableProvider<MethodNode> {
...
-              ),
-          );
-      }
-      return this.ctl.methodCategories(element.isMeta);
+      const rows: MethodNode[] = !this.ctl.groupMethodsByCategory()
+        ? this.ctl.flatMethods(isMeta, filter)
+        : this.ctl.methodCategories(isMeta, filter);
+      // While filtering, lead with a filter chip (funnel row + inline ✕) so the
+      // active filter reads distinctly from the method rows and can be cleared here.
+      return filter === undefined ? rows : [new FilterChipItem(VIEW_METHODS, filter), ...rows];
```
MatiasFernandez: nit: This is what `withFilterChip` already does — should MethodProvider call the helper too, so all four panes prepend the chip in one place? `MethodNode` includes `FilterChipItem`, so the types should line up.

package.json
```diff
@@ -1932,11 +1999,31 @@
           "when": "view == gemstoneExplorerClasses && viewItem == explorerClass",
           "group": "inline@2"
         },
+        {
+          "command": "gemstone.explorer.clearFilterChip",
+          "when": "viewItem == explorerFilterChip",
```
MatiasFernandez: nit: Every other entry in this menu scopes itself with a `view ==` / `view =~` clause; this one matches on `viewItem` alone, so it's the one contribution that would attach to any tree in the window whose item happened to use `explorerFilterChip`. Since the chip shows in all four panes, would `view =~ /^gemstoneExplorer/ && viewItem == explorerFilterChip` be closer to the convention here?

client/src/__tests__/explorerViewRegistration.test.ts
```diff
@@ -48,3 +48,48 @@ describe('GemStone Explorer views are registrable when created', () => {
...
+  const declared = new Set<string>(
+    (pkg.contributes.commands as Array<{ command: string }>).map((c) => c.command),
+  );
+  const menus = pkg.contributes.menus as Record<string, Array<{ command?: string }>>;
+  const menuCommands = new Set<string>();
+  for (const group of Object.values(menus)) {
+    for (const item of group) if (item.command) menuCommands.add(item.command);
+  }
+
+  it('declares every gemstone command that a menu references', () => {
```
MatiasFernandez, line 87: 💭 Nice guard. Would the mirror-image check for view ids be cheap to add — that every `view == <id>` referenced from `contributes.menus` resolves to an id declared in `contributes.views`? It's the same class of silent drift (a permanently-false `when` instead of "command not found"), and it would have caught the two entries still pointing at `gemstoneExplorerHierarchy` (see my comment on package.json).

MatiasFernandez (approved):

package.json
ericwinger: Fixed in `fb902d1a`. Both entries now gate on `gemstoneExplorerClassHierarchy`. These were failing silently — no view carries the old id, so the `when` was permanently false and the two items just vanished with no error to point at. Added the manifest test you suggested below (your last comment) so a renamed/mistyped view id trips a test instead of quietly dropping a menu item.

client/src/gemstoneExplorer.ts
ericwinger: Done in `fb902d1a`, and I took the "converging on the helper" argument a step further. Extracted a shared `parseMethodUri(uri)` (wraps `parseUri`) and pointed five hand-rolled path-splits at it: `ivarsToHighlight`, `syncToEditor`, the CodeLens provider, and the two *identical* private `parseMethodUri`s in `breakpointManager` + `selectorBreakpointManager`. Net effect: the `sel (base)` diff-label and slash-in-selector cases you flagged now work everywhere, `new-method` templates stop being parsed as real methods, and ~54 lines of duplicated parsing go away (one 42-line documented helper replaces them). Added helper unit tests plus a CodeLens regression test (no lenses on a template).

client/src/gemstoneExplorer.ts
ericwinger: Done in `fb902d1a` — the Methods pane calls `withFilterChip` now, so all four panes prepend the chip in exactly one place. And you're right that `MethodNode` already includes `FilterChipItem`, so the types lined up with no other change.

package.json
ericwinger: Done in `fb902d1a`: `view =~ /^gemstoneExplorer/ && viewItem == explorerFilterChip`, so it can't attach to some other tree in the window. I kept the regex over spelling out four `view ==`s; the new manifest test (your next comment) also checks that the pattern matches at least one real view, so it can't silently go dead either.

client/src/gemstoneExplorer.ts
ericwinger: Fixed in `fb902d1a` by letting the visible side win: `setMethodSide` now syncs `selectedIsMeta` to the toggled side and clears the now-stale category (otherwise the `selectedIsMeta === isMeta` guard would reuse the *other* side's category). Worth stressing it's symmetric — depending on which way the toggle and the last selection disagree, you'd get a class method when you wanted instance *or* the reverse. Added tests pinning both directions, the init default, the sync, and that a still-matching side keeps reusing its category.

client/src/gemstoneExplorer.ts
ericwinger: Fixed in `fb902d1a`. Added `onHierarchyVisibilityChanged` wired to `onDidChangeVisibility`: when the pane reappears it re-reveals the current class. It only acts on *becoming* visible, so a deliberately-collapsed pane is never forced open — the collapsed-by-default layout is preserved. Tests cover "navigated while the pane was hidden → catches up on show" and "no re-reveal on the hide event".

client/src/__tests__/explorerViewRegistration.test.ts
ericwinger: Great idea — added in `fb902d1a`. A new `menu view-id consistency` block asserts every `view ==` id referenced from `contributes.menus` resolves to a declared view, and every `view =~` pattern matches at least one. Verified it fails on an undeclared id — it would have caught the Hierarchy rename from your package.json comment above.

package.json
ericwinger: Eric: this is a good catch. Vanishing menu items is bad.

client/src/gemstoneExplorer.ts
ericwinger: Eric: I asked claude to make these two always agree. Can't understand why they shouldn't always be in sync.

client/src/gemstoneExplorer.ts
ericwinger: I expanded the scope a bit on this and tried to pass a number of very similar places in the code through the helper. Got rid of a bunch of duplication.

PR #384, opened 2026-08-06
Uninstall server plugin support; isolate Enhanced Inspector in GsEnhancedInspector (#318)

ericwinger:
## Summary

Implements #318: adds a way to **uninstall** the optional server-plugin support, and reworks the **Enhanced Inspector** so it can be removed cleanly.

Before this, neither the Enhanced Inspector nor the refactoring engine had any removal path — once filed into an image, their classes stayed forever. The Enhanced Inspector was also filed into the shared `Published` dictionary, commingled with unrelated content, so it couldn't be dropped as a unit. This change gives both an install/uninstall pair and isolates the Enhanced Inspector in its own `GsEnhancedInspector` dictionary — the same isolation the refactoring engine already uses with `GsRefactoring`.

## What's included

**Uninstall command (`GemStone: Uninstall Server Support`)**
- Mirrors the install flow: SystemUser elevation, progress, verify. Confirmation modal before it runs; a single consolidated `GemStone server support installed/uninstalled.` toast (not one per feature).
- Visible only when something is installed (`gemstone.serverSupportInstalled` context key); also reachable from the settings/walkthrough links.
- **Refactoring:** drops `GsRefactoring` from every user's symbol list, empties the dictionary object, removes the loader and the feature-detected `*ast-core-compat` kernel backports.
- **Enhanced Inspector:** drops `GsEnhancedInspector` from every user's symbol list and removes the `*GToolkit` extension methods, plus a legacy `Published` sweep so stones installed by the old build are fully cleaned.

**Enhanced Inspector → dedicated dictionary**
- The vendored payload is now filed into `GsEnhancedInspector` (build transform `apply_jasper_transforms.sh` retargeted; topaz loader updated to create/share the dictionary first). Install creates + shares the dictionary and migrates any legacy `Published`-resident classes.

**UI gating the uninstall exposed**
- Every engine-dependent Explorer refactoring affordance and the editor "Refactor…" code actions now gate on `gemstone.rbSupportAvailable`, so they disappear when the engine is absent and return after reinstall (Delete Method / Rename Method Category stay — they don't need the engine).
- Per-pane **Filter** buttons no longer require the pane to be focused (always visible, like Find Class).
- The Explorer clears the class/hierarchy/method panes when the selected dictionary is removed (e.g. uninstalling while sitting on `GsRefactoring`), instead of leaving orphaned classes.

## Testing

- **86 new tests.** Unit tests for both uninstall cores + command drivers, the consolidated toast, the dedicated-dictionary prepare/migrate step, the menu/code-action gating, the filter-visibility change, and the Explorer stale-pane reset.
- **Live-stone integration tests** for the uninstall on both engines (dedicated-dict isolation + detection flip), transient (harness auto-aborts — no commit, **no new `gci/` tests**).
- Full local CI matrix on **3.6.2 and 3.7.5**, both plugin worlds, green apart from two pre-existing in-stone-SUnit timeouts (change-signature suite, `querySunitRunLimit`) unrelated to this change.

### DIFF
--- a/client/src/__tests__/explorerRefreshSelectionGone.test.ts
+++ b/client/src/__tests__/explorerRefreshSelectionGone.test.ts
@@ -1,0 +1,57 @@
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+vi.mock('vscode', () => import('../__mocks__/vscode.js'));
+// refreshRetainingSelection re-resolves the selected dictionary by name and, when
+// present, reloads its classes. Stub just those two reads; the ivar-count reload
+// swallows its own query errors, so the rest can stay unmocked.
+vi.mock('../browserQueries', () => ({
+  getDictionaryNames: vi.fn(() => [] as string[]),
+  getClassesWithCategory: vi.fn(() => [] as unknown[]),
+}));
+
+import * as queries from '../browserQueries';
+import { ExplorerController } from '../gemstoneExplorer';
+import type { SessionManager, ActiveSession } from '../sessionManager';
+
+const getDictionaryNames = queries.getDictionaryNames as ReturnType<typeof vi.fn>;
+const getClassesWithCategory = queries.getClassesWithCategory as ReturnType<typeof vi.fn>;
+
+function makeController() {
+  const session = { id: 1 } as ActiveSession;
+  const sessionManager = { getSelectedSession: () => session } as unknown as SessionManager;
+  const ctl = new ExplorerController(sessionManager);
+  const state = (ctl as unknown as { state: Record<string, unknown> }).state;
+  // Pretend the user is sitting on the GsRefactoring dictionary (index 6).
+  state.dictName = 'GsRefactoring';
+  state.dictIndex = 6;
+  state.className = undefined;
+  const reset = vi.spyOn(ctl, 'reset').mockImplementation(() => {});
+  return { ctl, reset };
+}
+
+describe('ExplorerController.refreshRetainingSelection when the selected dictionary is gone', () => {
+  beforeEach(() => vi.clearAllMocks());
+
+  it('resets (clearing the class/hierarchy/method panes) when the selected dictionary no longer exists', async () => {
+    const { ctl, reset } = makeController();
+    getDictionaryNames.mockReturnValue(['UserGlobals', 'Globals', 'Published']);
+
+    await ctl.refreshRetainingSelection({ reveal: false });
+
+    expect(reset).toHaveBeenCalledTimes(1);
+    // The stale index is never used to fetch a different dictionary's classes.
+    expect(getClassesWithCategory).not.toHaveBeenCalled();
+  });
+
+  it('keeps the selection and reloads by the CURRENT index when the dictionary is still present but shifted', async () => {
+    const { ctl, reset } = makeController();
+    // GsRefactoring survives but earlier dictionaries were removed, so it is now at
+    // position 3 (1-based), not its stale index 6.
+    getDictionaryNames.mockReturnValue(['UserGlobals', 'Globals', 'GsRefactoring']);
+
+    await ctl.refreshRetainingSelection({ reveal: false });
+
+    expect(reset).not.toHaveBeenCalled();
+    expect(getClassesWithCategory).toHaveBeenCalledWith(expect.anything(), 3);
+  });
+});
--- a/client/src/__tests__/explorerRefreshSelectionGone.test.ts
+++ b/client/src/__tests__/explorerRefreshSelectionGone.test.ts
@@ -25,33 +25,55 @@ function makeController() {
   state.dictName = 'GsRefactoring';
   state.dictIndex = 6;
   state.className = undefined;
-  const reset = vi.spyOn(ctl, 'reset').mockImplementation(() => {});
-  return { ctl, reset };
+  // Seed the panes so "they end up empty" is an observable change rather than the initial
+  // condition — otherwise the assertions below would hold even if nothing ran.
+  state.classCategory = 'accessing';
+  (ctl as unknown as { classCategoryEntries: unknown[] }).classCategoryEntries = [
+    { category: 'accessing', className: 'Demo' },
+  ];
+  (ctl as unknown as { envLines: unknown[] }).envLines = [
+    { isMeta: false, envId: 0, category: 'accessing', selectors: ['at:'] },
+  ];
+  // NB `reset` is deliberately NOT stubbed: the behaviour under test is that the panes end up
+  // cleared, not that a particular private method was called. Stubbing it would make this test
+  // fail on a rework that cleared the panes inline while the behaviour stayed correct.
+  return { ctl, state };
 }
 
 describe('ExplorerController.refreshRetainingSelection when the selected dictionary is gone', () => {
   beforeEach(() => vi.clearAllMocks());
 
-  it('resets (clearing the class/hierarchy/method panes) when the selected dictionary no longer exists', async () => {
-    const { ctl, reset } = makeController();
+  it('clears the class/hierarchy/method panes when the selected dictionary no longer exists', async () => {
+    const { ctl, state } = makeController();
     getDictionaryNames.mockReturnValue(['UserGlobals', 'Globals', 'Published']);
 
     await ctl.refreshRetainingSelection({ reveal: false });
 
-    expect(reset).toHaveBeenCalledTimes(1);
-    // The stale index is never used to fetch a different dictionary's classes.
-    expect(getClassesWithCategory).not.toHaveBeenCalled();
+    // The observable outcome: the stale selection is gone (reset falls back to the default
+    // dictionary rather than leaving nothing selected) and the dependent panes are empty.
+    expect(state.dictName).not.toBe('GsRefactoring');
+    expect(state.className).toBeUndefined();
+    expect(state.classCategory).toBeUndefined();
+    expect((ctl as unknown as { classCategoryEntries: unknown[] }).classCategoryEntries).toEqual(
+      [],
+    );
+    expect((ctl as unknown as { envLines: unknown[] }).envLines).toEqual([]);
+    // The point of the guard: the stale index 6 is never used to fetch some other dictionary's
+    // classes. (Classes ARE loaded — reset selects the default dictionary — so asserting "not
+    // called at all" would only have held while `reset` was stubbed out.)
+    expect(getClassesWithCategory).not.toHaveBeenCalledWith(expect.anything(), 6);
   });
 
   it('keeps the selection and reloads by the CURRENT index when the dictionary is still present but shifted', async () => {
-    const { ctl, reset } = makeController();
+    const { ctl, state } = makeController();
     // GsRefactoring survives but earlier dictionaries were removed, so it is now at
     // position 3 (1-based), not its stale index 6.
     getDictionaryNames.mockReturnValue(['UserGlobals', 'Globals', 'GsRefactoring']);
 
     await ctl.refreshRetainingSelection({ reveal: false });
 
-    expect(reset).not.toHaveBeenCalled();
+    // Selection retained — the inverse of the reset case above.
+    expect(state.dictName).toBe('GsRefactoring');
     expect(getClassesWithCategory).toHaveBeenCalledWith(expect.anything(), 3);
   });
 });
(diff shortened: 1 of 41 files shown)


client/src/optionalSupportOffer.ts
```diff
@@ -195,3 +246,63 @@ export async function runInstallServerSupport(
...
+  }
+  // The refactoring engine removed its `GsRefactoring` dictionary from the stone,
+  // but the Explorer loaded its dictionary list on connect — before this ran — so
+  // the now-gone dictionary will linger in the tree until the list is reloaded.
+  // Reload it now so the Explorer reflects the stone without the user hitting
+  // Refresh. Harmless when nothing dictionary-shaped changed, and the command
+  // no-ops gracefully when the Explorer has no active session.
+  await vscode.commands.executeCommand('gemstone.explorer.refresh');
+  refreshServerSupportInstalledContext(base);
+  announceServerSupportChange('uninstalled', outcomes);
```
MatiasFernandez, line 268: 💭 `outcomes` here is what each `uninstall()` returned, and both drivers return `base.<latch> !== true` — a latch that only gets re-probed when `refreshWorkingSessionAfterInstall` returns true. That helper returns `false` when the session has uncommitted changes and the user picks "Later" on the refresh prompt.

So in that path the removal really did run, commit, and verify server-side, but the driver reports `false`, `announceServerSupportChange` stays silent, and the user gets neither the success toast nor an error — the operation just looks like it didn't happen. Same shape applies to `installFeatures` above, which is a change from before: the drivers used to toast unconditionally right after a verified install, so this consolidation quietly drops the message in the "Later" case.

Would it make sense to key the toast off the server-side result (`result.success` from the uninstall/install core) and keep the latch purely for the context keys / menu gating? That separates "did the change land on the stone" from "does this session see it yet".

client/src/serverPlugin/pluginFeatures.ts
```diff
@@ -65,6 +76,10 @@ export interface PluginFeature {
     payloadDir: string,
     onProgress?: ProgressReporter,
   ): Promise<PluginInstallResult>;
+  /** Remove the feature from the stone over a write-capable (SystemUser)
+   *  session. No payload directory: removal is expressed as inline Smalltalk,
+   *  so it needs nothing from disk. */
+  uninstall(session: ActiveSession, onProgress?: ProgressReporter): Promise<PluginUninstallResult>;
```
MatiasFernandez, line 82: 💭 I couldn't find a caller for this one. `PLUGIN_FEATURES` is only consumed by `installServerPlugin.ts` (install side), and the uninstall flow in `optionalSupportOffer.ts` goes through the command drivers (`uninstallRefactoringFeature` / `uninstallEnhancedInspectorFeature`) rather than the registry. So `uninstall` on `PluginFeature`, plus both implementations below and `PluginUninstallResult`, look like they have no consumers yet.

Is there a follow-up (a headless `uninstallServerPlugin.ts` mirroring the install script?) that will pick this up? If so a one-line note here would help; if not, might be worth leaving it out until something needs it, so the registry doesn't carry an untested-by-use branch.

client/src/enhancedInspector/__tests__/enhancedInspectorUninstall.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    const removalCode = String(
+      executeFetchStringMock.mock.calls.find((c) =>
+        String(c[1]).includes('GsEnhancedInspector'),
+      )?.[1],
+    );
+    expect(removalCode).toContain('dict keys asArray do:');
+    expect(removalCode).toContain('dict removeKey: k');
+  });
+
+  it('also sweeps any legacy GToolkit classes left in the shared Published dictionary', async () => {
```
MatiasFernandez: 💭 A few of the tests in this file (this one, plus the `#GsEnhancedInspector`/`AllUsers`/`removeDictionaryAt:` and `dict keys asArray do:` ones) assert on the *text* of `REMOVAL_SNIPPET` rather than on an outcome. They'll pass even if the snippet is semantically wrong, and they'll fail on a safe rewrite — e.g. extracting the Published sweep into its own statement, or renaming `dict` to `d`.

The integration test in `enhancedInspectorUninstall.integration.test.ts` is the one that actually proves the removal works, and the surrounding unit tests here (commit/abort/verify/progress) all assert real behavior. Should these three collapse into one "runs a removal snippet against the stone" check and let the integration test carry the correctness weight? Same question for the mirrored ones in `refactoringUninstall.test.ts`.

client/src/enhancedInspector/enhancedInspectorInstall.ts
```diff
@@ -93,6 +110,44 @@ export const ENHANCED_INSPECTOR_FILES: readonly string[] = [
...
+dict isNil ifTrue: [
+	dict := SymbolDictionary new name: sym; yourself.
+	dict at: sym put: dict.
+	prof insertDictionary: dict at: list size + 1 ].
+AllUsers do: [:p |
+	(p symbolList detect: [:d | d name == sym] ifNone: [nil]) isNil
+		ifTrue: [ p insertDictionary: dict at: p symbolList size + 1 ] ].
+pub := list detect: [:d | d name == #Published] ifNone: [nil].
+pub notNil ifTrue: [
+	pub keys asArray do: [:k |
```
MatiasFernandez, line 148: 💭 This legacy-migration sweep removes *any* class in the shared `Published` dictionary whose category begins with `GToolkit` — and it runs on every install, including a fresh one on a stone that never had the old build.

That's a broader fingerprint than the method sweep in the uninstall snippet, which anchors on `'*GToolkit'` (the leading `*` is what makes it provably Jasper's own extension category, as the doc comment there nicely explains). A bare `GToolkit` class-category prefix would also match a user's own class they happened to file into `Published` under a `GToolkit…` category — and since this runs inside the install transaction, it'd get unbound and committed along with the payload.

Would it be safer to scope this to the payload's own class names (we know them — they're what the `.gs` files declare), or at least to skip the sweep entirely when `GsEnhancedInspector` didn't already exist as a legacy install?

client/src/enhancedInspector/enhancedInspectorUninstallCommand.ts
```diff
@@ -0,0 +1,100 @@
...
+ * warning if the default is unavailable. Returns whether the support is gone
+ * afterward. Called by the unified optional-support bundle uninstall.
+ */
+export async function uninstallEnhancedInspectorFeature(
+  base: ActiveSession,
+  sessionManager: SessionManager,
+  interactive: boolean,
+): Promise<boolean> {
+  await performUninstall(base, sessionManager, interactive);
+  return base.enhancedInspectorAvailable !== true;
```
MatiasFernandez: nit: `uninstallRefactoringFeature`'s `performUninstall` returns the boolean itself, while this one returns `void` and the feature re-derives the answer from the latch. The rest of the PR is careful to make these two read as a matched pair — worth making `performUninstall` return `boolean` here too?

client/src/__tests__/explorerRefreshSelectionGone.test.ts
```diff
@@ -0,0 +1,57 @@
...
+function makeController() {
+  const session = { id: 1 } as ActiveSession;
+  const sessionManager = { getSelectedSession: () => session } as unknown as SessionManager;
+  const ctl = new ExplorerController(sessionManager);
+  const state = (ctl as unknown as { state: Record<string, unknown> }).state;
+  // Pretend the user is sitting on the GsRefactoring dictionary (index 6).
+  state.dictName = 'GsRefactoring';
+  state.dictIndex = 6;
+  state.className = undefined;
+  const reset = vi.spyOn(ctl, 'reset').mockImplementation(() => {});
```
MatiasFernandez: 💭 Stubbing `reset` means the test asserts "the controller called its own method" rather than the outcome the change is actually about — the class/hierarchy/method panes ending up empty. If someone reworked `refreshRetainingSelection` to clear `classCategoryEntries`/`envLines` inline instead of delegating to `reset`, this test would fail while the behavior stayed correct.

Could we let `reset` run for real and assert on the observable state instead (empty `classCategoryEntries`, cleared `className`)? The second test already asserts the good kind of thing — that `getClassesWithCategory` is called with the *current* index 3, not the stale 6.

client/src/enhancedInspector/enhancedInspectorInstall.ts
MatiasFernandez: We can also probably mark this as removable in the future. This should be code just needed for the devs who tried the enhanced inspector before the new installation method is merged, so we can ask the team to run this and avoid keeping this code around forever

client/src/serverPlugin/pluginFeatures.ts
MatiasFernandez: I think it would be good to get a `test:server:uninstall-plugin` command, so instead of removing this, it would probably be good to go the extra mile and complete the implementation

MatiasFernandez (approved):
Left some comment worth reviewing

client/src/enhancedInspector/enhancedInspectorInstall.ts
ericwinger: > "Would it be safer to scope this to the payload's own class names (we know them — they're what the .gs files declare), or at least to skip the sweep entirely when GsEnhancedInspector didn't already exist as a legacy install?"

The class names are in the .gs file but don't we need to install them to get the names?  Or would you do string searching in the .gs files?   ... I'm asking because the sweep has to occur before the GsEnhancedInspector dictionary is installed.

client/src/optionalSupportOffer.ts
ericwinger: Fixed in c32c4a23. The install drivers were still keying their return on the availability latch (`base.rbSupportAvailable` / `base.enhancedInspectorAvailable`), so in the deferred-refresh ("Later") path a verified, committed install reported `false` and the bundle showed neither a success toast nor an error — exactly the shape you described. Both install drivers now return the verified server-side result (EI's `performInstall` went `void`→`boolean` to carry it), and the latch is kept purely for the context keys / menu gating, as you suggested. The uninstall drivers already did this. Added deferred-refresh regression tests for both install drivers (negative-control-verified that they fail against the old latch-based return).

client/src/serverPlugin/pluginFeatures.ts
ericwinger: Went the extra mile as suggested (e97ca6fb) rather than removing it: added a `test:server:uninstall-plugin` npm script + `client/bin/uninstall-server-plugin.mjs` + `serverPlugin/uninstallServerPlugin.ts`, mirroring the install path. `PLUGIN_FEATURES` now has its uninstall-side consumer — `uninstallServerPlugin.ts` iterates `[...PLUGIN_FEATURES].reverse()` to remove each feature and then re-probes the remainder — so it stays.

client/src/enhancedInspector/__tests__/enhancedInspectorUninstall.test.ts
ericwinger: Fixed — these now assert on the effect (`result.success`/`verified`, that commit was called, and that the removal targets the `GsEnhancedInspector` dictionary) instead of the text of `REMOVAL_SNIPPET`. A comment in the file records that the old text-assertions passed even when the snippet did nothing.

client/src/enhancedInspector/enhancedInspectorInstall.ts
ericwinger: Scoped as we discussed. The sweep is now gated on the legacy marker `#GtRemotePhlowViewedObject` that the earlier build is known to have bound into `Published`, so on a fresh install the whole branch is skipped and only a stone that demonstrably carried the old placement is ever swept. Agreed it's legacy-migration-only — we can drop it entirely once the new install method has been out for a while.

client/src/enhancedInspector/enhancedInspectorUninstallCommand.ts
ericwinger: Fixed — `performUninstall` now returns `Promise<boolean>` (the verified server-side result) and `uninstallEnhancedInspectorFeature` returns it, matching `uninstallRefactoringFeature`. This is also what feeds the deferred-refresh fix in the optionalSupportOffer thread.

client/src/__tests__/explorerRefreshSelectionGone.test.ts
ericwinger: Fixed — the test no longer stubs `reset`; it asserts the observable outcome (the stale selection is gone and the dependent class/hierarchy/method panes end up empty). A comment notes `reset` is deliberately left real so the test verifies the effect rather than a private-method call.

PR #389, opened 2026-08-07
#362: refactoring-level wide-source regression test + asCount TSdoc

ericwinger:
Closes the two unresolved review threads from #361 (split out as #362 so #361 could land). **Low-risk: a test and a doc comment only — no production code changes.**

## 1. Refactoring-level wide-source regression test
`GsInstVarStructureRefactoringTest >> testConvertTempSucceedsWhenMethodSourceIsWide`

The #361 `GsRefactoringParseTest` pins the shim itself (`isSqueakSeparator` answers; `RBParser parseMethod:` survives wide source). But the thing that was actually broken for users sits one level up: callers wrap the parse in `[RBParser parseMethod: src] on: Error do: [:e | nil]` (e.g. `analyzeConvertTemp`), so pre-shim the DNU came back as `nil` and convert-temp hard-declined *"the method source does not parse."* That swallow is still there, so this adds a test that drives the **whole refactoring path**: a convert-temp over a method whose STORED SOURCE carries a wide (em-dash) comment, asserting it does **not** decline.

Confirmed with a negative control: with the shim it succeeds (`decline = nil`); with the shim removed it declines *"Cannot convert #w: the method source does not parse."* — so this is exactly the test that would have failed before #361. The em-dash is built via `#codePoint:` so the test's own source stays pure ASCII (3.6.2 discipline).

**Verified: `GsInstVarStructureRefactoringTest` 53/53 on both 3.6.2 and 3.7.5** (non-committing topaz file-in + abort).

## 2. TSdoc for `asCount`
`client/src/refactoring/methodRelocationPreview.ts` — the one export without a TSdoc, and the least guessable: it clamps non-numbers, negatives, `NaN`, and `Infinity` all to `0` rather than throwing or returning `undefined`. Added a one-line explanation.

## Testing
Pre-push `npm test` green (client + server + mcp); RB SUnit both boundaries as above; `tsc` + prettier clean.

Follow-up to #362.

### DIFF
--- a/client/src/refactoring/methodRelocationPreview.ts
+++ b/client/src/refactoring/methodRelocationPreview.ts
@@ -1,3 +1,4 @@
+import { asCount } from './previewCounts';
 /**
  * Shared "method-relocation" preview primitives (RB catalog C2). The move-method (M6)
  * and push-up / push-down (M7 / M8) client families model a relocation as a set of
@@ -80,16 +81,6 @@ export interface RelocationAnalysis<S extends BaseSelectorAnalysis> {
   globalDecline: string | null;
   movableCount: number;
   selectors: S[];
-}
-
-/**
- * Coerce a JSON value to a non-negative count. Anything that is not a finite, non-negative
- * number -- a non-number, a negative, `NaN`, or `Infinity` -- clamps to `0` (rather than
- * throwing or returning `undefined`), so a malformed engine payload degrades to a zero/empty
- * count instead of crashing the parse.
- */
-export function asCount(v: unknown): number {
-  return typeof v === 'number' && Number.isFinite(v) && v >= 0 ? v : 0;
 }
 
 /**
--- a/client/src/refactoring/methodRelocationPreview.ts
+++ b/client/src/refactoring/methodRelocationPreview.ts
@@ -82,6 +82,12 @@ export interface RelocationAnalysis<S extends BaseSelectorAnalysis> {
   selectors: S[];
 }
 
+/**
+ * Coerce a JSON value to a non-negative count. Anything that is not a finite, non-negative
+ * number -- a non-number, a negative, `NaN`, or `Infinity` -- clamps to `0` (rather than
+ * throwing or returning `undefined`), so a malformed engine payload degrades to a zero/empty
+ * count instead of crashing the parse.
+ */
 export function asCount(v: unknown): number {
   return typeof v === 'number' && Number.isFinite(v) && v >= 0 ? v : 0;
 }
--- a/gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
+++ b/gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
@@ -143,6 +143,25 @@ GsInstVarStructureRefactoringTest >> testConvertTempStagesIvarEditAndMethodRecom
 ]
 
 { #category : 'tests - V5 convert temp' }
+GsInstVarStructureRefactoringTest >> testConvertTempSucceedsWhenMethodSourceIsWide [
+	"Regression for the RBScanner wide-source parse gap (#361 / #362): converting a temporary in a
+	 method whose STORED SOURCE carries a non-ASCII (wide) comment must NOT decline. Before the
+	 Character>>isSqueakSeparator compat shim, RBParser parseMethod: DNU'd on wide source, the caller
+	 swallowed it to nil (analyzeConvertTemp: `[RBParser parseMethod: src] on: Error do: [:e | nil]`),
+	 and convert-temp hard-declined 'the method source does not parse'. This drives the WHOLE
+	 refactoring path (not just the parser, as GsRefactoringParseTest does), so a future regression
+	 anywhere along it is caught. The em-dash is built with #codePoint: so THIS test's own source
+	 stays pure ASCII (the 3.6.2 ComStrmSetCursor discipline)."
+	| wide ref |
+	wide := 'computeWide | w | "carry ', (String with: (Character codePoint: 8212)), ' forward" w := shared. ^ w'.
+	self compile: wide in: (self classNamed: 'GsVSBase').
+	ref := GsInstVarStructureRefactoring
+		class: (self classNamed: 'GsVSBase') convertTemporary: 'w' inMethod: #computeWide meta: false.
+	self assert: ref decline isNil.
+	self deny: ref analysisJsonString includesSubstring: 'does not parse'
+]
+
+{ #category : 'tests - V5 convert temp' }
 GsInstVarStructureRefactoringTest >> testConvertTempAppliesAddingIvarAndDroppingDecl [
 	| json |
 	json := (GsInstVarStructureRefactoring
--- a/gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
+++ b/gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
@@ -152,13 +152,37 @@ GsInstVarStructureRefactoringTest >> testConvertTempSucceedsWhenMethodSourceIsWi
 	 refactoring path (not just the parser, as GsRefactoringParseTest does), so a future regression
 	 anywhere along it is caught. The em-dash is built with #codePoint: so THIS test's own source
 	 stays pure ASCII (the 3.6.2 ComStrmSetCursor discipline)."
-	| wide ref |
-	wide := 'computeWide | w | "carry ', (String with: (Character codePoint: 8212)), ' forward" w := shared. ^ w'.
-	self compile: wide in: (self classNamed: 'GsVSBase').
+	| emDash wideSource ref cs edit recompile |
+	emDash := String with: (Character codePoint: 8212).
+	wideSource := 'computeWide | w | "carry ', emDash, ' forward" w := shared. ^ w'.
+	self compile: wideSource in: (self classNamed: 'GsVSBase').
+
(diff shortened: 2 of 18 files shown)


gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
```diff
@@ -142,6 +142,25 @@ GsInstVarStructureRefactoringTest >> testConvertTempStagesIvarEditAndMethodRecom
...
+	 method whose STORED SOURCE carries a non-ASCII (wide) comment must NOT decline. Before the
+	 Character>>isSqueakSeparator compat shim, RBParser parseMethod: DNU'd on wide source, the caller
+	 swallowed it to nil (analyzeConvertTemp: `[RBParser parseMethod: src] on: Error do: [:e | nil]`),
+	 and convert-temp hard-declined 'the method source does not parse'. This drives the WHOLE
+	 refactoring path (not just the parser, as GsRefactoringParseTest does), so a future regression
+	 anywhere along it is caught. The em-dash is built with #codePoint: so THIS test's own source
+	 stays pure ASCII (the 3.6.2 ComStrmSetCursor discipline)."
+	| wide ref |
+	wide := 'computeWide | w | "carry ', (String with: (Character codePoint: 8212)), ' forward" w := shared. ^ w'.
+	self compile: wide in: (self classNamed: 'GsVSBase').
```
MatiasFernandez: 💭 Should we assert the precondition that the stored source really is wide, before asserting the outcome?

Right now the only thing that makes this a wide-source test is the em-dash inside the string we hand to `compile:in:`. If `String with: (Character codePoint: 8212)` ever stops round-tripping through the compiler, or the fixture's `compile:in:` (which resumes `CompileWarning` with `nil`) quietly swallows a compile problem, the test still goes green and we've silently lost the regression — the method would just be plain ASCII and the shim wouldn't be exercised at all.

Something in the shape of line 154's assertion would pin it explicitly:

```smalltalk
self assert: (((self classNamed: 'GsVSBase') compiledMethodAt: #computeWide environmentId: 0 otherwise: nil) sourceString)
	includesSubstring: (String with: (Character codePoint: 8212)).
```

That way a lost em-dash fails loudly instead of turning this into a test of nothing.

gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
```diff
@@ -142,6 +142,25 @@ GsInstVarStructureRefactoringTest >> testConvertTempStagesIvarEditAndMethodRecom
...
+	ref := GsInstVarStructureRefactoring
+		class: (self classNamed: 'GsVSBase') convertTemporary: 'w' inMethod: #computeWide meta: false.
+	self assert: ref decline isNil.
+	self deny: ref analysisJsonString includesSubstring: 'does not parse'
```
MatiasFernandez: 💭 `self assert: ref decline isNil` already implies the decline text can't be in the analysis JSON, so the second assertion is close to a tautology. Would it be worth asserting the positive outcome instead — that the change set stages the ivar edit and the method recompile, the way `testConvertTempStagesIvarEditAndMethodRecompile` does just above?

The failure mode I'd most want covered on the wide path is the source parsing fine but the *recompiled* source coming back with the em-dash mangled or truncated. Today that scenario passes this test.

gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
```diff
@@ -142,6 +142,25 @@ GsInstVarStructureRefactoringTest >> testConvertTempStagesIvarEditAndMethodRecom
...
+	"Regression for the RBScanner wide-source parse gap (#361 / #362): converting a temporary in a
+	 method whose STORED SOURCE carries a non-ASCII (wide) comment must NOT decline. Before the
+	 Character>>isSqueakSeparator compat shim, RBParser parseMethod: DNU'd on wide source, the caller
+	 swallowed it to nil (analyzeConvertTemp: `[RBParser parseMethod: src] on: Error do: [:e | nil]`),
+	 and convert-temp hard-declined 'the method source does not parse'. This drives the WHOLE
+	 refactoring path (not just the parser, as GsRefactoringParseTest does), so a future regression
+	 anywhere along it is caught. The em-dash is built with #codePoint: so THIS test's own source
+	 stays pure ASCII (the 3.6.2 ComStrmSetCursor discipline)."
+	| wide ref |
+	wide := 'computeWide | w | "carry ', (String with: (Character codePoint: 8212)), ' forward" w := shared. ^ w'.
```
MatiasFernandez: nit: `wide` holds the method source, not the wide character — `wideSource` (or `src`, matching the `parseMethod: src` in the method comment) would read a little closer to what it actually contains.

client/src/refactoring/methodRelocationPreview.ts
```diff
@@ -82,6 +82,12 @@ export interface RelocationAnalysis<S extends BaseSelectorAnalysis> {
   selectors: S[];
 }
 
+/**
+ * Coerce a JSON value to a non-negative count. Anything that is not a finite, non-negative
+ * number -- a non-number, a negative, `NaN`, or `Infinity` -- clamps to `0` (rather than
+ * throwing or returning `undefined`), so a malformed engine payload degrades to a zero/empty
+ * count instead of crashing the parse.
```
MatiasFernandez: 💭 The doc enumerates exactly what clamps to `0`, which nudges the reader into assuming everything else comes back as a usable count — but a fractional value like `2.5` passes straight through. Since these feed `total`, `applied`, `nextOffset` and friends into the preview UI, is it worth either saying so here ("a finite non-negative number is returned as-is, including non-integers") or clamping with `Math.trunc`?

client/src/refactoring/methodRelocationPreview.ts
```diff
@@ -82,6 +82,12 @@ export interface RelocationAnalysis<S extends BaseSelectorAnalysis> {
...
+ */
```
MatiasFernandez: 💭 There are 13 copies of this same helper across `client/src/refactoring/*Preview.ts`, and this is the only exported one. Now that it's also the only documented one, would it make sense to point the others at it — or at least note here that this is the canonical copy?

As it stands, a reader (or an agent) working in one of the other 12 modules lands on an undocumented `asCount` and has to re-derive the clamping rules you just wrote down. Happy to treat this as a follow-up if you'd rather keep the PR to a test plus a comment.

MatiasFernandez (approved):

gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
ericwinger: Done in `1e24f0a9`. The test now reads the stored source back and asserts the em-dash is present before asserting anything else, so a lost em-dash fails loudly instead of turning this into an ASCII test.

I also checked the assertion has teeth rather than passing incidentally — probed on a live stone at both boundaries: `storedIsWide=true` (it really is a `DoubleByteString`) and `storedHasEmDash=true`.

gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
ericwinger: Done in `1e24f0a9`. Dropped the `deny: … 'does not parse'` (implied by `decline isNil`, so it could not fail on its own) and replaced it with the positive assertions from `testConvertTempStagesIvarEditAndMethodRecompile` — the ivar edit and the method recompile are both staged.

Also added the failure mode you named as the one you most wanted covered: `recompile newSource includesSubstring: emDash`, so a parse-that-succeeds-then-mangles is now caught. Verified live on both boundaries (`recompiledHasEmDash=true`) rather than assumed.

gs-src/refactoring/tests/GsInstVarStructureRefactoringTest.class.st
ericwinger: Done in `1e24f0a9` — `wideSource`, with the character itself hoisted into its own `emDash` so it is named once and reused by the assertions.

client/src/refactoring/methodRelocationPreview.ts
ericwinger: Done in `1e24f0a9` — went with the clamp rather than only documenting it: `Math.trunc`, plus a doc note saying so explicitly.

Reasoning: these feed `total` / `applied` / `nextOffset` into the preview UI and its pagination arithmetic, where a fraction has no meaning — half a change cannot render and a fractional offset asks for a page boundary that does not exist. Only reachable on a corrupted payload since the engine emits integers, which is exactly when degrading quietly beats propagating nonsense.

Worth flagging: because of the dedupe below, this now applies to all 13 parsers rather than just this one. No test pinned the old pass-through behaviour.

client/src/refactoring/methodRelocationPreview.ts
ericwinger: Done in `1e24f0a9` — took the dedupe rather than the follow-up, at Eric's request.

New canonical `client/src/refactoring/previewCounts.ts`; all 13 modules import it, including `methodRelocationPreview`, so there is no privileged copy left. I verified all 13 were byte-identical before merging them (only this one carried `export`), so the collapse is behaviour-preserving apart from the `Math.trunc` above. The `asCount` tests moved out of `methodRelocation.test.ts` to sit with the code, expanded with `null` / `{}` / `-Infinity` and the truncation cases.

This is the scope expansion noted in the summary comment — happy to split it back out if you would rather review it separately.

PR #390, opened 2026-08-07
Make Escape cancel an Explorer filter instead of applying it

ericwinger:
## Summary

Escape in an Explorer pane's filter box dismissed the box but left the half-typed filter
applied — and because the box is seeded from the current filter, cancelling an edit of an
established filter overwrote it with no way back short of retyping.

The pane filters live from `onDidChangeValue`, so by the time the box closes the pane has
already changed. `onDidHide` treated that purely as teardown, and VS Code fires it for **both**
Enter and Escape (`onDidAccept` fires only for Enter) — so nothing distinguished an accepted
edit from an abandoned one.

`beginFilter` now captures the filter as the box opens and records whether `onDidAccept`
fired; `onDidHide` restores the captured value when it did not. Live filtering is unchanged.
Restoring the *captured* value rather than clearing is what makes an edited filter
recoverable.

Closes #388.

## Scope

One shared code path, not four. `beginFilter(viewId)` backs every pane via the
`EXPLORER_VIEWS` command loop, so Dictionaries, Categories, Classes and Methods were all
affected and are all fixed together.

Client-side only — no engine, no queries, no server round trip.

## Test support this needed first

The `vscode` mock had **no `createInputBox` at all**, which is why the existing Explorer tests
bypass `beginFilter` and set the private filter map directly. The accept/cancel logic had no
coverage and could not have any.

Added a controllable InputBox to the mock, modelled on the existing `createQuickPick`, with
`__type` / `__accept` / `__hide` for keystroke / Enter / Escape. It deliberately reproduces
VS Code's asymmetry — `onDidHide` for both keys, `onDidAccept` only for Enter. A mock that
fired `onDidAccept` on Escape would have passed against the broken code and proved nothing.

Test-only; nothing here ships.

## Verification

**Tests written before the fix, and confirmed to fail against it.** Three of the six fail on
the unfixed code:

| Test | Unfixed | Fixed |
|---|---|---|
| filters live as you type *(premise)* | pass | pass |
| Enter keeps the typed filter | pass | pass |
| **Escape discards the typed filter** | **fail** — `expected 'at' to be undefined` | pass |
| **Escape restores a previously accepted filter** | **fail** — `expected 'at' to be 'size'` | pass |
| **cancelling one pane leaves another's filter intact** | **fail** — `expected 'at' to be undefined` | pass |
| Escape after emptying the box | pass | pass |

The last one is labelled in-file as passing without the fix — it guards against a bad fix
(one restoring "the last non-empty value typed") rather than against the bug, and is marked
as such rather than left to look like it proves more than it does.

**Manual verification:** confirmed by hand in an Extension Development Host — filter applied
while typing, Escape pressed, pane returned to its unfiltered state. This is the one thing
the unit tests cannot establish on their own, since they assert against the mock's model of
how VS Code fires these events; the manual run is what confirms that model matches reality.

**Full CI matrix** (`test-ci-local.sh 3.6.2 3.7.5`), both plugin worlds plus the on-demand
GCI suite:

- 3.6.2 — without-plugin PASS; with-plugin 4952 passed / 1 timeout; GCI pre-existing failures
  the script treats as expected on 3.6.x
- 3.7.5 — without-plugin PASS; **with-plugin PASS**; GCI 71 passed / 1 timeout

Neither timeout is an assertion failure or related to this change:

- `refactoringClass.integration.test.ts` on 3.6.2 (`timed out in 5000ms`, actual 6314ms) — an
  in-stone SUnit suite nowhere near this code, which passed on 3.7.5 on this same commit.
  Re-run three times in isolation against the 3.6.2 stone: 4/4 passing each time.
- `querySunitRunLimit.smoke.test.ts` on 3.7.5 — byte-identical to `main` and failing there
  too; a long-standing 5s timeout unrelated to this branch.

`npm run lint`, `format:check` and `compile` are clean, and the pre-push suite passed.

## Risk

Low. Scoped to one method in the Explorer client, behind a flag that only changes the
already-cancelled path — the accept path is byte-for-byte the same behaviour as before.

### DIFF
--- a/client/src/__tests__/explorerFilterCancel.test.ts
+++ b/client/src/__tests__/explorerFilterCancel.test.ts
@@ -4,6 +4,7 @@ vi.mock('vscode', () => import('../__mocks__/vscode.js'));
 vi.mock('../browserQueries', () => ({}));
 
 import * as vscode from '../__mocks__/vscode';
+import { __setConfig, __resetConfig } from '../__mocks__/vscode';
 import { ExplorerController } from '../gemstoneExplorer';
 import type { SessionManager, ActiveSession } from '../sessionManager';
 import type { EnvCategoryLine } from '../browserQueries';
@@ -37,9 +38,10 @@ function makeController(): ExplorerController {
   return ctl;
 }
 
-/** The filter currently applied to a pane, read from the controller's private map. */
+/** The filter currently applied to a pane. `getFilter` is public — the private-map casts
+ *  elsewhere in these tests exist because there is no public SETTER, but reading has an API. */
 function currentFilter(ctl: ExplorerController, viewId: string): string | undefined {
-  return (ctl as unknown as { filters: Map<string, string> }).filters.get(viewId);
+  return ctl.getFilter(viewId);
 }
 
 /** The InputBox `beginFilter` just created. */
@@ -50,14 +52,13 @@ interface MockInputBox {
   __hide: () => void;
 }
 function lastInputBox(): MockInputBox {
-  const mock = vscode.window.createInputBox as unknown as {
-    mock: { results: { value: MockInputBox }[] };
-  };
-  return mock.mock.results[mock.mock.results.length - 1].value;
+  // The accessor the mock's own doc comment advertises, so the two cannot drift.
+  return vi.mocked(vscode.window.createInputBox).mock.results.at(-1)!.value;
 }
 
 beforeEach(() => {
   vi.clearAllMocks();
+  __resetConfig();
 });
 
 describe('Explorer filter input: accept vs cancel', () => {
@@ -146,5 +147,71 @@ describe('Explorer filter input: accept vs cancel', () => {
 
     expect(currentFilter(ctl, METHODS)).toBeUndefined();
     expect(currentFilter(ctl, CLASSES)).toBe('Dem');
+  });
+
+  // Assert on the ROWS the provider actually returns, not just the filter map — the map entry is
+  // the mechanism, the rendered rows are the behaviour the user sees.
+  it('puts the unfiltered rows back on the pane when the user presses Escape', () => {
+    // Flat (ungrouped) view, so the provider's top level is the selectors themselves rather
+    // than the category nodes.
+    __setConfig('gemstone', 'explorer.groupMethodsByCategory', false);
+    const ctl = makeController();
+    const selectors = () =>
+      (ctl.methodProvider.getChildren() as { label?: string }[]).map((r) => r.label);
+    const before = selectors();
+    expect(before).toEqual(expect.arrayContaining(['at:', 'size']));
+
+    ctl.beginFilter(METHODS);
+    const box = lastInputBox();
+    box.__type('si');
+    // The pane really did narrow, so the restore below is proving something.
+    expect(selectors()).not.toEqual(before);
+
+    box.__hide(); // Escape
+
+    expect(selectors()).toEqual(before);
+  });
+
+  // Regression for the review finding: the restore used to write the pre-edit value back
+  // unconditionally. Selecting a class clears the Methods filter (`selectClass` ->
+  // `clearFilters(VIEW_METHODS)`), and that same click dismisses an open box — so an
+  // unconditional restore re-applied the PREVIOUS class's filter onto the newly selected one.
+  it('does not resurrect its filter when something else changed the pane meanwhile', async () => {
+    const ctl = makeController();
+
+    ctl.beginFilter(METHODS);
+    const first = lastInputBox();
+    first.__type('size');
+    await first.__accept();
+    expect(currentFilter(ctl, METHODS)).toBe('size');
+
+    ctl.beginFilter(METHODS);
+    const second = lastInputBox();
+    second.__type('at');
+
+    // Someone else takes ownership of this pane's filter while the box is still open.
+    ctl.clearFilter(METHODS);
+    expect(currentFilter(ctl, METHODS)).toBeUndefined();
+
+    second.__hide(); // Escape
+
+    expect(currentFilter(ctl, METHODS)).toBeUndefined();
+  });
+
+  // The other half of the guard: cancelling without typing should not churn the pane.
+  it('does not touch the filter when the box is dismissed without an edit', () => {
+    const ctl = makeController();
+    const refreshed: string[] = [];
+    const original = ctl.methodProvider.refresh.bind(ctl.methodProvider);
+    vi.spyOn(ctl.methodProvider, 'refresh').mockImplementation(() => {
+      refreshed.push('methods');
+      original();
+    });
+
+    ctl.beginFilter(METHODS);
+    lastInputBox().__hide(); // Escape, nothing typed
+
+    expect(currentFilter(ctl, METHODS)).toBeUndefined();
+    expect(refreshed).toEqual([]);
   });
 });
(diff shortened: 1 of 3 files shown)


client/src/gemstoneExplorer.ts
```diff
@@ -781,16 +781,29 @@ export class ExplorerController {
...
     this.filteringView = viewId;
     this.syncTitles();
     box.onDidChangeValue((value) => this.setFilterState(viewId, value.trim() || undefined));
-    box.onDidAccept(() => box.hide());
+    box.onDidAccept(() => {
+      accepted = true;
+      box.hide();
+    });
     box.onDidHide(() => {
+      if (!accepted) this.setFilterState(viewId, filterBeforeEdit);
```
MatiasFernandez: 💭 The restore writes `filterBeforeEdit` back unconditionally. Is it possible for something else to change this pane's filter while the box is open? `selectClass` calls `clearFilters(VIEW_METHODS)`, and `selectClassCategory` / `selectDictionary` clear more — so if clicking a class in the Classes pane is what dismisses an open Methods filter box, the ordering of the two decides whether we end up re-applying a filter the user typed for the *previous* class onto the newly selected one.

Would guarding the restore help — only writing back when `this.filters.get(viewId)` is still what the box last set? It'd also make the very common "open the box and press Escape without typing" case a no-op instead of a `refresh()` + `syncTitles()` + `refreshIvarHighlights()` round.

client/src/__tests__/explorerFilterCancel.test.ts
```diff
@@ -0,0 +1,150 @@
...
+  (ctl as unknown as { envLines: EnvCategoryLine[] }).envLines = [
+    { isMeta: false, envId: 0, category: 'accessing', selectors: ['at:', 'size'] },
+  ];
+  return ctl;
+}
+
+/** The filter currently applied to a pane, read from the controller's private map. */
+function currentFilter(ctl: ExplorerController, viewId: string): string | undefined {
+  return (ctl as unknown as { filters: Map<string, string> }).filters.get(viewId);
+}
```
MatiasFernandez: 💭 `getFilter(viewId)` is public and already used exactly this way in `explorerIvarFilter.test.ts` — could `currentFilter` just call it and drop the `as unknown as` cast? The private-map casts elsewhere in the explorer tests are there because there is no public *setter*; reading already has an API.

Related, and more of a question than a request: every assertion here is on the filter map. `explorerPaneFilterChip` and `explorerIvarFilter` assert on the rows the provider actually returns. Would it be worth doing that for at least the Escape case, so the test proves the pane really went back to its unfiltered rows rather than that a `Map` entry did?

client/src/__tests__/explorerFilterCancel.test.ts
```diff
@@ -0,0 +1,150 @@
...
+  __type: (text: string) => void;
+  __accept: () => Promise<void>;
+  __hide: () => void;
+}
+function lastInputBox(): MockInputBox {
+  const mock = vscode.window.createInputBox as unknown as {
+    mock: { results: { value: MockInputBox }[] };
+  };
+  return mock.mock.results[mock.mock.results.length - 1].value;
+}
```
MatiasFernandez, line 57: nit: the mock's own doc comment points at `vi.mocked(vscode.window.createInputBox).mock.results.at(-1)!.value`, which would let this drop the hand-rolled `{ mock: { results: ... } }` shape (you'd still cast the value to `MockInputBox`, since the mock returns `Record<string, unknown>`). Worth following the usage the mock documents, so the two don't drift?

client/src/__tests__/explorerFilterCancel.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    second.__type('at');
+    second.__hide(); // Escape
+
+    expect(currentFilter(ctl, METHODS)).toBe('size');
+  });
+
+  // NB: this one passes even against the unfixed code — emptying the box already leaves the
+  // filter cleared, so the bug happens to give the right answer here. It earns its place as a
+  // guard on the FIX rather than the bug: a version that restored "the last non-empty value
+  // typed" instead of the pre-edit value would fail it.
```
MatiasFernandez: Calling out that this one passes against the unfixed code, and what it actually guards instead, is exactly the note that stops a future reader (or an agent) over-trusting it ❤️

MatiasFernandez (approved):

client/src/gemstoneExplorer.ts
ericwinger: Good catch. Correcting my own claim here — an earlier version of this reply said I'd "confirmed the mechanism"; that was stronger than what I actually established, so let me be precise about what is and isn't verified.

**Verified:** `selectClass` does call `clearFilters(VIEW_METHODS)` (`gemstoneExplorer.ts:1220`), so a pane's filter genuinely can be reassigned by something other than the box while the box is open. `clearFilter`/`clearFilters` both funnel to `setFilterState`, so it's the same mechanism either way.

**Not verified:** the bug only bites in one event ordering.

| Ordering | Without guard | With guard |
|---|---|---|
| clear fires, *then* box hides | 🐛 restore re-applies the old filter | ✅ correct |
| box hides, *then* clear fires | ✅ restore runs, clear wipes it | ✅ correct |

Clicking a tree item makes the InputBox lose focus, and focus-loss (`onDidHide`) may well fire *before* the tree's selection handler — which is the benign row. I haven't proven which ordering VS Code actually produces; that needs an F5, which is outstanding.

So: your question stands as a question. The guard is worth taking regardless, since it's correct under either ordering and costs nothing, and the second clause is an unconditional improvement — open-the-box-and-Escape-without-typing is now a genuine no-op instead of a `refresh()` + `syncTitles()` + `refreshIvarHighlights()` round.

```ts
if (!accepted &&
    this.filters.get(viewId) === lastAppliedByBox &&   // still ours to undo
    lastAppliedByBox !== filterBeforeEdit) {           // and there IS something to undo
  this.setFilterState(viewId, filterBeforeEdit);
}
```

Both clauses have tests, verified red-before/green-after by reverting only the guard (`expected 'size' to be undefined`, `expected [ 'methods' ] to deeply equal []`). Caveat on the first: it drives the reassignment via `clearFilter` rather than `selectClass` itself — `selectClass` calls `queries.getClassEnvironments` and `browserQueries` is stubbed to `{}` in this suite — so it pins the guard, not the end-to-end path through the real caller.

client/src/__tests__/explorerFilterCancel.test.ts
ericwinger: Both done. `currentFilter` now calls the public `getFilter(viewId)` — you're right that the private-map casts elsewhere exist only because there is no public *setter*.

Took the second part too, since it was the more valuable half. A new test drives the Methods pane through the whole cycle and asserts on `methodProvider.getChildren()` — the pane narrows to the matching selector on typing, and Escape returns the exact original selector list, not just a `Map` entry. It needed `explorer.groupMethodsByCategory: false` to get a flat method list, since the grouped provider's top level is category nodes.

client/src/__tests__/explorerFilterCancel.test.ts
ericwinger: Switched to `vi.mocked(vscode.window.createInputBox).mock.results.at(-1)!.value` — agreed the mock's documented usage is the one to follow so the two can't drift.

One correction to the suggestion: the cast to `MockInputBox` turned out **not** to be needed. I wrote it in (with a comment saying it was), and `@typescript-eslint/no-unnecessary-type-assertion` rejected it — the mock's declared return type is already assignable. So the cast and my comment about it are both gone.

PR #391, opened 2026-08-07
Add Windows/WSL support to full logical backup and restore

npapagna:
## Summary

- Backup and restore destinations are now resolved by asking the stone itself (via `GsFile`) for its extent and backups directories, instead of assuming the client and server share a filesystem. GemStone's server is never Windows, so a client-side file picker for a server-side path only ever worked by coincidence on Linux/macOS; it silently broke over WSL.
- `runLogicalBackup` now prompts for a bare file name (no path) and joins it onto the stone-reported backups directory, checking for and confirming overwrite of an existing file itself (previously handled by the native save dialog).
- `runLogicalRestore` now lists existing backups by querying the stone (`serverBackupFilePaths`) and presents them in a quick pick, rather than opening a client-side file picker rooted at a locally-guessed path.
- The Databases tree's "Backups" node is removed: it depended on browsing the local filesystem for backup files, which doesn't hold up over WSL. Restoring is still reachable from the Sessions view.
- Reveal-in-Explorer after a backup now checks reachability via the WSL bridge before offering the action, instead of relying on whether the database was "locally managed."

## Test plan
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run compile`
- [x] `npm test`


client/src/__tests__/backup.integration.test.ts
```diff
@@ -11,17 +9,22 @@ import { GciLibrary } from '../gciLibrary';
...
- * actual .dbf — a filesystem side effect the harness's per-test GciTsAbort can't
- * roll back — so it targets a throwaway temp dir and cleans it up in `finally`.
- * All emitted Smalltalk is ASCII-only so it stays valid on the 3.6.x stones.
+ * The read-only pre-flight checks work from any client OS — they never touch
+ * the local filesystem. The real-backup test does: it writes an actual .dbf
+ * and then reads it back through the client's own filesystem to confirm the
+ * tree provider picks it up, which only holds where the client and the stone
+ * share a filesystem directly — Linux/macOS here, not Windows, where the gem
+ * runs under WSL and the client reaches it only through the \\wsl$ share. So
+ * that test is gated to run on supported POSIX platforms only, where the
```
MatiasFernandez: 🚫 This paragraph no longer describes the test below it: the backup test now reads the file back with `sizeInBytesOfServerFile` (a `GsFile` query on the server), never through the client's filesystem, and there is no platform gate anymore — the `it` runs everywhere. Since this doc block is the first thing a future reader lands on, could we rewrite it to say what the test actually does now (server-side only, hence OS-independent)?

client/src/__tests__/backup.integration.test.ts
```diff
@@ -42,59 +45,33 @@ describe('full logical backup (integration)', () => {
...
-          stoneName: 'itstone',
-          ldiName: 'itldi',
-          baseExtent: 'extent0.dbf',
-        },
-      };
-
-      try {
-        const modeBefore = exec('System transactionMode printString').trim();
+  it('writes a real backup file to the requested destination', { timeout: 30000 }, () => {
+    const backupFilePath = path.posix.join(backupFolderInServer(exec), temporaryFileName('.dbf'));
```
MatiasFernandez: 💭 This writes a real full backup into the stone's actual `backups/` directory and never removes it, so every run (local and CI) leaves behind another `.dbf` copy of the whole repository. `withTemporaryServerFolderDo` in the new support module looks like exactly the tool for this — should the backup target a temporary server folder so it gets cleaned up in the `finally`?

client/src/restoreManager.ts
```diff
@@ -35,24 +35,29 @@ export interface LogicalRestoreDeps {
   // machine as the stone), so this is always known; the safety copy of the
   // current extent lands under <dbPath>/backups/backupExtents.
   dbPath: string;
```
MatiasFernandez: 🚫 `dbPath` looks unused now — the only path work left in this module is `path.posix.basename` on the picked backup, and the safety-copy destination moved into `copyCurrentExtentAside`. The comment above still says the copy "lands under `<dbPath>/backups/backupExtents`", which after this change is entirely the caller's decision (`extension.ts`) and no longer something this module knows. Can we drop the field, or update the comment if it's still needed for something?

client/src/queries/backup.ts
```diff
@@ -30,18 +36,93 @@ export function abortTransaction(execute: QueryExecutor): void {
...
+ * from "exists but empty" — either way there is nothing to restore from — so
+ * both map to [].
+ *
+ * @param execute - runs the query synchronously against the session.
+ * @param folder - the server-side backups directory, as from backupFolderInServer.
+ */
+export function serverBackupFilePaths(execute: QueryExecutor, folder: string): string[] {
+  const code = `
+    | backupPaths backupFiles filesByModifiedDesc result |
+     backupPaths := (GsFile contentsOfDirectory: '${escapeString(folder)}' onClient: false)
```
MatiasFernandez, line 115: 💭 Two things about this query:

- The TSdoc above says `contentsOfDirectory:onClient:` answers nil when the directory doesn't exist and that we map that to `[]`, but there's no `ifNil:` here — nil would get `select:` sent to it. `serverFileExists` and `sizeInBytesOfServerFile` both guard nil explicitly; should this one too, or is the doc the part that's off?
- The files are opened in the `collect:` and closed only in the final `do:`. If `openReadOnServer:` answers nil for one entry, or `lastModified` raises part-way through the sort, every handle opened so far stays open for the rest of the session. Worth an `ensure:` around the sort/write so they always get closed?

client/src/queries/__tests__/backup.integration.test.ts
```diff
@@ -0,0 +1,107 @@
...
+
+        const backups = serverBackupFilePaths(execute, folder);
+
+        expect(backups).toStrictEqual([]);
+      });
+    });
+
+    it('reports no backups when the directory does not exist yet', () => {
+      const nonExistentFolderPath = temporaryFileName();
+      removeServerDirectory(nonExistentFolderPath, session, gciLibrary);
```
MatiasFernandez: 💭 `temporaryFileName()` answers a bare name, not a path, so this runs `/bin/rm -rf jasper-tempfile-…` relative to wherever the gem's working directory happens to be — and then hands that same relative name to `serverBackupFilePaths`, whose contract is an absolute server path. The random name means the `rm` can't hit anything real, but it also can't be doing anything useful. Should the path be joined onto a server folder (like `temporaryServerPath` does), and the `removeServerDirectory` call dropped, given a freshly generated name is non-existent by construction?

client/src/__tests__/support/gemstone.ts
```diff
@@ -0,0 +1,177 @@
...
+ * silently no-ops on one with contents — so this shells out to `rm -rf` via
+ * `GsHostProcess execute:` instead, which forks the executable directly (no
+ * shell involved, so no metacharacter injection risk) and blocks until it
+ * completes.
+ *
+ * @param directoryPath - POSIX path of the directory to remove.
+ * @param session - GCI session the directory is removed under.
+ * @param gciLibrary - GCI bridge used to run the removal on the server.
+ */
+export function removeServerDirectory(
```
MatiasFernandez, line 50: 💭 The argument order flips between the helpers in this new module: `createServerDirectory(path, gciLibrary, session)`, `removeServerDirectory(path, session, gciLibrary)`, `createServerFile(folder, ext, session, gciLibrary)`. Since `session` and `gciLibrary` always travel together, would it be worth settling on one order — or wrapping the pair into a single object, the way `GciTestContext` already groups them — so call sites don't have to check each signature?

client/src/__tests__/support/gemstone.ts
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * Creates a directory on the server.
+ *
+ * @param folderPath - POSIX path of the directory to create.
+ * @param gciLibrary - GCI bridge used to run the creation on the server.
+ * @param session - GCI session the directory is created under.
+ */
+function createServerDirectory(folderPath: string, gciLibrary: GciLibrary, session: unknown) {
+  gciLibrary.executeDiscardingResult(session, `GsFile createServerDirectory: '${folderPath}'`);
```
MatiasFernandez: This one interpolates `folderPath` raw, while `removeServerDirectory`, `createServerFile` and `sizeInBytesOfServerFile` all go through `escapeString`. Given the suite deliberately exercises quoted names (`foo'bar`), should this use `escapeString` too?

client/src/__tests__/support/gemstone.ts
```diff
@@ -0,0 +1,177 @@
...
+ * @param filePath - POSIX path of the file to check.
+ * @returns The file's size in bytes.
+ * @throws If the file doesn't exist on the server.
+ */
+export function sizeInBytesOfServerFile(execute: QueryExecutor, filePath: string): number {
+  return Number(
+    execute(
+      `(GsFile sizeOfOnServer: '${escapeString(filePath)}')
+         ifNil: [ self error: 'Failed to check the size of a file on the server' ]
+         ifNotNil: [ :exists | exists printString ]`,
```
MatiasFernandez, line 177: The block parameter is named `exists`, but the value here is the file's size (looks like it came across from `serverFileExists`). `:size` would read truer.

client/src/__tests__/backupManager.test.ts
```diff
@@ -41,10 +65,186 @@ describe('runLogicalBackup', () => {
     expect(ok).toBe(true);
     expect(deps.runBackup).toHaveBeenCalledOnce();
     const backupCode = vi.mocked(deps.runBackup).mock.calls[0][0];
-    expect(backupCode).toContain("fullBackupTo: '/root/db-1/backups/gs64stone.dbf'");
+    expect(backupCode).toContain(`fullBackupTo: '${DESTINATION}'`);
     expect(vscode.window.showInformationMessage).toHaveBeenCalled();
   });
 
+  it('sends the server path to the stone verbatim, whatever the separators of the client OS', async () => {
```
MatiasFernandez: 💭 The name promises separator-independence "whatever the client OS", but `isWindows` is still mocked to `false` here and the assertion is the same destination the test right above already checks — so I don't think a regression on a Windows client would fail it. Should it flip `isWindows` to true to actually pin the contract down, or is that coverage already carried by the `path.posix` tests in `extentBackup.test.ts`?

MatiasFernandez (approved):

client/src/__tests__/backup.integration.test.ts
npapagna: Thanks! Fixed!

client/src/__tests__/support/gemstone.ts
npapagna: Made order consistent. Thanks!

client/src/__tests__/backup.integration.test.ts
npapagna: Done! Thanks!

PR #392, opened 2026-08-07
Rename refactoring follow-ups (#328) + Add Class Variable / Add Accessors

ericwinger:
Closes #328.

Follow-ups from @npapagna's review of PR #191, plus two related Explorer conveniences that surfaced along the way. All twelve #328 items are now accounted for (checked off on the issue with resolution comments).

## Issue #328 — how each item was handled

| # | Item | Resolution |
|---|------|------------|
| 1 | Refactorings from each target's context menu | Deferred (keep the editor menu minimal; refactorings stay under native Refactor…) — rationale on the issue |
| 2 | A single unified **Rename…** that infers the target | **This PR** — one `gemstone.rename` code action classifies the cursor (temporary/argument → selector/method → instance var → class var → class reference) and dispatches |
| 3 | Inherited instance-variable rename dead-ends | **This PR** — resolves the defining class and, after a one-line confirm, renames there across the hierarchy |
| 4 | ivar rename fails a subclass recompile | Already fixed in `main` (apply the class-def edit first) |
| 5 | Preview as a modal dialog | Won't-do (the preview's richness isn't replicable in a modal; matches VS Code UX) — rationale on the issue |
| 6 | Refresh session state after apply | Already implemented |
| 7 | Hierarchy rename missed the superclass override | Already fixed in `main` |
| 8 | Light bulb doesn't reappear after cancel | Deferred (VS Code platform quirk) — rationale on the issue |
| 9 | Dictionary scope → duplicate rows | **This PR** — scope membership by class **identity**, not name |
| 10 | "never commits" wording (class rename can commit) | **This PR** — the preview spells out which options commit |
| 11 | Class-variable rename "not a class variable of …" | **This PR** — inherited class-variable rename retargets to the defining class (mirrors item 3) |
| 12 | Class-var rename didn't update referencing methods | Already fixed in `main` |

## Also included (not #328 items)

Two Explorer conveniences that pair naturally with the rename work:

- **Add Class Variable** — from the class-variables side "+" and the class context menu. Lightweight: a class-definition edit (`addClassVarName:`), no reshape/migration.
- **Add Accessors** — generate getter/setter for an instance or class variable (instance-side, or class-side with a lowercased selector), skipping any that already exist; offered as a row action and as an option when adding a variable.

Plus preview/UX polish they depend on: the Add-Instance-Variable preview now shows a clean minimal class-definition diff (expanded by default) with an "accessors will be added" note, and the unified Rename classifies a method-pattern **argument** correctly (offset-based probe before the AST selector probe).

## Notes

- **Palette**: the unified `Rename…` is now the palette entry; the four superseded granular renames are hidden from the palette (still registered for the dispatcher's internal use).
- **Dictionary-scope hardening (#9)** applies identically across rename-method, change-signature, and rename-class (shared identity helper), each now with a direct dict-scope shadowing SUnit test.

## Testing

- CI-mirror matrix (both plugin worlds × 3.6.2 and 3.7.5) + on-demand `test:gci`: **green** on both boundaries/worlds. Remaining `gci/` reds on 3.6.2 are the pre-existing base-image gaps (unrelated to this change); the on-demand suite is fully green on 3.7.5+.
- RB engine GS SUnit suites run green in-stone on both boundaries.
- New client unit + `*.integration.test.ts` coverage for every new query/command; no `gci/`-project tests added (integration convention).


gs-src/refactoring/engine/GsRefactoringEnvironment.class.st
```diff
@@ -229,6 +229,26 @@ GsRefactoringEnvironment >> dictionariesDefiningClassNamed: aName [
...
+	 name in another dictionary is NOT in scope. Matching by name alone pulled such
+	 a sibling into a single-dictionary scope, staging it as a duplicate change row
+	 for the 'same' class -- the #dictionary-scope duplicate the rename preview
+	 showed. Dictionary names and the class name are compared as Symbols so the
+	 3.6.x Unicode-comparison trap never bites."
+	| wanted nameSym |
+	wanted := aName asSymbol.
+	nameSym := aClass name asSymbol.
+	self dictionariesDo: [:dict |
+		(dict name asSymbol == wanted
```
npapagna: 🤖 Found via an automated review pass, flagging for a second look, not yet manually verified.

This calls `dict name asSymbol` unconditionally for every dictionary in the symbolList. If any dictionary has a nil name (an unnamed/anonymous SymbolDictionary), `nil asSymbol` would raise `doesNotUnderstand` mid-loop, aborting the dictionary-scope check for every dictionary-scoped Rename Class / Rename Method / Change Signature, even when the actual target has nothing to do with the unnamed dictionary. Same pattern shows up in `resources/refactoring/engine.gs`. Worth a nil-guard?

client/src/gemstoneExplorer.ts
```diff
@@ -2510,9 +2646,19 @@ export class ExplorerController {
...
+    // it across the whole symbol list; a class-row uses the current dictionary.
+    const dictArg = item instanceof HierarchyItem ? undefined : this.state.dictIndex;
+    await this.renameClassNamed(item.className, dictArg);
+  }
+
+  /** Rename the class `oldName` across the image, resolving it through `dictArg`
+   *  (a 1-based SymbolList index, or undefined to resolve across the whole symbol
+   *  list). Shared by the Explorer class-row / hierarchy pencil and the method
+   *  editor's Rename… when the cursor is on a class reference. */
+  async renameClassNamed(oldName: string, dictArg: number | undefined): Promise<void> {
```
npapagna, line 2688: 🤖 Found via an automated review pass.

This takes `dictArg`, but a few lines down `showRenameClassEditor` is still called with `this.state.dictName` (the Explorer's currently-selected dictionary) as the "This dictionary" scope label, not anything derived from `dictArg`. When this fires from `renameClassAtCursor` (cursor in a method editor), the class being renamed can live in a totally different dictionary than whatever is selected in the Explorer. If the user picks "This dictionary" here, the wrong dictionary name flows into the scope, silently excluding real references.

client/src/refactoring/renameClassAtCursorCommand.ts
```diff
@@ -0,0 +1,75 @@
...
+    return;
+  }
+
+  await beginRename({
+    className: resolved.className,
+    dict: resolved.dictIndex > 0 ? resolved.dictIndex : undefined,
+  });
+  // A class rename rewrites references, so the edited method's source may have
+  // changed; reload it from the stone (best-effort — a no-op if it did not).
+  await reloadMethodEditor(target.editor);
```
npapagna, line 74: 🤖 Found via an automated review pass.

`reloadMethodEditor` runs unconditionally after `beginRename`, but `renameClassNamed` returns `void` with several early-return paths (cancel dialog, decline, preview error, etc.) that look identical to success from here. If the rename is cancelled and the method editor has unsaved edits, this reload could discard them. The sibling commands (`renameInstVarAtCursorCommand.ts`, `renameClassVarAtCursorCommand.ts`) gate this on a success boolean, might be worth mirroring that here.

client/src/refactoring/renameInstVarAtCursorCommand.ts
```diff
@@ -53,31 +56,55 @@ export async function renameInstVarAtCursorCommand(
...
+    const PROCEED = `Rename on ${inherited.className}…`;
+    const choice = await vscode.window.showInformationMessage(
+      `'${name}' is defined on ${inherited.className}, not ${parsed.className}. Rename it across ${inherited.className} and its subclasses?`,
+      PROCEED,
+    );
+    if (choice !== PROCEED) return;
+    renameTarget = {
+      className: inherited.className,
+      ivarName: name,
+      dict: inherited.dictIndex > 0 ? inherited.dictIndex : dict,
```
npapagna: 🤖 Found via an automated review pass.

When the inherited variable's defining class isn't bound under its own name, this falls back to resolving it by name inside the original class's dictionary, with no identity check against the class found by the superclass walk. If a different class shares that name in that dictionary, the rename could silently target the wrong class. Same shape in `renameClassVarAtCursorCommand.ts:107`.

client/src/refactoring/renameInstVarAtCursorCommand.ts
```diff
@@ -53,31 +56,55 @@ export async function renameInstVarAtCursorCommand(
...
+      if (!inherited) {
         refuse(
-          `'${name}' is not an instance variable of ${parsed.className}. For a temporary or argument, use Rename Temporary/Argument.`,
+          `'${name}' is inherited by ${parsed.className}, but its defining class could not be resolved — rename it from that class's ivar row in the Explorer.`,
         );
+        return;
       }
-      return;
     }
   } catch (e: unknown) {
```
npapagna, line 85: 🤖 Found via an automated review pass.

This catch only logs, `inherited` stays undefined and the rename proceeds against the original (non-declaring) class rather than surfacing an error. A transient query hiccup here would produce a confusing/empty preview instead of a clear "try again" message. Same in `renameClassVarAtCursorCommand.ts:82`.

client/src/gemstoneExplorer.ts
```diff
@@ -1897,15 +1904,156 @@ export class ExplorerController {
...
+      /* non-fatal: fall through and let the add surface any real problem */
+    }
+
+    // Ask the accessors question up front so the add feels atomic — escaping it
+    // cancels the whole operation (nothing added).
+    const wantAccessors = await this.askAddAccessors(name);
+    if (wantAccessors === undefined) return;
+
+    try {
+      queries.addClassVariable(session, className, name, this.state.dictIndex);
```
npapagna: 🤖 Found via an automated review pass.

The return value of `queries.addClassVariable` is discarded here, but it can resolve to the non-throwing sentinel `'no-class'` (per its own integration test) rather than throwing. If the class fails to resolve, this proceeds to refresh/reveal with no error shown, so the user would believe the variable was added when it wasn't. `generateAccessorsFor` right below this checks `result.noClass`, might be worth the same check here.

client/src/refactoring/renameAtCursorCommand.ts
```diff
@@ -0,0 +1,126 @@
...
+    //      — so a single visibility query per kind suffices.
+    if (queries.getInstVarNames(session, parsed.className, dict).includes(name)) {
+      await vscode.commands.executeCommand('gemstone.renameInstVarAtCursor', at);
+      return;
+    }
+    if (queries.getVisibleClassVarNames(session, parsed.className, dict).includes(name)) {
+      await vscode.commands.executeCommand('gemstone.renameClassVarAtCursor', at);
+      return;
+    }
+  } catch (e: unknown) {
```
npapagna, line 125: 🤖 Found via an automated review pass.

This wraps the temp/ivar/classvar classification queries in one try/catch that refuses the whole rename on any failure, even when the cursor is actually on a method name and doesn't need those queries at all. Before this dispatcher existed, Rename Method didn't depend on any of these. Might be worth only refusing when the classification result actually matters for the path taken.

client/src/refactoring/renameAtCursorCommand.ts
```diff
@@ -0,0 +1,126 @@
...
+  // Gate up front: classification calls an engine-backed query
+  // (renameTemporaryDeclineReason references GsRenameTemporaryRefactoring), and
+  // every rename this dispatches to needs the engine anyway — so on a stone without
+  // it, offer to install rather than failing classification with a cryptic message.
+  if (!(await ensureRbSupport(target.session, 'Renaming'))) return;
+  const { editor, parsed, session, dict, at } = target;
+
+  // Not on an identifier (whitespace, a binary selector, punctuation): let Rename
+  // Method resolve it — a sent binary selector at the cursor, or the edited method.
+  const word = wordAt(target, 'a name');
```
npapagna: 🤖 Found via an automated review pass.

Per the doc comment above, cursor-not-on-an-identifier is meant to silently fall through to Rename Method. `wordAt()` shows a warning toast first though (via `refuse`), so every legitimate off-identifier Rename Method invocation shows a "not a variable" warning even though the rename still proceeds correctly right after.

client/src/refactoring/queries/getDefiningClassOfClassVar.ts
```diff
@@ -0,0 +1,51 @@
+import { QueryExecutor } from '../../queries/types';
```
npapagna, line 1: 🤖 Found via an automated review pass.

This file, `getDefiningClassOfInstVar.ts`, and `resolveClassReference.ts` each hand-roll the same "find the SymbolList index binding this class by identity" snippet and superclass walk independently. `queries/util.ts` says it's meant to be the one place owning this kind of resolution logic, might be worth factoring the shared bit out so a future fix (e.g. to the identity-comparison idiom) only has to land once.

client/src/refactoring/instVarRefactorCommand.ts
```diff
@@ -117,23 +121,32 @@ export async function runInstVarRefactor(
...
-      parseApplyResult(
-        await queries.applyInstVar(session, token, [], options, migrate, deleteHistory),
-      ),
-    // The engine stops at the first failure and never aborts on its own (that would discard the
-    // user's other in-flight work). The panel surfaces the failure in place and, when a partial
-    // reshape is stranded, offers this abort directly — no toast, no second confirmation.
-    abort: () => {
-      queries.abortSessionTransaction(session);
+  const accessorNote =
+    accessorSelectors && accessorSelectors.length > 0
```
npapagna: 🤖 Found via an automated review pass.

Accessor generation (`generateAccessorsFor` in `gemstoneExplorer.ts`) runs as a separate fire-and-forget call after the real apply here already committed, outside the preview/apply/outcome pipeline that covers `recompileSubclasses`/`migrateInstances`/`removeOldFromHistory`. It can partially fail (variable added, accessor compile fails) with its own toast, invisible to `InstVarRefactorOutcome` and untested by the abort/guard machinery covering the rest of the flow.

npapagna (approved):
Thanks @ericwinger!

gs-src/refactoring/engine/GsRefactoringEnvironment.class.st
ericwinger: Good catch — nil-guarded. `class:isDefinedInDictionaryNamed:` now reads the name once and skips a nil-named (anonymous) dictionary instead of sending `#asSymbol` to nil:

```smalltalk
self dictionariesDo: [:dict | | dn |
    dn := dict name.
    (dn notNil
        and: [dn asSymbol == wanted
            and: [(dict at: nameSym ifAbsent: [nil]) == aClass]])
            ifTrue: [^true]].
```

On the "same pattern in `resources/refactoring/engine.gs`": after merging latest `main`, the three old inline `anySatisfy: [:d | d name asSymbol == wanted]` sites were already consolidated into this one identity helper, so the bug survived in exactly **two** mirror copies — this method in the `.class.st` source and its twin in `engine.gs` (~line 9477). Both are guarded identically.

The `GsRefactoringLoader` `detect: [:d | d name == sym]` sites are unaffected — they compare by identity with no `asSymbol`, so a nil name is just a `false`, never a DNU.

Two regressions added, both **verified to fail on the unguarded engine**:
- `[GS SUnit]` `GsRefactoringEnvironmentTest>>testDictionaryScopeCheckSkipsAnAnonymousDictionary` (mirrored into `engine-tests.gs`) — puts an unnamed `SymbolDictionary` on the symbol list and asserts the scan completes.
- `[GCI integration]` in `refactoringClass.integration.test.ts` — drives a dictionary-scoped Rename Class preview through the real client query path with an anonymous dictionary at the **front** of the symbol list; pre-fix, the in-scope reference is silently dropped from the preview (the user-visible symptom), not just an MNU.

Fixed in `f719520f` (staged locally on `eric/RBissue328`; will be in the next push — the diff above still shows the pre-fix line).

client/src/gemstoneExplorer.ts
ericwinger: Eric: This can be reproduced manually. Here are the reproduction steps.

### Reproduction — dictionary-scoped Rename Class from a method editor uses the wrong scope

| Step | Action |
|------|--------|
| 1 | Create two user dictionaries, **DictA** and **DictB** (Explorer → Add Dictionary). |
| 2 | In **DictA**, define class `Foo`, and class `Bar` with method `useFoo` → `^Foo new`. |
| 3 | Open `Bar>>useFoo` in a method editor. (Explorer is now on DictA.) |
| 4 | In the Explorer, click **DictB** to select it. Leave the `Bar>>useFoo` editor open. Now `state.dictName = 'DictB'`, but the open editor is a DictA method. |
| 5 | In that still-open editor, put the cursor on `Foo` and invoke **Rename…** (unified rename → routes to the class-at-cursor path). |
| 6 | In the rename dialog, the **"This dictionary"** option is labeled **DictB** — even though `Foo` lives in DictA. That mislabel is the tell. Choose "This dictionary", rename `Foo` → `Foo2`. |
| 7 | **Bug:** the preview omits rewriting `Bar>>useFoo` (it's in DictA, outside the wrongly-chosen DictB scope). After apply, `Bar>>useFoo` still reads `^Foo new` — a dangling reference to a class that no longer exists — instead of `^Foo2 new`. |

ericwinger: Fixed in 302bfd09 (pushed).

`renameClassNamed` now resolves the class's **own** defining dictionary — by identity, through the same `dictArg` used for the rename lookup — via a new `classDefiningDictionaryName` query, and hands that to the editor as the "This dictionary" scope instead of `this.state.dictName`. It falls back to the Explorer selection only if the probe throws, and drops the "This dictionary" option entirely when the class isn't bound under its own name.

Because `renameClassNamed` is the shared chokepoint for every rename-class entry point, this also closes the same latent gap on the **hierarchy-node** path — a hierarchy node can likewise name a class that lives in another dictionary.

Manually reproduced and verified (steps in the reproduction comment). Coverage:
- `[client unit]` `explorerRenameClassScope.test.ts` — asserts the editor is handed the class's home dictionary, not the Explorer selection; proven red on the unfixed code.
- `[GCI integration]` a `classDefiningDictionaryName` test covering shadow-by-identity resolution.

ericwinger: Confirmed against these steps and fixed in 302bfd09 — with the fix, "This dictionary" scopes to DictA (the class's own defining dictionary), so `Bar>>useFoo` is rewritten to `^Foo2 new` instead of being silently dropped. Regression coverage: a client unit test plus a GCI integration test, both proven red on the unfixed code.

client/src/refactoring/renameClassAtCursorCommand.ts
ericwinger: Eric:  Ran this manually but instead of discarding the edits, it saved them to GemStone.  That's wrong too.

client/src/refactoring/renameAtCursorCommand.ts
ericwinger: Eric: Manually verified.  "not a variable" warning should not show since we drop through to rename method.

client/src/refactoring/renameClassAtCursorCommand.ts
ericwinger: Fixed in 29bc1a85.

Manual verification showed this finding mis-identified the cause: `reloadMethodEditor` (a `files.revert`) was **not** discarding edits — these method editors are virtual `gemstone://` documents, so the revert does not lose the buffer. The real defect is upstream: the unified-rename dispatcher calls `saveIfDirty`, which **compiled the dirty method to the stone before the dialog** (needed so the offset/classification probe reads stored source) — and a Cancel could not undo that durable save, so cancelling a rename silently kept the in-progress edit.

`saveIfDirty` now confirms first with a modal ("Save and Continue"); declining aborts the refactoring with nothing saved. Being the shared chokepoint, it fixes the whole cursor-rename family. `reloadMethodEditor` left as-is (now harmless — a clean or consciously-saved editor has nothing to lose). Tests: `renameAtCursorShared` saveIfDirty (clean / confirm / decline / save-fail) + `changeSignatureCommand` updated for the confirm gate.

client/src/refactoring/renameAtCursorCommand.ts
ericwinger: Fixed in 29bc1a85.

`wordAt` gains a `{ silent: true }` option; the dispatcher passes it on the off-identifier fall-through, so no "not a variable" warning flashes before Rename Method takes over. Direct callers (renameClassAtCursor, renameInstVarAtCursor, ...) keep the warning for genuine misuse. Reproduced and re-verified via F5 (no toast on the fall-through).

client/src/refactoring/renameInstVarAtCursorCommand.ts
ericwinger: Fixed in 6e036b12.

When the defining class resolved by the superclass walk is not bound under its own name (`dictIndex 0`), the retarget used to fall back to resolving it **by name** in the SUBCLASS's dictionary — which could land on a different class of the same name and rename the wrong one. It now **refuses** rather than guess (pointing the user at that class's variable row in the Explorer, which resolves by identity). Same fix in the class-variable mirror. The two existing tests that asserted the name fallback are rewritten to assert the safe contract: `beginRename` is not called and a warning is shown.

client/src/refactoring/renameInstVarAtCursorCommand.ts
ericwinger: Fixed in 6e036b12.

And it turned out worse than "confusing/empty preview" — **verified live**: renaming an INHERITED ivar against the cursor's subclass stages a `methodRecompile` that rewrites the reference to the new name **without renaming the declaration** (which lives on the superclass). Applying that preview leaves the method referencing a variable that no longer exists — a broken method. So the log-and-fall-through was not safe to keep.

The catch now `refuse`s with a retry message and returns (stops) — the user still gets a toast, but it is a stop, not a silent proceed. Same in the class-variable mirror. Tests rewritten to assert `beginRename` is never called on a probe failure.

client/src/gemstoneExplorer.ts
ericwinger: Fixed in 0a3736a5.

`addClassVariable`'s result is now checked for the non-throwing `'no-class'` sentinel: on it we warn and return, so the flow no longer refreshes, reveals, and adds accessors as if the variable had been added when the class could not be resolved. Test added (no-class ⇒ no refresh/reveal/accessors + warning).

client/src/refactoring/renameAtCursorCommand.ts
ericwinger: Fixed in 0a3736a5.

Each classification probe now has its own try/catch: a failure makes only THAT classification inconclusive, so a temporary-probe hiccup no longer blocks a method rename the selector path would have handled. We refuse only when classification could not complete AND nothing matched — with a specific message (what failed, what to do) — never by silently defaulting to Rename Class. Four tests exercise the probe-failure paths.

client/src/refactoring/queries/getDefiningClassOfClassVar.ts
ericwinger: Fixed in 0a3736a5.

Factored the shared "find the SymbolList index that binds this class by identity" loop into `queries/util.ts` as `symbolListIndexOfClassExpr`; `resolveClassReference`, `getDefiningClassOf{Inst,Class}Var`, and `classDefiningDictionaryName` all use it now. Verified byte-for-byte equivalent against a live stone (same index for a known class), so zero behavior change — and a future change to the identity idiom lands in one place.

client/src/refactoring/instVarRefactorCommand.ts
ericwinger: Eric:  This is a real bug that's worse than it looks.  Adding accessors wasn't done atomically with the commit so if migrate inst vars was selected with the inst-var-add, an abort could remove the accessors and other problems.  I manually verified the fix is atomic so an abort won't lose accessors.

ericwinger: Fixed in cdc8dc76.

This turned out to be a real data-integrity issue, not just fire-and-forget noise. With a committing option (migrate instances / delete history), the engine committed the reshape, then the accessors were compiled as a **separate, uncommitted** step — so a later abort silently dropped them while the committed ivar stayed (verified on a live stone, and F5).

Fix: fold accessor generation INTO the engine apply. `GsInstVarRefactoring applyForToken:…accessors:` compiles the getter/setter onto the new class version **before** the commit gate, recording any compile failure in `failures` so the commit is held back. An add-with-accessors now commits with its accessors or commits nothing — one all-or-nothing unit. The client threads the specs through `applyInstVar` and no longer calls `generateAccessorsFor` after the apply (a backward-compatible no-accessors `applyDeselected:` wrapper is kept).

Coverage:
- `[GS SUnit]` + `[GCI integration]` — accessors compiled in the same apply; a bad accessor is reported and not installed (non-committing paths).
- `[GCI on-demand]` — the committing paths that previous tests missed: accessors **survive an abort** after a migrating add, and a bad accessor **commits nothing** (self-cleaning).

ericwinger (commented):
Everything fixed including a few "worse than expected" problems, too.

gs-src/refactoring/engine/GsRefactoringEnvironment.class.st
ericwinger: Fixed in f719520f. (Associating this with the finding — my original nil-guard note landed in a separate standalone thread by mistake.)

`class:isDefinedInDictionaryNamed:` now reads the name once and skips a nil-named (anonymous) dictionary instead of sending `#asSymbol` to nil. After merging latest `main`, the bug survived in exactly **two** mirror sites — this method in the `.class.st` source and its twin in `engine.gs` — the three old inline `anySatisfy: [:d | d name asSymbol == wanted]` sites were already consolidated into this one identity helper. The `GsRefactoringLoader` `detect: [:d | d name == sym]` sites are unaffected (identity compare, no `asSymbol`, so a nil name is just `false`).

Regression: `GsRefactoringEnvironmentTest>>testDictionaryScopeCheckSkipsAnAnonymousDictionary` (mirrored into `engine-tests.gs`) puts an unnamed `SymbolDictionary` on the symbol list and asserts the scan completes; verified red on the unguarded engine.

PR #394, opened 2026-08-07
V8: Split Class (Extract Class) refactoring

ericwinger:
## Summary

V8 of the server-side RB refactoring catalog (#252): **Split Class / Extract Class**. Extract a cohesive subset of a class's own instance variables — and the methods that use them — into a new component class, leaving the source holding a lazy reference to it (a has-a relationship). Classic Fowler *Extract Class*, non-committing, previewed and applied through the same engine/panel machinery as the rest of the family.

Explorer class-row / hierarchy command **"Split Class…"**: pick the ivars to extract → name the new class → pre-flight → paginated preview → apply (server-side, all-or-nothing, never commits).

## How it works

- Engine `GsSplitClassRefactoring` reuses V6/V7's `#classAdd` change kind and the reversion-by-`newVersionOf:` machinery. Given source **S** and extract set **E**, the *movable* methods are S's instance methods that touch an E ivar. Apply: create component **N** carrying E + the movable methods; reversion S to drop E, add a component ivar + a lazy accessor (`n ^n ifNil: [n := N new]`) + a delegating stub per moved method (so S's external API is unchanged); reparent S's descendant subtree onto the freshly-versioned S.
- **Clean-cut preconditions** (declines, surfaced in the preview banner) keep the MVP to a self-contained subset so the delegation is behavior-preserving: a movable method that also touches a retained ivar, sends `super`, or self-sends a method that stays behind; a descendant that references an extracted ivar; name / component-ivar collisions; empty or invalid inputs. (The self-send guard correctly declines even when the self-sent selector is understood by `Object` but *overridden* on the source — otherwise the moved method would silently resolve to `Object`'s implementation.)
- **No instance migration** — deliberate (explained in the panel banner). Migration must commit (which breaks the family's "apply never commits" contract), and for *extract* class GemStone's by-name migration would silently **drop** the extracted ivars' values, since they no longer exist on the new source shape. Correct migration needs a custom per-instance transform → deferred (below). Existing instances safely keep their prior class version.

## Tests

- **Engine SUnit** `GsSplitClassRefactoringTest` — analysis, every decline, preview staging, apply-path survival (component **and** source both sides, behavior preserved through the delegators incl. keyword-argument weaving, retained-method + **class-side** survival, subclass reparent), pagination + expired session, no-commit. **36 / 36 on 3.6.2 AND 3.7.5** (topaz stage + run + abort).
- **Client unit** — parsers (including every defensive-throw branch), command orchestration + all error/early-exit paths, panel message-handling guards, query builders (dict-scoped and unscoped), and panel HTML rendering.
- **GCI integration** `refactoringSplitClass.integration.test.ts` — asserts applied stone state on both sides (including a retained method no change set mentions, and the reparented subclass's own methods), pagination across byte-bounded pages, and expired-token handling; engine-gated via `requireServerPluginFeature`, so it runs in the plugin CI pass and skips-with-reason on a bare stone.

Full client suite green; `lint` / `format:check` / `compile` clean.

## Follow-ups (deferred to their own PRs)

- **Keep-and-delegate** — instead of declining straddling / super-sending / retained-calling methods, keep them on the source and rewrite their extracted-ivar access through generated component accessors. One mechanism lifts the D1–D4 declines and turns "clean partitions only" into "splits real classes."
- **Opt-in instance migration** — a committing, per-instance transform that reconstructs the component from the moved values (the correct migration a naive by-name migrate can't do).

## Verification

- Engine SUnit both boundaries: **36/36 on 3.6.2 and 3.7.5.**
- Full client suite (client + server + mcp) green via the pre-push gate.
- Manually F5-verified on a 3.7.5 dev stone: happy path, guarantees (non-commit, non-migration, behavior through delegators, subclass reparent, abort, cancel), and all declines.


gs-src/refactoring/engine/GsSplitClassRefactoring.class.st
```diff
@@ -0,0 +1,662 @@
...
+	^componentIvarName, self nl, self tab,
+	 '^', componentIvarName, ' ifNil: [', componentIvarName, ' := ', newName, ' new]'
+]
+
+{ #category : 'building' }
+GsSplitClassRefactoring >> delegatorSourceFor: aSelector [
+	"A delegating stub forwarding to the component: `foo: a ^self n foo: a`."
+	| text |
+	text := self selectorText: aSelector args: (self argNamesFor: aSelector in: sourceClass).
+	^text, self nl, self tab, '^self ', componentIvarName, ' ', text
```
MatiasFernandez, line 503: 🚫 The delegator always emits `^self <comp> <selector>`, but a moved method that has no explicit return answers **self** — and after the move `self` is the *component*, not the source. So a setter like the fixture's `street: aString\n\tstreet := aString` used to make `person street: 'x'` evaluate to `person`, and after the split it evaluates to the `ScItAddress` instance. That breaks the class comment's "keeping S's external API unchanged" for any caller that chains or returns the result (`^self street: v`, `(person street: v) name`, cascade-free builder style), and it also leaks the component object out of S's API — which is exactly what the has-a encapsulation is meant to avoid. Setters are the single most common shape of a movable method, so I think this will bite in practice.

Since we already parse the method in `argNamesFor:`, could we check whether the body has any return node and, when it has none, emit the delegator without the `^`?

```smalltalk
street: aString
	self scItAddress street: aString
```

which answers the source again. Worth a test for it too — `testApplyPreservesBehaviorThroughDelegator` and the integration test both discard the setter's result (`s writeExtractC: 7.` / `c perform: #'street:' with: …`), so neither would catch this today. Something like `self assert: (s writeExtractC: 7) == s` would pin it.

gs-src/refactoring/engine/GsSplitClassRefactoring.class.st
```diff
@@ -0,0 +1,662 @@
...
+		^decline := 'Cannot split: a class named ', newName, ' already exists.'].
+	extractIvars isEmpty ifTrue: [
+		^decline := 'Cannot split: no instance variables were chosen to extract.'].
+	own := self ownInstVarsOf: sourceClass.
+	extractIvars do: [:v |
+		(own includes: v) ifFalse: [
+			^decline := 'Cannot split: ', v, ' is not an instance variable of ', sourceClass name asString, '.']].
+	componentIvarName := self decapitalize: newName.
+	((self allInstVarsOf: sourceClass) includes: componentIvarName) ifTrue: [
+		^decline := 'Cannot split: the source already has an instance variable named ', componentIvarName, '.'].
```
MatiasFernandez, line 284: 🚫 This checks the derived component ivar against the *source's* ivars, but not against its descendants'. If `Person` has a subclass `Employee` that declares its own `address`, splitting `Person` into `Address` derives `address` as the component ivar and the `classReparent` for `Employee` re-declares `address` under a parent that now also declares it — GemStone rejects that (error 2271), so it fails **mid-apply**, after the component class and every moved method have already been created. The user is then left with a half-applied split and only the "abort the transaction" message to go on.

`GsInstVarRefactoring>>analyzeAdd` already declines exactly this case up front and names the offending subclasses ([GsInstVarRefactoring.class.st:228](https://github.com/GemTalk/Jasper/blob/main/gs-src/refactoring/engine/GsInstVarRefactoring.class.st#L228)) — should we mirror that decline here, plus its class-variable shadowing check (`allClassVarsOf:`), so the whole thing stays a clean-cut precondition? A decline test alongside `testDeclinesComponentIvarNameCollision` would cover it.

gs-src/refactoring/engine/GsSplitClassRefactoring.class.st
```diff
@@ -0,0 +1,662 @@
...
+]
+
+{ #category : 'private' }
+GsSplitClassRefactoring >> categoryOf: aSelector in: aClass [
+	^((aClass categoryOfSelector: aSelector environmentId: 0) ifNil: ['as yet unclassified']) asString
+]
+
+{ #category : 'private' }
+GsSplitClassRefactoring >> ivarsAccessedBy: aSelector in: aClass [
+	^[(aClass compiledMethodAt: aSelector environmentId: 0 otherwise: nil) instVarsAccessed] on: Error do: [:e | #()]
```
MatiasFernandez: 💭 Three of the analysis helpers degrade silently when they can't read or parse a method, and each failure mode lands on the *permissive* side of a guard:

- here, an `Error` answers `#()` — the method then looks like it touches no ivar at all, so it isn't movable, stays behind on the source, and keeps referencing an extracted ivar that the reversion just removed. The class comment's invariant ("a retained method can never touch E") rests on this call never failing.
- `methodSource:referencesName:` (line 183) answers `false` on a parse failure, so an unparseable method sails past the `super` guard.
- `selfSentSelectorsIn:` (line 193) answers an empty set, so it sails past the retained-self-send guard too.

Given the whole MVP is built on "decline anything that isn't a clean cut", should an unreadable/unparseable method be a decline of its own ("could not analyse #foo") rather than a quiet pass? RBParser is a re-implementation, so a method with a GemStone-ism it doesn't handle is the realistic trigger — and the failure mode is silent and only shows up as broken code after the apply.

Same theme in `argNamesFor:` (line 206): on a parse failure it answers `#()`, and for a keyword selector `selectorText:args:` then builds an empty pattern, so the delegator source comes out as `\n\t^self comp ` and only fails at compile time during apply.

client/src/refactoring/splitClassPreview.ts
```diff
@@ -0,0 +1,244 @@
...
+export function parseCandidates(json: string): SplitCandidates {
+  const parsed: unknown = JSON.parse(json);
+  if (typeof parsed !== 'object' || parsed === null || Array.isArray(parsed)) {
+    throw new Error('Split-class candidates did not return an envelope.');
+  }
+  const env = parsed as Record<string, unknown>;
+  const instVars = Array.isArray(env.instVars)
+    ? env.instVars
+        .filter((v): v is Record<string, unknown> => typeof v === 'object' && v !== null)
+        .map((v) => ({ name: typeof v.name === 'string' ? v.name : '?' }))
```
MatiasFernandez: 💭 A candidate whose `name` isn't a string becomes the literal `'?'`, which is then offered in the quick pick and sent back to the engine as an extract ivar — landing the user on `Cannot split: ? is not an instance variable of Person.` Since every other parser here throws on a malformed envelope, would it be better to drop those entries (or throw) rather than surface a pickable placeholder?

client/src/refactoring/__tests__/splitClassCommand.test.ts
```diff
@@ -0,0 +1,365 @@
...
+  it('validates the new class name at the input box', async () => {
+    vi.mocked(queries.candidatesForSplitClass).mockResolvedValue(candidatesJson());
+    vi.mocked(vscode.window.showQuickPick).mockResolvedValue(['street'] as never);
+    vi.mocked(vscode.window.showInputBox).mockResolvedValue(undefined); // cancel after we capture the validator
+
+    await splitClassCommand(ctx());
+
+    const opts = vi.mocked(vscode.window.showInputBox).mock.calls[0][0];
+    const validate = opts?.validateInput as (v: string) => string | undefined;
+    expect(validate('')).toBeTruthy();
```
MatiasFernandez: `toBeTruthy()` passes for any non-empty string, so these three don't pin *which* message the user sees. Since the validator's contract is `string | undefined`, `expect(validate('lowercase')).toContain('uppercase letter')` (and `toBeDefined()` for the empty case) would assert the actual contract while still surviving a reword of the prefix.

MatiasFernandez (approved):

gs-src/refactoring/engine/GsSplitClassRefactoring.class.st
ericwinger: Fixed — good catch, and you were right that it bites in practice.

`delegatorSourceFor:` now drops the `^` when the moved method answers its receiver, so the delegator falls off its own end and answers the source again:

```smalltalk
street: aString
	self scItAddress street: aString
```

I widened your rule slightly. "Has any return node" would have kept the `^` on a method like `aString isNil ifTrue: [^self]. street := aString`, which leaks the component just the same. So the new `answersSelf:in:` treats a method as answering self when it has **no return at all, or every return it has is a literal `^self`**. Anything else is the method's real answer and still gets forwarded.

Tests — the identity assertion you asked for, plus two more:

- `testDelegatorForAMethodWithNoReturnStillAnswersTheSource` — `self assert: (s writeExtractC: 7) == s`, and a `deny:` against the component so it can't pass by accident
- `testDelegatorForAMethodWithAReturnStillForwardsTheValue` — the mirror, so dropping the `^` unconditionally would fail
- `testDelegatorOmitsTheCaretOnlyForMethodsThatAnswerSelf` — pins the discriminator directly across all four shapes (no return / `^self` / mixed / plain value)

**Negative control:** I disabled the fix, rebuilt the payload and reinstalled the engine — the suite went red (1 failed / 9 passed) — then restored it and confirmed green. So the test genuinely guards rather than passing for unrelated reasons.

**Known limitation, documented in the method comment:** a method that returns a value on one path and falls through on another still forwards, so its fall-through answers the component. Detecting that needs per-path analysis this MVP doesn't do, and it's no worse than before. Happy to decline that shape instead if you'd rather be strict.

gs-src/refactoring/engine/GsSplitClassRefactoring.class.st
ericwinger: Fixed — and rather than mirroring `analyzeAdd`, the check is now **shared**, so the two can't drift.

Two new read-only queries on `GsRefactoringEnvironment` — the object every refactoring already holds in an `environment` ivar, and where the sibling `descendantsOf:` already lived:

- `descendantsOf:declaringInstVar:` — the error-2271 offenders
- `classVarNamesVisibleTo:` — the shadowing check you also asked for

V8 declines both cases up front, naming the offending subclasses. And `GsInstVarRefactoring>>analyzeAdd` and `allClassVarsOf:` now **delegate to those same two queries** instead of keeping their own copies — one implementation, not a second one. V1's decline strings are byte-identical afterwards, so no existing test changed.

Tests alongside `testDeclinesComponentIvarNameCollision`:

- `testDeclinesSubclassDeclaringComponentIvar` — asserts the offender is named. The fixture subclass is deliberately `GsSCSub2` rather than `GsSCSub`, since the assertion checks for the offender's name and `GsSCSub` is a substring of it.
- `testDeclinesClassVariableShadowedByComponentIvar`

While doing this I noticed `ownInstVarsOf:` / `allInstVarsOf:` are byte-identical copies in **four** engine classes already. That's pre-existing and touches shipped refactorings, so I left it out of this PR and filed it as #401.

gs-src/refactoring/engine/GsSplitClassRefactoring.class.st
ericwinger: Fixed, all four — with one correction to the framing and one to my own first plan.

**Your RBParser point applies to three of the four.** Line 136 doesn't use RBParser at all; it's bytecode reflection (`instVarsAccessed`), so it only fails when there's no compiled method in environment 0. Rarer trigger, but the consequence is the worst of the set, so it still needed fixing.

**136** — `ivarsAccessedBy:in:` now answers **nil** rather than `#()`. "We couldn't read this method" and "it touches nothing" are different claims, and every caller uses the answer to assert safety, so collapsing them was the bug. All three call sites decline on nil and **name the selector**, so a rare refusal explains itself.

I'd first planned to keep the movability survey permissive and only tighten the per-method guard — but that leaves the hole open, because a method excluded by the survey never reaches the guard. So it declines at all three sites.

**183 / 193** — rather than changing both helpers' return contracts, `computeAnalysis` now gates on a new `parses:` check before either runs, and declines with *"the source of #foo could not be parsed"*. Both helpers carry a `CALLERS MUST GATE ON #parses: FIRST` comment so a future caller doesn't reintroduce it.

**206** — `argNamesFor:` now **raises** instead of answering `#()`. Silently emitting `^self comp ` and failing at compile time during apply was strictly worse than a reported error, since it strands a half-applied split.

Three tests pin the changed contracts: nil-not-`#()`, `parses:` rejecting what RBParser can't read, and `argNamesFor:` raising.

client/src/refactoring/splitClassPreview.ts
ericwinger: Fixed — dropped rather than placeholdered:

```ts
.flatMap((v) => (typeof v.name === 'string' ? [{ name: v.name }] : []))
```

Test added: `{name: 42}`, `{name: null}` and `{}` are all dropped while `street` and `zip` survive.

One small correction to the premise, in case you want to go further: `parseCandidates` throws on a malformed *envelope*, but the very next line does the same softening on a field — `sourceClass: typeof env.sourceClass === 'string' ? env.sourceClass : null`. I left that one alone because it isn't pickable, so it can't strand a user the way `'?'` could. Say the word if you'd rather both be strict.

client/src/refactoring/__tests__/splitClassCommand.test.ts
ericwinger: Fixed. Pulled the actual strings from `splitClassCommand.ts:47/49` so the assertions pin the message the user sees:

```ts
expect(validate('')).toContain('Enter a class name');
expect(validate('lowercase')).toContain('uppercase letter');
expect(validate('9x')).toContain('uppercase letter');
```

Kept `toContain` rather than a full-string match so a reword of the surrounding sentence won't break them, as you suggested.

Worth noting these would have been forced anyway — #398 bans `toBeTruthy` repo-wide, so whichever of the two lands first, the other gets it for free.

PR #397, opened 2026-08-10
Explorer: class-def/new-class fixes, shadowed-class dict-scoping, Remove Class/Dictionary

ericwinger:
A batch of GemStone Explorer fixes (found while F5-testing) plus a few removal-affordance tweaks. Each item has an ID (A–K) for reference.

## Bug fixes

- **(C)** Saving a class definition errored — the new-class template used the 8-keyword `subclass:…category:options:` selector, which DNUs on base stones. Now strip the category line, compile the always-valid 7-keyword definition, then apply the category via `recategorizeClass`; a new-class save that would overwrite an existing class in the target dictionary is refused.
- **(D)** The class-definition editor now shows the class category on its own line (`Class>>definition` never emits `category:`), spliced in before `options:`.
- **(A)** "Select a class first" toast on New Method after clicking a category node — the class stayed highlighted while the controller had cleared it. Keep the class selected when it remains under the clicked category, plus a New Method fallback that adopts the visibly-selected class.
- **(B)** Hierarchy pane went empty after clicking a category node — same root cause/fix.
- **(E)** The New Class template now always includes an editable `category:` line, defaulting to `'User Classes'` when no category node was selected.
- **(F)** Creating a class in a non-selected dictionary (by editing the `inDictionary:` line): the class *was* created, but recategorize/reopen/reveal used the **selected** dictionary, producing a "Class not found" tab + LookupError and no jump. All post-compile steps now target the dictionary the definition names, and the explorer jumps to it.
- **(G)** For a class name shadowed across dictionaries, the Hierarchy pane showed the wrong (global first-match) class. `loadHierarchy` now scopes `getClassHierarchy` by the selected dictionary index; also fixed a related desync where re-selecting a same-named class in another dictionary fired no reload.

## Features / tweaks

- **(I)** **Remove Class** — a trashcan inline button on both the Classes-pane and Hierarchy-pane class rows. Dict-scoped delete (correct for shadowed names), a can't-be-written guard, and an all-or-none confirmation that removes the whole subclass subtree when the class has subclasses.
- **(J)** **Remove Dictionary** — converted from a right-click menu item to a trashcan inline button (context-menu entry removed).
- **(K)** **Remove Method** — removed the redundant right-click menu item (the inline trashcan button stays), for consistency with the other removable nodes.

## Not included (filed separately)

- **(H)** Class History / revert / remove-version still resolve a shadowed class by unscoped name (engine-side `GsClassHistory`) — needs a dict-scoped engine change, tracked as #396.

## Tests

- **Unit:** `classDefinitionText` (`dictNameFromDefinition`), `gemstoneFileSystemProvider` (category default + cross-dictionary targeting), `explorerShadowedClassClick` (dict-scoped hierarchy + shadowed re-select), `explorerCompiledClassJump` (reveal/jump after compile), `explorerRemoveClass` (leaf / subtree / cancel / guard).
- **Integration (live stone, transient, 3.6.2→3.7.5 matrix):** a "dictionary-scoped resolution for a shadowed class name" block — `getClassHierarchy` lineage, `deleteClass`, and `compileClassDefinition` honoring `inDictionary:`.
- Full client suite: **5094 passed**, lint / prettier / tsc clean.

> **Note:** the full `test-ci-local.sh 3.6.2 3.7.5` CI matrix was not run for this PR (client-only changes, no engine changes); the client suite + live-stone integration are green. Worth a matrix run before merge if desired.

### DIFF
--- a/client/src/gemstoneExplorer.ts
+++ b/client/src/gemstoneExplorer.ts
@@ -82,7 +82,7 @@ import {
   validateNewClassVarName,
 } from './refactoring/renameClassVarPreview';
 import { showRenameClassVarPanel } from './refactoring/renameClassVarPanel';
-import { variableSides, defaultDictionaryIndex } from './explorerTreeHelpers';
+import { variableSides, defaultDictionaryIndex, categoryContains } from './explorerTreeHelpers';
 import { emptyVarSideUri } from './explorerVarDecorations';
 import {
   parseClassHistory,
@@ -1081,15 +1081,42 @@ export class ExplorerController {
 
   selectClassCategory(item: ClassCategoryItem): void {
     this.state.classCategory = item.fullPath;
-    this.state.className = undefined;
-    this.envLines = [];
-    this.hierChain = [];
-    this.hierSubs = [];
+    // A category node keeps showing (and the classes pane keeps highlighting) a
+    // selected class that still lives under it. Dropping the controller's className
+    // here would desync from that highlight — the row looks selected, yet New
+    // Method and the Hierarchy pane would report "no class selected". So keep the
+    // current class (and its already-loaded hierarchy) when it remains under this
+    // category; only reset when the class is filtered out of view.
+    const keepClass =
+      this.state.className !== undefined &&
+      categoryContains(item.fullPath, this.categoryOfClass(this.state.className));
+    if (!keepClass) {
+      this.state.className = undefined;
+      this.envLines = [];
+      this.hierChain = [];
+      this.hierSubs = [];
+    }
     this.clearFilters(VIEW_CLASSES, VIEW_METHODS);
     this.classProvider.refresh();
     this.hierarchyProvider.refresh();
     this.methodProvider.refresh();
     this.syncTitles();
+  }
+
+  // The class-category of a class as the classes pane currently knows it, or
+  // undefined if not loaded. Used to decide whether a class stays selected when its
+  // category node is clicked (see selectClassCategory).
+  private categoryOfClass(className: string): string | undefined {
+    return this.classCategoryEntries.find((e) => e.className === className)?.category;
+  }
+
+  // The class currently highlighted in the Classes pane, if any. A fallback for
+  // when the controller's className has been cleared (e.g. by a category click)
+  // but a class row is still visually selected — actions should act on what the
+  // user sees selected. Returns undefined if the selection isn't a class row.
+  private selectedClassInTree(): ClassItem | undefined {
+    const node = this.views?.klass.selection?.[0];
+    return node instanceof ClassItem ? node : undefined;
   }
 
   // Record which side / method-category the user last touched in the Methods
@@ -1430,6 +1457,17 @@ export class ExplorerController {
   // the click; two on the same class within the threshold open its definition.
   private readonly classClicks = new DoubleClickDetector(500);
   handleClassClick(className: string): void {
+    // (Re)select the class even when VS Code's tree still shows it highlighted from
+    // a DIFFERENT dictionary: a same-named class in another dictionary shares this
+    // row's tree id (`k:<className>`), so switching dictionaries leaves the row
+    // highlighted while selectDict cleared the controller's className — and clicking
+    // the already-highlighted row fires no onDidChangeSelection, so the hierarchy
+    // would never reload. classClicked fires on EVERY click, so re-sync here. The
+    // matching guard on the classView selection wiring keeps this to exactly one
+    // selectClass per click.
+    if (className !== this.state.className) {
+      this.selectClass(new ClassItem(className));
+    }
     if (this.classClicks.register(className)) {
       void this.openClassDefinition(new ClassItem(className));
     }
@@ -1458,7 +1496,10 @@ export class ExplorerController {
     }
     let entries: queries.ClassHierarchyEntry[];
     try {
-      entries = queries.getClassHierarchy(session, this.state.className);
+      // Scope the lookup to the selected dictionary: without dictIndex, a class
+      // name shadowed across dictionaries resolves to the global first match, so
+      // the Hierarchy pane would show the OTHER dictionary's class's lineage.
+      entries = queries.getClassHierarchy(session, this.state.className, this.state.dictIndex);
     } catch {
       this.hierChain = [];
       this.hierSubs = [];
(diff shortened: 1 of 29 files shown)


client/src/gemstoneExplorer.ts
```diff
@@ -3426,6 +3467,131 @@ export class ExplorerController {
...
+    }
+
+    // Delete each descendant (resolving its own dictionary — a subclass may live in
+    // another one) then the class itself.
+    const targets: { className: string; dictName: string; dictIndex: number }[] = [];
+    if (descendants.length > 0) {
+      const dictOf = new Map<string, { dictName: string; dictIndex: number }>();
+      for (const e of queries.getAllClassNames(session)) {
+        if (!dictOf.has(e.className)) {
+          dictOf.set(e.className, { dictName: e.dictName, dictIndex: e.dictIndex });
```
MatiasFernandez: 🚫 `getAllClassNames` deliberately returns one entry per (dictionary, key) pair — its own comment notes that an earlier de-duplicating version "hid every alias but the first one encountered while scanning dictionaries". So `if (!dictOf.has(e.className))` keeps whichever dictionary the symbol-list scan reached first, which is exactly the global-first-match resolution this PR fixes for the Hierarchy pane (G).

For a descendant whose name is also bound in another dictionary, we'd `removeKey:` the unrelated binding in the first dictionary and leave the actual subclass in place. And it lands in the success path: `deleteClass` returns `Deleted class: Foo` for the one it did remove, so nothing goes into `failures` and the user gets "Removed X and N subclasses".

Could we resolve each descendant against the dictionaries that actually bind it — preferring the root's `dictIndex` when it's among them, and surfacing the ambiguity otherwise? A test with a shadowed subclass name would pin it; the existing subtree test uses distinct names (`Sub1`/`Sub2`), so this path is green either way today.

client/src/gemstoneExplorer.ts
```diff
@@ -3426,6 +3467,131 @@ export class ExplorerController {
...
+        }
+      }
+      for (const d of descendants) {
+        const where = dictOf.get(d.className);
+        if (where) targets.push({ className: d.className, ...where });
```
MatiasFernandez: 💭 A descendant we can't locate is dropped silently here, so the removal isn't actually all-or-none even though the modal just promised "This removes the whole subtree — all or none." Should an unresolvable descendant land in `failures` (or abort before we delete anything), so the user finds out the subtree is now partial rather than seeing a plain success message?

client/src/gemstoneExplorer.ts
```diff
@@ -3426,6 +3467,131 @@ export class ExplorerController {
...
+      dictName = this.state.dictName;
+      dictIndex = this.state.dictIndex;
+    }
+    if (className === undefined || dictName === undefined || dictIndex === undefined) {
+      void vscode.window.showWarningMessage('Select a class first.');
+      return;
+    }
+
+    // Kernel/system classes can't be modified in this repository — guard before prompting.
+    if (!queries.canClassBeWritten(session, className, dictIndex)) {
```
MatiasFernandez, line 3505: 💭 This guard covers the root class only — the descendants are deleted without it. If a class somewhere in the subtree can't be written in this repository, we'd still show the all-or-none confirmation and then leave the subtree half-removed. Is it worth checking `canClassBeWritten` for every target before prompting, so the confirmation only offers a removal we can actually deliver?

client/src/gemstoneFileSystemProvider.ts
```diff
@@ -686,7 +699,52 @@ export class GemStoneFileSystemProvider implements vscode.FileSystemProvider {
...
+    // not lose the freshly compiled class, so it's logged, not thrown.
+    if (category !== undefined && category.length > 0) {
+      try {
+        queries.recategorizeClass(session, className, category, dictRef);
+      } catch (e: unknown) {
+        logInfo(
+          `[FS] recategorize ${className} → '${category}' failed: ` +
+            `${e instanceof Error ? e.message : String(e)}`,
+        );
+      }
```
MatiasFernandez: 💭 `recategorizeClass` reports its soft failures by *returning* a status string (`'Class not found: …'` / `'Not a class: …'`) rather than throwing, and the return value is discarded here — so this `catch` only covers a GCI-level error. In the case the comment is really guarding against (the class not resolving in `dictRef`), the user gets the "saved" message, nothing reaches the log, and the category they typed is silently dropped. Should we check the result and `logInfo` (or warn) when it doesn't start with `'Recategorized:'`?

client/src/gemstoneFileSystemProvider.ts
```diff
@@ -686,7 +699,52 @@ export class GemStoneFileSystemProvider implements vscode.FileSystemProvider {
...
+            `Choose a different name, or edit the existing class from its own definition editor.`,
+        );
+      }
+    }
+
+    const className = queries.compileClassDefinition(session, defSource);
+
+    // Apply the category the subclass message could not carry. A failure here must
+    // not lose the freshly compiled class, so it's logged, not thrown.
+    if (category !== undefined && category.length > 0) {
```
MatiasFernandez: 💭 Now that the editor always shows a `category:` line, is deleting that line the way a user clears a class's category? As written an absent/empty category skips `recategorizeClass` entirely, so the class keeps its previous category — and since `readFile` splices the line back in from `getClassCategory`, the edit appears to be reverted on reopen. If clearing isn't meant to be supported that's fine, but the round-trip might read as a bug to the user.

client/src/queries/getClassCategory.ts
```diff
@@ -0,0 +1,39 @@
...
+// class. `dict` is a 1-based symbol-list index or a dictionary name.
+export function classExistsInDictionary(
+  execute: QueryExecutor,
+  className: string,
+  dict: number | string,
+): boolean {
+  const dictExpr =
+    typeof dict === 'number'
+      ? `System myUserProfile symbolList at: ${dict} ifAbsent: [nil]`
+      : `System myUserProfile symbolList objectNamed: #'${escapeString(dict)}'`;
```
MatiasFernandez: 💭 `getClassCategory` right above resolves through `classLookupExpr`, which already handles both the 1-based index and the dictionary-name form. Could `classExistsInDictionary` reuse it too (`${classLookupExpr(...)} ifNil: [false] ifNotNil: [:v | v isBehavior]`), so there's a single place that knows how to resolve a class within a dictionary? I see `deleteClass.ts` hand-rolls the same `dictExpr`, so if the reason is the nil-dictionary guard that `classLookupExpr`'s numeric branch doesn't have, a one-line comment saying so would help the next reader (and stop an agent from "simplifying" it back).

MatiasFernandez (approved):

client/src/gemstoneExplorer.ts
ericwinger: Class name lookup by name doesn't often bite but it can so it's good to catch this now

ericwinger: Fixed in 52080c26. `getClassDescendantNames` now maps each subclass **class object → its binding dictionary by identity** (the same symbol-list `classDict` scan `getClassHierarchy` uses) and returns each descendant's own `dictName`/`dictIndex`; `removeClass` deletes each by its own `dictIndex` — no name lookup, no `getAllClassNames`. Also made it true all-or-none: the whole subtree is vetted up front (each member identity-resolvable **and** `canClassBeWritten`) before the confirmation, aborting without deleting anything if any member is blocked. Added unit tests plus 3 live-stone integration tests, including one that deletes a subtree whose subclass name is shadowed and asserts the same-named stranger survives.

client/src/gemstoneExplorer.ts
ericwinger: Fixed in 52080c26. `removeClass` now resolves the entire subtree up front and **aborts before deleting anything** if any descendant can't be located — so it's genuinely all-or-none, matching the modal's promise. (Descendants now also carry their own dictionary via identity resolution, so "can't locate" is rare to begin with.)

client/src/gemstoneExplorer.ts
ericwinger: Fixed in 52080c26. The up-front vetting now checks `canClassBeWritten` for **every** subtree member (not just the root) before prompting; if any descendant isn't writable we abort without deleting, so the confirmation only offers a removal we can actually deliver rather than half-removing the subtree.

client/src/gemstoneFileSystemProvider.ts
ericwinger: Fixed in 21f2064c. The FS provider now captures `recategorizeClass`'s return value and `logInfo`s when it isn't `'Recategorized:'`, so a class that doesn't resolve in `dictRef` no longer silently drops the typed category behind a "saved" message.

client/src/gemstoneFileSystemProvider.ts
ericwinger: Fixed in 21f2064c. Clearing is now supported: the save **always** recategorizes, and an empty/absent category clears it (`Class>>category: ''`). So deleting the `category:` line actually clears the category and round-trips, instead of the old category reappearing on reopen.

client/src/queries/getClassCategory.ts
ericwinger: Fixed in 21f2064c — added a single `dictLookupExpr(dict)` helper in `queries/util.ts` and refactored `classExistsInDictionary` to reuse `classLookupExpr`, as you suggested.

On scope: I went with a **full sweep** rather than just this call site, and I think that was the right call:

- Grepping turned up the *identical* hand-rolled ternary (`typeof dict === 'number' ? 'symbolList at: N' : 'symbolList objectNamed: #name'`) in **~10** query files (`getClassNames`, `getClassesWithCategory`, `getDefinedInstVarCounts`, `getDictionaryEntries`, `removeDictionary`, `getClassVersions`, `deleteClass`, …), plus two more about to be added on a sibling branch. Fixing only the flagged site would have left the same pattern scattered.
- It's a **zero-behavior-change** refactor: the helper's emitted Smalltalk is byte-identical to the old inline expression, so every existing `toContain('symbolList at: N')` assertion still passes and the live-stone integration suite is green.
- Most importantly, consolidating to one helper made a **guard test** possible (`noHandRolledDictLookup`) that fails if any file outside `util.ts` re-inlines the dict-by-parameter lookup — so the duplication can't creep back, and it'll flag the sibling branch's two new copies once it rebases on `main`. A partial fix would have left the guard test unable to pass.

So the "one-line comment" you suggested turned into one helper + a test that keeps it that way.

PR #398, opened 2026-08-10
Ban toBeTruthy/toBeFalsy in vitest tests

MatiasFernandez:
## Summary

- `toBeTruthy()`/`toBeFalsy()` are vague matchers: they pass for *any* truthy/falsy value, so a misuse can assert the wrong thing without ever failing (e.g. a typo'd getter returning `''` instead of `undefined` still reads as "falsy"), and when they do fail the message is just "expected truthy, got falsy" instead of showing the actual value.
- Enables `vitest/no-restricted-matchers` to ban both matchers outright, with a message pointing to the intended replacement.
- I initially tried `vitest/prefer-strict-boolean-matchers` (autofixable), but it isn't type-aware — it rewrote every `toBeTruthy`/`toBeFalsy` call to `toBe(true)`/`toBe(false)` regardless of the underlying type, which broke several existing tests (checks against strings, DOM nodes, and optional booleans where `undefined` should count as false). Went with `no-restricted-matchers` instead, which has no autofix, so each site was fixed by hand with the matcher that actually states its intent.
- Fixed the 18 existing call sites this newly caught: `toBeDefined()`/`not.toBeNull()` for presence checks, `Boolean(x).toBe(false)` where an optional boolean's absence should read as false.

## Test plan

- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run test:client` (324 files, 4973 tests passing)

### DIFF
--- a/client/src/__tests__/debuggerPanel.test.ts
+++ b/client/src/__tests__/debuggerPanel.test.ts
@@ -3903,7 +3903,7 @@ describe('DebuggerPanel', () => {
         (f: { receiverClass?: string }) => f.receiverClass === 'JasperDebugDemo',
       );
       expect(selfFrames.length).toBeGreaterThan(0);
-      for (const f of selfFrames) expect(Boolean(f.overridable)).toBe(false);
+      for (const f of selfFrames) expect(f.overridable).toBe(false);
     });
 
     it('opens a new-method template for the receiver class + frame selector, with banner help (no init)', async () => {
@@ -4325,7 +4325,7 @@ describe('DebuggerPanel', () => {
         (f: { receiverClass?: string }) => f.receiverClass === 'JasperMakerBase class',
       );
       expect(selfFrames.length).toBeGreaterThan(0);
-      for (const f of selfFrames) expect(Boolean(f.overridable)).toBe(false);
+      for (const f of selfFrames) expect(f.overridable).toBe(false);
     });
 
     it('implementInReceiver opens a CLASS-side new-method template for a metaclass receiver', async () => {
--- a/client/src/__tests__/debuggerPanel.test.ts
+++ b/client/src/__tests__/debuggerPanel.test.ts
@@ -3854,7 +3854,7 @@ describe('DebuggerPanel', () => {
 
     it('keeps the Create button suppressed once a create is underway (re-detect returns it)', async () => {
       const panel = openWithDnu();
-      expect(initPayload(panel).dnu).toBeTruthy(); // shown initially
+      expect(initPayload(panel).dnu).toBeDefined(); // shown initially
       sendMessage(panel, { command: 'createDnuMethod' });
       await flush();
       // A refresh while editing must NOT re-offer the button (method-in-progress).
@@ -3903,7 +3903,7 @@ describe('DebuggerPanel', () => {
         (f: { receiverClass?: string }) => f.receiverClass === 'JasperDebugDemo',
       );
       expect(selfFrames.length).toBeGreaterThan(0);
-      for (const f of selfFrames) expect(f.overridable).toBeFalsy();
+      for (const f of selfFrames) expect(Boolean(f.overridable)).toBe(false);
     });
 
     it('opens a new-method template for the receiver class + frame selector, with banner help (no init)', async () => {
@@ -4169,7 +4169,7 @@ describe('DebuggerPanel', () => {
       });
       const payload = initPayload(openPanel());
       expect(payload.subclassResp).toBeUndefined();
-      expect(payload.dnu).toBeTruthy();
+      expect(payload.dnu).toBeDefined();
     });
 
     it('QuickPicks the chain BOUNDED at the abstract definer, then opens a stub', async () => {
@@ -4325,7 +4325,7 @@ describe('DebuggerPanel', () => {
         (f: { receiverClass?: string }) => f.receiverClass === 'JasperMakerBase class',
       );
       expect(selfFrames.length).toBeGreaterThan(0);
-      for (const f of selfFrames) expect(f.overridable).toBeFalsy();
+      for (const f of selfFrames) expect(Boolean(f.overridable)).toBe(false);
     });
 
     it('implementInReceiver opens a CLASS-side new-method template for a metaclass receiver', async () => {
--- a/client/src/__tests__/explorerInstVar.test.ts
+++ b/client/src/__tests__/explorerInstVar.test.ts
@@ -74,9 +74,9 @@ describe('ExplorerController add instance variable', () => {
     const validate = opts?.validateInput as (v: string) => string | undefined;
     expect(validate('goodName')).toBeUndefined();
     expect(validate('_ok9')).toBeUndefined();
-    expect(validate('9bad')).toBeDefined();
-    expect(validate('has space')).toBeDefined();
-    expect(validate('  ')).toBeDefined();
+    expect(validate('9bad')).toContain('lowercase');
+    expect(validate('has space')).toContain('lowercase');
+    expect(validate('  ')).toContain('Enter a name');
     // Kept in step with the engine's isValidIvarName: an UPPERCASE first letter reads as a global
     // and must be rejected here too, not just declined after a round trip (#360 item 5). Assert the
     // *word* the engine's testAddDeclinesUppercaseFirstLetter also pins, so the two stay in step and
--- a/client/src/__tests__/explorerInstVar.test.ts
+++ b/client/src/__tests__/explorerInstVar.test.ts
@@ -74,9 +74,9 @@ describe('ExplorerController add instance variable', () => {
     const validate = opts?.validateInput as (v: string) => string | undefined;
     expect(validate('goodName')).toBeUndefined();
     expect(validate('_ok9')).toBeUndefined();
-    expect(validate('9bad')).toBeTruthy();
-    expect(validate('has space')).toBeTruthy();
-    expect(validate('  ')).toBeTruthy();
+    expect(validate('9bad')).toBeDefined();
+    expect(validate('has space')).toBeDefined();
+    expect(validate('  ')).toBeDefined();
     // Kept in step with the engine's isValidIvarName: an UPPERCASE first letter reads as a global
     // and must be rejected here too, not just declined after a round trip (#360 item 5). Assert the
     // *word* the engine's testAddDeclinesUppercaseFirstLetter also pins, so the two stay in step and
--- a/client/src/refactoring/__tests__/refactoringInstVar.integration.test.ts
+++ b/client/src/refactoring/__tests__/refactoringInstVar.integration.test.ts
@@ -292,7 +292,7 @@ r := (System myUserProfile symbolList objectNamed: #GsInstVarRefactoringTest) su
       await analyzeInstVar(asyncExec, 'add', BASE, 'mine', userIndex()),
     );
 
-    expect(analysis.decline).toBeTruthy();
+    expect(analysis.decline).toBeDefined();
     expect(analysis.decline).toContain(SUB);
   });
 
@@ -304,6 +304,6 @@ r := (System myUserProfile symbolList objectNamed: #GsInstVarRefactoringTest) su
     const analysis = parseAnalysis(
       await analyzeInstVar(asyncExec, 'add', BASE, 'count', userIndex()),
     );
-    expect(analysis.decline).toBeTruthy();
+    expect(analysis.decline).toBeDefined();
   });
 });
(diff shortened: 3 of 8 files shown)


client/src/refactoring/__tests__/refactoringInstVar.integration.test.ts
```diff
@@ -244,6 +244,6 @@ r := (System myUserProfile symbolList objectNamed: #GsInstVarRefactoringTest) su
     const analysis = parseAnalysis(
       await analyzeInstVar(asyncExec, 'add', BASE, 'count', userIndex()),
     );
-    expect(analysis.decline).toBeTruthy();
+    expect(analysis.decline).toBeDefined();
```
ericwinger: 🚫 This test can no longer fail.

`decline` is `string | null`, not `string | undefined` — `instVarRefactorPreview.ts:43` declares it, and `parseAnalysis` explicitly normalizes anything non-string to `null` at line 150:

```ts
decline: typeof env.decline === 'string' ? env.decline : null,
```

`toBeDefined()` asserts `!== undefined`, and `null !== undefined`, so this passes whether the engine declines or not. It's the only assertion in the test, so "declines adding a duplicate instance variable" now passes when the engine *doesn't* decline — which is the precise failure mode this PR exists to prevent.

Two things needed here. The matcher has to be null-aware, and per the house rule for decline tests the reason text is the actual contract, not non-nil-ness. The engine's message is built at `GsInstVarRefactoring.class.st:225`:

```smalltalk
^decline := 'Cannot add ', varName, ': ', definingClass name asString,
    ' already has an instance variable of that name.'
```

so:

```ts
    expect(analysis.decline).not.toBeNull();
    expect(analysis.decline).toContain('already has an instance variable');
```

That second line makes `not.toBeNull()` redundant, honestly — `toContain` on `null` throws — so just the `toContain` would do.

eslint.config.mjs
```diff
@@ -137,6 +137,19 @@ export default tseslint.config(
...
+      // reports "expected truthy, got falsy" instead of showing the value.
+      // Ban them and require a matcher that states the actual intent.
+      'vitest/no-restricted-matchers': [
+        'error',
+        {
+          toBeTruthy:
+            'Prefer a specific matcher: toBeDefined(), not.toBeNull(), toBe(true), toContain(...), etc.',
+          toBeFalsy: 'Prefer a specific matcher: toBeUndefined(), toBeNull(), toBe(false), etc.',
+        },
+      ],
```
ericwinger, line 177: 🚫 This doesn't ban the matchers outright — the negated forms escape.

`no-restricted-matchers` matches on the whole modifier chain. From `isChainRestricted` in `@vitest/eslint-plugin`:

```js
const isChainRestricted = (chain, restriction) => {
  if (Object.prototype.hasOwnProperty.call(ModifierName, restriction) || restriction.endsWith(".not"))
    return chain.startsWith(restriction);
  return chain === restriction;
};
```

`toBeTruthy` is neither a modifier name nor `.not`-suffixed, so it falls to **exact equality**. The chain for `expect(x).not.toBeTruthy()` is the string `"not.toBeTruthy"`, which is not `"toBeTruthy"` — no report.

I ran it to be sure, ESLint API with exactly this config over a synthetic file:

```
expect(a).toBeTruthy();                  ← reported
expect(b).toBeFalsy();                   ← reported
expect(c).not.toBeTruthy();              ← NOT reported
expect(d).not.toBeFalsy();               ← NOT reported
await expect(e).resolves.toBeTruthy();   ← NOT reported
await expect(f).rejects.toBeFalsy();     ← NOT reported
```

`expect(x).not.toBeTruthy()` is semantically identical to the `toBeFalsy()` being banned, so the loophole is a drop-in replacement for the thing we're removing. Nothing exploits it today (zero occurrences on the branch), but the rule needs to say what the PR title says.

Enumerating the chains fixes it — I re-ran with these keys and all six lines above are caught:

```js
      'vitest/no-restricted-matchers': [
        'error',
        Object.fromEntries(
          ['toBeTruthy', 'toBeFalsy'].flatMap((m) =>
            ['', 'not.', 'resolves.', 'rejects.', 'resolves.not.', 'rejects.not.'].map((p) => [
              p + m,
              `Prefer a matcher that states the intent: ...`,
            ]),
          ),
        ),
      ],
```

Spelling out all twelve literally is fine too if the generated form reads as too clever for a config file — your call which.

eslint.config.mjs
```diff
@@ -137,6 +137,19 @@ export default tseslint.config(
...
+      // so a misuse can assert the wrong thing without failing (e.g. a typo'd
+      // getter that returns '' instead of undefined), and a real failure just
+      // reports "expected truthy, got falsy" instead of showing the value.
+      // Ban them and require a matcher that states the actual intent.
+      'vitest/no-restricted-matchers': [
+        'error',
+        {
+          toBeTruthy:
+            'Prefer a specific matcher: toBeDefined(), not.toBeNull(), toBe(true), toContain(...), etc.',
+          toBeFalsy: 'Prefer a specific matcher: toBeUndefined(), toBeNull(), toBe(false), etc.',
```
ericwinger: 💭 The guidance leads with the weakest option, and that's not a coincidence.

`toBeDefined()` is listed first for `toBeTruthy`, and it's the one replacement that silently changes meaning: it only means the same thing when the value is `T | undefined`. On a `T | null` it admits `null`, which is how `refactoringInstVar.integration.test.ts:247` ended up with an assertion that can't fail.

Since the message is what the next person reads at 5pm, it could do the type dispatch for them:

```js
          toBeTruthy:
            'Prefer a matcher that states the intent: not.toBeNull() for `T | null`, toBeDefined() for `T | undefined`, toBe(true) for a boolean, toContain(...) for a message.',
          toBeFalsy:
            'Prefer a matcher that states the intent: toBeNull(), toBeUndefined(), toBe(false), or toHaveLength(0).',
```

Cheap, and it turns the rule into the thing that prevents the mistake instead of the thing that suggests it.

client/src/refactoring/__tests__/refactoringInstVar.integration.test.ts
```diff
@@ -232,7 +232,7 @@ r := (System myUserProfile symbolList objectNamed: #GsInstVarRefactoringTest) su
       await analyzeInstVar(asyncExec, 'add', BASE, 'mine', userIndex()),
     );
 
-    expect(analysis.decline).toBeTruthy();
+    expect(analysis.decline).toBeDefined();
```
ericwinger: 💭 Same `null` hole as the duplicate-decline test, just survivable here.

`decline` is `string | null`, so this line passes on `null` too. The difference is that line 236's `toContain(SUB)` throws on `null`, so the test as a whole still guards — this assertion is simply doing nothing.

Either drop it and let `toContain` carry the test, or make it null-aware:

```ts
    expect(analysis.decline).not.toBeNull();
```

I'd drop it. `toContain(SUB)` is a strictly stronger statement of the same contract, and one assertion that means something beats two where the first is decorative.

client/src/__tests__/explorerInstVar.test.ts
```diff
@@ -74,9 +74,9 @@ describe('ExplorerController add instance variable', () => {
     const validate = opts?.validateInput as (v: string) => string | undefined;
     expect(validate('goodName')).toBeUndefined();
     expect(validate('_ok9')).toBeUndefined();
-    expect(validate('9bad')).toBeTruthy();
-    expect(validate('has space')).toBeTruthy();
-    expect(validate('  ')).toBeTruthy();
+    expect(validate('9bad')).toBeDefined();
+    expect(validate('has space')).toBeDefined();
+    expect(validate('  ')).toBeDefined();
```
ericwinger: 💭 These three now sit slightly below the standard this very block sets.

`validate` returns `string | undefined` with no empty-string path, so `toBeDefined()` is faithful — no bug here. But four lines down, the `'Tally'` case asserts `toContain('lowercase')` with a comment explaining why: the *word* is what keeps this test in step with the engine's `testAddDeclinesUppercaseFirstLetter`, and a regression to a generic message has to fail here.

By that standard, `9bad` / `has space` / `'  '` currently pass on *any* non-empty message, including the generic one the comment is guarding against. Since you're already in here:

```ts
    expect(validate('9bad')).toContain('identifier');
    expect(validate('has space')).toContain('identifier');
    expect(validate('  ')).toContain('Enter');
```

(worth checking against the real strings — I read `validateNewIvarName`, not this validator). Fine to leave for a follow-up; it's just that the file already argues against itself here.

client/src/__tests__/debuggerPanel.test.ts
```diff
@@ -3903,7 +3903,7 @@ describe('DebuggerPanel', () => {
         (f: { receiverClass?: string }) => f.receiverClass === 'JasperDebugDemo',
       );
       expect(selfFrames.length).toBeGreaterThan(0);
-      for (const f of selfFrames) expect(f.overridable).toBeFalsy();
+      for (const f of selfFrames) expect(Boolean(f.overridable)).toBe(false);
```
ericwinger: 💭 Does `Boolean(x).toBe(false)` actually state an intent, or just satisfy the rule?

`overridable` is `boolean | undefined` (`debuggerPanel.ts:519`, `:669`), so this is exactly equivalent to the `toBeFalsy()` it replaced — same semantics, same failure message, one more wrapper. It reads as "assert this is falsy", which is the phrasing the rule set out to eliminate. Same pattern at `:4328` and `localVersion.test.ts:421`.

Not wrong, and I don't want to over-rotate on three lines. But if the contract is "no self-frame is ever marked overridable", `expect(f.overridable ?? false).toBe(false)` says that at the same cost, and if the field is genuinely never set on these frames then `expect(f.overridable).toBeUndefined()` says something the old assertion couldn't — it would catch a regression that starts writing an explicit `false`.

Mostly raising it because these three are the sites where the rule bought nothing, which is useful to know if the ban ever gets revisited.

ericwinger (approved):
Looks like at least one test can no longer fail and should be fixed before merging. 
The rest of the claude comments are inlined as usual.

client/src/refactoring/__tests__/refactoringInstVar.integration.test.ts
MatiasFernandez: Fixed — the duplicate-ivar test now asserts `toContain('already has an instance variable')`, and I dropped the redundant `toBeDefined()` above the `toContain(SUB)` in the subclass test. Heads up that I couldn't confirm the substring against a live engine: the refactoring plugin isn't installed in my stone, so both decline tests skip on the feature gate.

eslint.config.mjs
MatiasFernandez: Fixed — the rule now enumerates all six chains per matcher (bare, `not.`, `resolves.`, `rejects.`, and both `.not.` combinations), so all twelve keys are restricted.

I went with a named `banChain(matcher, message)` helper over the inline `flatMap` since you left the shape open: it keeps one message per matcher, which the separate request to make the advice type-specific needs. Verified by linting your six-line file — 2 of 6 reported before, all report now.

eslint.config.mjs
MatiasFernandez: Done — the messages now do the type dispatch, using your wording verbatim.

Went one step further: `banChain` takes a second message for the `not.`-bearing chains, so `expect(x).not.toBeTruthy()` gets the falsy advice rather than being told to write the thing it already wrote inverted. Verified against ESLint that all six chains per matcher report the advice matching their effective assertion.

client/src/refactoring/__tests__/refactoringInstVar.integration.test.ts
MatiasFernandez: Dropped it — `toContain(SUB)` carries the test.

client/src/__tests__/explorerInstVar.test.ts
MatiasFernandez: Applied, though your guessed strings didn't survive contact with the real validator — this test drives the inline `validateInput` in `addInstVarOnClass` (gemstoneExplorer.ts), not `validateNewIvarName`. It has only two messages, so `'identifier'` doesn't exist anywhere and `'9bad'`/`'has space'`/`'Tally'` all produce the same one.

Used `toContain('lowercase')` for `'9bad'` and `'has space'`, and `toContain('Enter a name')` for `'  '` — that last one is the real win, since `toBeDefined()` passed even if whitespace fell through to the shape message. Confirmed all three go red against a generic message.

client/src/__tests__/debuggerPanel.test.ts
MatiasFernandez: You're right, the wrappers bought nothing — but the fix isn't uniform across the three. In the debugger, `overridable` is computed as a pure boolean expression (`debuggerPanel.ts:3860`) and copied verbatim into the init payload, so those frames always carry an explicit `true`/`false`; the `| undefined` is the declaration at `:519`/`:669`, not the producer. `toBeUndefined()` would fail there and `?? false` would be dead code, so both sites are now `expect(f.overridable).toBe(false)` — which does catch the field going missing.

`localVersion.test.ts:421` is your case: `local` is only ever assigned in the symlink loop (`versionManager.ts:62`), never on the locally-built-directory path under test, so that one is now `toBeUndefined()`.

PR #402, opened 2026-08-10
Explorer: Rename Class Category + Rename Dictionary (R6)

ericwinger:
Adds two GemStone Explorer operations (catalog item **R6**): **Rename Class Category** and **Rename Dictionary**. Neither recompiles or commits — the user commits the session.

## Rename Class Category
Right-click / pencil on a category node. Renames that single node of the dash-segmented category tree and carries its whole subtree along, server-side via `Class>>category:` (`queries/renameClassCategory.ts`). Category names are compared by **Symbol identity** (`asSymbol ==`), never `String =`, to dodge the 3.6.x wide-vs-narrow `ArgumentError 2718`. Non-ASCII names are blocked below 3.7 (same boundary as server UTF-8 file-in). An empty overlay-only category is carried across in the client overlay.

## Rename Dictionary
Right-click / pencil on a dictionary node. A `SymbolDictionary` has no name slot — its name is a self-referential entry it holds by identity — so the rename is a reflective swap (add the new self-entry, drop the old); the symbol-list index is unchanged and the classes it holds are untouched (`queries/renameDictionary.ts`). System dictionaries (Globals / Published / UserGlobals) and name collisions are refused. Source that names the dictionary literally is **not** rewritten (rare; the confirmation says so).

## Consistency with `main`
Both new queries resolve their dictionary through the shared `dictLookupExpr` helper (from #397) rather than hand-rolling `symbolList at:/objectNamed:`, so the `noHandRolledDictLookup` guard test passes. `dictLookupExpr`'s numeric branch is made uniformly nil-on-absent (matching its name branch) so an out-of-range index answers nil instead of raising.

## Review addressed (pre-PR)
A review of this branch was folded in before opening:
- **MED-1** — after a dictionary rename, stale editor tabs still name the old dictionary in their URIs; close the clean ones and warn about unsaved ones.
- **MED-2** — parse `renamed: <n>`; an unrecognised payload is an error, and 0-moved-while-classes-expected surfaces a warning, not false success.
- **MED-3** — always run the category-rename query rather than gating on a possibly-stale client cache.
- **LOW-1** — a category named exactly `<old>-` now moves with its subtree (`>=`).
- **LOW-2** — classes whose category can't be read are counted and reported (`renamed: N skipped: M`).
- **LOW-3/4/5/6** — category-rename right-click entry; a tautological test assertion; retain selection when renaming the current dictionary; document the `classLookupExpr` index-branch guard asymmetry.

## Tests
Unit tests for both queries and both command handlers (including the MED-2 warn/error paths and the MED-1 tab sweep), plus live-stone integration tests that drive real renames. Emitted Smalltalk is ASCII-only (a non-ASCII char in a doit trips the 3.6.x `ComStrmSetCursor` compiler bug). Full local suite: **client 5267 · server 322 · mcp 92**; lint / prettier / tsc clean.

> **Note:** verified locally on a 3.6.2 stone; a 3.7.5 pass is worth it in CI since the category path leans on `Class>>category:` and symbol interning, which shifted between those releases.


client/src/queries/renameClassCategory.ts
```diff
@@ -0,0 +1,58 @@
...
+prefixSym := prefix asSymbol.
+count := 0.
+skipped := 0.
+dict keysAndValuesDo: [:k :v |
+  v isBehavior ifTrue: [
+    | cat |
+    cat := [v category asString] on: Error do: [:e | skipped := skipped + 1. nil].
+    cat ifNotNil: [
+      (cat size = oldCat size and: [cat asSymbol == oldSym])
+        ifTrue: [v category: newCat. count := count + 1]
```
MatiasFernandez: 💭 The *read* of `v category` is wrapped in `on: Error do:` (nice — that's LOW-2), but the *write* `v category: ...` isn't. If one class in the subtree can't be reassigned (authorization, a read-only segment, a corrupt class object), the whole doit unwinds mid-loop and the client lands in the `catch` — which shows the error and deliberately does **not** refetch (`explorerRenameClassCategory.test.ts` asserts that). So the server ends up with the category *half* renamed while the tree still shows the old one, and the only way to notice is a manual refresh.

Since the read already established the pattern, would it make sense to guard the assignment the same way and count it into `skipped`? Then a stubborn class degrades to `renamed: N skipped: M` (which the client already handles) instead of aborting the batch. Nothing is committed either way, so it's recoverable — but the mismatch between what the server did and what the tree shows is the part that worries me.

client/src/queries/util.ts
```diff
@@ -28,15 +28,16 @@ export function compiledMethodExpr(
...
+// an unbound name) — both branches are uniformly nil-on-absent, so callers just
+// `ifNil:` to handle the missing case. This is the ONE place that knows how to
+// resolve a dictionary within the symbol list — do not hand-roll
+// `System myUserProfile symbolList at:/objectNamed:` for a caller-supplied dict
+// elsewhere (see the duplication guard test). To resolve a CLASS within a
+// dictionary, use classLookupExpr instead.
 export function dictLookupExpr(dict: number | string): string {
   return typeof dict === 'number'
-    ? `System myUserProfile symbolList at: ${dict}`
+    ? `System myUserProfile symbolList at: ${dict} ifAbsent: [nil]`
```
MatiasFernandez, line 54: 💭 I'm a bit torn on making the numeric branch nil-on-absent, mostly because of the note you added ~20 lines below for `classLookupExpr`:

> a numeric dict index always comes from Jasper's own tree (a live SymbolList position), so out-of-range means a programming error worth surfacing, not a user-supplied name that may legitimately be absent

That reasoning seems to apply just as well here — the numeric `dict` reaching `dictLookupExpr` comes from the same place. And the practical effect isn't neutral: I checked the callers and they all `ifNil:`, but most of them answer `''` on nil (`getClassNames`, `getClassesWithCategory`, `getDictionaryEntries`, `getDefinedInstVarCounts`, `getClassVersions`, …). So a stale/out-of-range index that used to blow up loudly now renders as *an empty dictionary* — which looks exactly like a dictionary that legitimately has no classes.

The new queries do need a nil-able lookup for the string branch, which they already had. Is the numeric relaxation actually needed by `renameDictionary` / `renameClassCategory`, or did it come along for symmetry? If it's for symmetry, I'd lean toward keeping the raise and having the two docs agree; if it's needed, could the comment here say why the two helpers deliberately land on opposite answers for the same kind of input?

client/src/queries/renameDictionary.ts
```diff
@@ -0,0 +1,40 @@
...
+export function renameDictionary(
+  execute: QueryExecutor,
+  dict: number | string,
+  newName: string,
+): string {
+  const dictExpr = dictLookupExpr(dict);
+  const code = `| sl d newSym oldKey |
+sl := System myUserProfile symbolList.
+d := ${dictExpr}.
+d ifNil: [^ 'Dictionary not found'].
```
MatiasFernandez, line 30: 💭 `dictLookupExpr` with a string resolves through `objectNamed:`, which answers **whatever** is bound under that symbol — not necessarily a `SymbolDictionary`. `renameDictionary` is exported from `browserQueries` with `dict: number | string`, and the integration tests do call it by name, so `renameDictionary(session, 'Object', 'Foo')` is reachable. Today that gets past the `ifNil:` and the system-dictionary guard and then DNUs on `keyAtValue:` — so it fails loudly, which is fine, but the user sees a raw doesNotUnderstand rather than one of the tidy refusal strings the rest of this doit is careful to produce.

Worth widening the guard to something like `(d isKindOf: SymbolDictionary) ifFalse: [^ 'Not a dictionary']` right after the nil check?

client/src/queries/renameDictionary.ts
```diff
@@ -0,0 +1,40 @@
...
+((d == Globals) or: [(d == Published) or: [d == UserGlobals]])
+  ifTrue: [^ 'Cannot rename a system dictionary (Globals, Published, or UserGlobals)'].
+newSym := #'${escapeString(newName)}'.
+d name == newSym ifTrue: [^ 'ok'].
+(sl objectNamed: newSym) ifNotNil: [:o | ^ 'The name ${escapeString(newName)} is already in use in the symbol list'].
+oldKey := d keyAtValue: d ifAbsent: [nil].
```
MatiasFernandez, line 37: nit: the module comment describes the swap as "add the new one, drop the old", but the code reads `d name: newSym` then `removeKey:` — the "add" step is hidden behind `name:`. Could the comment name the selector ("`name:` installs the new self-entry; then drop the old key")? It'd save the next reader a trip into the kernel to confirm `name:` is what does the adding.

client/src/__tests__/explorerRenameDictionary.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  it('renames by index, sweeps stale tabs, refreshes and reports on success', async () => {
+    const { ctl, refresh, selectDict, closeStaleTabs } = makeController({} as ActiveSession);
+    vi.mocked(vscode.window.showInputBox).mockResolvedValue('NewDict');
+    vi.mocked(vscode.window.showWarningMessage).mockResolvedValue('Rename' as never);
+    vi.mocked(queries.renameDictionary).mockReturnValue('ok');
+    await ctl.renameDictionary(NODE);
+    expect(queries.renameDictionary).toHaveBeenCalledWith(expect.anything(), 3, 'NewDict');
+    // Stale editor tabs for the old name are swept (MED-1).
+    expect(closeStaleTabs).toHaveBeenCalledWith(expect.anything(), 'MyDict');
```
MatiasFernandez, line 121: 💭 `closeStaleTabsForRenamedDictionary` is spied out in `makeController`, so every test in this file runs against the stub and the only thing asserted is that it *was called with the old name*. That means the actual MED-1 behaviour — filtering by `sessionId` + `dictName`, skipping tabs whose `parseUri` throws, leaving dirty tabs open, and the singular/plural warning — has no coverage at all, and it's the part most likely to regress (it's the only place a rename can silently eat someone's open editor).

Since it's a private method, would it be worth one focused test that stubs `listOpenGemstoneTabs` / `tabGroups.close` and drives the sweep directly — asserting a clean matching tab is closed, a dirty one is *not*, and a tab from another session is untouched? The two existing sweeps (`refreshRenamedSelectorEditors`, `closeStaleSourceMethodEditors`) may already have a test setup you can borrow.

client/src/__tests__/explorerRenameClassCategory.test.ts
```diff
@@ -0,0 +1,251 @@
...
+
+    await ctl.renameClassCategory(NODE);
+
+    expect(vscode.window.showErrorMessage).toHaveBeenCalledWith(
+      expect.stringContaining('unexpected server reply'),
+    );
+    // Bailed before refetch / success.
+    expect(queries.getClassesWithCategory).not.toHaveBeenCalled();
+    expect(vscode.window.setStatusBarMessage).not.toHaveBeenCalled();
+  });
```
MatiasFernandez, line 257: 💭 Two behaviours the handler implements don't seem to be covered here: carrying the `newClassCategories` overlay entries across the rename (suffix-preserving, which the PR description calls out as a feature), and remapping `this.state.classCategory` when the selected category is inside the renamed subtree. Both are pure client-side bookkeeping with a suffix-slicing calculation that's easy to get off by one, and neither shows up in the integration tests (the server doesn't know about the overlay).

Worth adding a case that seeds `newClassCategories` with `Announcements` and `Announcements-Core-Empty`, renames the node, and asserts the set now holds `Events` and `Events-Core-Empty`?

client/src/__tests__/explorerQueries.integration.test.ts
```diff
@@ -384,6 +384,78 @@ describe('explorer queries (integration)', () => {
...
+    it('renames a dictionary in place, keeping its symbol-list position', () => {
+      q.addDictionary(session(), 'JasperItRenameSrc');
+      const before = dictIndexOf('JasperItRenameSrc');
+      expect(before).toBeGreaterThan(0);
+
+      const result = q.renameDictionary(session(), 'JasperItRenameSrc', 'JasperItRenameDst');
+
+      expect(result).toBe('ok');
+      expect(dictIndexOf('JasperItRenameSrc')).toBe(0); // old name gone
+      expect(dictIndexOf('JasperItRenameDst')).toBe(before); // new name, same index
```
MatiasFernandez, line 397: 💭 This proves the symbol-list *index* survives, but the claim that makes the reflective swap interesting — "the classes it holds are untouched" and the dictionary resolves under the new name — isn't asserted. As written the test would still pass if the self-entry swap left the dictionary's contents unreachable.

Could we file a class into `JasperItRenameSrc` first and then, after the rename, assert it's still found under `JasperItRenameDst` (e.g. via `getClassNames` on the new name, or `objectNamed:`)? That's the one thing only a live stone can tell us, so it feels like the assertion this integration test exists for.

client/src/__tests__/explorerRenameClassCategory.test.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  it('does nothing without a session', async () => {
+    const { ctl } = makeController(undefined);
+    await ctl.renameClassCategory(NODE);
+    expect(vscode.window.showInputBox).not.toHaveBeenCalled();
+    expect(queries.renameClassCategory).not.toHaveBeenCalled();
+  });
+
+  it('does nothing when the prompt is cancelled or unchanged', async () => {
```
MatiasFernandez: nit: this one packs two scenarios (cancelled, and unchanged) into a single test, and they share the mock — if the first `await` doesn't consume its `mockResolvedValueOnce`, the second call silently picks it up and the assertion still passes. `explorerRenameDictionary.test.ts` splits these into 'does nothing when the name prompt is cancelled' / 'does nothing when the name is unchanged'; worth matching that here?

package.json
```diff
@@ -2008,6 +2028,26 @@
...
+        },
+        {
+          "command": "gemstone.explorer.renameClassCategory",
+          "when": "view == gemstoneExplorerCategories && viewItem == explorerCategory",
+          "group": "1_modification@0"
+        },
+        {
+          "command": "gemstone.explorer.renameDictionary",
+          "when": "view == gemstoneExplorerDicts && viewItem == explorerDict",
+          "group": "inline@0"
```
MatiasFernandez, line 2044: nit: the new inline entries use `inline@0` while the existing `removeDictionary` on the same view/item stays at a bare `"inline"` (no order). Mixing ordered and unordered entries in one group leaves the trash-vs-pencil button order up to VS Code's fallback sort. Should `removeDictionary` get an explicit `inline@1` so the pair is pinned?

MatiasFernandez (approved):

ericwinger (commented):
Thanks for the thorough pass. All nine comments addressed — replies inline on each. Summary: guarded the category write (#1) and widened the dictionary type guard (#3); kept dictLookupExpr nil-on-absent for user-friendliness and reconciled both helper docs (#2); comment/manifest polish (#4, #9); and added the missing coverage — the tab-sweep suite (#5), the overlay-carry + classCategory remap case (#6), a live-stone contents-reachability assertion (#7), and split the cancelled/unchanged test (#8).

client/src/queries/renameClassCategory.ts
ericwinger: (re: guarding the write) Fixed. Both `v category:` writes are now wrapped in `on: Error do: [:e | skipped := skipped + 1]`, so a class that can't be reassigned counts into `skipped` and the loop continues — matching the read guard. A stubborn class degrades to `renamed: N skipped: M` (already handled client-side) instead of unwinding the batch and leaving the server half-renamed against a stale tree. Nothing committed, so it's recoverable. LOW-2 note extended to cover the write.

client/src/queries/util.ts
ericwinger: (re: numeric nil-on-absent) Kept it nil on purpose and reconciled the docs rather than reverting — we didn't want to raise, because a raise here isn't a tidy 'programming error surfaced': the out-of-range `at:` unwinds the doit before the caller's `ifNil:` guard, and executeFetchString rethrows the raw GemStone index-out-of-range as a walkback notification. And the trigger is a stale tree (a dictionary earlier in the list was removed, node not yet refreshed), which a refresh recovers — not really a code bug. So nil -> clean refusal wins. I rewrote both comments to agree: `dictLookupExpr` explains why nil, and `classLookupExpr`'s NB now says both prefer a clean answer and diverge only because guarding its numeric branch needs an extra `ifNotNil:` wrap. The empty-dictionary masking you flagged is the one cost of nil (also self-correcting) and is now called out.

client/src/queries/renameDictionary.ts
ericwinger: Done — added `(d isKindOf: SymbolDictionary) ifFalse: [^ 'Not a dictionary']` right after the nil check, so `renameDictionary(session, 'Object', ...)` now gives a tidy refusal instead of a raw DNU on `keyAtValue:`.

client/src/queries/renameDictionary.ts
ericwinger: Done — the comment now names the selectors: "`name:` installs the new self-entry, then `removeKey:` drops the old key".

client/src/__tests__/explorerRenameDictionary.test.ts
ericwinger: Done — added a focused `closeStaleTabsForRenamedDictionary` sweep suite that stubs `listOpenGemstoneTabs`/`parseUri` and drives the private method directly: a clean matching tab is closed, a dirty one is left open (with the singular/plural warning), and other-session / unparseable tabs are skipped.

client/src/__tests__/explorerRenameClassCategory.test.ts
ericwinger: Done — added a case seeding `newClassCategories` with `Announcements` + `Announcements-Core-Empty` and `state.classCategory = 'Announcements-Core'`; after renaming to `Events` it asserts the overlay holds `Events` + `Events-Core-Empty` (unrelated entry untouched) and the selection remaps to `Events-Core`. Plus a guard case that a prefix-sharing-but-outside-subtree selection (`Announcementss-Core`) is left alone.

client/src/__tests__/explorerQueries.integration.test.ts
ericwinger: Done — added an integration case that files a class into `JasperItRenameSrc`, renames, then asserts the class is still reachable under `JasperItRenameDst` (via `getClassNames` on the new index and `includesKey:` on the dictionary resolved by the new name). That's the live-stone-only assertion this test exists for.

client/src/__tests__/explorerRenameClassCategory.test.ts
ericwinger: Done — split into 'does nothing when the name prompt is cancelled' / 'does nothing when the name is unchanged', matching explorerRenameDictionary.test.ts so neither shares a mock.

package.json
ericwinger: Done — `removeDictionary` is now `inline@1`, pinning it after the `inline@0` rename pencil so the pencil/trash order is fixed.

PR #404, opened 2026-08-10
Deduplicate ownInstVarsOf: / allInstVarsOf: onto GsRefactoringEnvironment (#401)

ericwinger:
Closes #401.

`ownInstVarsOf:` and `allInstVarsOf:` were byte-identical one-liners copy-pasted into four
engine classes — `GsExtractSuperclassRefactoring`, `GsInstVarRefactoring`,
`GsInstVarStructureRefactoring`, `GsSplitClassRefactoring` — which is how they could quietly
drift apart. This consolidates them.

## What changed

| | |
|---|---|
| **Added** | `GsRefactoringEnvironment >> ownInstVarNamesOf:` / `allInstVarNamesOf:`, next to the sibling structural queries (`descendantsOf:`, `descendantsOf:declaringInstVar:`, `classVarNamesVisibleTo:`) |
| **Removed** | all 8 copies from the four refactorings |
| **Rewired** | 40 call sites, `self ownInstVarsOf:` → `environment ownInstVarNamesOf:` |
| **Bonus dedup** | `descendantsOf:declaringInstVar:` had the same one-liner inlined in its `select:` block; it now calls `ownInstVarNamesOf:` too |
| **Regenerated** | `engine.gs`, `engine-tests.gs`, `manifest.gs` (920 methods: −2 on each of the four classes, +2 on the environment) |

No behaviour change — these are pure read-only queries with no state.

### Two things worth a reviewer's eye

**Why the call sites talk to `environment` rather than keeping a thin local forwarder.** The
`allClassVarsOf:` precedent from #394 kept a one-line delegator on the refactoring. Going that
route here would still leave four byte-identical one-liners in four classes, just shallower
ones — which is the thing the issue asks to remove. Deleting them outright is a wider diff but
leaves nothing to drift. All 40 sites are instance-side, and `environment` is assigned at
instance creation and lives on the instance the preview token holds in `SessionTemps`, so it is
never nil on any path — including the resumed `applyForToken:` path.

**Why they were renamed instead of moved verbatim.** The environment's own convention is
`...NamesOf...` (`classVarNamesVisibleTo:`), and these answer *names*, not variables. The doc
comments now also record why the names are normalised to Strings, which none of the four copies
said: callers compare them against a name that arrived as a String from the client, and on 3.6.x
a Symbol/String comparison can silently answer false (the Unicode-comparison trap).

## Tests — [GS SUnit], `GsRefactoringEnvironmentTest` 9 → 16

The fixture subclass now declares its own `'beta'` and `'gamma'` so own-vs-inherited is
distinguishable.

- own excludes inherited and preserves declaration order; all lists inherited-then-own
- own is empty for a subclass that declares none, while all still sees the inherited name;
  empty for a class with no instance variables
- results are **Strings, not Symbols** — load-bearing, and verified as such against a live
  stone: raw `#instVarNames` elements answer `isSymbol` true and normalised ones false, so
  dropping the `asString` fails this test
- `descendantsOf:declaringInstVar:` — which had **no coverage at all** before this PR — finds
  only redeclaring subclasses, not merely-inherited names
- the read-only/commit-state test now exercises all three queries
- a **drift guard**: each query has exactly one implementor, and `#ownInstVarsOf:` /
  `#allInstVarsOf:` have none — so a re-introduced copy fails loudly instead of diverging again

> Note on the drift guard: it asks the *running stone* for implementors, so it fails against a
> stale installed engine. That is deliberate (a stale engine has masked a test before), but it
> does mean a local run wants `npm run test:server:install-plugin` first. CI installs from the
> branch, so it is always consistent there.

## Verification

| Layer | Result |
|---|---|
| RB engine SUnit — **3.6.2** (topaz, stage + run + abort) | **578 run, 0 failures+errors** |
| RB engine SUnit — **3.7.5** | **578 run, 0 failures+errors** |
| `GsRefactoringEnvironmentTest` alone | 16 run (was 9), 0 bad |
| lint / format:check / compile | clean |
| client (`--project default`) | 336 files, 5225 passed, 12 skipped |
| server / mcp-server | 322 passed / 92 passed |
| refactoring integration, serial | 126 files, 1209 passed, 1 skipped |

The engine was reinstalled on the `.env.test` stone from this branch before the client runs, so
nothing was measured against a stale engine.

**Not run:** the full CI matrix (`test-ci-local.sh 3.6.2 3.7.5` + the on-demand GCI suite). Both
RB SUnit boundaries and the whole fast gate are green above; happy to run the matrix on request.

### DIFF
--- a/gs-src/refactoring/engine/GsInstVarRefactoring.class.st
+++ b/gs-src/refactoring/engine/GsInstVarRefactoring.class.st
@@ -152,14 +152,6 @@ GsInstVarRefactoring >> optionsOf: aClass [
 ]
 
 { #category : 'private' }
-GsInstVarRefactoring >> allClassVarsOf: aClass [
-	"aClass's own and inherited class-variable names, as a Set of Strings. Used to decline an
-	 #add whose name would shadow a class variable inside method bodies. Shared with V8 split --
-	 the walk itself lives on the environment so both refactorings ask the same question."
-	^environment classVarNamesVisibleTo: aClass
-]
-
-{ #category : 'private' }
 GsInstVarRefactoring >> isValidIvarName: aString [
 	"A syntactically valid Smalltalk instance-variable name: a LOWERCASE letter or an underscore,
 	 followed by letters, digits, or underscores.
@@ -207,7 +199,7 @@ GsInstVarRefactoring >> analyzeAdd [
 		^decline := 'Cannot add ', varName printString, ': an instance-variable name must start with a lowercase letter or an underscore, followed by letters, digits, or underscores.'].
 	((environment allInstVarNamesOf: definingClass) includes: varName) ifTrue: [
 		^decline := 'Cannot add ', varName, ': ', definingClass name asString, ' already has an instance variable of that name.'].
-	((self allClassVarsOf: definingClass) includes: varName) ifTrue: [
+	((environment classVarNamesVisibleTo: definingClass) includes: varName) ifTrue: [
 		^decline := 'Cannot add ', varName, ': a class variable of that name is visible to ', definingClass name asString, ', which it would shadow.'].
 	"A subclass that already declares varName as its OWN instance variable would end up with two
 	 of that name once the new one is inherited. That fails mid-apply (error 2271), so decline up
--- a/gs-src/refactoring/engine/GsInstVarRefactoring.class.st
+++ b/gs-src/refactoring/engine/GsInstVarRefactoring.class.st
@@ -146,17 +146,6 @@ GsInstVarRefactoring >> varName [
 ]
 
 { #category : 'private' }
-GsInstVarRefactoring >> ownInstVarsOf: aClass [
-	"aClass's OWN instance-variable names (not inherited), as an Array of Strings."
-	^aClass instVarNames collect: [:e | e asString]
-]
-
-{ #category : 'private' }
-GsInstVarRefactoring >> allInstVarsOf: aClass [
-	^aClass allInstVarNames collect: [:e | e asString]
-]
-
-{ #category : 'private' }
 GsInstVarRefactoring >> optionsOf: aClass [
 	"aClass's own class-creation options, as an Array of Strings (e.g. #('selfCanBeSpecial'))."
 	^(aClass _optionsArray ifNil: [#()]) collect: [:e | e asString]
@@ -216,7 +205,7 @@ GsInstVarRefactoring >> analyzeAdd [
 	| conflicts names |
 	(self isValidIvarName: varName) ifFalse: [
 		^decline := 'Cannot add ', varName printString, ': an instance-variable name must start with a lowercase letter or an underscore, followed by letters, digits, or underscores.'].
-	((self allInstVarsOf: definingClass) includes: varName) ifTrue: [
+	((environment allInstVarNamesOf: definingClass) includes: varName) ifTrue: [
 		^decline := 'Cannot add ', varName, ': ', definingClass name asString, ' already has an instance variable of that name.'].
 	((self allClassVarsOf: definingClass) includes: varName) ifTrue: [
 		^decline := 'Cannot add ', varName, ': a class variable of that name is visible to ', definingClass name asString, ', which it would shadow.'].
@@ -232,7 +221,7 @@ GsInstVarRefactoring >> analyzeAdd [
 			(conflicts size = 1 ifTrue: [' already declares'] ifFalse: [' already declare']),
 			' an instance variable of that name; adding it to ', definingClass name asString,
 			' would duplicate that variable.'].
-	newIvarLists at: definingClass name asString put: ((self ownInstVarsOf: definingClass) copyWith: varName).
+	newIvarLists at: definingClass name asString put: ((environment ownInstVarNamesOf: definingClass) copyWith: varName).
 	self computeAffectedFrom: (Array with: definingClass).
 	self recordWillNotRecompileShadowedBy: definingClass
 ]
@@ -241,9 +230,9 @@ GsInstVarRefactoring >> analyzeAdd [
 GsInstVarRefactoring >> analyzeRemove [
 	"V1 remove: varName must be an OWN instance variable of definingClass. Every method (defining
 	 class and descendants) that accesses it will lose it -- those are the willNotRecompile set."
-	((self ownInstVarsOf: definingClass) includes: varName) ifFalse: [
+	((environment ownInstVarNamesOf: definingClass) includes: varName) ifFalse: [
 		^decline := 'Cannot remove ', varName, ': it is not an instance variable declared in ', definingClass name asString, '.'].
-	newIvarLists at: definingClass name asString put: ((self ownInstVarsOf: definingClass) reject: [:n | n = varName]).
+	newIvarLists at: definingClass name asString put: ((environment ownInstVarNamesOf: definingClass) reject: [:n | n = varName]).
 	self computeAffectedFrom: (Array with: definingClass).
 	self recordWillNotRecompileLosing: definingClass
 ]
@@ -368,7 +357,7 @@ GsInstVarRefactoring >> previewDefinitionFor: aClass oldDef: defString [
 	 are not surfaced for editing in the panel."
 	^self replaceListClause: 'instVarNames:'
 		in: defString
-		with: (newIvarLists at: aClass name asString ifAbsent: [self ownInstVarsOf: aClass])
+		with: (newIvarLists at: aClass name asString ifAbsent: [environment ownInstVarNamesOf: aClass])
 ]
 
 { #category : 'building' }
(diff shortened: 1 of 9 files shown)


gs-src/refactoring/engine/GsRefactoringEnvironment.class.st
```diff
@@ -168,6 +168,31 @@ GsRefactoringEnvironment >> classVarNamesVisibleTo: aClass [
...
+
+	 Strings, not the Symbols #instVarNames answers, because every caller compares them
+	 against a name that arrived as a String from the client. Comparing a Symbol against
+	 such a String can silently answer false on 3.6.x (the Unicode-comparison trap), so
+	 normalising here is what makes those comparisons hold across releases.
+
+	 Shared by every refactoring that reasons about instance-variable structure --
+	 add/remove, push up/down/move, extract superclass, split class -- so the
+	 normalisation lives in one place. Read-only."
+	^(aClass instVarNames ifNil: [#()]) collect: [:e | e asString]
```
MatiasFernandez, line 194: 💭 The four copies this replaces sent `instVarNames collect:` with no nil guard, so a class whose `instVarNames` answered nil used to raise; now it quietly answers `#()`. `descendantsOf:declaringInstVar:` already had that guard, so I read this as deliberate — but is nil actually reachable? If it is, it's worth a test pinning the empty answer, since the description says "no behaviour change" and this is a small widening. If it isn't, the guard hides an invariant violation from whoever hits it next. Either way, a clause in the doc comment saying which one it is would save the next reader the same question.

gs-src/refactoring/engine/GsRefactoringEnvironment.class.st
```diff
@@ -151,7 +151,7 @@ GsRefactoringEnvironment >> descendantsOf: aClass declaringInstVar: aName [
 	 Shared by every refactoring that introduces an instance variable on an existing class (V1 add,
 	 V8 split), so the check and its cross-version behaviour live in one place. Read-only."
 	^(self descendantsOf: aClass)
-		select: [:d | ((d instVarNames ifNil: [#()]) collect: [:e | e asString]) includes: aName]
+		select: [:d | (self ownInstVarNamesOf: d) includes: aName]
```
MatiasFernandez: 💭 `aName` is compared against the normalised Strings, so a caller passing a Symbol gets a silently empty answer — the exact Unicode-comparison trap the new `ownInstVarNamesOf:` comment describes, just on the argument side rather than the result side. Both current callers pass Strings so nothing is broken today, but `instanceMethodsAccessing:inClass:` in this same class accepts either (there's a test for it), so the environment is inconsistent about it. Should this normalise `aName asString` too, now that the comparison it leans on is documented?

gs-src/refactoring/tests/GsRefactoringEnvironmentTest.class.st
```diff
@@ -141,6 +176,44 @@ GsRefactoringEnvironmentTest >> testInstVarNameArgumentAcceptsAStringOrASymbol [
...
+	 makes those comparisons hold, so it is asserted rather than assumed."
+	| env |
+	env := GsRefactoringEnvironment new.
+
+	(env ownInstVarNamesOf: self subFixture) do: [:each |
+		self assert: each isString.
+		self deny: each isSymbol].
+	(env allInstVarNamesOf: self subFixture) do: [:each |
+		self assert: each isString.
+		self deny: each isSymbol].
```
MatiasFernandez, line 210: 💭 Both loops pass vacuously if the query ever answers an empty collection: a regression that made `ownInstVarNamesOf:` return `#()` would leave only the last line catching anything, and the `allInstVarNamesOf:` half would be unguarded entirely. Worth asserting the precondition first (`self deny: (env ownInstVarNamesOf: self subFixture) isEmpty`, same for `all`), so the test can't quietly stop proving the thing its comment says is load-bearing?

gs-src/refactoring/tests/GsRefactoringEnvironmentTest.class.st
```diff
@@ -151,10 +224,29 @@ GsRefactoringEnvironmentTest >> testReadOnlyQueriesDoNotChangeCommitState [
...
+	 assert that, so a re-introduced copy (or a revived #ownInstVarsOf: / #allInstVarsOf:)
+	 fails here instead of silently diverging again."
+	| env |
+	env := GsRefactoringEnvironment new.
+
+	#(#ownInstVarNamesOf: #allInstVarNamesOf:) do: [:sel |
+		self assert: ((env implementorsOf: sel) collect: [:m | m inClass])
+			equals: (Array with: GsRefactoringEnvironment)].
+	#(#ownInstVarsOf: #allInstVarsOf:) do: [:sel |
+		self assert: (env implementorsOf: sel) isEmpty]
```
MatiasFernandez: 💭 This asks the running stone via `ClassOrganizer`, so it asserts the state of the *installed* engine rather than the source on this branch — the stale-engine caveat you call out in the description. Two things that would soften it:

- A description on the asserts naming that cause. Otherwise the next person sees "0 implementors of #ownInstVarNamesOf:" and debugs it as a real regression before working out they just need `npm run test:server:install-plugin`.
- The second loop asserts that *nothing in the whole image* implements `#ownInstVarsOf:` / `#allInstVarsOf:`. Those are generic enough names that an unrelated class adopting one would fail this test for a reason that has nothing to do with the drift it's guarding. Would scoping the implementor check to the four engine classes make it fail only for the case it's actually about?

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
```diff
@@ -216,7 +205,7 @@ GsInstVarRefactoring >> analyzeAdd [
 	| conflicts names |
 	(self isValidIvarName: varName) ifFalse: [
 		^decline := 'Cannot add ', varName printString, ': an instance-variable name must start with a lowercase letter or an underscore, followed by letters, digits, or underscores.'].
-	((self allInstVarsOf: definingClass) includes: varName) ifTrue: [
+	((environment allInstVarNamesOf: definingClass) includes: varName) ifTrue: [
```
MatiasFernandez, line 200: 💭 Two lines apart we now have `environment allInstVarNamesOf:` and, just below, `self allClassVarsOf:` — and that one (line 166) is exactly the thin forwarder the description argues against keeping for the ivar queries. It's the only one left now, so the drift argument applies to it just as well. Is it worth inlining it to `environment classVarNamesVisibleTo:` here too so the file reads one way?

MatiasFernandez (approved):

gs-src/refactoring/engine/GsRefactoringEnvironment.class.st
ericwinger: Went the doc route — #ownInstVarNamesOf: now says the ifNil: is defensive-only: a live class always answers an Array from #instVarNames, so nil is not reachable here; the guard just mirrors the one #descendantsOf:declaringInstVar: already carried and keeps an empty Array rather than raising if that ever changes. No test, since there is no reachable path to pin. (931c1785)

gs-src/refactoring/engine/GsRefactoringEnvironment.class.st
ericwinger: Good catch — normalized `aName asString` in #descendantsOf:declaringInstVar: so a Symbol argument cannot hit the trap on the argument side, matching #instanceMethodsAccessing:inClass:. Added `testDescendantsDeclaringInstVarAcceptsAStringOrASymbol` pinning Symbol == String parity. (931c1785)

gs-src/refactoring/tests/GsRefactoringEnvironmentTest.class.st
ericwinger: Added — `testInstVarNameQueriesAnswerStringsNotSymbols` now `deny:`s empty on both #ownInstVarNamesOf: and #allInstVarNamesOf: before the loops, so a regression to #() can no longer pass vacuously. (931c1785)

gs-src/refactoring/tests/GsRefactoringEnvironmentTest.class.st
ericwinger: Both done: the comment now names the installed-engine/stale caveat (reinstall via `npm run test:server:install-plugin` before reading a 0-implementors result as a real regression), and the revived-copy check is scoped to the four engine classes that once held the copies (GsExtractSuperclass / GsInstVar / GsInstVarStructure / GsSplitClass) instead of the whole image. (931c1785)

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
ericwinger: Agreed — inlined it. Removed GsInstVarRefactoring>>allClassVarsOf: (single sender) and now call `environment classVarNamesVisibleTo:` directly in #analyzeAdd, so the file reads one way. (931c1785)

PR #405, opened 2026-08-10
Fix Unicode7 doit-literal comparisons on GemStone 3.6.x (#399, #400)

ericwinger:
## Summary

On GemStone 3.6.x, every string literal inside a GCI-executed doit compiles to `Unicode7`, so comparing an image-derived value against a **string literal** misbehaves in one of two ways:

- `String = 'literal'` → **raises** `ArgumentError` 2718 ("Unicode argument disallowed in String comparison")
- `Symbol = 'literal'` → silently answers **false**

Both hazards had a live client site. 3.7.x is unaffected either way.

## Fixes

**#399 — class sync/export failed outright on 3.6.x.** `syncClassBuildExpr` resolved the target dictionary with `(sl at: i) name asString = '<dict>'`, which raised 2718, so the whole sync/export path died. Now compares interned Symbols: `(sl at: i) name asSymbol == #'<dict>'`.

**#400 — Rowan project diff labelled every row `M` on 3.6.x.** `diffRowanProject` classified operations with `op class name = 'CypressAddition'`. Since `op class name` is a `Symbol`, `Symbol = Unicode7` answered `false` and both the add/remove branches fell through to `'M'` — silently wrong, no error. Now compares `op class name asSymbol == #CypressAddition` / `== #CypressRemoval`.

Both changes are no-ops on 3.7.x (worked before and after) and correct on 3.6.x.

## Tests

- **Unit guards** on the generated Smalltalk assert the Symbol idiom for both sites (`client/src/sync/__tests__/syncProtocol.test.ts` for #399; a new case in `rowanQueries.test.ts` for #400). Each was confirmed to **fail on the pre-fix source and pass on the fix**, so they genuinely bite.
- The `exportManager` fake-executor regex is retargeted to the new idiom.
- **Verified end-to-end against a live 3.6.2 stone by hand:** the pre-fix `syncClassBuildExpr` raised exactly `ArgumentError (error 2718) ... UserGlobals`; the fix returns the class payload.

A committed live integration test was deliberately omitted — it destabilized the parallel RB-SUnit integration tests by adding a concurrent GCI session to the shared stone (variable victims, always the in-stone SUnit suites). The deterministic unit guards plus the manual live proof cover the fix without making the suite flaky.

## Scope

Client-only (two query-builder strings + tests); no engine `.gs` changes. Gate: lint / format:check / compile clean, full client suite 337/337.

Fixes #399
Fixes #400

### DIFF
--- a/client/src/__tests__/exportManager.test.ts
+++ b/client/src/__tests__/exportManager.test.ts
@@ -68,7 +68,7 @@ const h = vi.hoisted(() => {
 
   // syncClassBuildExpr: resolve dict by name, return `idx \t hash \n source`.
   const classPayload = (img: FakeImage, code: string): string => {
-    const dn = code.match(/name asString = '((?:[^']|'')*)'/);
+    const dn = code.match(/name asSymbol == #'((?:[^']|'')*)'/);
     const cn = code.match(/at: #'((?:[^']|'')*)'/);
     if (!dn || !cn) return '';
     const dictName = dn[1].replace(/''/g, "'");
--- a/client/src/__tests__/exportManager.test.ts
+++ b/client/src/__tests__/exportManager.test.ts
@@ -68,9 +68,17 @@ const h = vi.hoisted(() => {
 
   // syncClassBuildExpr: resolve dict by name, return `idx \t hash \n source`.
   const classPayload = (img: FakeImage, code: string): string => {
-    const dn = code.match(/name asSymbol == #'((?:[^']|'')*)'/);
+    const dn = code.match(/name asString asSymbol == #'((?:[^']|'')*)'/);
     const cn = code.match(/at: #'((?:[^']|'')*)'/);
-    if (!dn || !cn) return '';
+    // This fake tracks the generated class-build idiom by hand. If it drifts (e.g. syncClassBuildExpr
+    // changes how it names the dictionary or class), fail loudly HERE rather than returning '' — an
+    // empty payload silently reads as "class not found" and surfaces as a puzzling sync result several
+    // assertions later. classPayload is only called for 'class'-labeled codes, so a miss is a harness
+    // bug, not a valid input.
+    if (!dn || !cn)
+      throw new Error(
+        `exportManager test fake: class-build idiom not recognized — retarget classPayload's regex to match syncClassBuildExpr. Code was:\n${code}`,
+      );
     const dictName = dn[1].replace(/''/g, "'");
     const className = cn[1].replace(/''/g, "'");
     const c = img.classes.find(
--- a/client/src/queries/rowan/diffRowanProject.ts
+++ b/client/src/queries/rowan/diffRowanProject.ts
@@ -37,6 +37,9 @@ ws := WriteStream on: Unicode7 new.
 patches do: [:assoc | | pkg |
   pkg := assoc key.
   assoc value operations do: [:op | | c |
+    "Unicode7 idiom (see escapeString in queries/util.ts): a string literal inside a GCI doit is
+     Unicode7 on 3.6.x, so compare the image-derived class name as an interned Symbol via asSymbol
+     == #Lit -- a plain = against a string literal raises ArgumentError 2718 on a 3.6.x client."
     c := op class name asSymbol == #CypressAddition
       ifTrue: ['I']
       ifFalse: [op class name asSymbol == #CypressRemoval ifTrue: ['D'] ifFalse: ['M']].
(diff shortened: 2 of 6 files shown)


client/src/sync/syncProtocol.ts
```diff
@@ -112,7 +112,7 @@ export function syncClassBuildExpr(dictName: string, className: string): string
   | sl idx |
   sl := System myUserProfile symbolList.
   idx := 0.
-  1 to: sl size do: [:i | (idx = 0 and: [(sl at: i) name asString = '${escapeString(dictName)}']) ifTrue: [idx := i]].
+  1 to: sl size do: [:i | (idx = 0 and: [(sl at: i) name asSymbol == #'${escapeString(dictName)}']) ifTrue: [idx := i]].
```
MatiasFernandez: 💭 Is it possible for a dictionary in the symbol list to have a nil `name`? Before this change `nil asString` answered `'nil'`, so an unnamed dictionary simply didn't match and the loop moved on; now `nil asSymbol` would be a doesNotUnderstand and the whole sync/export dies — the same failure mode #399 was about, just from a different cause. And since the loop scans every dictionary until it finds a match, one unnamed dictionary sitting ahead of the target is enough to trigger it.

Would `name asString asSymbol == #'...'` keep both properties (interned-Symbol comparison *and* nil-tolerant)?

client/src/__tests__/exportManager.test.ts
```diff
@@ -68,7 +68,7 @@ const h = vi.hoisted(() => {
 
   // syncClassBuildExpr: resolve dict by name, return `idx \t hash \n source`.
   const classPayload = (img: FakeImage, code: string): string => {
-    const dn = code.match(/name asString = '((?:[^']|'')*)'/);
+    const dn = code.match(/name asSymbol == #'((?:[^']|'')*)'/);
```
MatiasFernandez: 💭 This fake's regex has to be retargeted by hand in lockstep with the generated idiom, and if it were ever missed, `classPayload` returns `''` — so the fake silently behaves as "class not found" instead of failing. Should the `if (!dn || !cn)` branch throw instead of returning `''`? Then the next idiom change breaks loudly right here, rather than showing up as a puzzling sync result several assertions later.

client/src/queries/rowan/diffRowanProject.ts
```diff
@@ -37,9 +37,9 @@ ws := WriteStream on: Unicode7 new.
 patches do: [:assoc | | pkg |
   pkg := assoc key.
   assoc value operations do: [:op | | c |
-    c := op class name = 'CypressAddition'
+    c := op class name asSymbol == #CypressAddition
```
MatiasFernandez, line 43: 💭 The *why* for `asSymbol ==` lives only in the tests. Here and at syncProtocol.ts:115, the next reader (or an agent doing a cleanup pass) sees this and a plain `=` looks like a harmless simplification — and on 3.7.x both spellings work, so reverting it would pass on most dev images and only break at a 3.6.x client site again.

Would a one-line comment at each of the two sites help ("a string literal inside a GCI doit is Unicode7 on 3.6.x — never compare an image-derived value against one")? Since this bites every time someone writes a new query builder, it might also be worth recording next to `escapeString` in `queries/util.ts` or in CLAUDE.md, so new generated Smalltalk starts from the safe idiom.

MatiasFernandez (approved):

client/src/sync/syncProtocol.ts
ericwinger: Adopted — and confirmed nil is reachable: `SymbolDictionary new name` is nil, and in-stone (3.6.2) `nil asSymbol` is a doesNotUnderstand (2010) that kills the whole scan, while `nil asString asSymbol` answers #nil and moves on. syncClassBuildExpr now compares `name asString asSymbol == #...`; added a regression guard test pinning the asString hop. (cffe4f74)

client/src/__tests__/exportManager.test.ts
ericwinger: Done — classPayload now throws a descriptive error when it cannot recognize the class-build idiom, instead of returning '' (which read as "class not found" far from the cause). Also had to retarget its regex to the new `name asString asSymbol` idiom in lockstep — exactly the drift you flagged, and the new throw is what would catch the next one. (cffe4f74)

client/src/queries/rowan/diffRowanProject.ts
ericwinger: Added — a one-line note at both production sites (syncProtocol.ts and diffRowanProject.ts) plus a durable block next to `escapeString` in queries/util.ts spelling out the idiom (interned-Symbol compare, `asString` first when the value may be nil, never a plain `= 'literal'`). Left CLAUDE.md out of it per Eric. (cffe4f74)

PR #406, opened 2026-08-11
Fix Windows path-separator assertions in tests

npapagna:
## Summary
- Several test files compared a real `vscode.Uri#fsPath` value against a hardcoded forward-slash literal or a POSIX-only regex, which breaks on Windows: `fsPath` uses `\` separators and lowercases drive letters, neither of which `path.join`/raw string literals reproduce.
- Adds a shared `uriFsPath()` test helper (`client/src/__tests__/support/uri.ts`) that calls the real `Uri.file(...).fsPath` so expected values are built the same way production builds them, then fixes each affected assertion in `databaseManager`, `debuggerPanel`, `seasideServer`, `extentBackupManager`, `systemBrowser`, `rowanLoad`, and `rowanProjectView` tests.
- Along the way, `extentBackupManager.test.ts`'s mock-call assertions were switched from unsafe `as unknown as` casts to `vi.mocked()`, and an implicit argument-count check that got dropped when a `toHaveBeenCalledWith` was replaced by manual destructuring was restored explicitly.

## Test plan
- [x] `npx vitest run` passes for every touched test file (individually verified during review)
- [x] `npm run lint`, `npm run format:check`, and `npx tsc --noEmit` all clean on the changed files
- [x] Verified the restored arity check in `extentBackupManager.test.ts` actually fails against a deliberately broken (3-arg) `copyFile` call before confirming it passes against the real code

### DIFF
--- a/client/src/__tests__/databaseManager.test.ts
+++ b/client/src/__tests__/databaseManager.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, beforeEach, vi } from 'vitest';
+import * as path from 'path';
 
 vi.mock('vscode', () => import('../__mocks__/vscode.js'));
 vi.mock('../wslFs');
@@ -20,6 +21,8 @@ import { SysadminStorage } from '../sysadminStorage';
 import { ProcessManager } from '../processManager';
 
 // ── Helpers ────────────────────────────────────────────────
+
+const DB_DATA_DIR = path.join('/root/db-1', 'data');
 
 function makeDb(overrides?: Partial<GemStoneDatabase['config']>): GemStoneDatabase {
   return {
@@ -56,6 +59,29 @@ function pickItem(index: number): void {
   vi.mocked(vscode.window.showQuickPick).mockImplementation(async (items) => (await items)[index]);
 }
 
+/**
+ * Mirrors `vscode.Uri#fsPath`'s platform-dependent output, for building the
+ * "expected" side of an assertion against a path that came from a
+ * `vscode.Uri`'s `fsPath` getter rather than from `path.join`.
+ *
+ * @remarks
+ * `Uri#fsPath` is a separate implementation from `path.join`. Both swap `/`
+ * for `\` on Windows, but they are not interchangeable in general:
+ * `fsPath` lowercases a leading drive letter and treats a leading `//` as
+ * a network-share authority, neither of which `path.join`/`normalize`
+ * does. Asserting against `path.join`'s output instead of this helper
+ * would happen to match for simple inputs, recreating the same "two
+ * different functions, hope they agree" risk this helper exists to
+ * eliminate — call the exact getter production reads, not a lookalike, so
+ * expected tracks actual by construction, not by luck.
+ *
+ * @param rawPath - The POSIX-style path passed to `vscode.Uri.file(...)` when building the mocked value.
+ * @returns The value `.fsPath` returns for that URI on this platform.
+ */
+function uriFsPath(rawPath: string): string {
+  return vscode.Uri.file(rawPath).fsPath;
+}
+
 describe('DatabaseManager.replaceExtent', () => {
   beforeEach(() => {
     vi.clearAllMocks();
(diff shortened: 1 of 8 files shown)


client/src/__tests__/extentBackupManager.test.ts
```diff
@@ -70,12 +69,19 @@ describe('runOnlineExtentBackup', () => {
     const ok = await runOnlineExtentBackup(deps);
 
     expect(ok).toBe(true);
-    expect(deps.copyFile).toHaveBeenCalledWith(
-      '/db/data/extent0.dbf',
-      expect.stringMatching(/\/chosen\/gs64stone_extents_[\d-]+_[\d-]+\/extent0\.dbf$/),
+    const copyFile = vi.mocked(deps.copyFile);
+    expect(copyFile.mock.calls[0]).toHaveLength(2);
+    const [src, dest] = copyFile.mock.calls[0];
+    expect(src).toBe('/db/data/extent0.dbf');
```
MatiasFernandez: 🚫 I think this one still breaks on Windows. `src` comes from `path.join(deps.dataDir, f)` in `extentBackupManager.ts:119`, so on Windows the actual value is `\db\data\extent0.dbf` while this expects `/db/data/extent0.dbf`. The sibling test at line 172 got the `path.join('/db/data', 'extent0.dbf')` treatment for the same value — should this one match?

client/src/__tests__/extentBackupManager.test.ts
```diff
@@ -70,12 +69,19 @@ describe('runOnlineExtentBackup', () => {
...
+    // path.basename/dirname parse whatever separator path.join used to build
+    // `dest`, so decomposing it this way stays correct on every platform,
+    // unlike a regex with a hardcoded `/`, which only matches on POSIX.
+    expect(path.basename(dest)).toBe('extent0.dbf');
+    expect(path.basename(path.dirname(dest))).toMatch(
+      /^gs64stone_extents_\d{4}-\d{2}-\d{2}_\d{2}-\d{2}-\d{2}$/,
     );
-    const copyOrder = (deps.copyFile as unknown as { mock: { invocationCallOrder: number[] } }).mock
-      .invocationCallOrder[0];
+    expect(path.dirname(path.dirname(dest))).toBe(path.join('/chosen'));
```
MatiasFernandez: 💭 `dest`'s root here traces back to `showOpenDialog`, which the `beforeEach` mocks as a hand-built `{ fsPath: '/chosen' } as vscode.Uri` — so production never sees a value a real `Uri` would produce, and the expectation has to use `path.join('/chosen')` (a single-arg `path.join`, i.e. `normalize`) to compensate. That's the "two different functions, hope they agree" pattern your `uriFsPath` docstring argues against, just moved into the mock.

Would it be cleaner to mock the dialog with a real `vscode.Uri.file('/chosen')` and assert `uriFsPath('/chosen')` here? Then the fixture and the expectation both go through the same getter production reads, and it stays consistent with how `databaseManager.test.ts` mocks its dialogs.

client/src/__tests__/extentBackupManager.test.ts
```diff
@@ -70,12 +69,19 @@ describe('runOnlineExtentBackup', () => {
     const ok = await runOnlineExtentBackup(deps);
 
     expect(ok).toBe(true);
-    expect(deps.copyFile).toHaveBeenCalledWith(
-      '/db/data/extent0.dbf',
-      expect.stringMatching(/\/chosen\/gs64stone_extents_[\d-]+_[\d-]+\/extent0\.dbf$/),
+    const copyFile = vi.mocked(deps.copyFile);
+    expect(copyFile.mock.calls[0]).toHaveLength(2);
```
MatiasFernandez, line 76: nit: the comment below explains the `basename`/`dirname` decomposition nicely, but this line reads a bit cryptic on its own — it's an arity check standing in for what `toHaveBeenCalledWith` used to enforce implicitly. Worth a short `// copyFile takes exactly (src, dest) — ...` so a future reader doesn't delete it as redundant?

client/src/__tests__/support/uri.ts
```diff
@@ -0,0 +1,24 @@
...
+ * for `\` on Windows, but they are not interchangeable in general:
+ * `fsPath` lowercases a leading drive letter and treats a leading `//` as
+ * a network-share authority, neither of which `path.join`/`normalize`
+ * does. Asserting against `path.join`'s output instead of this helper
+ * would happen to match for simple inputs, recreating the same "two
+ * different functions, hope they agree" risk this helper exists to
+ * eliminate — call the exact getter production reads, not a lookalike, so
+ * expected tracks actual by construction, not by luck.
+ *
+ * @param rawPath - The POSIX-style path passed to `vscode.Uri.file(...)` when building the mocked value.
```
MatiasFernandez: 💭 Two things I'd add here, since this is a new shared helper and the docstring is otherwise doing real teaching work:

- `rowanProjectView.test.ts:140` passes `path.join(dir, 'rowan', ...)` where `dir` is an `mkdtempSync` path — so on Windows the argument arrives with `\` separators and a drive letter, not "POSIX-style". `Uri.file` handles both, so the call is right; the contract just reads narrower than the actual usage. Worth widening it to something like "a filesystem path, in either separator style"?
- The `import { Uri } from 'vscode'` at the top only resolves once the *importing* test file has hoisted a `vi.mock('vscode', ...)` — there's no global alias in `client/vitest.config.ts`. Could we note that in the docstring? It'd save the next person (or agent) a confusing module-resolution failure when they reach for this helper from a file that doesn't mock vscode.

client/src/__tests__/databaseManager.test.ts
```diff
@@ -77,14 +79,17 @@ describe('DatabaseManager.replaceExtent', () => {
     expect(ok).toBe(true);
     expect(vscode.window.showOpenDialog).toHaveBeenCalledTimes(1);
     expect(wslImportFileSync).toHaveBeenCalledWith(
-      '/seed/mydata.dbf',
-      '/root/db-1/data/extent0.dbf',
+      uriFsPath('/seed/mydata.dbf'),
+      path.join('/root/db-1', 'data', 'extent0.dbf'),
```
MatiasFernandez: 💭 The `path.join` granularity is mixed across this file — `path.join('/root/db-1', 'data', ...)` here, but `path.join('/root', 'db-1', 'conf', ...)` at line 230 and `path.join('/gs', 'data', 'system.conf')` at line 229 versus `path.join('/gs/bin', ...)`-style splits elsewhere. They're all equivalent (`join` normalizes embedded `/` too), but the inconsistency suggests the split points are meaningful when they aren't.

Since `path.join('/root/db-1', 'data', ...)` shows up five times in this describe block, would a `const DB_DATA_DIR = path.join('/root/db-1', 'data')` near the existing helpers tidy this up and settle the style question at the same time? `systemBrowser.test.ts` already does this with `SESSION_ROOT`.

MatiasFernandez (approved):

PR #407, opened 2026-08-11
Fix Windows CI failures caused by CRLF checkout and path separators

npapagna:
There are currently three genuine Windows OS semantics the tests didn't account for.

Root cause 1: git's core.autocrlf on Windows rewrites LF to CRLF.
No .gitattributes existed to prevent this, so:
- jsonHelperDedup.test.ts's hand-rolled parser scanned checked-out gs-src/**/*.class.st files for an exact ] line to close a method body. A CRLF'd ]\r line never matched, so the scan ran to EOF and silently mis-parsed the file, undercounting definitions (3 failing assertions).
- rowanLoad.test.ts's updateGitRepo suite writes \n-terminated content, commits it, and reads it back after a real git clone/checkout round trip through its own throwaway fixture repos. The checkout re-materialized it as \r\n, failing a hardcoded .toBe('...\n').

Root cause 2: a backslash is a Windows path separator, not a valid filename character.
rowanCreate.test.ts tested a project name containing a literal \, expecting it to produce a file literally named a\b.ston. That's impossible on Windows by design (Win32 reserves \ in filenames outright), so the scenario can never pass there. It's also unreachable in production: every real caller of createRowanProject already blocks \ and / in the name before it gets here.

Root cause 3: a client-platform path.join isn't the same thing as the stone's own path.
extentBackupManager.test.ts asserted the extent path handed to copyFile against path.join('/db/data', 'extent0.dbf'), which renders backslashes on Windows. The production code keeps that path verbatim, since it's the remote GemStone stone's own POSIX path (the stone always runs on Linux/WSL), not a value ever run through the client's path module.

Changes

- Added .gitattributes (* text=auto eol=lf) so checkout normalizes text files to LF everywhere, fixing jsonHelperDedup.test.ts at the source with no test changes needed.
- rowanLoad.test.ts: added its own .gitattributes to the throwaway seed repo its setup() builds, so every clone made from it (remote, clone, each pusher) checks out as LF regardless of the host's core.autocrlf. A core.autocrlf override on the test's own git commands wouldn't have worked here, since updateGitRepo's internal checkout runs as a separate git process that doesn't inherit it.
- rowanCreate.test.ts: gated the backslash-in-project-name test to POSIX only (onSupportedPosixIt), since the scenario is impossible on Windows and unreachable through any real caller.
- extentBackupManager.test.ts: compare against path.posix.join instead of path.join, matching what the code under test actually returns.

No production code changed.

Test plan

- [x] npm run lint && npm run format:check && npm run compile pass
- [x] All four touched test files pass locally (macOS/LF, where the fixes are no-ops, confirming no regression)
- [x] Verified against a real Windows run on the throwaway windows-health-check.yml POC branch (not part of this PR): failures dropped from 8 across 5 files to 2, both in localVersion.test.ts, which this PR doesn't touch
- [ ] localVersion.test.ts's 2 remaining Windows failures (a test/production platform-key mismatch, unrelated to CRLF or path separators) are tracked separately and not fixed by this PR

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# Windows CI checks out CRLF by default; several tests parse repo files
+# line-exactly and assume LF.
+* text=auto eol=lf
--- a/client/src/__tests__/extentBackupManager.test.ts
+++ b/client/src/__tests__/extentBackupManager.test.ts
@@ -75,7 +75,11 @@ describe('runOnlineExtentBackup', () => {
     // enforce that implicitly; destructuring below doesn't, so check it here.
     expect(copyFile.mock.calls[0]).toHaveLength(2);
     const [src, dest] = copyFile.mock.calls[0];
-    expect(src).toBe(path.join('/db/data', 'extent0.dbf'));
+    // The stone reports its own extent paths verbatim (see extentFolderInServer's
+    // doc comment) -- POSIX since the server is never Windows, regardless of
+    // the client's OS -- so this literal is exactly what the mock returned,
+    // not a path.join reassembly.
+    expect(src).toBe('/db/data/extent0.dbf');
     // path.basename/dirname parse whatever separator path.join used to build
     // `dest`, so decomposing it this way stays correct on every platform,
     // unlike a regex with a hardcoded `/`, which only matches on POSIX.
--- a/client/src/__tests__/rowanCreate.test.ts
+++ b/client/src/__tests__/rowanCreate.test.ts
@@ -3,6 +3,7 @@ import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
 import { createRowanProject } from '../rowanCreate';
+import { onSupportedPosixIt } from './platformGates';
 
 const dirs: string[] = [];
 function tmp(): string {
@@ -117,7 +118,13 @@ describe('createRowanProject', () => {
     expect(spec).toContain("#projectName : 'O\\'Hara',");
   });
 
-  it('backslash-escapes a backslash in the project name', () => {
+  // A backslash is a plain filename character on POSIX but a path separator
+  // on Windows, so a name containing one can't be used as a spec's filename
+  // there. Not a real-world gap: every caller of createRowanProject already
+  // rejects '\' and '/' in the name before it gets here (see
+  // extension.ts's rowanNewProject input validation and rowanInitHere's use
+  // of path.basename).
+  onSupportedPosixIt('backslash-escapes a backslash in the project name', () => {
     const dir = tmp();
 
     createRowanProject(dir, 'a\\b');
--- a/client/src/__tests__/rowanLoad.test.ts
+++ b/client/src/__tests__/rowanLoad.test.ts
@@ -178,6 +178,12 @@ describe('updateGitRepo', () => {
     g(['init', '-q', '-b', 'main'], seed);
     g(['config', 'user.email', 't@t'], seed);
     g(['config', 'user.name', 'T'], seed);
+    // Force LF checkouts regardless of the host's core.autocrlf: this file is
+    // tracked, so it travels with every clone (remote, clone, each pusher)
+    // and applies no matter which process performs the checkout, unlike a
+    // core.autocrlf override on the git commands here, which wouldn't reach
+    // updateGitRepo's own checkout.
+    fs.writeFileSync(path.join(seed, '.gitattributes'), '* text=auto eol=lf\n');
     fs.writeFileSync(path.join(seed, 'a.txt'), 'one\n');
     g(['add', '-A'], seed);
     g(['commit', '-qm', 'one'], seed);
--- a/client/src/__tests__/platformGates.ts
+++ b/client/src/__tests__/platformGates.ts
@@ -4,13 +4,13 @@ const isSupportedPosixPlatform = process.platform === 'linux' || process.platfor
 
 /**
  * Wraps a `describe` block that only makes sense on a POSIX platform we've
- * validated socket behavior on (Linux/macOS), e.g. behavior backed by real
- * Unix domain sockets or filesystem paths that have no equivalent on Windows
- * named pipes. This is an allow-list, not a POSIX-detection check: other
- * POSIX platforms (freebsd, aix, sunos, ...) are deliberately excluded
- * because we haven't validated socket behavior there, not because they
- * aren't POSIX. Runs the block's tests normally on Linux/macOS; reports them
- * as skipped elsewhere.
+ * validated on (Linux/macOS): behavior backed by real Unix domain sockets or
+ * filesystem paths that have no equivalent on Windows (named pipes instead of
+ * sockets; `\` as a path separator instead of a legal filename character).
+ * This is an allow-list, not a POSIX-detection check: other POSIX platforms
+ * (freebsd, aix, sunos, ...) are deliberately excluded because we haven't
+ * validated this behavior there, not because they aren't POSIX. Runs the
+ * block's tests normally on Linux/macOS; reports them as skipped elsewhere.
  */
 export const onSupportedPosixDescribe: ReturnType<typeof describe.runIf> =
   describe.runIf(isSupportedPosixPlatform);
--- a/client/src/refactoring/__tests__/jsonHelperDedup.test.ts
+++ b/client/src/refactoring/__tests__/jsonHelperDedup.test.ts
@@ -41,13 +41,18 @@ const headerRe = /^(.+?) >> (jsonEscape:|jsonQuote:|hex2:) (\w+) \[\s*$/;
 function collectDefs(): Def[] {
   const defs: Def[] = [];
   for (const f of fs.readdirSync(engineDir).filter((n) => n.endsWith('.class.st'))) {
-    const lines = fs.readFileSync(path.join(engineDir, f), 'utf8').split('\n');
+    // Tolerate CRLF here rather than relying on .gitattributes' eol=lf: that
+    // normalizes fresh checkouts, but not an already-checked-out working
+    // copy, a hand-edited file, or a later narrowing of the pattern. A plain
+    // split('\n')/'!== ]' would then undercount definitions instead of
+    // failing loudly.
+    const lines = fs.readFileSync(path.join(engineDir, f), 'utf8').split(/\r?\n/);
     for (let i = 0; i < lines.length; i++) {
       const m = headerRe.exec(lines[i]);
       if (!m) continue;
       const bodyLines: string[] = [];
       let j = i + 1;
-      for (; j < lines.length && lines[j] !== ']'; j++) bodyLines.push(lines[j]);
+      for (; j < lines.length && lines[j].trim() !== ']'; j++) bodyLines.push(lines[j]);
       const body = bodyLines.join('\n').trim();
       defs.push({
         file: f,


client/src/rowanLoad.ts
```diff
@@ -145,11 +145,20 @@ export function cloneGitRepo(url: string, dest: string): Promise<void> {
   });
 }
 
-function git(args: string[], cwd: string, timeoutMs = 300_000): Promise<string> {
+// A checked-out repo's content must match what's committed byte-for-byte,
+// since callers (e.g. updateGitRepo) read files back afterward. Without this,
+// Windows' git default rewrites LF to CRLF on checkout.
+const GIT_NO_AUTOCRLF_ARGS = ['-c', 'core.autocrlf=false'] as const;
```
MatiasFernandez: 🚫 The comment promises that "a checked-out repo's content must match what's committed byte-for-byte", but `cloneGitRepo` (right above) still spawns git directly and never gets these args. On Windows the *first* checkout of a tracked Rowan project is the clone, so a freshly cloned repo lands with CRLF, and only later `updateGitRepo` pulls get LF. That leaves a repo whose files change line endings depending on how they got there.

Should `cloneGitRepo` pass `GIT_NO_AUTOCRLF_ARGS` too? It can't reuse `git()` as-is (no `-C` working copy yet), but `['clone', ...]` → `[...GIT_NO_AUTOCRLF_ARGS, 'clone', ...]` would keep the guarantee consistent.

client/src/rowanLoad.ts
```diff
@@ -145,11 +145,20 @@ export function cloneGitRepo(url: string, dest: string): Promise<void> {
...
+
+// Run git against an arbitrary working copy, stripped of ambient git env and
+// with autocrlf disabled (see above and gitEnv()). Exported so tests needing
+// the same throwaway-fixture-repo guarantees can reuse it instead of
+// duplicating this setup.
```
MatiasFernandez: 💭 The PR description says "No production code changed", but this does change production behavior: every `updateGitRepo` call now overrides the user's own `core.autocrlf` on *their* working copy, not just on throwaway test fixtures. For a Windows user who deliberately runs `autocrlf=true`, Jasper's pull will now materialize LF files into a worktree that the rest of their tooling expects to be CRLF.

Is that the intended contract — Jasper-tracked clones are always LF regardless of user config? If so, worth saying that explicitly here (and in the PR body), since it's a decision future readers will want the rationale for. If the real requirement is just "the bytes we read back match the bytes we committed", another option is leaving checkout alone and normalizing on read.

client/src/rowanLoad.ts
```diff
@@ -145,11 +145,20 @@ export function cloneGitRepo(url: string, dest: string): Promise<void> {
...
+export function git(args: string[], cwd: string, timeoutMs = 300_000): Promise<string> {
```
MatiasFernandez: nit: now that this is exported, `git` is a very generic name at the import site — `import { ..., git } from '../rowanLoad'` reads like the library, not like "run a git subprocess in this cwd". Would `runGit` be clearer?

client/src/__tests__/rowanCreate.test.ts
```diff
@@ -117,7 +118,13 @@ describe('createRowanProject', () => {
...
   });
 
-  it('backslash-escapes a backslash in the project name', () => {
+  // A backslash is a plain filename character on POSIX but a path separator
+  // on Windows, so a name containing one can't be used as a spec's filename
+  // there. Not a real-world gap: every caller of createRowanProject already
+  // rejects '\' and '/' in the name before it gets here (see
+  // extension.ts's rowanNewProject input validation and rowanInitHere's use
+  // of path.basename).
+  onSupportedPosixIt('backslash-escapes a backslash in the project name', () => {
```
MatiasFernandez, line 127: 💭 `onSupportedPosixIt`'s doc frames the allow-list as being about *socket* behavior we've validated on Linux/macOS ("other POSIX platforms are deliberately excluded because we haven't validated socket behavior there"). This test has nothing to do with sockets — it's about `\` being a legal filename character, which is true on every POSIX platform including the excluded ones.

The gate still produces the right outcome, so this isn't blocking, but should the helper's TSdoc be broadened to cover "POSIX filesystem semantics" as well? Otherwise the next reader hits a rationale that doesn't match the use.

MatiasFernandez (approved):

.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto eol=lf
```
MatiasFernandez, line 3: 💭 Could we add a `#` comment line above this? A one-liner repo-wide `.gitattributes` with no rationale is the kind of thing a future reader (or agent) can't evaluate — it silently changes checkout behaviour for everyone, and the reason it exists (test scanners under `client/src` assume LF, and Windows CI checks out CRLF without it) lives only in this PR's description. Something like `# Windows CI checks out CRLF by default; several tests parse repo files line-exactly and assume LF.` would keep it from being "cleaned up" later.

MatiasFernandez, line 3: 💭 This fixes `jsonHelperDedup.test.ts` at checkout time, but the test's own parser is still the fragile part: it splits on `'\n'` and closes a method body on `lines[j] !== ']'` (`client/src/refactoring/__tests__/jsonHelperDedup.test.ts`, lines 44 and 50). With CRLF it doesn't fail loudly, it scans to EOF and *undercounts* definitions — which is exactly the failure mode of a structural guard silently going soft.

Would it be worth also making the scan CRLF-tolerant (`split(/\r?\n/)`, or comparing `lines[j].trim()`)? It's a couple of characters, and it means the guard can't be disarmed by a checkout setting, a hand-edited file, or someone narrowing this `.gitattributes` later.

client/src/__tests__/extentBackupManager.test.ts
```diff
@@ -75,7 +75,11 @@ describe('runOnlineExtentBackup', () => {
     // enforce that implicitly; destructuring below doesn't, so check it here.
     expect(copyFile.mock.calls[0]).toHaveLength(2);
     const [src, dest] = copyFile.mock.calls[0];
-    expect(src).toBe(path.join('/db/data', 'extent0.dbf'));
+    // The stone reports its own extent paths verbatim (see extentFileNames'
+    // doc comment). It's always POSIX, since the server runs on Linux/WSL
+    // regardless of the client's OS, so compare with path.posix rather than
+    // the client-platform path.join, which would render backslashes on Windows.
+    expect(src).toBe(path.posix.join('/db/data', 'extent0.dbf'));
```
MatiasFernandez: 💭 The comment points at `extentFileNames`' doc for the "verbatim, always POSIX" claim, but that doc only says *"The stone's extent file paths (absolute, as the stone sees them)"* — nothing about the stone always being Linux/WSL, which is the part the assertion actually leans on. Should the POSIX note go there instead (or as well)? That's where the next caller — say, one that tries to `path.join` the result on a Windows client — will look for the contract.

MatiasFernandez: nit: since the fake stone hands back the literal `'/db/data/extent0.dbf'` and the point is that the code passes it through untouched, would `expect(src).toBe('/db/data/extent0.dbf')` say it more directly? `path.posix.join` reassembles the exact string the mock returned, so it reads as if some joining were expected here.

client/src/__tests__/rowanCreate.test.ts
MatiasFernandez, line 127: 💭 Agreed the *filename* half can't work on Windows, but the assertion this test actually makes is about STON escaping (`#projectName : 'a\b'`), which is platform-independent — so gating the whole `it` drops that escaping rule from Windows coverage entirely, and `\` is the one escape the sibling `O'Hara` test doesn't cover.

Two options depending on how you read it: if the escaping is worth pinning everywhere, testing the escaper directly (rather than through `createRowanProject`'s file write) would keep it green on every platform; if the rule is really "no caller can ever pass a `\`", then per your own comment the case is unreachable and maybe the test should just go, rather than living on as POSIX-only. What do you think?

client/src/__tests__/rowanLoad.test.ts
```diff
@@ -178,6 +178,12 @@ describe('updateGitRepo', () => {
     g(['init', '-q', '-b', 'main'], seed);
     g(['config', 'user.email', 't@t'], seed);
     g(['config', 'user.name', 'T'], seed);
+    // Force LF checkouts regardless of the host's core.autocrlf: this file is
+    // tracked, so it travels with every clone (remote, clone, each pusher)
+    // and applies no matter which process performs the checkout, unlike a
+    // core.autocrlf override on the git commands here, which wouldn't reach
+    // updateGitRepo's own checkout.
+    fs.writeFileSync(path.join(seed, '.gitattributes'), '* text=auto eol=lf\n');
```
MatiasFernandez, line 186: 💭 Just to close the loop on my earlier comment about production line endings: with the fixture forcing LF, this test now guarantees LF by construction, so nothing in the suite says anything about what a *real* Windows user's Rowan clone gets — with `core.autocrlf=true` (the Windows default) their checked-out `.st` files will be CRLF.

Happy for that to be out of scope here (this PR is about CI), but do we know CRLF working copies are harmless on the Rowan load path, or is it untested territory? If it's the latter, worth an issue so the pivot away from touching `updateGitRepo` doesn't quietly close the question.

MatiasFernandez (approved):

.gitattributes
npapagna: Added a rationale comment above the `.gitattributes` line: `# Windows CI checks out CRLF by default; several tests parse repo files line-exactly and assume LF.` Folded into the `Force LF checkout for text files on Windows` commit.

npapagna: Good catch, the guard could go soft without ever failing loudly. Made the scanner CRLF-tolerant: `split(/\r?\n/)` instead of `split('\n')`, and `lines[j].trim() !== ']'` instead of the exact comparison, with a comment explaining why this shouldn't lean solely on `.gitattributes` staying correct. Folded into the same commit.

client/src/__tests__/extentBackupManager.test.ts
npapagna: You're right that `extentFileNames`'s own doc makes no POSIX claim, so there was nothing to move there. The POSIX contract is already documented on `extentFolderInServer` (and `backupManager.ts`), so I pointed the comment there instead. Also caught that 'the server runs on Linux/WSL' was inaccurate, WSL is how a Windows client reaches the server, not a server OS, so reworded it to match the 'never Windows, so always POSIX' phrasing used everywhere else in the codebase. Folded into `Fix extent path assertion for Windows CI`.

npapagna: Applied, changed to `expect(src).toBe('/db/data/extent0.dbf')`. Folded into the same commit.

client/src/__tests__/rowanCreate.test.ts
npapagna: Agreed this is a real gap, but keeping it out of scope for this CI-fix PR. Filed as a follow-up: https://github.com/GemTalk/Jasper/issues/418

client/src/__tests__/rowanLoad.test.ts
npapagna: Agreed this needs answering too, but separately from this PR. Filed as a follow-up: https://github.com/GemTalk/Jasper/issues/419

PR #411, opened 2026-08-11
Report post-rename recompile failures, and make them reachable

ericwinger:
## What

A rename that leaves methods un-recompiled now says which ones — and, for three of the four rename
refactorings, notices that it happened at all.

Two halves:

**1. The failure list survives.** The apply toast could only ever name `failed[0]`, and then it
vanished. The full list now goes to the persistent **GemStone GCI** output channel, headed by the
rename it belongs to, and the toast carries a **Show Details** button that reveals the channel —
the same `'Show Details'` idiom `logJasperError` already uses in `extension.ts`.

**2. Three engines could not report a failure at all.** `Behavior>>compileMethod:dictionaries:category:`
**answers an Array of error tuples instead of raising**. An unchecked send therefore reads as
success: the method is never installed, but the apply counts it as applied and the user is told
nothing. Only `GsRenameInstanceVariableRefactoring` inspected the result. Verified live on 3.7.5:

```smalltalk
(SomeClass compileMethod: 'bar ^MissingGlobal' dictionaries: System myUserProfile symbolList category: 'p')
  "→ anArray( anArray( 1031, 6, 'undefined symbol', nil, 'MissingGlobal'))  — no exception raised"
```

What that cost, per engine:

| Engine | Silent consequence |
|---|---|
| `GsRenameClassVariableRefactoring>>applyMethodRecompile:` | method left bound to the OLD, now-detached association — the dangling reference its own all-or-nothing comment says the design prevents |
| `GsRenameMethodRefactoring>>applyChange:` | **method loss** — the old selector is removed *after* the compile, so a failed compile deleted the method and installed nothing |
| `GsRenameClassRefactoring>>applyMethodRecompile:` | external referencer left naming a class that no longer exists |

New `GsRefactoringEnvironment>>compile:into:category:` checks the result and raises, putting the
failure onto the `on: Error do:` guard each apply loop already has — so it lands in `failed` and is
reported like everything else.

## Testing

- **[GS SUnit]** one test per fixed engine, each building a method compiled against a scaffold
  global that is then removed (error 1031 on recompile, while the installed method still runs).
  **Negative-controlled:** reverting the engine and rebuilding fails exactly one test per suite
  (31→30, 16→15, 20→19), so none of them are vacuous.
  **100/100 on BOTH 3.6.2 and 3.7.5** across the four rename suites + `GsRefactoringEnvironmentTest`.
- **[client .ts]** `explorerRenameFailureLog.test.ts` drives `renameMethodNamed` end to end:
  every failure reaches the channel, the block carries the action label, **Show Details** reveals
  the channel, dismissing leaves it alone, a clean apply logs nothing. Plus formatter unit tests.
  Full client suite **5312 passed**, 12 skipped.
- **[GCI auto]** included in the client suite above (engine reinstalled on the test stone first).
- **F5** verified on `gs64stone_375` against a purpose-built fixture: renaming an instance variable
  with 5 un-recompilable methods lists all five (including the `Class class>>sel` form), the button
  opens the channel, two renames in one session stay distinguishable, a clean rename logs nothing,
  and the class-variable rename — which previously reported plain success — now reports the failure.

## Not in this PR — tracked in #412

- `GsRenameClassRefactoring>>copyMethod:from:to:` (the class re-version copy-forward) is unchecked
  too. Raising there would abort a whole class rename over one bad method; it needs R1's
  record-and-continue treatment, which means threading a failures collection through
  `copyMethodsFrom:to:`. Separate change.
- The three fixed engines label a failure with `change className`, so the log names a class where
  R1 names the method. Cosmetic, but it undercuts "here is what to go fix".

Both are filed as #412 so neither is lost.

Part of #252.


client/src/refactoring/renameFailureLog.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export function formatRenameFailureLog(
+  action: string,
+  failed: RenameApplyResult['failed'],
+): string | undefined {
+  if (failed.length === 0) return undefined;
+  return (
+    `${action}: ${failed.length} method(s) did not recompile onto the new class version ` +
+    '(compiled but NOT committed — abort if this was not intended):\n' +
+    failed.map((f) => `    • ${f.label}: ${f.error}`).join('\n')
```
MatiasFernandez, line 22: 💭 I know #412 tracks this, but I want to make sure we've weighed what it does to *this* feature before it ships. `applyDeselected:` builds each failure's label from `change className`, so for the three engines fixed here the log renders one line per failure that all say the same thing:

```
Rename class variable 'Counter' → 'Tally': 3 method(s) did not recompile ...
    • GsRCVBase: did not recompile: undefined symbol (GsRCVGone)
    • GsRCVBase: did not recompile: undefined symbol (GsRCVGone)
    • GsRCVBase: did not recompile: undefined symbol (GsRCVGone)
```

The whole pitch of the durable channel is "here is the complete set to go fix" — but three identical lines naming a class don't tell the user which three methods to open, which is roughly where they were with the toast. The PR body calls this cosmetic; I'd argue it's the difference between the list being actionable and being a count.

Is threading the selector into the failure tuple bigger than it looks? From what I can see `applyChange:`/`applyMethodRecompile:` already have `aChange` in hand, so it reads like adding the selector to the `Array with: change id with: change className with: e messageText` triple and to the label the JSON emits. If it's genuinely bigger, no objection to deferring — but then should the follow-up land before we tell users about the channel?

client/src/gemstoneExplorer.ts
```diff
@@ -2332,7 +2364,8 @@ export class ExplorerController {
 
     if (result.failed.length > 0) {
       const first = result.failed[0];
-      void vscode.window.showErrorMessage(
+      this.logRenameFailures(`Rename method '${oldSelector}' → '${newSelector}'`, result.failed);
+      this.showRenameFailureToast(
         `Rename applied ${result.applied} change(s); ${result.failed.length} failed: ` +
```
MatiasFernandez: 💭 Now that the log block has a formatter of its own, should the toast message get the same treatment? This exact template is repeated character-for-character at three call sites (method, class, class variable), and the instance-variable one at ~2211 is a fourth variant — the only one that carries the `Compiled but NOT committed — abort if this is not what you wanted.` caveat, even though `formatRenameFailureLog` puts that caveat in *every* log block.

A `formatRenameFailureToast(action, result)` sitting next to `formatRenameFailureLog` in `renameFailureLog.ts` would take the same two inputs the log already takes, drop the triplication, and make it hard for the toast and the log to disagree about what they tell the user. Worth doing while the call sites are already being touched?

client/src/gemstoneExplorer.ts
```diff
@@ -2124,6 +2129,29 @@ export class ExplorerController {
...
+  // After a rename apply reports methods that could not be recompiled onto the new
+  // class version, list ALL of them in the persistent "GemStone GCI" output channel.
+  // The apply toast can only name the first (a notification truncates and vanishes);
+  // this gives the user the complete set to go fix, per the rename-family hardening
+  // goal of a durable warning surface rather than a disappearing one. `action` must
+  // name WHAT was renamed — the channel is durable and shared, so two renames in a
+  // session otherwise leave two indistinguishable blocks.
+  private logRenameFailures(action: string, failed: RenameApplyResult['failed']): void {
+    const line = formatRenameFailureLog(action, failed);
+    if (line) logInfo(line);
```
MatiasFernandez: 💭 Should this go through `logError` rather than `logInfo`? Two things it costs us in a channel that's shared with ordinary GCI traffic:

- No `ERROR:` marker. The user opens the channel from a red toast and the block reads like routine chatter — nothing scannable distinguishes it from the surrounding log lines.
- `logInfo` doesn't append the trailing blank line `logError` does, so two rename failures in one session butt directly against each other (and against whatever GCI logged in between). The action header keeps them *parseable*, but the visual separation the comment above is aiming for isn't there.

I see the catch: `logError` wants a `sessionId` and this call site doesn't obviously have one. If plumbing the session through isn't worth it, would a `logWarning(message)` in `gciLog.ts` — same shape as `logInfo` plus a marker and the trailing blank line — be the cheaper path?

gs-src/refactoring/engine/GsRefactoringEnvironment.class.st
```diff
@@ -305,6 +305,41 @@ GsRefactoringEnvironment >> instanceMethodsShadowing: anInstVarName inClass: aCl
...
+	 recorded in `failed` and reported like any other."
+	| result |
+	result := aBehavior
+		compileMethod: source
+		dictionaries: System myUserProfile symbolList
+		category: aCategory.
+	result isNil ifTrue: [^self].
+	(result isKindOf: Array)
+		ifFalse: [^self error: 'did not recompile'].
+	result isEmpty ifTrue: [^self].
```
MatiasFernandez, line 328: 💭 This branch throws away the one piece of information we have. `compileErrorTextFrom:` already opens with `(aTuple isKindOf: Array) ifFalse: [^aTuple printString]` — it's built to handle exactly the unexpected shape this guard is catching — but here we raise a bare `'did not recompile'` instead of asking it.

Should this be `^self error: 'did not recompile: ', (self compileErrorTextFrom: result)`? If `compileMethod:` ever answers something we didn't anticipate, that's precisely the case where the user (and whoever debugs it) most needs to see what came back, and the fallback is already written.

(Same line in the generated `resources/refactoring/engine.gs`, naturally.)

gs-src/refactoring/tests/GsRenameClassRefactoringTest.class.st
```diff
@@ -194,6 +194,25 @@ GsRenameClassRefactoringTest >> testNewNameCollisionDetected [
...
+	| json |
+	UserGlobals at: #GsRCGone put: 7.
+	self compile: 'usesBase "makes a GsRCBase" ^(Array with: GsRCBase new) , GsRCGone'
+		in: (UserGlobals at: #GsRCOther).
+	UserGlobals removeKey: #GsRCGone.
+
+	json := (self renameTo: 'GsRCRenamed' scope: #wholeSystem) applyDeselected: #().
+
+	self deny: json includesSubstring: '"failed":[]'.
+	self assert: json includesSubstring: 'did not recompile'
```
MatiasFernandez: 💭 These two assertions don't pin down *which* change failed. `#wholeSystem` scope stages a change per referencing method, so "some change failed with 'did not recompile'" would still hold if `usesBase` recompiled fine and something else in the fixture broke — the test would go green for the wrong reason, and the negative control wouldn't catch it (reverting the engine also stops reporting the unrelated failure).

The sibling tests both do better: the class-variable one adds `'"applied":8'` to pin that exactly one of nine failed, and the method one asserts `caller` and `moveY:x:` are still installed. Could this one assert the failure entry names `GsRCOther` (or `usesBase` once #412 lands), and/or the applied count?

client/src/__tests__/explorerRenameFailureLog.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    await Promise.resolve();
+
+    expect(gciChannel.show).toHaveBeenCalledWith(true);
+  });
+
+  it('leaves the channel alone when the user dismisses the toast', async () => {
+    applyReturns([{ id: 'a', label: 'Account>>balance', error: 'undeclared variable' }]);
+
+    await renameMethod(makeController());
+    await Promise.resolve();
```
MatiasFernandez, line 125: 💭 Is this test able to fail? `showErrorMessage` resolves to `undefined`, so `gciChannel.show` is never called whether or not the `.then` ran at all — drop the `await Promise.resolve()` and it still passes. That means it's asserting "the microtask hasn't run yet" just as happily as "the user dismissed it", and if the `.then` ever stopped firing entirely this is the test that would have to notice.

Would asserting the toast actually went out with the button (`expect(vscode.window.showErrorMessage).toHaveBeenCalledWith(expect.any(String), 'Show Details')`) alongside the `not.toHaveBeenCalled()` tie it down? Then the pair says "we offered the button, the user declined, nothing was revealed" rather than just "nothing happened."

client/src/__tests__/explorerRenameFailureLog.test.ts
```diff
@@ -0,0 +1,145 @@
...
+import type { SessionManager, ActiveSession } from '../sessionManager';
+
+/**
+ * Pins the WIRING between a rename apply that reports un-recompilable methods and the
+ * persistent "GemStone GCI" output channel: the toast can name only the first failure
+ * and then vanishes, so the full set must reach the durable log. `formatRenameFailureLog`
+ * is unit-tested on its own (renameFailureLog.test.ts); this drives the rename-method
+ * call site end to end so the wiring — and the action label that says WHICH rename the
+ * block belongs to — cannot be dropped unnoticed.
+ */
```
MatiasFernandez: nit: This docblock describes the whole file (what's pinned, why the toast isn't enough, where the unit tests live), but sitting here it attaches to `makeController` — so hovering that helper, or an agent reading the symbol, gets the file's rationale as the helper's contract. Could it move above the imports as a file-level comment, or onto the `describe`?

MatiasFernandez (approved):

ericwinger (commented):
Thanks — all seven taken, including the label one, which turned out to be a one-method change rather than a deferral.

The two that changed the shape of the feature:

- **Failure labels now name the method.** `GsRefactoringChange>>failureLabel` renders `Class>>selector` / `Class class>>selector`, so the channel lists what to go open instead of the same class name N times. #412 is narrowed to the remaining 13 engines plus the unchecked class-rename copy-forward.
- **One reporter for log and toast.** `reportRenameFailures(action, result)` builds both from the same inputs, so they can't disagree; the caveat and the applied count are now consistent across all four renames.

Re-verified: RB SUnit **108/108 on both 3.6.2 and 3.7.5**, client suite **5312**.

client/src/refactoring/renameFailureLog.ts
ericwinger: Agreed, and fixed here rather than deferred — you're right that a list of three identical class names is a count, not an artifact.

It was smaller than it looked. `GsRefactoringChange` already carries `selector` and `isMeta`, so this is now a single `failureLabel` on the change itself:

```smalltalk
GsRefactoringChange >> failureLabel
	| cls |
	cls := className asString.
	selector isNil ifTrue: [^cls].
	^isMeta == true
		ifTrue: [cls, ' class>>', selector asString]
		ifFalse: [cls, '>>', selector asString]
```

The three engines this PR touches now build their failure tuple with `change failureLabel` instead of `change className`, so your example reads:

```
    • GsRCVBase>>getIt: did not recompile: undefined symbol (GsRCVGone)
```

Structural changes (a class-definition edit, no selector) still answer the class name, which is the right label for them.

Putting it on the change rather than per-engine means the other 13 engines get it by changing one word each — that part stays in #412, which I've narrowed accordingly.

client/src/gemstoneExplorer.ts
ericwinger: Done, and it collapsed further than a formatter. `formatRenameFailureToast(action, result)` now sits next to `formatRenameFailureLog` and takes the same two inputs, and the four call sites go through one private `reportRenameFailures(action, result)` that does both:

```ts
private reportRenameFailures(action: string, result: RenameApplyResult): void {
  if (result.failed.length === 0) return;
  const block = formatRenameFailureLog(action, result.failed);
  if (block) logWarning(block);
  void vscode.window
    .showErrorMessage(formatRenameFailureToast(action, result), SHOW_RENAME_DETAILS)
    .then((choice) => { if (choice === SHOW_RENAME_DETAILS) getGciLog().show(true); });
}
```

Each site is now one line. The caveat inconsistency you spotted is gone too — every rename's toast carries the not-committed warning, matching what the log block already said, and the toast reports `applied` for all four (the instance-variable one previously didn't).

client/src/gemstoneExplorer.ts
ericwinger: Took the `logWarning` option — plumbing a session id through four call sites to satisfy `logError`'s signature would have been the tail wagging the dog.

```ts
export function logWarning(message: string): void {
  const log = getGciLog();
  log.appendLine(`${stamp()} WARNING: ${message}`);
  log.appendLine('');
}
```

Same shape as `logInfo`, plus the marker and the trailing blank line, so consecutive blocks separate. It also reads more honestly than `logError`: this reports on an operation, not on a session, which is exactly why it has no session id to give.

gs-src/refactoring/engine/GsRefactoringEnvironment.class.st
ericwinger: You're right — that branch was throwing away the one thing worth having, and the fallback was already sitting in `compileErrorTextFrom:`. Fixed:

```smalltalk
(result isKindOf: Array)
	ifFalse: [^self error: 'did not recompile: ', (self compileErrorTextFrom: result)].
```

Payloads rebuilt, so `resources/refactoring/engine.gs` carries it too.

gs-src/refactoring/tests/GsRenameClassRefactoringTest.class.st
ericwinger: Good catch — it would have gone green for the wrong reason, and the negative control wouldn't have caught it. Now that the label names the method (your first comment), it can pin exactly which change failed:

```smalltalk
self assert: json includesSubstring: 'GsRCOther>>usesBase'.
self deny: json includesSubstring: '"applied":0'
```

The first names the failing method rather than any failure; the second pins that the rest of the rename still went through. That brings it in line with the class-variable and method tests.

client/src/__tests__/explorerRenameFailureLog.test.ts
ericwinger: It wasn't able to fail — you're right, dropping the `await` left it passing. Added your suggestion:

```ts
expect(vscode.window.showErrorMessage).toHaveBeenCalledWith(
  expect.any(String),
  'Show Details',
);
expect(gciChannel.show).not.toHaveBeenCalled();
```

and renamed it to `offers the button but reveals nothing when the user dismisses the toast`, so the name states both halves. If the `.then` ever stops firing, the paired press-test fails and this one still asserts the button went out.

client/src/__tests__/explorerRenameFailureLog.test.ts
ericwinger: Moved it above the imports as a file-level comment. It was describing the file, so attaching it to `makeController` was misleading for exactly the reason you give.

PR #414, opened 2026-08-11
C3: one apply-result parser for the whole refactoring family

ericwinger:
## What

Stage C of the RB catalog is the de-duplication pass over the refactoring family
(C1 = engine JSON helpers, C2 = the move↔push relocation abstraction, both merged in
#361). C3 is the client half.

Every refactoring's server-side apply returns the same envelope — `applied`, `failed[]`,
optional `error` — and 14 families each carried their own copy of the parser for it.

**First slice (Aug 7):** extracted `previewEnvelope.ts` with the canonical `ApplyResult` +
`parseApplyResult`, and migrated the 8 byte-identical simple families onto it
(extractMethod, extractTemporary, inlineMethod, inlineTemporary, renameTemporary,
renameClassVar, renameMethod, changeSignature). Each re-exports, so call sites and their
tests are untouched.

**This PR completes it.** Six families still had their own copy, and V8 (split class)
landed after the first slice with a fresh one — the duplication was regrowing faster than
it was being removed. Added:

```ts
parseApplyResultWith(json, extend)
```

The shared parser validates the envelope and reads `applied`/`failed`/`error` once; a
family contributes only its extra fields. Same shape `makeParseChange` already uses for
relocation changes (C2).

- `methodRelocation` re-exports it — its copy was byte-identical
- `splitClass`, `renameClass`, `extractSuperclass`, `instVarStructure`, `instVarRefactor`
  go through the hook for `committed` / `migratedFailures` / `dropped` / `partiallyApplied`
- `renameInstVar` is normalized onto it

**No private apply-result parser remains outside `previewEnvelope.ts`.**

## ⚠️ One behavior change, worth a look

`renameInstVar`'s copy had **drifted**: a missing failure `id` answered `''` rather than
`'?'`, and `applied` was read with a raw `typeof === 'number'` check instead of `asCount`,
so a negative or non-numeric count was never clamped to 0. That count drives the
"Renamed … N classes re-versioned" toast and the abort advice.

The file had **no imports at all**, which is why it could not reach `asCount` in the first
place — isolation caused the drift. No test pinned the old defaults; two new tests pin the
normalized ones.

## Not in scope

The larger part of C3 — a shared preview *panel*, panel-HTML builder, and start/page
parsers parametrized by a `parseChange` hook, across all 16 families — is a much bigger
sweep with no user-facing gain. Deferred with a written plan.

## Testing

- **[client .ts]** full client suite **5312 passed**, 12 skipped. Existing `*Preview.test.ts`
  files unchanged — the point is that the public API did not move. Two new tests for the
  normalized `renameInstVar` defaults.
- **F5** rename instance variable (the one behavior change) and extract superclass
  (exercises `committed` through the new hook) — both normal.

Part of #252.


client/src/refactoring/changeSignaturePreview.ts
```diff
@@ -86,11 +88,6 @@ export interface StartPreview {
 }
 
 /** The result of a server-side apply. */
```
MatiasFernandez: 💭 The interface moved out but its doc comment stayed, so we now have `/** The result of a server-side apply. */` followed by a blank line and then the pre-flight analysis doc — dead text a reader has to skip past. Same thing at line 213 with `/** Parse a server-side apply result. */`. Should we drop both, since the re-exported symbols carry their docs from `previewEnvelope.ts` now?

client/src/refactoring/renameMethodPreview.ts
```diff
@@ -87,11 +89,6 @@ export interface StartPreview {
 }
 
 /** The result of a server-side apply. */
```
MatiasFernandez: 💭 Same two orphaned doc comments as in `changeSignaturePreview.ts` (here at lines 91 and 195) — worth sweeping in the same pass.

client/src/refactoring/previewEnvelope.ts
```diff
@@ -0,0 +1,56 @@
...
+ * the envelope validation, the `failed` filtering and the `asCount` clamp have
+ * exactly one implementation. No `vscode` dependency, so it unit-tests directly.
+ */
+export interface ApplyResult {
+  applied: number;
+  failed: { id: string; label: string; error: string }[];
+  error?: string;
+}
+
+export function parseApplyResult(json: string): ApplyResult {
```
MatiasFernandez, line 33: 💭 Could `parseApplyResult` get a short JSDoc of its own? Eight families now reach it through a bare `export { parseApplyResult } from './previewEnvelope'`, and a re-export doesn't carry a doc comment — so where the old sites had `/** Parse a server-side apply result. */`, hovering the symbol from any of those modules now shows nothing. The interface above is well documented; a two-liner here (what it throws on, that counts are clamped rather than rejected) would keep that from regressing for readers and for agents deciding whether to call it.

client/src/refactoring/methodRelocationPreview.ts
```diff
@@ -232,27 +233,7 @@ export function makeParsePage<C extends BaseMethodChange>(
...
-          error: typeof f.error === 'string' ? f.error : 'unknown error',
-        }))
-    : [];
-  return {
-    applied: asCount(env.applied),
-    failed,
-    error: typeof env.error === 'string' ? env.error : undefined,
-  };
-}
+export { parseApplyResult } from './previewEnvelope';
```
MatiasFernandez, line 235: 💭 `/** Parse an apply result. Identical across families. */` is now sitting above a re-export statement, where TS won't attach it to anything — so this doc is invisible at every call site. Ties into my note on `previewEnvelope.ts`: if the doc lands on the definition there, this one can just go away.

client/src/refactoring/renameInstVarPreview.ts
```diff
@@ -61,34 +62,21 @@ export function parseRenamePreview(json: string): RenamePreview {
...
-    error: typeof env.error === 'string' ? env.error : undefined,
-  };
-}
+/**
+ * Was a private copy that had DRIFTED from every other family: a missing failure `id`
+ * defaulted to '' rather than '?', and `applied` was read with a raw number check rather
+ * than `asCount`, so a negative or non-numeric count was never clamped to 0. Normalized
+ * onto the shared parser (C3) — a deliberate behaviour change, pinned by tests.
+ */
+export const parseRenameApplyResult = parseApplyResult;
```
MatiasFernandez, line 80: 💭 There are two stacked JSDoc blocks here (and on `RenameApplyResult` above), and only the last one attaches — so hovering `parseRenameApplyResult` now shows the C3 refactoring history rather than `/** Parse the engine's apply envelope, {"applied":N,"failed":[..]} ... */`, which is the part a caller actually needs. Should we merge them, contract first and the drift note as a trailing sentence? Same for the type alias, where `/** The apply result: how many classes were re-versioned ... */` is the domain-specific bit worth keeping visible.

client/src/refactoring/__tests__/previewEnvelope.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    expect(r.applied).toBe(0);
+    expect(r.failed).toEqual([]);
+  });
+
+  it('throws on a non-envelope payload', () => {
+    expect(() => parseApplyResult('[]')).toThrow(/result envelope/);
+    expect(() => parseApplyResult('42')).toThrow(/result envelope/);
+  });
+});
```
MatiasFernandez, line 48: 💭 The new tests all go through `parseApplyResult`, so `parseApplyResultWith` — the actual new abstraction this PR adds — is only covered indirectly through the existing family tests. Is it worth one test on the hook directly? Two things it would pin that nothing else does: that `extend`'s fields land on the result alongside the base, and the merge precedence (`{ ...base, ...extend(env) }` means `extend` wins). The second is invisible at the type level, since `Omit<T, keyof ApplyResult>` stops a family from *declaring* a base field, but a runtime spread would still let one through.

client/src/refactoring/__tests__/previewEnvelope.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    expect(r.failed).toEqual([{ id: '?', label: '?', error: 'unknown error' }]);
+  });
+
+  it('surfaces a top-level error string', () => {
+    const r = parseApplyResult('{"applied":0,"failed":[],"error":"preview session expired"}');
+
+    expect(r.error).toBe('preview session expired');
+  });
+
+  it('clamps a missing/negative applied count to zero and ignores non-array failures', () => {
```
MatiasFernandez, line 34: The name says "missing/negative", but the payload `{"failed":"nope"}` only exercises the missing case — the negative one is pinned over in `renameInstVarPreview.test.ts`. Either add `applied: -1` here or trim the name to what it checks?

MatiasFernandez (approved):

client/src/refactoring/changeSignaturePreview.ts
ericwinger: Dropped both, here and in `renameMethodPreview.ts` — dead text either way, and the re-exported symbols do carry their docs now (see the `previewEnvelope.ts` thread; `parseApplyResult` has a real block on it as of this round, since a re-export carries nothing).

client/src/refactoring/renameMethodPreview.ts
ericwinger: Swept in the same pass — both gone.

client/src/refactoring/previewEnvelope.ts
ericwinger: Added. It documents what a caller actually needs to decide whether to wrap the call:

> Throws only when the payload is not a JSON object (a malformed envelope is a bug, not a user-facing outcome); everything inside it is coerced rather than rejected — `applied` is clamped to a non-negative count, a non-array or malformed `failed` degrades to an empty list, and a missing field in a failure entry falls back to `'?'` / `'unknown error'`. A whole-apply failure the engine reports (an expired preview token, say) arrives as `error`, not as a throw.

Plus a closing line noting most families reach it through a bare re-export, so this block is what their call sites see — which is the thing that would otherwise rot again.

client/src/refactoring/methodRelocationPreview.ts
ericwinger: Removed — you're right that it attaches to nothing there, and with the doc now living on the definition it has no reason to exist.

client/src/refactoring/renameInstVarPreview.ts
ericwinger: Merged both, contract first. `parseRenameApplyResult` now opens with the envelope it parses and points at `parseApplyResult` for the coercion detail, then the drift note follows as background. The alias keeps `The apply result: how many classes were re-versioned...` as its first sentence, with the C3 aliasing as a trailing clause — the domain-specific part stays visible, as you suggested.

client/src/refactoring/__tests__/previewEnvelope.test.ts
ericwinger: Added four tests on the hook directly. The collision one is the interesting one, and it turned out sharper than expected:

```ts
const r = parseApplyResultWith<ApplyResult>('{"applied":1,"failed":[]}', () => ({
  applied: 99,
}));
expect(r.applied).toBe(99);
```

No cast is needed — with `T = ApplyResult`, `Omit<T, keyof ApplyResult>` is `{}`, so the compiler accepts an object carrying `applied` and the spread silently wins. That's exactly the hole you described, and it's now pinned with a comment saying why the absence of a cast is the point.

The other three: extras land alongside the base; the base gets the same coercion through the hook as through the plain function; and a non-envelope payload throws **before** `extend` is consulted (asserted with a spy, so the ordering is pinned rather than assumed).

client/src/refactoring/__tests__/previewEnvelope.test.ts
ericwinger: Kept the name and made it true — added `applied: -1` and a non-numeric case alongside the missing one, so all three coercions live in the test that claims them. The `renameInstVarPreview.test.ts` pair still pins the same behaviour at that call site, which is worth keeping since that one is the family whose defaults changed.

PR #416, opened 2026-08-11
Fix Windows CI failures in VersionManager catalog tests

npapagna:
## What

Two tests in `VersionManager.fetchAvailableVersions`'s describe block failed on Windows without WSL.

`excludes remote versions older than the minimum supported gemstone version` built its mock catalog HTML with `storage.getPlatformKey()`, which is `undefined` on Windows without WSL. Production (`versionManager.ts:22`) matches the catalog against `storage.getCatalogPlatformKey()` instead, which falls back to `x86_64.Linux` for exactly that case. Switching the test to the same method fixes it on every platform, with no gating needed.

`keeps a local version even when it is older than 3.6.2` genuinely needs a local server platform to exist (`hasLocalServer`), which this test file's `wslBridge` mock rules out on Windows. Gated it to POSIX with `onSupportedPosixIt`, matching its siblings in the same describe block. Real Windows+WSL coverage for that scenario is tracked in #413.

## Why

These were the last two Windows-only test failures found while validating Windows CI on top of #407.

## Test plan

- [x] `npm run lint && npm run format:check && npm run compile` pass
- [x] `localVersion.test.ts` passes locally (25/25)
- [x] Verified against a real Windows run of this exact diff, via the `windows-health-check.yml` workflow on a separate validation branch (that workflow isn't part of this PR): these were the only 2 failures remaining after #407's fixes, and both are now green there

### DIFF
--- a/client/src/__tests__/localVersion.test.ts
+++ b/client/src/__tests__/localVersion.test.ts
@@ -360,7 +360,11 @@ describe('VersionManager.fetchAvailableVersions', () => {
   it('excludes remote versions older than the minimum supported gemstone version', async () => {
     const storage = new SysadminStorage();
     const manager = new VersionManager(storage);
-    const platformKey = storage.getPlatformKey();
+    // getCatalogPlatformKey, not getPlatformKey: production matches the
+    // catalog against the former, which falls back to 'x86_64.Linux' when
+    // there's no local server platform (e.g. Windows without WSL). Building
+    // the fixture with getPlatformKey would diverge from that on such hosts.
+    const platformKey = storage.getCatalogPlatformKey();
     const ext = storage.getDownloadExtension();
 
     const html = [
@@ -376,7 +380,11 @@ describe('VersionManager.fetchAvailableVersions', () => {
     expect(versions.map((v) => v.version)).toEqual(['3.7.0', '3.6.2']);
   });
 
-  it('keeps a local version even when it is older than 3.6.2', async () => {
+  // Gated on POSIX not for symlinks but because it needs hasLocalServer to be
+  // true, and this file's wslBridge mock makes getPlatformKey() undefined
+  // (so hasLocalServer is false) on Windows. Real Windows+WSL coverage is
+  // tracked in #413.
+  onSupportedPosixIt('keeps a local version even when it is older than 3.6.2', async () => {
     const storage = new SysadminStorage();
     const manager = new VersionManager(storage);
     const suffix = storage.getPlatformSuffix();
--- a/client/src/__tests__/sysadminStorage.test.ts
+++ b/client/src/__tests__/sysadminStorage.test.ts
@@ -49,4 +49,11 @@ describe('SysadminStorage.getPlatformKey on Darwin', () => {
       expect(new SysadminStorage().getCatalogPlatformKey()).toBe('i386.Darwin');
     });
   });
+
+  it('falls back to x86_64.Linux when there is no local server platform', () => {
+    withPlatform('win32', 'x64', () => {
+      expect(new SysadminStorage().getPlatformKey()).toBeUndefined();
+      expect(new SysadminStorage().getCatalogPlatformKey()).toBe('x86_64.Linux');
+    });
+  });
 });


client/src/__tests__/localVersion.test.ts
```diff
@@ -360,7 +360,11 @@ describe('VersionManager.fetchAvailableVersions', () => {
   it('excludes remote versions older than the minimum supported gemstone version', async () => {
     const storage = new SysadminStorage();
     const manager = new VersionManager(storage);
-    const platformKey = storage.getPlatformKey();
+    // getCatalogPlatformKey, not getPlatformKey: production matches the
+    // catalog against the former, which falls back to 'x86_64.Linux' when
+    // there's no local server platform (e.g. Windows without WSL). Building
+    // the fixture with getPlatformKey would diverge from that on such hosts.
+    const platformKey = storage.getCatalogPlatformKey();
```
MatiasFernandez, line 367: 💭 This is the right call — matching production's method is exactly the fix. But it does mean the fixture and the code under test now both derive the key from `getCatalogPlatformKey()`, so this test can no longer fail if that method's fallback regresses.

And the fallback is the interesting half: `sysadminStorage.test.ts:49` only covers the Intel-Mac case where `getPlatformKey()` *does* resolve ("no Linux fallback"), so the `?? 'x86_64.Linux'` branch — the one this whole PR depends on for Windows-without-WSL — isn't pinned anywhere. Worth adding a sibling there for "falls back to x86_64.Linux when there's no local server platform"?

client/src/__tests__/localVersion.test.ts
```diff
@@ -376,7 +380,7 @@ describe('VersionManager.fetchAvailableVersions', () => {
     expect(versions.map((v) => v.version)).toEqual(['3.7.0', '3.6.2']);
   });
 
-  it('keeps a local version even when it is older than 3.6.2', async () => {
+  onSupportedPosixIt('keeps a local version even when it is older than 3.6.2', async () => {
```
MatiasFernandez, line 387: 💭 Could we get a one-line comment here on *why* it's gated, plus the `#413` reference? Everywhere else in this file the gate means "needs symlinks", and a reader will reasonably assume the same thing here — but the actual reason is different (this test needs `hasLocalServer` to be true, which the file's `wslBridge` mock rules out on Windows). You explained it well in the PR body; it'd be a shame for that to only live there. The comment you added in `rowanCreate.test.ts` is a good model.

client/src/__tests__/rowanCreate.test.ts
```diff
@@ -117,7 +118,13 @@ describe('createRowanProject', () => {
...
   });
 
-  it('backslash-escapes a backslash in the project name', () => {
+  // A backslash is a plain filename character on POSIX but a path separator
+  // on Windows, so a name containing one can't be used as a spec's filename
+  // there. Not a real-world gap: every caller of createRowanProject already
+  // rejects '\' and '/' in the name before it gets here (see
+  // extension.ts's rowanNewProject input validation and rowanInitHere's use
+  // of path.basename).
+  onSupportedPosixIt('backslash-escapes a backslash in the project name', () => {
```
MatiasFernandez, line 127: 💭 The comment says every caller already rejects `\` and `/` before reaching `createRowanProject`, which means this test is pinning behavior for an input production can't produce. Given that, is POSIX-gating the right move rather than dropping it? As it stands we keep a test that (a) never runs on Windows and (b) covers an unreachable case — and the escaping logic it exercises is plain string manipulation that the `O'Hara` test above already covers on the reachable path.

If the backslash escaping is worth protecting, the more valuable test might be at the caller level: that `rowanNewProject`'s input validation actually rejects a name with `\`. Happy either way, just want the choice to be deliberate.

client/src/__tests__/extentBackupManager.test.ts
```diff
@@ -75,7 +75,11 @@ describe('runOnlineExtentBackup', () => {
     // enforce that implicitly; destructuring below doesn't, so check it here.
     expect(copyFile.mock.calls[0]).toHaveLength(2);
     const [src, dest] = copyFile.mock.calls[0];
-    expect(src).toBe(path.join('/db/data', 'extent0.dbf'));
+    // The stone reports its own extent paths verbatim (see extentFileNames'
+    // doc comment). It's always POSIX, since the server runs on Linux/WSL
+    // regardless of the client's OS, so compare with path.posix rather than
+    // the client-platform path.join, which would render backslashes on Windows.
+    expect(src).toBe(path.posix.join('/db/data', 'extent0.dbf'));
```
MatiasFernandez: nit: Since `src` is the mock's `extents: '/db/data/extent0.dbf\n'` value handed back verbatim, would `expect(src).toBe('/db/data/extent0.dbf')` say it more directly? `path.posix.join` reads as "we're constructing a platform-specific path" when the point is the opposite — the string never goes through `path` at all. That would also let the four-line comment shrink to roughly "the stone reports POSIX paths verbatim; don't rebuild this with `path.join`".

MatiasFernandez (approved):

client/src/__tests__/localVersion.test.ts
npapagna: Good catch — added a sibling test in `sysadminStorage.test.ts` pinning the `getPlatformKey() ?? 'x86_64.Linux'` fallback directly: "falls back to x86_64.Linux when there is no local server platform". Amended into this commit (now `43b2399d` after rebasing onto main, which already had PR #407's fixes).

client/src/__tests__/localVersion.test.ts
npapagna: Added a one-line comment explaining the real gating reason (needs `hasLocalServer` true, which this file's `wslBridge` mock rules out on Windows) plus the `#413` reference, modeled on the `rowanCreate.test.ts` comment. Amended into this commit.

client/src/__tests__/rowanCreate.test.ts
npapagna: Same thread as the earlier PR #407 discussion on this — already deferred and tracked as Jasper#418, not re-litigating here.

client/src/__tests__/extentBackupManager.test.ts
npapagna: Turns out this exact literal-string fix already landed on main via PR #407 (37419cbd) before this branch was rebased — this branch had independently reinvented the same fix with `path.posix.join` instead. Rebasing onto main picked up the merged fix directly, so there's nothing left to change here.

PR #420, opened 2026-08-12
Make integration test sessions refuse commits outright

MatiasFernandez:
## What

Groundwork for the next slice of the GCI-suite migration: making the integration harness's isolation promise real, and writing down the test tiers it belongs to.

**The guard.** `useIntegrationTest` wraps every test in a transaction it always aborts — but an abort cannot undo a commit, so the promise only held as long as nothing (test, or production code the test exercises) committed. Every session the harness creates now arms GemStone's own session-level commit guard (`System disableCommitsWithReason:`) before handing it to a test, so a commit fails at the commit site with `TransactionError 2249` naming the harness, instead of silently leaking state into the next test, the parallel workers sharing the stone, and the second CI pass within a matrix cell.

- Armed in `login()` **and** in `withTransientSession()` — the latter previously went straight through `gciLibrary.login` without arming anything.
- Session-scoped and irreversible: there is no `enableCommits`, the only exit is `logout`. That's why it's armed once per session rather than per test.
- Arming is required to be *this harness's* doing. GemStone reports whether the call is what disabled commits; "they were already disabled" is a hard failure, because at that point the invariant holds for a reason the harness didn't establish and can't vouch for (a `UserProfile` with `disableCommits`, a stone mid-restore). A loud stop beats a whole matrix pass under a guard we didn't arm.
- `afterEach` asserts the guard is still armed *before* cleaning up, and treats a dropped guard as a harness bug that poisons the rest of the file — cleanup isn't trustworthy on a session that managed to commit.
- Arming is a doit, so it caches an oop and dirties the `PureExportSet`; `resetNonTransactionalSessionState` runs right after, so every session is handed over in the state a bare login leaves behind.

**The tests.** A new `useIntegrationTest.integration.test.ts` covers the invariant the only way it can be covered — by attempting real commits and asserting GemStone refuses them: on the shared session, through `GciTsCommit` directly, on a `withTransientSession` session, on a session re-established mid-file via `logout()`/`login()`, and that the session stays usable after a refusal. Assertions are on error number `2249` and the exported `COMMIT_GUARD_REASON`, never on GemStone's prose, which varies across the releases in the integration matrix.

**The docs.** Three new pages, indexed in `docs/README.md`:

- `reference/integration-test-harness.md` — the `GciTestContext`, hook order, the commit invariant, the failure modes you'll actually hit.
- `explanation/integration-test-isolation.md` — why transaction-abort is the isolation mechanism, where it's weaker than it reads, why the guard is session-level and irreversible.
- `explanation/test-tiers.md` — unit / integration / GCI / acceptance: what each proves, what it needs, what CI runs, what each runs *inside* (Node vs jsdom vs Electron), and the four name collisions that keep costing people time (`test:server` vs `test:server:start`, "integration" ≠ "several modules", "GCI test" vs the legacy `gci` suite, `.test.ts` vs `.spec.ts`).

`.claude/rules/client/tests.md` now points at those pages, and its "must commit" escape-to-the-`gci`-suite reason is restated as what it now is: harness sessions refuse commits outright, with no opt-out.

## Why

Tests keep moving out of the on-demand `gci/` suite (#326, #334, #344) into the CI integration suite. That's only safe if the integration harness actually isolates — and its central promise had a hole big enough that "it must commit" was a legitimate reason to leave a test behind in a directory CI never runs. Closing the hole first, before the next batch of migrations, makes the remaining escape hatches genuinely narrow. The opt-in `commitStrategy` for tests that truly need to commit lands in a later phase; the reference doc marks the spot.

## Test plan

- [x] `npm run lint && npm run format:check && npm run compile` pass
- [x] `npm test` passes against a live stone, including the new harness suite
- [x] Guard verified positively, not just by "tests still pass": each commit path is asserted to be refused with `2249` + the harness reason

### DIFF
--- a/client/src/__tests__/useIntegrationTest.ts
+++ b/client/src/__tests__/useIntegrationTest.ts
@@ -119,11 +119,7 @@ export function useIntegrationTest(callback: UseIntegrationTestCallback) {
     try {
       // Checked before the cleanup below on purpose: if the commit guard is no longer armed, that means a test
       // committed and left the session dirty, so the cleanup below is not guaranteed to succeed. Bail out early
-      const commitGuardStillArmed = gciLibrary.executeAndRelease(
-        session,
-        'System commitsDisabledUntilLogout',
-        (oop) => gciLibrary.isTrueOop(oop),
-      );
+      const commitGuardStillArmed = gciLibrary.areCommitsDisabledUntilLogout(session);
 
       if (!commitGuardStillArmed) {
         throw new Error(
@@ -225,11 +221,7 @@ export function useIntegrationTest(callback: UseIntegrationTestCallback) {
   function armCommitGuard(aSession: unknown) {
     let armedNewly: boolean;
     try {
-      armedNewly = gciLibrary.executeAndRelease(
-        aSession,
-        `System disableCommitsWithReason: '${COMMIT_GUARD_REASON}'`,
-        (oop) => gciLibrary.isTrueOop(oop),
-      );
+      armedNewly = gciLibrary.disableCommitsUntilLogout(aSession, COMMIT_GUARD_REASON);
     } catch (error) {
       // `cause` keeps the original error — stack included — linked to this one, which
       // vitest prints under a "Caused by:" heading. Its message is also inlined here so
--- a/client/src/__tests__/useIntegrationTest.ts
+++ b/client/src/__tests__/useIntegrationTest.ts
@@ -109,6 +109,20 @@ export function useIntegrationTest(callback: UseIntegrationTestCallback) {
     if (!session) return;
 
     try {
+      // Checked before the cleanup below on purpose: if the commit guard is no longer armed, that means a test
+      // committed and left the session dirty, so the cleanup below is not guaranteed to succeed. Bail out early
+      const commitGuardStillArmed = gciLibrary.executeAndRelease(
+        session,
+        'System commitsDisabledUntilLogout',
+        (oop) => gciLibrary.isTrueOop(oop),
+      );
+
+      if (!commitGuardStillArmed) {
+        throw new Error(
+          `useIntegrationTest: the commit guard was no longer armed at the end of "${expect.getState().currentTestName}" — this is a harness bug, not a mid-test abort (the guard cannot be dropped). Failing the rest of this file since every remaining test would run against an unguarded session.`,
+        );
+      }
+
       gciLibrary.abortTransaction(session);
       gciLibrary.resetNonTransactionalSessionState(session);
     } catch (error) {
@@ -131,6 +145,7 @@ export function useIntegrationTest(callback: UseIntegrationTestCallback) {
       options?.user ?? process.env.VITE_GEMSTONE_USER!,
       process.env.VITE_GEMSTONE_PASSWORD!,
     );
+    armCommitGuard(session);
 
     callback({
       gciLibrary,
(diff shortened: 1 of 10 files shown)


client/src/__tests__/useIntegrationTest.ts
```diff
@@ -109,6 +117,20 @@ export function useIntegrationTest(callback: UseIntegrationTestCallback) {
...
+        'System commitsDisabledUntilLogout',
+        (oop) => gciLibrary.isTrueOop(oop),
+      );
+
+      if (!commitGuardStillArmed) {
+        throw new Error(
+          `useIntegrationTest: the commit guard was no longer armed at the end of "${expect.getState().currentTestName}" — this is a harness bug, not a mid-test abort (the guard cannot be dropped). Failing the rest of this file since every remaining test would run against an unguarded session.`,
+        );
+      }
+
```
npapagna, line 123: Maybe we can extract this to its own method, to make it more declarative?
Something like `assertCommitGuardIsarmed()`

client/src/__tests__/useIntegrationTest.ts
```diff
@@ -109,6 +117,20 @@ export function useIntegrationTest(callback: UseIntegrationTestCallback) {
     if (!session) return;
 
     try {
+      // Checked before the cleanup below on purpose: if the commit guard is no longer armed, that means a test
+      // committed and left the session dirty, so the cleanup below is not guaranteed to succeed. Bail out early
+      const commitGuardStillArmed = gciLibrary.executeAndRelease(
+        session,
```
npapagna: Claude 🤖 : This "commit guard armed" check (`gciLibrary.executeAndRelease(session, 'System commitsDisabledUntilLogout', (oop) => gciLibrary.isTrueOop(oop))`) is duplicated verbatim in `commitGuardIsArmedOn` over in `useIntegrationTest.integration.test.ts`. Worth exporting it once (e.g. as a `GciTestContext` helper) so a future change to the check expression doesn't get applied to one copy and forgotten in the other.

client/src/__tests__/useIntegrationTest.ts
```diff
@@ -176,12 +199,67 @@ export function useIntegrationTest(callback: UseIntegrationTestCallback) {
...
+   *
+   * Deliberately strict: arming is required to be *this* harness's doing. If
+   * GemStone reports commits were already disabled -- by this user's profile,
+   * or by a stone that is mid-restore -- the commit invariant happens to hold,
+   * but it holds for a reason the harness did not establish and cannot vouch
+   * for. Rather than run a whole matrix pass under a guard it did not arm, the
+   * harness fails loudly so the environment gets explained.
+   */
+  function armCommitGuard(aSession: unknown) {
+    let armedNewly: boolean;
```
npapagna, line 216: Claude 🤖 : `armCommitGuard`'s arm/verify logic lives entirely in this test-only harness file. `GciLibrary` already hosts the analogous `resetNonTransactionalSessionState` session-lifecycle primitive, so it might be worth putting "arm the commit guard" (and the armed-check) on `GciLibrary` itself. That way other consumers (mcp-server, extension.ts) that want the same "this session can't commit" guarantee outside tests have a reusable, documented primitive instead of copying the raw Smalltalk strings and oop coercion by hand.

npapagna (approved):
This is great work @MatiasFernandez!
Really liked what you did at the doc level, too.

 A couple more things I noticed, but couldn't attach inline since they're in files this PR doesn't touch:                                                                      
                                                                                                                                                                                
  -  Claude 🤖 .claude/rules/refactoring-tests.md:19 says "Only a scenario that must COMMIT belongs in the on-demand client/src/__tests__/gci/ project, since the integration harness      
  aborts every test." That's now stale: this PR's own .claude/rules/client/tests.md update says harness sessions refuse commits outright (TransactionError 2249), not just abort
  afterward. Worth updating this rationale too.                                                                                                                                 
  - Claude 🤖 client/src/__tests__/gci/gciInstVar.e2e.test.ts:33 repeats the same stale "an abort cannot undo a commit" framing in its own doc comment justifying why its committing paths
  can't move into the harness yet.

client/src/__tests__/useIntegrationTest.ts
MatiasFernandez: Good call. Pulled the check onto `GciLibrary` itself as `areCommitsDisabledUntilLogout(session)`, so the harness's `afterEach` and `commitGuardIsArmedOn` both go through one definition. Put it next to `abortTransaction`, where the arming primitive from your other comment can sit alongside it.

client/src/__tests__/useIntegrationTest.ts
MatiasFernandez: Done — added `GciLibrary.disableCommitsUntilLogout(session, reason)` next to `areCommitsDisabledUntilLogout`, and `armCommitGuard` is now just that call. The reason goes through `escapeString` now, so an apostrophe can't break the doit. The reset and the strict "already disabled" failure stay in the harness — those are harness policy, and another consumer would plausibly want already-disabled to be fine.

client/src/__tests__/useIntegrationTest.ts
MatiasFernandez: Extracted to `assertCommitGuardIsStillArmed(session)`, sitting next to `armCommitGuard` — `afterEach` now reads as assert-guard, abort, reset. Went with "IsStillArmed" since it's a teardown check.

PR #422, opened 2026-08-12
Harden the RB refactoring paths: webview scope guard, prefix matcher, extract-superclass collision

ericwinger:
Hardening pass over the RB refactoring paths, combining two small branches that touch
disjoint files, plus the pre-merge checklist and acceptance scaffold that accompany them.
Three of the four code items close a gap at a trust boundary where a check existed but was
too narrow to catch the case.

## Client

**`isRenameClassScope` — validate the webview message instead of casting it.**
The rename-class panel returns a scope (`class`, `hierarchy`, `wholeSystem`, or `dictionary`
with a name). The handler took it through an unchecked `as RenameClassScope` cast, so a
malformed shape flowed through into a generated server query. A runtime guard now validates
the shape before `validate()` runs; a rejected scope posts `invalid` back to the panel and
leaves it open rather than resolving. Webview input is treated as untrusted, which is the
right posture for a boundary the extension does not control.

**`matchesClassPrefix` — extract an untestable predicate.**
The class-by-prefix QuickPick deliberately disables VS Code's fuzzy substring matching in
favour of strict prefix semantics, so typing `Z` surfaces only classes beginning with `Z`.
That logic was inlined in an `onDidChangeValue` closure and therefore unreachable from tests
— it carried zero coverage. It is now a named, exported predicate with six tests pinning
case-insensitivity, query trimming, and empty-query-matches-all. Pure refactor, no behaviour
change.

## Engine — follow-up to #379

`GsExtractSuperclassRefactoring>>computeAnalysis` screened the proposed superclass name with
`environment classNamed:`, which resolves only Class bindings. A name bound to a **non-class
global** therefore passed the precondition, and `applyClassAdd:`'s
`subclass:...inDictionary:` then silently rebound — destroyed — the existing value.

The precondition now also consults `symbolList objectNamed:` and declines any pre-existing
binding for the name. The client already guarded this case, so the exposure was confined to
callers with no UI: MCP, SUnit, and direct engine use.

Covered by `testInsertDeclinesWhenNameIsANonClassGlobal`, which asserts the decline message,
an empty change set (nothing created), and survival of the original global.
`resources/refactoring/engine.gs` and `engine-tests.gs` regenerated.

## Docs and scaffolding

~~**`.claude/rules/refactoring-pre-merge.md`** — a path-scoped checklist for
`gs-src/refactoring/**` and `client/src/refactoring/**`, codifying what this family has been
verifying ad hoc: SUnit green on both boundaries, payload regeneration, 3.6.2 dialect hazards,
method survival across reshape, and the client invariants (disposal, CSP/nonce, escaping at
both boundaries, `ensureRbSupport` gating, one-shot apply).~~

**Dropped from this PR** — a path-scoped rule file loads for anyone merely *reading* under those
directories, so the content moved to **#430** to become an on-demand skill instead.

**`acceptance/tests/refactoring.spec.ts`** — ⚠️ **scaffold only, it does not execute.** Both
tests are `test.fixme` with `TODO(selector)` / `TODO(assert)` stubs. The webview locators must
be captured against a running window, and the acceptance container does not yet provision a
stone with the RB engine. It is landed alongside the checklist that calls for it so the
intended F5-layer coverage is tracked in-tree — please read it as intent, not as coverage.

## Testing

| Layer | Result |
|---|---|
| [GS SUnit] RB engine, **3.6.2** — staged from this branch, aborted after | **583 run / 0 fail / 0 err** across all 22 suites |
| [GS SUnit] RB engine, **3.7.5** — same | **583 run / 0 fail / 0 err**, identical per-class counts |
| [GS SUnit] `GsExtractSuperclassRefactoringTest` | 34/34 on both boundaries |
| [client .ts] full suite | 350 files, 5430 passed, 8 skipped |
| [client .ts] server / mcp-server | 322 passed / 92 passed |
| [GCI auto] RB integration suites | 128 files, 1228 passed |
| lint / format:check / compile | clean |

Engine SUnit was run via topaz on both boundaries with `stageFromServerDir:` so the tests
exercise *this branch's* `engine.gs` rather than the stone's installed engine, followed by
`System abortTransaction` — both stones left untouched. Rebuilding the payloads from source
produces no diff against what is committed.


gs-src/refactoring/engine/GsExtractSuperclassRefactoring.class.st
```diff
@@ -184,6 +184,13 @@ GsExtractSuperclassRefactoring >> computeAnalysis [
 		^decline := 'Cannot extract a superclass above ', anchorClass name asString, ': it has no superclass.'].
 	(environment classNamed: newName) notNil ifTrue: [
 		^decline := 'Cannot extract superclass ', newName, ': a class of that name already exists.'].
+	"classNamed: only matches a Class binding; a name bound to a NON-class global
+	 would otherwise slip past and be silently clobbered by applyClassAdd:'s
+	 subclass:...inDictionary:. Decline any existing binding for the name (#379
+	 review: the client already guards this, but a non-UI caller -- MCP, SUnit,
+	 direct engine use -- must be protected here too)."
+	(environment symbolList objectNamed: newName asSymbol) notNil ifTrue: [
```
MatiasFernandez, line 200: 💭 `symbolList objectNamed:` asks the question system-wide, but `applyClassAdd:` creates the class in ONE dictionary (`dictObjectForNewClass` — the explicit `inDictionary:` or the anchor's dict). So a non-class global named `Pet` sitting in a *lower-priority* dictionary now declines an extract into the target dict, where the creation would only have shadowed it, not clobbered it.

Is the system-wide reach deliberate — for consistency with `GsRenameClassRefactoring>>newNameCollision`, which is also symbol-list-wide — or should this ask about the target dictionary only? `resolveSibling:` right below goes out of its way to scope its lookup ("a same-named class in another dictionary is never grabbed"), so I couldn't tell which convention this one is meant to follow. Happy either way, but if it's intentional it might be worth a word in the comment, since the decline message ("already in use as a global") reads as if the collision were in the class's own dictionary.

gs-src/refactoring/engine/GsExtractSuperclassRefactoring.class.st
```diff
@@ -184,6 +184,13 @@ GsExtractSuperclassRefactoring >> computeAnalysis [
...
+		^decline := 'Cannot extract superclass ', newName, ': the name ', newName, ' is already in use as a global.'].
```
MatiasFernandez: 💭 `GsSplitClassRefactoring>>computeAnalysis` has the identical narrow precondition (`(environment classNamed: newName) notNil ifTrue: [^decline := 'Cannot split: a class named ...']`, around line 264) and then creates the component class with `Object subclass: newName ... inDictionary: (self dictObjectForNewClass)` in its own `applyClassAdd:` — so a `newName` bound to a non-class global looks like it slips past there for exactly the same reason, with the same silent rebind at apply time.

Since the exposure argument here is "non-UI callers (MCP, SUnit, direct engine use) must be protected in the engine", split-class seems to be in the same position. Worth closing in this pass while the reasoning is fresh, or filing it as the next item in the hardening family?

client/src/refactoring/queries/previewRenameClass.ts
```diff
@@ -9,6 +9,19 @@ import { classLookupExpr, escapeString } from '../../queries/util';
 export type RenameClassScope =
   { kind: 'class' | 'hierarchy' | 'wholeSystem' } | { kind: 'dictionary'; dictName: string };
 
+// Runtime guard for a RenameClassScope arriving from the webview (untrusted). The
+// editor casts the incoming message; validate it here so a malformed shape is
```
MatiasFernandez: "The editor casts the incoming message" describes the world before this PR — after your change the editor calls this guard instead of casting. Maybe "the editor validates the incoming message with this before using it"?

client/src/refactoring/queries/previewRenameClass.ts
```diff
@@ -9,6 +9,19 @@ import { classLookupExpr, escapeString } from '../../queries/util';
...
+// rejected rather than flowing through the cast into a server query.
+export function isRenameClassScope(x: unknown): x is RenameClassScope {
+  if (typeof x !== 'object' || x === null) return false;
+  const kind = (x as { kind?: unknown }).kind;
+  if (kind === 'class' || kind === 'hierarchy' || kind === 'wholeSystem') return true;
+  if (kind === 'dictionary') {
+    return typeof (x as { dictName?: unknown }).dictName === 'string';
```
MatiasFernandez: 💭 This accepts `{ kind: 'dictionary', dictName: '' }`, which then reaches `scopeClauseOf` as `dictionaryScope: ''`. The tests pin the missing and non-string cases nicely, but an empty string is the shape most likely to actually arrive from a webview (an unselected/blank dropdown value) and it's the one that produces a syntactically valid query naming a dictionary that cannot exist.

Should the guard require a non-empty `dictName` (and get a case in the test alongside `dictName: 5`)? Or does the engine already decline an unknown dictionary scope with something legible?

client/src/refactoring/renameClassEditor.ts
```diff
@@ -77,7 +81,15 @@ export function showRenameClassEditor(
     panel.webview.onDidReceiveMessage((message) => {
       if (message?.command === 'ok') {
         const newName = typeof message.newName === 'string' ? message.newName.trim() : '';
-        const scope = message.scope as RenameClassScope;
+        if (!isRenameClassScope(message.scope)) {
```
MatiasFernandez, line 89: 💭 The module doc at the top (lines 6–9) still describes the submit path as "the host runs `validate(newName)` ... and, if it returns an error, posts it back" — with this change there's now an earlier rejection that happens *before* `validate` and also keeps the panel open. Could you add a sentence there so the header still matches the contract? It's the first thing anyone (or any agent) reads before touching this handler.

client/src/explorerTreeHelpers.ts
```diff
@@ -39,6 +39,19 @@ export function defaultDictionaryIndex(names: string[]): number {
   return userGlobals >= 0 ? userGlobals : 0;
 }
 
+/**
+ * Whether a class-picker label matches the typed query, for the class-by-prefix
+ * QuickPick (move-method / move-instance-variable destination pickers). The
```
MatiasFernandez: The doc says "move-method / move-instance-variable destination pickers", but `pickClassByPrefix` has a single caller today — the move-methods-to-class flow in `gemstoneExplorer.ts`. `moveInstVar` is the up/down reorder, which never picks a destination class. Drop the second one, or say "the move-to-class destination picker" so it doesn't send a reader looking for a call site that isn't there?

.claude/rules/refactoring-pre-merge.md
```diff
@@ -0,0 +1,64 @@
+---
+paths:
```
MatiasFernandez: 🚫 Rules are pre-loaded as soon as the agent reads these files, even when it is not modifying them. That means the agent context for any dev on the team would get these instructions loaded even when the developer is not working on a change here. There's also some duplication against other context files, which is problematic. I also think some of the rules can be enforced by other mechanisms, like an automatic build script. Until we get such automations in place, I'm ok introducing agent-targeted instructions/rules, but I think this would be better suited as a skill. This describes a specific task/workflow that is a verification checklist before considering a new change on the refactorings plugin. 

I would encourage we don't merge this rule file, and we either replace it with a skill within this PR or add a skill as a follow-up. Let me know if you want help with that

acceptance/tests/refactoring.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { test, expect } from '../helpers/vscode';
+
+/**
+ * End-to-end coverage for the refactoring UI — the "live F5" layer that unit and
+ * GCI tests cannot reach: the Explorer pencil → preview panel → Apply round-trip,
+ * driven through a real editor window.
+ *
+ * STATUS: SCAFFOLD. The steps below are marked `test.fixme` because the concrete
```
MatiasFernandez: 💭 If this is not really functional, what's the value of committing it?

MatiasFernandez (changes requested):
I'm just requesting changes because I'm worried about the new Claude rule introduced in the PR. The rest is non-blocking but still worth reviewing

.claude/rules/refactoring-pre-merge.md
ericwinger: You're right, and the file is now deleted from this branch.

**Why you're right.** The `paths:` header at the top meant this file loaded into the agent's context for anyone who merely *read* a file under `gs-src/refactoring/**` or `client/src/refactoring/**` — that is the whole refactoring area, so every developer paid for it even when they were only looking around. A skill loads only when someone asks for it, which is the right fit for a checklist you run once, before merging.

**What was in it.** It was a list of things I check by hand on every refactoring change, written down so I stop forgetting them. On the engine side: run the SUnit suites on both 3.6.2 and 3.7.5; rebuild the generated `resources/refactoring/*.gs` payloads from source; don't use selectors that 3.6.2 lacks; make sure a class reshape carries every method forward and reports anything it could not recompile; build a change set without committing. On the client side: dispose panels and listeners; give each webview a strict CSP, a fresh nonce, and a runtime check of every incoming message; escape strings going into Smalltalk and into HTML; check the engine is installed before previewing; make Apply one-shot; reselect the affected row afterwards.

Three of its items only repeated things already written in `CLAUDE.md` and `.claude/rules/refactoring-tests.md`, which is the duplication you spotted.

**Where it went.** All of the above is now in **#430**: https://github.com/GemTalk/Jasper/issues/430 — nothing was thrown away. The issue keeps the engine and client items word for word, points at the existing files for the three duplicated ones instead of repeating them, and records your point that some of these are better checked by a build or CI script than by an agent reading a list. It is assigned to both of us, to be written as a skill later. Happy to take you up on your offer of help when we get to it.

gs-src/refactoring/engine/GsExtractSuperclassRefactoring.class.st
ericwinger: Deliberate — and you're right that the message didn't say so. `GsRenameClassRefactoring>>newNameCollision` asks the same symbol-list-wide question, and while a binding in a lower-priority dictionary would only be *shadowed* rather than clobbered, shadowing an existing global with a new class is its own trap, so declining is the behaviour I want in both.

The comment now spells that out, including why `resolveSibling:` scopes its lookup instead (it is about FINDING an existing class, where grabbing a same-named class from another dictionary would be wrong — the opposite question). The decline text now names the symbol list:

> Cannot extract superclass Pet: the name Pet is already in use as a global on the symbol list.

`testInsertDeclinesWhenNameIsANonClassGlobal` asserts the fuller wording, so it can't drift back. Same treatment in split-class (see the next thread). [GS SUnit] 591 run / 0 fail / 0 err on both 3.6.2 and 3.7.5.

gs-src/refactoring/engine/GsExtractSuperclassRefactoring.class.st
ericwinger: Confirmed and fixed in this pass — good catch. Split-class was in fact *worse* off than extract-superclass: `splitClassCommand.ts` validates only the SHAPE of the name (`validateClassName`), so there was no client-side guard either. The Explorer path could clobber a global, not just MCP / SUnit / direct engine use.

Reproduced under F5 before fixing: bind `UserGlobals at: #SplitDemoBucket put: 'DO NOT LOSE ME'`, split a class into a component named `SplitDemoBucket` — the preview offered no decline, and after Apply the string was gone, the name rebound to the new class.

`GsSplitClassRefactoring>>computeAnalysis` now asks the same `symbolList objectNamed:` question, declining with `Cannot split: the name X is already in use as a global on the symbol list.` Covered by `testDeclinesWhenNameIsANonClassGlobal`, which asserts the decline text, an empty change set (nothing created) and survival of the original global.

Verified the test actually catches it: with the guard backed out and the payloads rebuilt, the split suite reports exactly one failure — the new test. [GS SUnit] 591 run / 0 fail / 0 err on both 3.6.2 and 3.7.5.

client/src/refactoring/queries/previewRenameClass.ts
ericwinger: Right — that sentence described the world before the change. Reworded to "the editor validates the incoming message with this before using it, so a malformed shape is rejected rather than flowing into a server query."

client/src/refactoring/queries/previewRenameClass.ts
ericwinger: Agreed — that's the shape most likely to actually arrive, and it's the one that produces a valid query naming an impossible dictionary. The guard now requires a non-blank name:

```ts
const dictName = (x as { dictName?: unknown }).dictName;
return typeof dictName === 'string' && dictName.trim() !== '';
```

Test written first (it failed against the old guard), covering `''`, `'   '` and `'\t\n'` alongside the existing `dictName: 5` and missing-key cases. The comment now records why the trim is there rather than leaving it looking like defensive noise.

client/src/refactoring/renameClassEditor.ts
ericwinger: Added — the header now describes the earlier rejection before it describes `validate`:

> On submit the host first validates the SHAPE of the posted scope with `isRenameClassScope` — a webview message is untrusted — and a scope that fails is rejected before `validate` runs, posting `invalid` back and leaving the panel open just as a bad name does.

Agreed on the reasoning: the header is the first thing anyone (or any agent) reads before touching the handler, so it has to match the contract.

client/src/explorerTreeHelpers.ts
ericwinger: Correct — `pickClassByPrefix` has the single caller in `gemstoneExplorer.ts`, and `moveInstVar` is the up/down reorder, which never picks a destination class. Changed to "the move-to-class destination picker" so it doesn't send anyone hunting for a call site that isn't there.

acceptance/tests/refactoring.spec.ts
ericwinger: Fair — it wasn't coverage, it was intent wearing a test file's clothes, and it counted as a file in the suite while executing nothing. Removed from this PR.

The intent is now tracked as **#431**: https://github.com/GemTalk/Jasper/issues/431 — the three scenarios (ivar rename end-to-end through the Explorer pencil → preview → Apply; the CSP/nonce view-source assertion; the Class Definition History viewer), plus the two things that actually block writing them: capturing the webview locators against a running window, and provisioning a stone with the RB engine in the acceptance container.

gs-src/refactoring/engine/GsExtractSuperclassRefactoring.class.st
ericwinger: Eric: I verified this.  If we split a class to a name that already exists as a global we risk clobbering the global which is bad.

.claude/rules/refactoring-pre-merge.md
ericwinger: (Edited my reply above to say this more plainly, and edits don't send a notification — so, the short version: I agree, the file is deleted from this branch, and every check it contained is written down in **#430** to be turned into a skill later. Nothing was lost.)

gs-src/refactoring/tests/GsInstVarRefactoringTest.class.st
```diff
@@ -554,6 +554,39 @@ GsInstVarRefactoringTest >> testApplyAddInstallsNewVersionWithIvar [
...
+GsInstVarRefactoringTest >> testApplyReportsAnAccessorThatCannotCompileAndInstallsNothing [
+	"A getter/setter that will not compile is surfaced in `failed` -- which holds the commit back --
+	 and is not installed: the mechanism that makes an add-with-accessors all-or-nothing."
+	| refactoring result |
+	refactoring := self add: 'tally'.
+	result := refactoring applyDeselected: #() options: nil migrate: false deleteHistory: false
+		accessors: (Array with: (Array with: 'tally' with: 'tally
+^ )( not parseable')).
+	"failed non-empty -> the empty-array marker is absent"
+	self assert: (result indexOfSubCollection: '"failed":[]') = 0.
```
MatiasFernandez: nit: this class already defines `deny:includesSubstring:` right at the top of the file for exactly this (line 41-43), and the rest of this test file uses it consistently (e.g. `self deny: result includesSubstring: 'keeper'.`). Could this be `self deny: result includesSubstring: '"failed":[]'` instead of the raw `indexOfSubCollection: ... = 0`? Same assertion, but consistent with the established helper.

client/src/refactoring/queries/previewRenameClass.ts
```diff
@@ -9,6 +9,24 @@ import { classLookupExpr, escapeString } from '../../queries/util';
...
+// A #dictionary scope needs a NON-BLANK name: an unselected dropdown posts
+// dictName: '', which would otherwise reach scopeClauseOf as dictionaryScope: '' --
+// a syntactically valid query naming a dictionary that cannot exist.
+export function isRenameClassScope(x: unknown): x is RenameClassScope {
+  if (typeof x !== 'object' || x === null) return false;
+  const kind = (x as { kind?: unknown }).kind;
+  if (kind === 'class' || kind === 'hierarchy' || kind === 'wholeSystem') return true;
+  if (kind === 'dictionary') {
+    const dictName = (x as { dictName?: unknown }).dictName;
+    return typeof dictName === 'string' && dictName.trim() !== '';
```
MatiasFernandez, line 29: 💭 This guards against blank (`''`, `'   '`) but a `dictName` with just leading/trailing whitespace around a real name (e.g. `' UserGlobals'`) passes this check and then flows untrimmed into `scopeClauseOf`'s `dictionaryScope: '...'`. Low-probability from a dropdown, but since the whole point of this guard is "don't trust what the webview posts," should the guard trim-and-store rather than just trim-to-check? Something like returning/using `dictName.trim()` downstream, or trimming in `scopeClauseOf` itself.

MatiasFernandez (approved):
Looks good! just left some minor recommendations

gs-src/refactoring/tests/GsInstVarRefactoringTest.class.st
ericwinger: Good catch — switched to the helper:

```smalltalk
	"failed non-empty -> the empty-array marker is absent"
	self deny: result includesSubstring: '"failed":[]'.
```

That was the only raw `indexOfSubCollection:` left in the file outside the two helper definitions themselves (lines 37 and 42), so the file is now fully consistent — 9 uses of `deny:includesSubstring:` and no hand-rolled equivalents.

One thing worth noting since it isn't visible from this file alone: `resources/refactoring/engine-tests.gs` is generated from these `.st` sources by `gs-src/refactoring/build/build-refactoring.sh`, so I regenerated it rather than hand-editing. The rebuild produced exactly one changed line in the payload, which also confirms there's no source/payload drift on this branch.

client/src/refactoring/queries/previewRenameClass.ts
ericwinger: Agreed, and I took the `scopeClauseOf` option — it's the single choke point every caller funnels through, so the guard stays a pure predicate and the normalization happens where the value actually becomes query text:

```ts
function scopeClauseOf(scope: RenameClassScope): string {
  return scope.kind === 'dictionary'
    ? `dictionaryScope: '${escapeString(scope.dictName.trim())}'`
    : `scope: #${scope.kind}`;
}
```

I also spelled out the split in the comment above `isRenameClassScope`, since "trims to check but doesn't trim to use" was exactly the thing that read as intentional and wasn't:

> Blank-vs-non-blank is all this guard decides; it does not rewrite the value. A name with surrounding whitespace (' UserGlobals') passes here, so scopeClauseOf trims at the point of use rather than trusting the caller.

Plus a unit test asserting `' \tMyDict\n '` reaches the engine as `dictionaryScope: 'MyDict'`.

Scope note: `previewRenameMethod.ts` and `previewChangeSignature.ts` have the identical `dictionaryScope: '...'` line, but neither has a runtime scope guard at all (`isRenameClassScope` is the only one, added in this PR) — so tightening those is a different and larger change than this note, and I left them alone rather than quietly widening the PR. Filed as [#440](https://github.com/GemTalk/Jasper/issues/440) so the other two get brought up to the same standard separately.

PR #423, opened 2026-08-12
GemStone Search (#378): global search UI — bottom panel + editor tab

ericwinger:
Closes #378.

**GemStone Search** — a global "search anything browsable" for the IDE (the Jasper answer to Pharo's Spotter / IntelliJ's Search Everywhere). This is the **first cut**; follow-up issues will track improvements and fixes.

### What you get

Open it with **Shift+Enter** (anywhere in a session; `when: hasActiveSession && !terminalFocus`), the **GemStone Search** status-bar button, or the command palette. A `gemstone.omniSearch.ui` setting picks the experience:

- **`panel`** (default) — a webview docked in the bottom panel (next to Terminal/Output); revealed + focused on login. Results open in the editor area above it.
- **`spotter`** — the same webview in an editor tab (pin = native tab pin, open-beside, dialog-style close on focus-out).
- **`quickpick`** — a native `vscode.QuickPick` fallback.

### Search scopes (provider-per-category)

- Always-on fan-out: **Classes, Methods, Dictionaries, Globals**.
- Explicit "start a search" scopes (heavier, run only when you pick them): **Source** (full method-source substring), **Literals** (methods using a compilable `#symbol`/`'string'` literal — exact), **Categories** (class-category names).
- **Reference search** pivot: a class → *references to* it, a method → *senders* (per-row ↗, or **Alt+Enter**; **←** to go back).

### Ranking

A pure, unit-tested matcher (`fuzzy` | `substring` | `prefix`, case-sensitivity) plus a global cross-category ranker:
- a **contiguous run** (real substring) beats a scattered subsequence;
- **prefix** matches lead; the **first letter's case** picks the kind (upper → class/global names first, lower → methods first);
- same-selector implementors sort **A→Z by class name**.

### UI details

Labeled scope tabs (filters vs. searches, visually split), a flat relevance-ranked list with per-row category tags and our own case-correct highlighting, an always-on case indicator, a clear button, an always-on source-preview pane (with the query highlighted), and a footer with an exact/`N+` count, Load more/all, and an Enter/Alt+Enter legend.

### One Explorer change (looks out of scope, isn't)

The Categories scope surfaced a latent Explorer bug: `revealCategoryByPath` could silently land nowhere. It now focuses the Class Categories view before the nested reveal (a hidden panel view can swallow `reveal()`) and reports a not-found / reveal error instead of swallowing it.

### Architecture

`omniEngine.ts` is a pure, `vscode`-free search engine reused by both webview hosts (`omniSearchPanel.ts`, `omniSearchViewProvider.ts`) via `omniSearchShared.ts`; the DOM lives in a jsdom-tested `omniSearchView.js`; providers/queries are stone-injected so they unit-test without a stone. See `client/src/omniSearch/DESIGN.md`.

### Testing

~130 new/updated unit tests for the engine, matcher, providers, config, and the webview DOM (jsdom). lint / format / compile / `npm test` green (client 5581 · server 322 · mcp 92). Also run through the local CI-mirror matrix on 3.6.2 + 3.7.5 (both plugin worlds, both boundaries) + on-demand GCI: 3.7.5-with-plugin fully green; the only failures were the known load-flaky in-stone SUnit timeouts (rename-class / inline-method / sunit-run-limit) — the very same tests pass on the other boundary — which this client-only change doesn't touch.

---


srbaker (approved):

client/src/omniSearch/DESIGN.md
```diff
@@ -0,0 +1,167 @@
...
+  - `omniMatch`, scoped to a class/subtree. Built here to be reusable; #377 will be a **separate
+    worktree/branch** that calls the scoped backend + adds the Explorer title button. No code moved
+    between issues without a note in both.
+- **#387 items 1–5 (funnel=filter, magnifier=find, wording):** the "find" affordance #387 wants to
+  free up is the same entry point #377 adds. Kept out of this branch; noted for the #377/#387 work.
+
+## Phase 2 — the webview "Spotter" (`ui: spotter`, default)
+
+Phase 2 replaces the QuickPick CHROME with a webview panel while keeping the exact search behaviour.
+The setting **`gemstone.omniSearch.ui`** switches between `spotter` (default) and `quickpick` so the
```
MatiasFernandez, line 119: 🚫 This section is out of date with what the PR actually ships, and since it's the entry point for anyone (or any agent) picking up the follow-ups, the wrong version is worse than none:

- the heading + this line say `ui` switches between `spotter` (default) and `quickpick`, but the shipped setting has three values and defaults to `panel`;
- the module map (lines 84–94) predates half the feature — `omniEngine.ts`, `omniSearchPanel.ts`, `omniSearchViewProvider.ts`, `omniSearchShared.ts`, `omniActions.ts`, `rank.ts`, `references.ts` and the globals/source/literals/categories providers are all missing;
- the Settings list (99–103) is missing `ui` and `methodMinQueryLength`;
- "Deferred / follow-ups (explicitly NOT in this issue)" still lists the Phase-2 webview Spotter and "result highlight ranges rendering", both of which land here.

Could you bring it in sync before merging? The Phase-2 prose lower down is great — it's mainly the pre-Phase-2 sections that need reconciling.

client/src/omniSearch/providers/globalsProvider.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Globals provider: enumerate every non-class symbol-list entry once when the picker opens
+ * (`getAllGlobalNames`), then match client-side on each keystroke — the variables/constants half of
+ * "search any name," alongside classes and method selectors. Activating a global reveals its home
```
MatiasFernandez, line 4: 🚫 "Activating a global reveals its home dictionary" is no longer what happens — `buildOmniHandlers().revealGlobal` runs `gemstone.explorer.findClass` on the *value's* class, and there's even a test named "jumps a global to the class of its value, **not** to the dictionary" asserting the opposite of this doc. Could you update it?

Same issue one file over in [categoriesProvider.ts:6](client/src/omniSearch/providers/categoriesProvider.ts#L6): "Activating a hit reveals its home dictionary (a precise category reveal is a follow-up)" — the follow-up shipped in this PR (`revealCategoryByPath` now selects and reveals the category node).

client/src/gemstoneExplorer.ts
```diff
@@ -3309,6 +3309,75 @@ export class ExplorerController {
...
+    const names = queries.getDictionaryNames(session);
+    const idx = names.indexOf(dictName);
+    if (idx < 0) {
+      void vscode.window.showWarningMessage(`No dictionary matching "${dictName}".`);
+      return;
+    }
+    this.selectDict(new DictItem(dictName, idx + 1));
+    const segment = categoryPath.split('-').pop() ?? categoryPath;
+    const catItem = new ClassCategoryItem(segment, categoryPath, false);
+    this.selectClassCategory(catItem);
```
MatiasFernandez, line 3534: 💭 Should the "is this category actually here?" check happen *before* we mutate the cascade state? Right now `selectClassCategory(catItem)` runs first, and only then does the `allCategoryPaths()` membership test below bail out — so for an unknown category the user gets the warning *and* an Explorer that has already been cascaded to a phantom category (classes pane filtered to nothing). Moving the check above `selectClassCategory` would make the failure path leave the panes untouched, which I think is what the "reported strange spot" comment is trying to avoid.

client/src/gemstoneExplorer.ts
```diff
@@ -3309,6 +3309,75 @@ export class ExplorerController {
...
+    if (idx < 0) {
+      void vscode.window.showWarningMessage(`No dictionary matching "${name}".`);
+      return;
+    }
+    const item = new DictItem(name, idx + 1);
+    this.selectDict(item);
+    try {
+      await this.views?.dict.reveal(item, { select: true, focus: true });
+    } catch {
+      /* ignore */
```
MatiasFernandez, line 3515: 💭 Worth applying the same reasoning you wrote two methods down here? `revealCategoryByPath` deliberately stopped swallowing its reveal failure ("log it so a future failure is diagnosable from the GCI log"), but this one still ends in a bare `/* ignore */` — so a dictionary jump that silently lands nowhere would be exactly as undiagnosable as the category one was. A `logWarning` here too?

client/src/omniSearch/providers/methodsProvider.ts
```diff
@@ -0,0 +1,89 @@
...
+  term: string,
+  limit: number,
+  ignoreCase: boolean,
+) => SelectorSearchResult[];
+
+/** Over-fetch factor: request this many × the displayed cap from the server, so ranking has a
+ *  wider pool to pick the best matches from (see search()). */
+export const SERVER_OVERFETCH = 4;
+/** Hard ceiling on the server slice, to bound the scan cost regardless of maxResultsPerCategory. */
+export const MAX_SERVER_LIMIT = 200;
```
MatiasFernandez, line 25: 💭 `MAX_SERVER_LIMIT` is exported but I don't see a test asserting the clamp, and it's the constant that matters most on the "Load all" path: the engine sets `maxResultsPerCategory` to `LOAD_ALL_LIMIT` (100 000), so without this `Math.min` the provider would ask the stone for 400 000 rows. A test alongside the existing over-fetch one (`p.search(term, cfg({ maxResultsPerCategory: 100_000 }), …)` → `expect(runSearch).toHaveBeenCalledWith(term, MAX_SERVER_LIMIT, true)`) would pin the ceiling that keeps Load-all from running away.

Related user-visible consequence: after Load all the engine reports `exact: true` (omniEngine.ts:336) and the footer renders "200 results" — but for Methods that 200 *is* this server cap, not the real total. Is it worth keeping `N+` when a provider came back exactly at its own fetch ceiling?

client/src/omniSearch/omniSearchViewProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+  private async ensureEngine(): Promise<boolean> {
+    const ctx = await this.resolveContext();
+    if (!ctx) {
+      this.engine = undefined;
+      this.deps = undefined;
+      this.builtForSession = undefined;
+      this.post({ command: 'error', message: 'Log in to a GemStone session to search.' });
+      return false;
+    }
+    if (this.engine && this.builtForSession === ctx.sessionId) return true;
```
MatiasFernandez, line 102: 💭 The docked panel is the default UI, and I think it's the only one that never picks up a settings change. `resolveContext()` re-reads `readOmniConfig(...)` on every call, but this early return throws that fresh config away whenever the session id is unchanged — and `this.deps` isn't refreshed either, so `postView`'s `categories`/`config` stay stale too. The QuickPick and the Spotter both read config at open time so they're fine; the panel keeps the config from first login until the session changes or the window reloads.

So a user who flips `matchMode` to `substring` or raises `maxResultsPerCategory` sees no effect and reasonably concludes the setting is broken. Would an `onDidChangeConfiguration('gemstone.omniSearch')` that clears `this.engine` (letting the next message rebuild it) be enough? The Explorer already does something similar around gemstoneExplorer.ts:5074.

client/src/omniSearch/omniEngine.ts
```diff
@@ -0,0 +1,391 @@
...
+ * Transport-agnostic Omni Search engine for the Phase-2 webview Spotter (issue #378).
+ *
+ * The Phase-1 UI is a native `vscode.QuickPick` driven by `omniSearchController.ts`. Phase 2 replaces
+ * that chrome with a webview panel (`omniSearchPanel.ts`), but the SEARCH behaviour is identical:
+ * hold the active scope + case flag + result cap, fan out to the in-scope providers, rank/group the
+ * results by category, and support the reference pivot. That behaviour lives here — a pure engine
+ * with NO `vscode` dependency, so it unit-tests with plain fake providers and is reused by the panel.
+ *
+ * The engine deliberately does not import the controller (which pulls in `vscode`); the two tiny
+ * pure helpers it shares with the controller (`providersInScope` / `gatherResults`) are re-stated
```
MatiasFernandez, line 11: 💭 This is honest about the trade-off, but by the end there's more restated than "two tiny pure helpers": `providersInScope`, `gatherResults`, `filterPivot`, `LOAD_ALL_LIMIT`, the `effectiveConfig` overlay and the per-category `hasMore` computation all exist in both this file and `omniSearchController.ts`, and `SCOPE_LABEL` exists in a third copy in `omniSearchShared.ts`. None of them touch `vscode` — so a shared `vscode`-free module (next to `rank.ts`, say) could hold them and both hosts could import it, without the engine ever reaching for the controller.

Since the two UIs are meant to be behaviourally identical, "kept in lockstep by hand" is the part that worries me — a future ranking or cap tweak has to be remembered in two places with no test that would catch the drift. Worth consolidating now while the two are still identical, or is the plan to retire the QuickPick path soon (which would make this moot)?

client/src/omniSearch/omniSearchCommand.ts
```diff
@@ -0,0 +1,374 @@
...
+    syncStatus();
+    maybeTip();
+    if (nowActive && !hadSession) {
+      const ui = vscode.workspace
+        .getConfiguration('gemstone.omniSearch')
+        .get<string>('ui', 'panel');
+      // Defer the reveal: `selectSession` fires THIS event BEFORE it sets the `gemstone.hasActiveSession`
+      // context that gates the view's `when` clause, so focusing synchronously finds no view to reveal.
+      // A short delay lets the context propagate (and the login flow settle) before we switch tabs.
+      if (ui === 'panel') setTimeout(() => void viewProvider.focus(), 400);
```
MatiasFernandez, line 359: 💭 A fixed 400 ms is a race we can't win reliably — on a slow login (or a big workspace restore) the context still won't have propagated and the reveal silently does nothing; on a fast one we've inserted a visible delay. Is there a deterministic signal we can wait on instead? Ideas: retry `viewProvider.focus()` once if `this.view` is still undefined, or hook the same `gemstone.hasActiveSession` context that gates the `when` clause so we act after it's set rather than guessing how long it takes. If the timeout is genuinely the only option, could the comment say what happens when 400 ms isn't enough (the panel just doesn't switch — harmless, user presses Shift+Enter)?

client/src/omniSearch/providers/literalsProvider.ts
```diff
@@ -0,0 +1,58 @@
...
+        if (term.startsWith('#') && term.length > 1) {
+          rows = runSymbol(term);
+        } else if (term.startsWith("'")) {
+          const content = parseStringLiteral(term);
+          if (content === null) return []; // still typing an unclosed/empty string
+          rows = runString(content, !cfg.caseSensitive);
+        } else {
+          return []; // not a #symbol or 'string' — the placeholder instructs the user
+        }
+      } catch {
```
MatiasFernandez, line 72: 💭 This `catch` is documented as "not (yet) a compilable expression", but it swallows *every* failure the same way — a dropped GCI connection, a stone error, an `abort` mid-query. In all of those the user has typed a perfectly valid `#at:put:`, sees an empty result list, and has no way to tell "nothing uses this literal" apart from "the search failed". The two other hosts already have an `onError` channel wired (`deps.onError` → `showErrorMessage`); would it be worth letting a non-compile error through to it, or at least distinguishing the compile-error case so the silent path stays narrow?

client/src/omniSearch/__tests__/omniMatch.test.ts
```diff
@@ -0,0 +1,165 @@
...
+    .sort(compareMatches)
+    .map((x) => x.label);
+}
+
+describe('omniMatch — fuzzy (default)', () => {
+  it('matches a subsequence in order and rejects out-of-order / missing chars', () => {
+    expect(match('oc', 'OrderedCollection', opts('fuzzy'))).not.toBeNull();
+    expect(match('ordc', 'OrderedCollection', opts('fuzzy'))).not.toBeNull();
+    // 'co' is NOT a subsequence of 'OrderedCollection' in order? c(9) then o... there is 'o' after
+    // c at index 12 ('Collection' -> C-o-l-l...), so 'co' DOES match. Use a genuine non-match:
```
MatiasFernandez, line 21: This comment reads like a note-to-self from working the case out ("is NOT a subsequence … in order? c(9) then o… so 'co' DOES match") rather than something a future reader needs. Since the assertions below already use the genuine non-matches, could it collapse to something like `// 'co' does match ('C-o' in "Collection"), so use chars that really aren't there.`?

Also `[7, 1 + 7]` on line 33 — is the arithmetic carrying meaning I'm missing, or can it just be `[7, 8]` to match the `[0, 1]` above it?

client/src/omniSearch/omniEngine.ts
```diff
@@ -0,0 +1,391 @@
...
+  }
+
+  async function runSearch(rawValue: string): Promise<OmniViewData | null> {
+    // In the reference view, typing filters the loaded rows client-side (no provider fan-out); don't
+    // touch `lastRawValue` (it holds the search to restore when the pivot is dismissed).
+    if (pivot) {
+      current = filterPivot(pivot.results, rawValue.trim());
+      return pivotView();
+    }
+    lastRawValue = rawValue;
```
MatiasFernandez: 🚫 The display cap never resets when the term changes, and I think that's a real divergence from the QuickPick rather than a deliberate difference. The controller resets it on every keystroke (`scopeLimit = config.maxResultsPerCategory; // a new term starts back at the base cap`, omniSearchController.ts:460); here `runSearch` only ever *reads* it — `setScope` resets, `loadMore`/`loadAll` grow, and a fresh query silently inherits whatever the previous one left behind.

Concretely, in the panel (the default UI): type `si`, click **Load all** (cap → 100 000), then keep typing. Every subsequent keystroke now fans out with `maxResultsPerCategory: 100_000`, so the Methods provider asks the stone for its full `MAX_SERVER_LIMIT` slice on every character instead of ~80, and `exact: scopeLimit >= LOAD_ALL_LIMIT` (line 336) keeps the footer claiming an exact count for a list that is still capped by the providers' server limits. Should `runSearch` reset `scopeLimit` when `rawValue` differs from `lastRawValue`?

Worth a test either way — the load-more test (omniEngine.test.ts:221) grows the cap and then never types again, so nothing pins this direction.

package.json
```diff
@@ -2667,6 +2766,21 @@
         "mac": "cmd+k c",
         "when": "gemstone.hasActiveSession"
       },
+      {
+        "command": "gemstone.omniSearch",
+        "key": "shift+enter",
```
MatiasFernandez: 🚫 Gating `shift+enter` on only `!terminalFocus` means it also shadows the notebook execute-cell binding. We ship a tutorial notebook (`gemstone.openTutorialNotebook`, and the extension activates on `onNotebook:jupyter-notebook`), and inside a notebook cell Shift+Enter is VS Code's `notebook.cell.executeAndSelectBelow`. Extension keybindings take precedence over the built-in defaults when their `when` matches, so with a session logged in — the normal state while following the tutorial — Shift+Enter in a cell would open GemStone Search instead of running the cell.

Should the `when` exclude the notebook (and similar input) contexts too, e.g. `gemstone.hasActiveSession && !terminalFocus && !notebookEditorFocused`? The PR description already calls out the SCM-commit-field case as an accepted trade-off, but that one only shadows a text gesture; this one breaks a feature we ship, so it feels different in kind.

client/src/omniSearch/omniSearchViewProvider.ts
```diff
@@ -0,0 +1,156 @@
...
+  constructor(private readonly resolveContext: () => Promise<OmniViewContext | null>) {}
+
+  resolveWebviewView(view: vscode.WebviewView): void {
+    this.view = view;
+    view.webview.options = { enableScripts: true, localResourceRoots: [] };
+    view.webview.html = renderOmniHtml({ showPin: false });
+    view.webview.onDidReceiveMessage((m: PanelInbound) => void this.onMessage(m));
+    // The engine is session-bound; rebuild it if the session changed while the view was hidden.
+    view.onDidChangeVisibility(() => {
+      if (view.visible) void this.ensureEngine();
```
MatiasFernandez, line 58: 💭 Building on my note about the early return in `ensureEngine` below — I think the same shape leaves the engine bound to a stale *session*, not just stale config.

`ensureEngine` is the only thing that compares `builtForSession`, and it is reached from exactly two places: the `ready` message and this visibility change. The third call site, `if (!this.engine && !(await this.ensureEngine())) return;`, short-circuits as soon as an engine exists. So if the user switches sessions (or logs out and logs back in) while the panel stays visible, no visibility change fires, `ensureEngine` never runs, and the engine keeps searching — and activating results — against the previous `sessionId`.

The class comment says it rebuilds "when the session changes"; should it hook `sessionManager.onDidChangeSelection` (which `registerOmniSearch` already subscribes to) and clear `this.engine` there? That would cover the config-staleness case in the same move.

client/src/queries/methodSearch.ts
```diff
@@ -141,3 +141,70 @@ ${methodSerialization(environmentId)}`;
...
+// `referencesToObject:`) find both — a selector send puts the symbol in the literal frame too — so
+// we subtract the senders. e.g. #size: ~2104 literal-frame refs, ~2090 senders → ~14 pure literal
+// uses. `symbolExpr` is a raw, compilable `#...` expression (evaluated on the server).
+export function literalSymbolReferences(
+  execute: QueryExecutor,
+  symbolExpr: string,
+  environmentId: number = 0,
+): MethodSearchResult[] {
+  const code = `| symLit lit sends methods stream limit classDict sl |
+symLit := ${symbolExpr}.
```
MatiasFernandez, line 174: 💭 This interpolates a string the user is *typing into a search field* straight into evaluated Smalltalk, and the only gate on the way in is `term.startsWith('#') && term.length > 1` (literalsProvider.ts:43). So in the Literals scope, typing `#foo. System abortTransaction` sends both statements to the stone — and it runs per keystroke as you type, so you don't even have to finish the thought.

A Jasper user can of course already evaluate anything in a workspace, so this isn't an escalation of what they're allowed to do; what bothers me is that it's *unintentional* — nobody expects a search box to discard their uncommitted work. The string branch already validates its shape with `parseStringLiteral` before anything reaches the stone; would a matching selector-shaped check on the symbol branch (a unary/keyword/binary selector regex) close this cheaply? The comment here is explicit that the rawness is deliberate, so maybe the validation belongs in the provider rather than in the query.

client/src/queries/methodSearch.ts
```diff
@@ -141,3 +141,70 @@ ${methodSerialization(environmentId)}`;
...
   return parseMethodSearchResults(execute(code));
 }
+
+// Methods that reference the VALUE of a user-typed, compilable literal expression — e.g. `#at:put:`,
+// `42`, `$a`, `#{Globals.Object}`. The expression is compiled and evaluated on the server (it is
+// intentionally raw, not escaped — it IS Smalltalk source), then `referencesToObject:` finds the
+// literal frame. Interned literals (symbols, SmallIntegers, characters, specials, globals) match;
+// a fresh String/Array literal is a distinct object each compile and so matches nothing. A
+// malformed expression makes `execute` raise a compile error, which the caller handles.
+export function referencesToLiteral(
```
MatiasFernandez, line 151: 💭 Is `referencesToLiteral` reachable from anywhere? The only references I can find are its own test and the DESIGN.md line claiming the Literals scope uses it — the provider actually goes through `literalSymbolReferences` / `stringLiteralReferences`.

If it's a deliberate building block for a follow-up, could the comment say so? As it stands it reads as a supported query while being exercised only by its own test — and it's the one with no shape validation whatsoever, since nothing narrows `literalExpr` on the way in.

client/src/omniSearch/providers/classesProvider.ts
```diff
@@ -0,0 +1,42 @@
...
+export function createClassesProvider(
+  sessionId: number,
+  loadEntries: () => ClassNameEntry[],
+): OmniProvider {
+  let entries: readonly ClassNameEntry[] = [];
+  return {
+    category: CATEGORY_BY_ID.classes,
+    prime() {
+      entries = loadEntries();
+    },
```
MatiasFernandez, line 18: 💭 `prime()` loads the class corpus exactly once per engine, which was the right call when the only UI was a QuickPick that opened and closed on every search — but the docked panel is long-lived and its engine is built once per session. So if I compile a new class (or add a global or a dictionary) while the panel is open, it never shows up in search until I switch sessions or reload the window, with no clue as to why.

Is there a signal we could re-prime on? `explorer.onClassCompiled` already fires when a class definition is saved, and `SystemBrowser.refresh(sessionId)` runs after commit/abort. Even a "re-prime on the next query if the corpus is older than the last compile" would avoid a puzzling stale-results report. (`categoriesProvider.ts:21` has the same shape — its lazy cache is never invalidated either.)

package.json
```diff
@@ -132,6 +132,84 @@
...
+          "gemstone.omniSearch.matchMode": {
+            "type": "string",
+            "enum": ["fuzzy", "substring", "prefix"],
+            "enumDescriptions": [
+              "Subsequence match: the typed characters appear in order (the Quick-Open feel).",
+              "The typed text appears as a contiguous run.",
+              "The result starts with the typed text."
+            ],
+            "default": "fuzzy",
+            "markdownDescription": "How Omni Search matches your query against classes, methods, dictionaries and open editors."
```
MatiasFernandez, line 158: nit: "classes, methods, dictionaries and **open editors**" — the Open Editors provider was dropped (DESIGN.md notes it "shipped initially but was **dropped**"), and the list is now also missing globals, source, literals and categories. Since this is a settings-UI string, would something like "…against the enabled search categories" be better, so it doesn't need editing every time the provider set changes?

client/src/omniSearch/omniMatch.ts
```diff
@@ -0,0 +1,192 @@
...
+  }
+  return firstAt < 0 ? null : Array.from({ length: q.length }, (_, i) => firstAt + i);
+}
+
+/**
+ * Fuzzy subsequence match. First prefers a contiguous occurrence (a real substring, see
+ * `contiguousIndices`) so exact runs rank above scattered matches; only when `q` is not a substring
+ * does it fall back to a greedy left-to-right subsequence (predictable and cheap).
+ */
+function matchFuzzy(q: string, t: string): number[] | null {
```
MatiasFernandez, line 133: 💭 Not for this PR — but worth a follow-up issue to explore [`fuzzysort`](https://github.com/farzher/fuzzysort) instead of hand-rolling this. It's the library behind the Sublime/Quick-Open feel we're chasing here, and it's tiny and dependency-free, so it's not much of a supply-chain ask.

The reason I think it's worth a look rather than just "one less thing to maintain": the two places this file compromises are exactly the ones a mature matcher has already solved. `matchFuzzy` falls back to a **greedy left-to-right** subsequence, which the `contiguousIndices` comment openly describes as a workaround (greedy grabs the stray `c` in "Colle**c**tions" and misses a clean later run) — fuzzysort backtracks instead of greedily committing, so it finds the best subsequence rather than the first one. And the weights above (`W_CONTIGUOUS`, `W_WORD_START`, `W_SHORTNESS`, …) are a tuning surface we'd own forever; its scores are normalised 0–1, which also makes `threshold` a natural way to drop junk matches that today only sort to the bottom.

Two things that map onto our shape nicely: `result.indexes` is the same match-position data as our `ranges`, so the webview's case-correct highlighting keeps working unchanged; and `fuzzysort.prepare()` pre-processes a target list once for repeated searches — which is exactly what the load-once corpora (classes, globals, dictionaries) are already doing in `prime()`.

Two things to check before committing to it, hence "explore": whether the `caseSensitive` setting survives (fuzzysort prefers exact-case matches but I don't think it exposes a strict mode), and that `substring`/`prefix` would stay ours either way — so this shrinks `omniMatch.ts` rather than deleting it.

client/src/omniSearch/DESIGN.md
MatiasFernandez: Is the intention to keep this around as documentation? If it was just a support doc while building the feature, we can keep it outside of the repo and later ensure what's important is documented.

package.json
MatiasFernandez: I also think that `shift+enter` may be too common a keyboard shortcut to use as a default for this.

MatiasFernandez (approved):
Nice work! this is really useful. I left some comments worth reviewing and also created this GitHub issue for feedback I got from manually using the feature: https://github.com/GemTalk/Jasper/issues/428

client/src/omniSearch/omniEngine.ts
ericwinger: Eric: Manually verified.  Select Methods in the filter.  Type in a broad search, Load All, then clear the search. Type in a new broad search.  Notice that Load More/Load All don't appear even though they should.

ericwinger: Fixed in 91689d87.

`runSearch` now resets `scopeLimit` to the base per-category cap whenever the typed term actually changes, so Load-all's raised cap no longer leaks into the next search — no more 100k fan-out per keystroke, and the footer no longer falsely reads "exact". The re-run paths (loadMore/loadAll/setScope/toggleCase/exitPivot) pass the unchanged `lastRawValue`, so they keep their raised cap.

It also resets on clear, which is the shared root of the stale-result items in #428 (the "switching scope" / "clear x shows previous results" reports). Regression test added: a fresh term after Load-all comes back capped (hasMore, not exact).

ericwinger: Manually verified fix, too.

package.json
ericwinger: Fixed by moving the binding off `shift+enter` entirely rather than just scoping it: `gemstone.omniSearch` now defaults to **`ctrl+shift+a`** (`cmd+shift+a` on macOS), `when: gemstone.hasActiveSession && !terminalFocus`. A single simultaneous modifier chord can't shadow the notebook cell-run gestures (`shift+enter`/`ctrl+enter`/`alt+enter`), so the tutorial notebook 'execute cell' works again — F5-verified. A `keybindings.test.ts` guard pins the new chord and asserts it is never one of the notebook cell chords. (In the working tree; rides the next push.)

client/src/omniSearch/omniSearchViewProvider.ts
ericwinger: Good catch — fixed. The command layer now registers an `onDidChangeConfiguration` listener for `gemstone.omniSearch` that calls a new `onConfigChanged()` on the provider; that drops the cached engine (and `deps`/`builtForSession`) so the next build re-reads `readOmniConfig(...)`. If the view is visible it rebuilds immediately, so a settings edit shows up live (scope tabs, case flag, limits); if it's hidden the next webview message rebuilds it lazily. Mirrors the Explorer pattern you pointed at.

Regression test added in `omniSearchViewProvider.test.ts` (rebuild-while-visible + defer-while-hidden). Landing in the next push.

client/src/omniSearch/omniSearchViewProvider.ts
ericwinger: Agreed the same early-return shape leaves the engine bound to a stale *session* — but I'm deferring this half to #437 rather than fixing it here, and here's the reasoning:

`gemstone.sessionMode` defaults to `"single"`, so the "switch sessions while the panel stays visible" path only exists once a user opts into `"multiple"`. In single mode the only exposure is keeping the docked panel visible across a full logout→login with no visibility flip in between — a narrow edge that self-heals on the next hide/show.

More to the point, "which session is current, and keeping every tool bound to it" is exactly what #437 ("Multiple sessions have no visible current session — Explorer, Search and other tools each pick their own") is about, so the session-rebinding belongs there with the rest of that work rather than as a one-off here. The config-staleness half you flagged in the sibling thread is fixed now via `onConfigChanged()`.

ericwinger: Eric:  Short answer is that multiple sessions are really well supported in the tools yet.  Deferring to fix this problem when search can support multiple sessions better.

client/src/queries/methodSearch.ts
ericwinger: Done — went with the selector-shaped gate in the provider, as you suggested. The symbol branch now runs a new `isSymbolLiteral(term)` check before anything reaches the stone: the whole term must be a complete, well-formed symbol literal — `#` followed by exactly one of a keyword selector (`#at:put:`), a unary identifier (`#foo`), a binary selector (`#<=`), or a quoted symbol (`#'with spaces'`), anchored `^…$`. So `#foo. System abortTransaction` (trailing statement), `#at:put: bar` (trailing token), and a bare `#` all fail the anchors and never get evaluated — the scope just shows nothing, same as the string branch does for an unclosed `'`.

The raw-eval comment in `methodSearch.ts` stays accurate (the query is still intentionally raw source); the validation now lives in the provider where it belongs. I also made the Literals search hint spell out what's allowed: "Type a symbol literal (#at:put:) or a string literal ('no such element') to find its literal uses".

Unit tests cover the accept set (unary/keyword/binary/quoted incl. `''` escapes) and the reject set (`#`, trailing-token, trailing-statement, unterminated `#'`, `#'a'; evil`).

client/src/omniSearch/omniSearchCommand.ts
ericwinger: Moved to #428 to keep this PR mergeable — tracked there as **400 ms reveal race**, under the *From the #423 code review* section. Follow-up will happen against the issue.

client/src/omniSearch/omniEngine.ts
ericwinger: Moved to #428 to keep this PR mergeable — tracked there as **Consolidate the engine/controller duplication into a shared, vscode-free module**, under the *From the #423 code review* section. Follow-up will happen against the issue.

client/src/omniSearch/providers/methodsProvider.ts
ericwinger: Moved to #428 to keep this PR mergeable — tracked there as **`MAX_SERVER_LIMIT` clamp is untested; footer reads "exact" at the fetch ceiling**, under the *From the #423 code review* section. Follow-up will happen against the issue.

client/src/gemstoneExplorer.ts
ericwinger: Moved to #428 to keep this PR mergeable — tracked there as **Guard the category check before `selectClassCategory`**, under the *From the #423 code review* section. Follow-up will happen against the issue.

client/src/omniSearch/providers/globalsProvider.ts
ericwinger: Moved to #428 to keep this PR mergeable — tracked there as **Globals/categories provider header docs describe old behavior**, under the *From the #423 code review* section. Follow-up will happen against the issue.

client/src/omniSearch/providers/classesProvider.ts
ericwinger: Moved to #428 to keep this PR mergeable — tracked there as **`prime()` loads the corpus once — a new class/global/dict never appears**, under the *From the #423 code review* section. Follow-up will happen against the issue.

client/src/omniSearch/DESIGN.md
ericwinger: Moved to #428 to keep this PR mergeable — tracked there as **Decide DESIGN.md's home — in-repo vs support doc**, under the *From the #423 code review* section. Follow-up will happen against the issue.

client/src/omniSearch/__tests__/omniMatch.test.ts
ericwinger: Moved to #428 to keep this PR mergeable — tracked there as **omniMatch.test.ts comment cleanup + `[7, 1+7]` → `[7, 8]`**, under the *From the #423 code review* section. Follow-up will happen against the issue.

client/src/omniSearch/providers/literalsProvider.ts
ericwinger: Moved to #428 to keep this PR mergeable — tracked there as **Literals `catch` swallows all errors**, under the *From the #423 code review* section. Follow-up will happen against the issue.

package.json
ericwinger: Moved to #428 to keep this PR mergeable — tracked there as **Settings description still says "open editors" and omits scopes**, under the *From the #423 code review* section. Follow-up will happen against the issue.

client/src/queries/methodSearch.ts
ericwinger: Moved to #428 to keep this PR mergeable — tracked there as **`referencesToLiteral` looks dead — document or remove it**, under the *From the #423 code review* section. Follow-up will happen against the issue.

client/src/omniSearch/omniMatch.ts
ericwinger: Moved to #428 to keep this PR mergeable — tracked there as **Explore `fuzzysort` vs the hand-rolled matcher**, under the *From the #423 code review* section. Follow-up will happen against the issue.

client/src/gemstoneExplorer.ts
ericwinger: Moved to #428 to keep this PR mergeable — tracked there as **`logWarning` on a dictionary-jump reveal failure**, under the *From the #423 code review* section. Follow-up will happen against the issue.

client/src/omniSearch/DESIGN.md
ericwinger: Moved to #428 to keep this PR mergeable — tracked there as **DESIGN.md is out of sync**, under the *From the #423 code review* section. Follow-up will happen against the issue.

PR #424, opened 2026-08-13
Fix installer and refactoring-test paths on Windows+WSL

npapagna:
## Summary
- On Windows the gem always runs inside WSL, so it only sees a Windows-drive checkout through WSL's DrvFs mount (`/mnt/<drive>/...`), never the raw `D:\...` path the client resolves. Add `toLocalGemPath()` and use it in the refactoring and enhanced-inspector installers before payload paths cross the GCI connection.
- Every refactoring integration test built its own `GsFileIn` call with a fallback to a GemStone 3.7-only signature. That fallback never fired when client and gem shared a filesystem, but on the Windows client + WSL server topology the first attempt fails (unresolvable path) and the fallback's `to:` argument raises `ImproperOperation 2085` on our 3.6.2 CI stone. Extract one shared helper (`fileInEngineTestsExpr`) that translates the path the same way the production installers do and drops the broken fallback, since `fromServerPath:` alone works on every supported release.

## Test plan
- [x] `npm run lint && npm run format:check && npm run compile && npm test` (against a local 3.6.2 stone)
- [x] New unit tests for `toLocalGemPath`, and WSL-path-translation cases added to the refactoring/enhanced-inspector installer test suites

### DIFF
--- a/client/src/refactoring/__tests__/refactoringInstall.test.ts
+++ b/client/src/refactoring/__tests__/refactoringInstall.test.ts
@@ -5,8 +5,13 @@ vi.mock('../../browserQueries', () => ({
   checkRefactoringSupportAvailable: vi.fn(),
 }));
 
+vi.mock('../../wslBridge', () => ({
+  needsWsl: vi.fn(() => false),
+}));
+
 import { ActiveSession } from '../../sessionManager';
 import { executeFetchString, checkRefactoringSupportAvailable } from '../../browserQueries';
+import { needsWsl } from '../../wslBridge';
 import {
   installRefactoringSupport,
   isRefactoringSupportInstalled,
@@ -149,6 +154,29 @@ describe('installRefactoringSupport', () => {
       String(c[1]).includes('loadFromServerDir'),
     );
     expect(ranLoader).toBe(false);
+  });
+
+  it('translates a Windows-style payload path to its WSL mount before it reaches the gem', async () => {
+    vi.mocked(needsWsl).mockReturnValueOnce(true);
+    const { session } = createMockSession();
+
+    await installRefactoringSupport(session, 'D:\\a\\Jasper\\Jasper\\resources\\refactoring');
+
+    const loaderCode = String(
+      executeFetchStringMock.mock.calls.find((c) =>
+        String(c[1]).includes('loadFromServerDir'),
+      )?.[1],
+    );
+    expect(loaderCode).toContain('/mnt/d/a/Jasper/Jasper/resources/refactoring');
+    expect(loaderCode).not.toContain('D:\\');
+
+    const fileInCode = String(
+      executeFetchStringMock.mock.calls.find((c) => String(c[1]).includes('GsFileIn'))?.[1],
+    );
+    expect(fileInCode).toContain(
+      '/mnt/d/a/Jasper/Jasper/resources/refactoring/refactoring-loader.gs',
+    );
+    expect(fileInCode).not.toContain('D:\\');
   });
 
   it('rolls back and reports failure when the loader file-in raises', async () => {
--- a/client/src/refactoring/refactoringInstall.ts
+++ b/client/src/refactoring/refactoringInstall.ts
@@ -39,6 +39,7 @@ import {
   gsStringLiteral,
   messageOf,
   safeAbort,
+  toLocalGemPath,
   yieldToEventLoop,
 } from '../serverPlugin/installHelpers';
 
@@ -106,8 +107,9 @@ export function isRefactoringSupportInstalled(session: ActiveSession): boolean {
  * engine is committed and the loader's completeness check has already passed.
  *
  * @param session     a session with write access to kernel classes (SystemUser).
- * @param payloadDir  absolute path to `resources/refactoring/`, readable by the
- *                    gem (a local stone).
+ * @param payloadDir  absolute client-side path to `resources/refactoring/`;
+ *                    translated to the gem's local path (see `toLocalGemPath`)
+ *                    before use, so callers must pass it untranslated.
  * @param onProgress  optional incremental progress callback.
  */
 export async function installRefactoringSupport(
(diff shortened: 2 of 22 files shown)


client/src/refactoring/refactoringInstall.ts
```diff
@@ -115,8 +116,9 @@ export async function installRefactoringSupport(
   payloadDir: string,
   onProgress: ProgressReporter = () => {},
 ): Promise<RefactoringInstallResult> {
-  const sep = payloadDir.endsWith('/') ? '' : '/';
-  const serverPath = (file: string): string => `${payloadDir}${sep}${file}`;
+  const gemPayloadDir = toLocalGemPath(payloadDir);
```
MatiasFernandez, line 120: 💭 The JSdoc above (`@ param payloadDir  absolute path to 'resources/refactoring/', readable by the gem (a local stone)`) now describes the old contract — after this change the caller passes the *client-side* path and the function translates it, so the argument itself no longer has to be gem-readable. Could we update it (and the equivalent `@ param payloadDir` in `enhancedInspectorInstall.ts`) to say the path is client-side and gets translated for the gem? Otherwise a future caller may pre-translate and end up double-converting.

client/src/refactoring/refactoringInstall.ts
```diff
@@ -127,7 +129,7 @@ export async function installRefactoringSupport(
       report: '',
       message:
         `The database's gem cannot read the payload files (${unreadable.join(', ')}) under ` +
-        `${payloadDir}. Server-side install requires a local stone whose gem shares this ` +
+        `${gemPayloadDir}. Server-side install requires a local stone whose gem shares this ` +
```
MatiasFernandez, line 133: 💭 Should the user-facing message keep the *client* path? On a Windows client the payload dir the user knows about is `D:\...\resources\refactoring`; after this change the failure says the gem cannot read `/mnt/d/...`, a path they never configured and can't check in Explorer. Since `needsWsl()` is just `isWindows()`, this also fires for a genuinely remote stone reached from Windows, where `/mnt/d/...` is doubly meaningless. Would `${payloadDir}` (or both, e.g. `${payloadDir} (as ${gemPayloadDir} inside WSL)`) read better here and in `enhancedInspectorInstall.ts`?

client/src/serverPlugin/__tests__/installHelpers.test.ts
```diff
@@ -0,0 +1,34 @@
...
+describe('toLocalGemPath', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('translates a Windows drive path to its WSL mount when the gem runs inside WSL', () => {
+    vi.mocked(needsWsl).mockReturnValue(true);
+
+    expect(toLocalGemPath('D:\\a\\Jasper\\Jasper\\resources\\refactoring')).toBe(
+      '/mnt/d/a/Jasper/Jasper/resources/refactoring',
```
MatiasFernandez, line 27: 💭 Is it worth a third case for the `\\wsl$\<distro>\...` UNC form? `toWslPath` handles it (it maps to `/home/...`, not `/mnt/...` — see `wslFs.test.ts`), and that's the shape a checkout living on the WSL side has when it's opened from Windows, so it's a realistic second input to this helper. Cheap to add and it pins the behaviour that `toLocalGemPath` isn't only about drive letters.

client/src/refactoring/__tests__/refactoringInstall.test.ts
```diff
@@ -151,6 +156,21 @@ describe('installRefactoringSupport', () => {
...
+    const { session } = createMockSession();
+
+    await installRefactoringSupport(session, 'D:\\a\\Jasper\\Jasper\\resources\\refactoring');
+
+    const loaderCode = String(
+      executeFetchStringMock.mock.calls.find((c) =>
+        String(c[1]).includes('loadFromServerDir'),
+      )?.[1],
+    );
+    expect(loaderCode).toContain('/mnt/d/a/Jasper/Jasper/resources/refactoring');
```
MatiasFernandez, line 170: 💭 This asserts the `loadFromServerDir:` call, but the install also sends the loader's own `GsFileIn` with `serverPath(REFACTORING_LOADER_FILE)` a step earlier. Given the bug class this PR fixes is "one of the paths crossing GCI wasn't translated", should the test also assert on that first call? Today both derive from the same `gemPayloadDir`, so a regression that translates only one of them would still pass here.

MatiasFernandez (approved):

client/src/refactoring/refactoringInstall.ts
npapagna: Updated the JSDoc in both files to describe the client-side contract: `payloadDir` is now documented as the client-side path, translated via `toLocalGemPath` before it's used, so a future caller knows not to pre-translate it. Thanks for flagging this.

client/src/refactoring/refactoringInstall.ts
npapagna: Good catch on the WSL case, but I'd rather keep this one as `${gemPayloadDir}`: the message should report whatever path was actually handed to the gem, since that's the one relevant to diagnosing an unreadable-file failure regardless of how it got there. Branching the message on `needsWsl()` (or echoing "as X inside WSL") would leak an implementation detail of `toLocalGemPath` into the install layer, and it doesn't even hold up for the genuinely-remote-stone case you mentioned, where "inside WSL" wouldn't be accurate anyway. Reporting the path that was actually used keeps the message correct in both cases without the install function needing to know why the path looks the way it does. Leaving this one as-is.

client/src/serverPlugin/__tests__/installHelpers.test.ts
npapagna: Added a third case asserting `toLocalGemPath('\\\\wsl$\\Ubuntu\\home\\x')` returns `/home/x` when `needsWsl()` is true, matching `toWslPath`'s UNC behavior in `wslFs.test.ts`. Thanks for pointing out the gap.

client/src/refactoring/__tests__/refactoringInstall.test.ts
npapagna: Added an assertion on the loader file-in call too, so a regression that translates only one of the two paths would now be caught here rather than slipping through. Good catch.

PR #425, opened 2026-08-13
Make the GemStone Explorer the browser users meet first (#421)

ericwinger:
Closes #421.

The classic **System Browser** was advertised by two always-present buttons — an inline action
on every session row in **Logins & Sessions**, and a `$(book)` item in the status bar shown
whenever a session was active. Between them, the older five-column browser was the most
discoverable way into a stone's code, ahead of the GemStone Explorer that is now the supported
experience.

## What changed

| | |
|---|---|
| **Buttons removed** | the session-row inline action, and the status-bar `$(book)` item |
| **Feature kept** | `Cmd+K B` (`Ctrl+K B`) and the Command Palette still open it — nothing inside it changed |
| **Command retitled** | **Open System Browser (Classic)**, so anyone finding it in the palette or the keyboard-shortcuts list knows which browser it opens |
| **Remaining row buttons** | renumbered `inline@1`–`@5` rather than left with a hole |

The palette route needed no work: the handler already falls back to `resolveSession()` when
invoked without a tree item, and surfaces *"No GemStone sessions are active. Please log in
first."* when there is none.

## Documentation

Redirecting mentions is only useful if the destination is findable, so this does both.

- **README** gains a **GemStone Explorer** section — panes, editing, filtering, the
  refactorings, senders/implementors, drag-and-drop — placed *before* the System Browser
  section, which now opens with a note that the browser is **frozen**: it works today, it is not
  being extended, and how long it keeps working is not guaranteed. The reasoning lives once, in
  the new `docs/explanation/system-browser-and-explorer.md`, rather than being restated here.
- **The onboarding walkthrough** and the **"Welcome to GemStone Smalltalk"** tutorial notebook
  both sent new users to the System Browser; the tutorial's closing *"where to go next"*
  pointed straight at it. Both now point at the Explorer — **and say where it is.** The
  Explorer has its own activity-bar icon rather than sitting in the GemStone sidebar, so
  naming it alone was sending people hunting in the wrong sidebar.
- **Class sync, the inspector, Jupyter** and the single/multi-session prose now reference the
  Explorer.
- **Two setting descriptions corrected:** `gemstone.classSync.readOnlyMirror` no longer routes
  users to the System Browser to edit, and `gemstone.maxEnvironment` no longer describes itself
  as browser-only when six modules read it (Explorer, browser, `extension.ts`, and the hover /
  CodeLens / Go-to-Definition providers).

**Deliberately not touched:** `CHANGELOG` history (a record of shipped releases — rewriting it
would falsify it), `ROADMAP`'s browser-convergence item (accurate, and about both browsers), and
the `TODO` / agent-rule references to `systemBrowser.ts` as a *source file* rather than as a
user-facing feature.

## Tests

`sessionMenuOrder.test.ts` is updated for the removed inline button. There is no new test file:
the first version of this PR added one, and @MatiasFernandez was right to push back — it carried
the issue ID in its filename and every `describe`, hand-parsed `package.json`, and pinned things
like README section *order* that no edit should have to preserve. For a change that removes UI,
the existing menu-order test plus the type checker carry it.

## Review round

All seven of @MatiasFernandez's points are addressed in `d47fd593`, and the three factual ones
were verified in code first — all three were correct:

- **"run SUnit tests on a class" was wrong.** Only `systemBrowser.ts` dispatches the SUnit
  commands; the Explorer has no such action. Claim dropped, gap filed as
  [#427](https://github.com/GemTalk/Jasper/issues/427).
- **The class-variable claim was generous.** `addClassVar`/`removeClassVar` exist nowhere;
  narrowed to instance variables plus renaming class variables. Having got two wrong, every other
  bullet in that section was re-audited — the rest hold up.
- **The command's `icon` was dead config** with no menu entry left to render it. Dropped.
- Location wording no longer promises a position the user can reorder or overflow.
- **"Frozen"** replaces the contradictory "still maintained, but not supported" — and Eric
  sharpened it further: there is no guarantee, for now, that the browser keeps working, so an
  earlier draft's "its bugs are still bugs" promised maintenance nobody has committed to. That
  claim had reached four places and is corrected in all of them.

## Verification

**Eric manually verified this in a dev host (F5):** the session row shows its five remaining
buttons with no browser action, the status bar no longer carries the book icon, and `Cmd+K B`
still opens the browser.

Automated: lint / format:check / compile clean; client **350 files / 5416 passed**, server 322,
mcp 92.

No engine, payload or `.gs` file is touched, so the RB SUnit and GCI layers cannot be affected —
the full CI matrix was not run for that reason.


client/src/__tests__/browserDiscoverability421.test.ts
```diff
@@ -0,0 +1,185 @@
+import { describe, it, expect } from 'vitest';
```
MatiasFernandez: 🚫  This is a very weird test, clearly generated by Claude trying to fill in tests here. It contains the Issue ID everywhere (even in the file name) and also asserts weird stuff like manually parsing the package.json. It is very fragile and tbh, not super useful. I would remove the file completely. We don't need tests for a change that is only removing stuff.

client/src/__tests__/sessionMenuOrder.test.ts
```diff
@@ -46,6 +45,10 @@ describe('session row inline button order', () => {
     ]);
   });
 
+  // The classic System Browser's absence from this row (#421) is guarded in
```
MatiasFernandez: 🚫  Same note as above: I would remove this comment; it's not useful. If we want to ensure devs understand the System Browser is not being maintained or not considered the main UX, we can add a short doc to the docs folder explaining the rationale.

README.md
```diff
@@ -165,9 +164,26 @@ By default, **Display It** shows its result as a non-destructive inline overlay
...
+
+Selecting down the panes narrows what the next one shows. Click a method to open its source; **Cmd+S** (Ctrl+S) compiles it back to GemStone. Class definitions and comments are editable the same way.
+
+Beyond browsing, the Explorer is where the code-changing operations live:
+
+- Filter any pane by name, with `*` as a wildcard, plus `reads:`/`writes:`/`accesses:` in the Methods pane to find the methods touching an instance variable
+- Group methods by category, or list them flat
+- Add, rename, and delete dictionaries, class categories, classes, methods, and instance/class variables
+- The refactorings — rename, extract/inline method and temporary, change signature, move/push up/push down method, instance-variable structure changes, extract superclass, split class — each previewed before it is applied
+- Browse senders, implementors, references, and the class hierarchy; run SUnit tests on a class
```
MatiasFernandez: 🚫 I think this bullet promises something the Explorer doesn't have. `gemstone.runSunitClass` / `runSunitClasses` / `runSunitMethods` are only ever dispatched from `systemBrowser.ts` (plus `extension.ts` where they're registered) — there's no "Run SUnit Tests" action anywhere in the Explorer's class context menu, and `gemstoneExplorer.ts` never mentions SUnit.

That makes this the one place where "prefer the Explorer" actually costs the user a feature. Should we drop the claim from this bullet (and track this missing feature as an issue)?

Related, one line up: the bullet says we can add/rename/delete "instance/class variables", but I only see `addInstVar` / `removeInstVar` / `renameClassVariable` — no add/remove for class variables. Is that meant to be covered by "class definitions are editable", or is that bullet a bit generous too?

client/src/tutorialNotebook.ts
```diff
@@ -300,7 +300,7 @@ Run these read-only cells to look around. (This lesson changes nothing permanent
 
 **Where to go next:**
 - Open a **Workspace** (command: *GemStone: Open Getting Started Workspace*) for free-form experimenting with *Display It*, *Inspect It*, and *Debug It*.
-- Browse the image in the **System Browser** and the **Globals Browser**.
+- Browse and edit the image in the **GemStone Explorer** (its own icon in the activity bar, beside the GemStone one), and inspect globals in the **Globals Browser**.
```
MatiasFernandez: 💭 "beside the GemStone one" is a nice touch for findability, but is it something we can promise? Activity-bar containers get reordered by the user (drag, or pushed into the `···` overflow when the window is short), and the position also depends on what other extensions contribute. A first-timer whose bar doesn't match could end up more confused than if we'd just said where to look generically.

Would something like "its own **GemStone Explorer** icon in the activity bar (the far-left strip)" be safer? Same note applies to the matching sentence in `resources/walkthrough/inspectIt.md`.

package.json
```diff
@@ -874,7 +874,7 @@
       },
       {
         "command": "gemstone.openBrowser",
-        "title": "Open System Browser",
+        "title": "Open System Browser (Classic)",
         "category": "GemStone",
         "icon": "$(list-tree)"
```
MatiasFernandez: Now that this command has no menu contribution at all, the `icon` is dead config — the Command Palette and keybindings don't render it. Worth dropping so nobody later reads the icon as evidence there's still a button somewhere?

README.md
```diff
@@ -165,9 +164,26 @@ By default, **Display It** shows its result as a non-destructive inline overlay
...
+- Drag and drop methods between categories, and classes between dictionaries
+
 ### System Browser
 
-Open with **Cmd+K B** (Ctrl+K B) or from a session's inline button. The browser provides a five-column layout:
+> **Note:** the System Browser is the older, five-column browser that predates the **GemStone Explorer** above. It still works and is still maintained, but the Explorer is the supported experience and gets the new features — prefer it unless you specifically want this layout.
```
MatiasFernandez: 💭 "It still works and is still maintained, but the Explorer is the supported experience" reads a little contradictory to me — if it's maintained, in what sense is it not supported? Between this, the `(Classic)` retitle, and removing both buttons, I read the actual intent as "frozen: it keeps working, but new work goes to the Explorer".

Should we say that outright ("it is no longer being extended — new features land in the Explorer") so a reader knows what they're opting into? Also ties into the point I left on `sessionMenuOrder.test.ts`: whatever we decide the story is, this note is a better home for it than a test comment.

MatiasFernandez (approved):
I'm approving this one but I would love for that test to not be included

client/src/__tests__/browserDiscoverability421.test.ts
ericwinger: Agreed and done — file deleted in `d47fd593`. You're right about the fragility specifically: two of its assertions were scanning `extension.ts` as text and pinning the *order* of README sections, neither of which would survive an innocuous edit. Test count moved by exactly −1 file / −14 tests, and nothing references it.

client/src/__tests__/sessionMenuOrder.test.ts
ericwinger: Removed. The comment only existed to point at the file you asked me to delete, so it went with it — and I took your `docs/` suggestion: the rationale now lives in `docs/explanation/system-browser-and-explorer.md`, indexed from `docs/README.md` and the README's Documentation table. Much better home than a test comment, and it's now the single place that explains the decision rather than it existing only in this PR description.

README.md
ericwinger: You're right on both counts — I checked before changing anything, and both were my errors.

**SUnit:** `gemstone.runSunitClass` / `runSunitClasses` / `runSunitMethods` are dispatched only from `systemBrowser.ts` (plus registration in `extension.ts`), `gemstoneExplorer.ts` contains **zero** SUnit references, and there's no Explorer-scoped SUnit menu entry. Claim dropped, and filed as #427 — with your framing of the two sizes (a run action on a test class with a results webview now, a proper SUnit Runner later). It's also recorded as the known browser-only gap in the new doc.

**Class variables:** also generous — `addClassVar` / `removeClassVar` exist in zero files; only `addInstVar`, `removeInstVar`, `renameClassVariable`. Narrowed to "…and instance variables; rename class variables".

Since I got two claims wrong I re-audited **every** other bullet in that section — dictionaries (they use `new`/`remove`/`rename`, which is why my first grep for `addDictionary` came back empty and looked like a third error), filtering, grouping, the refactorings, senders/implementors/references/hierarchy, drag-and-drop. The rest hold up.

client/src/tutorialNotebook.ts
ericwinger: Fair — I over-specified. Reworded to your phrasing in both places (`tutorialNotebook.ts` and `resources/walkthrough/inspectIt.md`): *"its own icon in the activity bar (the far-left strip)"*. Position isn't ours to promise once the user drags things around or the `···` overflow kicks in.

package.json
ericwinger: Dropped. Confirmed it's unreachable — no menu entry references the command any more, and the palette and keyboard-shortcuts list don't render command icons. Your reasoning is the better half of the argument: leaving it invites a later reader to infer a button still exists somewhere.

README.md
ericwinger: Good catch, and you read the intent correctly — **frozen** is what it is. Eric then sharpened it further: there's no guarantee, *for now*, that the browser keeps working, so my first attempt ("it keeps working and its bugs are still bugs") promised maintenance nobody has committed to. That claim had propagated to four places and is corrected in all of them.

It mattered most in the new doc's *"why frozen rather than deprecated or maintained"* section, because the **argument** rested on the false promise. Now it claims only what's been decided: *maintained* overclaims (neither feature parity nor repair of breakage is committed), *deprecated* overclaims (no removal decision, no date), and frozen is what's true — **no new work goes in, and no guarantee comes out**.

And agreed on placement: that story lives in `docs/explanation/system-browser-and-explorer.md` now, with a short pointer from the README note, rather than in a test comment.

PR #426, opened 2026-08-13
Clean up docs/: version-mapped GCI header snapshots, Enhanced Inspector rationale, dead-file removal

MatiasFernandez:
Cleanup pass over `docs/`: everything under it that wasn't documentation moved to where it belongs, the stale files went away, and the one genuinely undocumented subsystem (the Enhanced Inspector install) got a written explanation.

## Vendor snapshots of the GCI headers (`vendor/gci-headers/`)

`docs/3.7/` held a single, ambiguously-named copy of the GCI headers — no record of which GemStone release it came from, and nothing to check it against. That's now `vendor/gci-headers/`, with **one folder per distinct header content revision**, named after the earliest GemStone version that shipped it. `versions.md` maps every locally-installed version (3.6.2 → 3.7.5) to its folder with a sha256 per file, so several patch releases that ship byte-identical headers share one folder instead of duplicating it.

Two reasons this is worth the ~23k lines:

- **Correctness for agents.** The headers are the authoritative reference for GCI function signatures, struct layouts, and constants. With one unlabeled snapshot, an agent reading them had no way to know whether a signature it found actually applies to the stone in front of it. Version-mapped folders make that answerable, and `.claude/rules/client/gci.md` now points at the map.
- **Version gating in GCI tests.** These snapshots are the input to a separate project that consolidates the GCI library's cross-version limitations — which calls, structs, and constants exist in which releases. The output of that work is better version gating in our GCI-related integration tests, which today guess at version floors rather than deriving them.

`vendor/**` is excluded from the packaged `.vsix` — reference material, not shipped.

## Enhanced Inspector: tooling out of `docs/`, rationale into `docs/`

`docs/enhancedInspectorSupport/` was build scripts and generated-payload notes, not documentation.

- **Dead content deleted**: the unused Pharo/GT perf tracker, the AcmeSystem demo `.gs` (that schema lives outside this repo), a stray `.gitignore`, and the standalone `load_enhanced_inspector_support.sh` — which had drifted from the shipping TS installer (it never gained the legacy-`Published` migration) and which nothing automated called. The **Install Server Support** command is the supported path.
- **Build tooling moved** to `gs-src/enhancedInspector/build/`, mirroring `gs-src/refactoring/build/`, with a source-of-truth README naming the four upstream feenk/GemTalk repos it builds from — and flagging that `apply_jasper_transforms.sh` is **broken on macOS** (GNU `sed -i` syntax; it destroys the committed payload and still prints `Update complete.`). I tracked that issue [here](https://github.com/GemTalk/Jasper/issues/429). A generated-payload README landed in `resources/enhancedInspector/` to mirror `resources/refactoring/`.
- **New explanation doc**: [`docs/explanation/enhanced-inspector.md`](docs/explanation/enhanced-inspector.md), registered in `docs/README.md`. It consolidates rationale that previously existed only as TS comments — why the payload gets its own `GsEnhancedInspector` dictionary, why it's vendored and transformed instead of Rowan-loaded, why `ENHANCED_INSPECTOR_FILES` order is load-bearing, the real cause of the 3.7.5 version gate (pre-3.7.5 GCI-compiled string literals are Unicode and won't `=`-compare against the payload's byte-`String` keys, so lookups silently return nothing), and why the legacy-`Published` migration is gated on a marker. The TS comments now point at the doc rather than duplicating it.

## Also

Deleted three orphaned reference files nothing referenced: `docs/BNF.txt`, `docs/Tokens.txt`, `docs/export_classes.txt`.

No runtime behavior changes — the only non-doc code edit is doc comments in `enhancedInspectorInstall.ts`.


gs-src/enhancedInspector/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Regenerating
+
+```sh
+export ROWAN_PROJECTS_HOME=/path/to/your/checkouts   # gt4gemstone, gtoolkit-remote,
+                                                       # gtoolkit-wireencoding, RemoteServiceReplication
+gs-src/enhancedInspector/build/update_enhanced_inspector_support.sh
+```
+
+> ⚠️ **Known broken on macOS.** Both `sed -i` calls in `apply_jasper_transforms.sh` (lines 47 and 51)
```
npapagna: The macOS-breakage warning cites "lines 47 and 51" for the two `sed -i` calls in `apply_jasper_transforms.sh`, but this diff's own edit to that script (expanding a comment block by 2 lines) shifted those calls to lines 49 and 53.

gs-src/enhancedInspector/README.md
```diff
@@ -0,0 +1,37 @@
...
+> use GNU syntax; BSD sed consumes the next argument as a backup suffix. Line 47 silently no-ops the
+> placement rewrite and leaves a stray `<file>.gs-E`; line 51 hard-errors. Because `update_…sh` copies
+> pristine upstream over the payload *before* transforming and has no `set -e`, running it on macOS
+> **destroys the committed payload and still prints `Update complete.`** Run on Linux until this is
+> fixed. See issue #NNN.
```
npapagna: Ends with a literal, unfilled placeholder "See issue #NNN" rather than a real issue number. GemTalk/Jasper#429 ("Payload transform scripts are broken on macOS and corrupt the committed payload") already tracks this.

npapagna (approved):
I really like the direction in this PR.
I think is essential to have these gci headers, too. 👏 

Left a couple of comments from Claude.

vendor/gci-headers/README.md
```diff
@@ -0,0 +1,14 @@
+# GCI headers
+
+Vendor snapshots of the GCI (GemStone C Interface) header files — `gci.ht`, `gcicmn.ht`, `gcits.hf`, and (from 3.7.2 onward) `gcits.ht` — copied verbatim from a GemStone server distribution's `include/` directory. They are the authoritative reference for GCI function signatures, struct layouts, and constants; not authored by this project, so don't edit them.
```
ericwinger: 💭 The snapshot stops one level short of the headers whose constants we actually hand-transcribe.

All four vendored files `#include` headers that aren't here — `gci.ht` pulls in `gcioc.ht`, `gcioop.ht`, `gcierr.ht`, `flag.ht`, `l2unix.hf`; `gcicmn.ht` adds `version.ht`; `gcits.hf` adds `gcirtl.hf` and `gcistring.hf`. So "the authoritative reference for GCI function signatures, struct layouts, and **constants**" holds for signatures and structs, but following the include chain for a constant dead-ends.

That matters concretely for the agent-correctness argument in the PR body, because `client/src/gciConstants.ts` names the missing header as its source ~13 times:

```ts
// OOP tag bits (from gcioop.ht)
// Well-known OOP values (from gcioop.ht)
export const OOP_ILLEGAL = 0x01n;
// Class OOPs (from gcioop.ht)
export const OOP_CLASS_STRING = 74753n;
export const OOP_CLASS_UTF8 = 154113n;
```

Those literal class OOPs are exactly the kind of value where "does this apply to the stone in front of me?" is a real question and a version-mapped snapshot pays off — and they're the one thing a reader can't check here. `gcierr.ht` is the same story for the GCI error numbers we branch on in `client/src/queries/util.ts` and friends.

Adding `gcioop.ht`, `gcioc.ht`, and `gcierr.ht` per folder would close it (they're right there in the same `include/` dir, and small). If they're deliberately out of scope, worth narrowing the sentence to signatures/structs and saying which headers were left out, so the next reader doesn't conclude the constants are absent from the vendor distribution the way `gcits.ht` genuinely is pre-3.7.2.

gs-src/enhancedInspector/build/update_enhanced_inspector_support.sh
```diff
@@ -92,8 +93,9 @@ cp "$ROWAN_PROJECTS_HOME/gtoolkit-remote/src-gs/gtoolkit-remote.gs"   "$PAYLOAD_
 # no headers). See apply_jasper_transforms.sh.
 echo ""
 echo "Applying Jasper transforms (attribution headers + Globals->Published)..."
```
ericwinger, line 95: Two leftovers from the old `Published` placement in this file — everything else in the PR (the sibling script, all the TS doc comments, the new docs) moved to `GsEnhancedInspector`, but these didn't:

```
90:#   - class placement rewrite from Globals to Published
95:echo "Applying Jasper transforms (attribution headers + Globals->Published)..."
```

Line 95 is the one that stings: the script prints a claim about what it just did that is wrong, and `apply_jasper_transforms.sh` right next to it now rewrites to `GsEnhancedInspector` (and even carries a `sed` whose whole job is scrubbing the stale `Globals to Published` note out of already-headered payload files). Suggest matching the sibling's wording:

```sh
#   - class placement rewrite from Globals to GsEnhancedInspector
...
echo "Applying Jasper transforms (attribution headers + Globals->GsEnhancedInspector)..."
```

resources/enhancedInspector/README.md
```diff
@@ -0,0 +1,37 @@
...
+|---|---|
+| `Announcements.gs` | feenk's Announcement framework, used by the remote-inspector notifications |
+| `RemoteServiceReplication.gs` | GemTalk's service-replication bootstrap connecting the stone to the remote-inspector transport |
+| `STON.gs` | Smalltalk Object Notation serialization, used to wire-encode inspector data |
+| `patch-gemstone.gs` | Small kernel patches GT needs that can't be shipped as extension methods |
+| `gtoolkit-wireencoding.gs` | The wire-encoding scheme for GToolkit's remote protocol |
+| `gt4gemstone.gs` | The core GT4GemStone remote-inspector views and bindings |
+| `gtoolkit-remote.gs` | The GToolkit remote-inspector protocol (`RemotePhlow` objects and friends) |
+
+This order is load-bearing — it must match `ENHANCED_INSPECTOR_FILES` in
```
ericwinger, line 32: 💭 This table re-introduces the duplication the rest of the PR just removed.

`load_enhanced_inspector_support.sh` went away partly because it carried a second copy of the load order that had drifted, and the doc comment on `ENHANCED_INSPECTOR_FILES` was rewritten in this same PR to say it "is the sole authority on load order" — dropping the old "must match the `input` order in ..." cross-reference. This table is now a second hand-maintained copy of that order, and the sentence below it re-establishes exactly the must-match coupling that was just retired. Nothing enforces it.

(I checked — the seven rows are in the right order against `ENHANCED_INSPECTOR_FILES` today.)

The per-file "what it is" column is genuinely useful and worth keeping; it's the ordering claim that's the liability. Sorting the table alphabetically and describing order only by pointing at the array, or adding "listed here for orientation; the array is authoritative," would keep the value without the second source of truth.

vendor/gci-headers/versions.md
```diff
@@ -0,0 +1,27 @@
...
+| 3.6.8 | [3.6.8](3.6.8/) | `f60dd662` | `58cf1136` | `44353b0d` | — |
+| 3.7.0 | [3.7.0](3.7.0/) | `ec9cdc27` | `239fb201` | `29c62182` | — |
+| 3.7.1 | [3.7.1](3.7.1/) | `7e8614c5` | `bff40429` | `537643af` | — |
+| 3.7.2 | [3.7.2](3.7.2/) | `41f7c518` | `03d82497` | `6c9ff61e` | `d287f74c` |
+| 3.7.2.2 | [3.7.2](3.7.2/) | `41f7c518` | `03d82497` | `6c9ff61e` | `d287f74c` |
+| 3.7.4.1 | [3.7.4.1](3.7.4.1/) | `b4883655` | `cb4fdb1d` | `5ca21076` | `7058fb85` |
+| 3.7.4.3 | [3.7.4.1](3.7.4.1/) | `b4883655` | `cb4fdb1d` | `5ca21076` | `7058fb85` |
+| 3.7.5 | [3.7.5](3.7.5/) | `c3af9c02` | `ec9dd5cb` | `46e936d5` | `98381241` |
+
+Hashes are truncated to the first 8 hex characters for readability; run `shasum -a 256` on a file to verify the full digest.
```
ericwinger, line 21: 💭 Nothing checks this map against the folders it describes, and the map is the whole load-bearing artifact — an agent that trusts a wrong row reads the wrong header and gets a confidently wrong answer, which is worse than having no snapshot.

For the record, I verified all 33 hashes in this table against the committed files and every one matches, including the four byte-identical pairs in the Notes. So it's correct as landed; the concern is only about staying that way as rows get added by hand.

A ~10-line check would pin it — for each row, hash the four files in the named folder and assert the truncated digests match. Worth considering as a repo script (or a `format:check`-style step) so a mistyped row fails loudly rather than misleading silently.

Two nits while here:

- Step 2 and step 4 of the README's "Adding a new version" both say "four header files", but every pre-3.7.2 folder holds three (`gcits.ht` doesn't exist yet, as the first Note says). Only bites someone back-filling an older release, but the two docs contradict each other.
- The install path pattern on line 27 is macOS/arm64-only (`GemStone64Bit<version>-arm64.Darwin`). Devs and CI on Linux find them at `~/Documents/GemStone/GemStone64Bit<version>-x86_64.Linux/include/` — worth generalizing to `GemStone64Bit<version>-<arch>.<platform>/include/` since whoever adds the next row may not be on a Mac.

docs/explanation/enhanced-inspector.md
```diff
@@ -0,0 +1,34 @@
...
+
+The tradeoff, made explicit in `gs-src/enhancedInspector/README.md`, is that the source of truth for this payload lives outside this repo, in `$ROWAN_PROJECTS_HOME`, and regenerating it is a human-run, non-CI step.
+
+## File load order is load-bearing
+
+`ENHANCED_INSPECTOR_FILES` in `enhancedInspectorInstall.ts` lists the payload files in dependency order — earlier files define classes and behavior that later files depend on — and files in that order over a single session, so a later file can compile against classes an earlier file just installed. That array is the sole authority on load order (there is no other manifest to keep in sync with it); reordering it without checking the payload's actual dependencies risks a file-in failure partway through, which aborts the whole install rather than leaving a partial one committed.
```
ericwinger, line 22: nit: this sentence lost its subject in the middle — "`ENHANCED_INSPECTOR_FILES` ... lists the payload files in dependency order ... **and files in that order** over a single session". The array doesn't file anything; the installer does. Reads like an edit that dropped a clause.

Suggest: "... — and the installer files them in that order over a single session, so a later file can compile against classes an earlier file just installed."

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ All notable changes to the **GemStone Smalltalk** extension will be documented i
...
 
+### Removed
+
+- **The standalone `load_enhanced_inspector_support.sh` developer script.** It had drifted from the shipping installer (it never gained the legacy-`Published` migration) and nothing automated called it; the **Install Server Support (Enhanced Inspector + Refactoring)** command is the supported install path.
```
ericwinger, line 13: 💭 Should this be in the CHANGELOG at all? The file opens with "All notable changes to the **GemStone Smalltalk** extension", and this script was never in a user's hands — `docs/**` is excluded from the packaged `.vsix`, so no released extension ever shipped it. A "Removed" entry tells readers of the release notes that something they never had is gone.

The removal is well-justified in the PR body and the `gs-src/enhancedInspector/README.md` / `resources/enhancedInspector/README.md` pair documents the supported path for the developers this actually affects, which feels like the right audience. Your call — if the convention here is that internal tooling changes get logged too, then it's fine as-is and I'd leave it.

gs-src/enhancedInspector/README.md
MatiasFernandez: Fixed

gs-src/enhancedInspector/README.md
MatiasFernandez: Fixed

resources/enhancedInspector/README.md
```diff
@@ -0,0 +1,37 @@
+# Enhanced Inspector payload (generated — do not edit by hand)
+
+These `*.gs` files are **vendored, generated build artifacts**, not source written for this repo.
```
ericwinger, line 3: 💭 This README is the right place to notice that the shipped Enhanced Inspector payload isn't covered by `THIRD-PARTY.md`.

`THIRD-PARTY.md` has exactly one section — "Pharo AST-Core (Refactoring Browser AST)" — covering the refactoring engine's vendored source, and it inlines the full MIT License text at line 34 to do it. These seven files are third-party MIT code too (feenk gmbh for six of them, GemTalk's RemoteServiceReplication for the seventh, per the table in `apply_jasper_transforms.sh`), and unlike the AST they are **shipped verbatim in the VSIX** — `resources/**` isn't in `.vscodeignore`, as this README itself points out on line 7.

What the payload carries today is a generated per-file pointer:

```
! License: MIT - Copyright (c) feenk gmbh. See LICENSE in the origin repository.
```

That names the license and the holder, which is most of the way there, but it directs the reader *out* of the distribution for the license text — whereas the MIT terms ask that the notice and permission text travel with the copies. `THIRD-PARTY.md` is already the mechanism this repo uses for exactly that, so the gap is really just that the Enhanced Inspector never got a section.

Not asking you to fix it in a docs-cleanup PR, and it's pre-existing rather than introduced here. But this PR is the first time the vendored payload gets a "what is this and where does it come from" README, and the four upstream repos + copyright holders are now written down in two places (`gs-src/enhancedInspector/README.md` and the transform script's table) — so the raw material for a `THIRD-PARTY.md` section exists as of this PR and won't be easier to assemble later. Either add the section here or a follow-up issue would close it out; a one-line "third-party provenance is recorded in `THIRD-PARTY.md`" pointer from this README would then keep the two in each other's orbit.

resources/enhancedInspector/README.md
```diff
@@ -0,0 +1,37 @@
...
+They are the GemStone-side [feenk GToolkit](https://github.com/feenkcom) remote-inspector support
+that [`enhancedInspectorInstall.ts`](../../client/src/enhancedInspector/enhancedInspectorInstall.ts)
+files into a stone (server-side, via `GsFileIn`) when a user runs the extension's "Install Server
```
ericwinger, line 6: nit: "when a user runs the extension's 'Install Server Support' command" is the user-facing path, but it isn't the only one — and the omission matters more than usual in this PR, because the script being deleted here (`load_enhanced_inspector_support.sh`) was the command-line installer, and both this README and the CHANGELOG entry point the reader only at a GUI command in its place.

`npm run test:server:install-plugin` → `client/bin/installServerPluginMain.ts` → `installServerPlugin` reaches the same `installEnhancedInspectorSupport` through `serverPlugin/pluginFeatures.ts`, headlessly and outside the extension host, with the 3.7.5 gate applied ("Version-gated features (Enhanced Inspector) are skipped below their minimum, exactly like the client's own auto-install path").

For what it's worth I went looking for a capability regression here and didn't find one — that script really is fully covered, including the shell-driven "provision a stone without opening the editor" use case, so the deletion is well-justified. It's only the docs that don't say so. Something like "…via the extension's *Install Server Support* command, or headlessly with `npm run test:server:install-plugin` for a test stone" would spare the next person the same search.

gs-src/enhancedInspector/build/apply_jasper_transforms.sh
```diff
@@ -21,15 +21,17 @@
...
 # USAGE:
 #   ./apply_jasper_transforms.sh [target-dir]   # defaults to the payload dir
 
 set -e
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-TARGET_DIR="${1:-$SCRIPT_DIR/../../resources/enhancedInspector}"
+REPO="$(CDPATH= cd -- "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)"
+TARGET_DIR="${1:-$REPO/resources/enhancedInspector}"
 
 SENTINEL="! Jasper Enhanced Inspector vendored source"
```
ericwinger, line 37: Anchoring here because `SENTINEL` is the mechanism, but the issue is the header block below at lines 66–69: the header this script emits doesn't match the header the committed payload actually carries. All seven files in `resources/enhancedInspector/` currently say

```
! installer. DO NOT EDIT BY HAND - regenerated from upstream by
! update_gemstone_gt_support.sh, which re-applies this header and rewrites
```

and `update_gemstone_gt_support.sh` exists nowhere in the repo — not before this PR, not after. So a reader who follows the new `resources/enhancedInspector/README.md` into a payload `.gs` is told to regenerate with a script that isn't there, while this script (line 68) emits the correct `update_enhanced_inspector_support.sh`.

It stays wrong because of the sentinel check on line 58: the header is prepended only when the sentinel is *absent*, so on an already-headered file the whole block is skipped and a stale line inside it is never refreshed. You already hit that exact limitation for the placement note and worked around it with the targeted `sed` on line 53 — the generator name needs the same one-liner next to it:

```sh
    # Refresh the stale generator name in an already-headered file (same reason
    # as the placement note above: the header block itself is never re-emitted).
    sed -i 's/^! update_gemstone_gt_support\.sh,/! update_enhanced_inspector_support.sh,/' "$path"
```

Worth doing in this PR specifically because this is the change that renames the tooling's home — otherwise the payload keeps citing a script name that now predates two relocations. (A real regeneration from upstream would also fix it, but that's a human-run non-CI step, so it could be a long while.)

vendor/gci-headers/versions.md
ericwinger: I'm not sure that an agent confidently giving the answer is anything new but if we can avoid it, it might be worth doing.

ericwinger (approved):
Overall, pretty clean. Mostly looked at Enhanced Inspector changes and the stuff claude found looks legit, if relatively minor.

gs-src/enhancedInspector/build/apply_jasper_transforms.sh
MatiasFernandez: Good catch — confirmed, all seven payload files cited `update_gemstone_gt_support.sh`, which has never existed in this repo. Fixed the seven files directly; they now say `update_enhanced_inspector_support.sh`, matching what the generator emits.

vendor/gci-headers/versions.md
MatiasFernandez: All three done. README now distinguishes three files before 3.7.2 from four after, and the install path is generalized to `GemStone64Bit<version>-<arch>.<platform>/include/` (matching the three platform keys `sysadminStorage.ts` already computes).

Added `scripts/lint-gci-header-versions.mjs`, wired as `npm run lint:gci-headers` and run in the health-check `lint` job — it hashes each row's files and also catches a row pointing at a missing folder or a folder no row references. Went with a standalone `lint:*` script rather than `format:check`, since Prettier's glob is ts/js only.

vendor/gci-headers/README.md
MatiasFernandez: Good catch — done, full back-fill rather than narrowing the sentence. `gcioop.ht`, `gcioc.ht`, and `gcierr.ht` are now in every folder, `versions.md` has three more digest columns, and the checker passes unchanged (it reads file names from the table header row).

Two things worth flagging from doing it:

- **3.6.2 and 3.6.3 were not actually the same snapshot.** They shared a folder because their `gci.ht`/`gcicmn.ht`/`gcits.hf` are byte-identical — but `gcioc.ht` and `gcierr.ht` differ between them. So 3.6.3 got split into its own folder. That's a nice argument for your point: the old snapshot would have silently handed a 3.6.3 user 3.6.2's error numbers.
- **The include chain now closes.** `gcioop.ht` and `gcioc.ht` include nothing at all, and `gcierr.ht` includes only `gcioop.ht`, so there's no second dead-end. `flag.ht`, `l2unix.hf`, `version.ht`, `gcirtl.hf`, and `gcistring.hf` are still out — they're build-flag/platform/version-stamp plumbing, not constants we transcribe — and `versions.md` now says so by name, per your suggestion.

Small corrections on the specifics: `gcioop.ht` is named 3 times in `gciConstants.ts`, not ~13, and `queries/util.ts` doesn't branch on error numbers. But `gcierr.ht` is a stronger case than you made it — the two magic numbers we hardcode are defined there under their real names: `debuggerPanel.ts:722` `GS_ERR_UNCONTINUABLE = 6011` is `RT_ERR_UNCONTINUABLE_ERROR`, and `gciCompile.integration.test.ts:123` `2213` is `RT_ERR_MUST_BE_SYSTEMUSER`. Both are now checkable against the right version.

Cost is ~1.6 MB of vendored text, mostly `gcierr.ht` at ~113 KB a copy. Seemed worth it for the values where "does this apply to the stone in front of me?" is the actual question.

resources/enhancedInspector/README.md
MatiasFernandez: Fixed — this was a real second-source-of-truth risk (verified the table still matches `ENHANCED_INSPECTOR_FILES` today, and confirmed the array's own doc comment no longer references the table). Kept the table's dependency order for its orientation value, but reframed the coupling: heading is now just "## The payloads", and the paragraph below the table now reads:

> Rows above are listed in current dependency order for orientation only. `ENHANCED_INSPECTOR_FILES` is the sole authority on load order — nothing keeps this table in sync with it, so if the two ever disagree, trust the array.

No enforcement claim left to drift out of sync.

resources/enhancedInspector/README.md
MatiasFernandez: Agreed — tracking this as a follow-up issue rather than doing it in this docs-cleanup PR.

gs-src/enhancedInspector/build/update_enhanced_inspector_support.sh
MatiasFernandez: Good catch — fixed. Updated both lines in `update_enhanced_inspector_support.sh` to say `GsEnhancedInspector` instead of `Published`, matching `apply_jasper_transforms.sh`:

```diff
-#   - class placement rewrite from Globals to Published
+#   - class placement rewrite from Globals to GsEnhancedInspector
 ...
-echo "Applying Jasper transforms (attribution headers + Globals->Published)..."
+echo "Applying Jasper transforms (attribution headers + Globals->GsEnhancedInspector)..."
```

resources/enhancedInspector/README.md
MatiasFernandez: Fair point — the README (and the CHANGELOG) only mentioned the GUI path. Updated the wording to also call out the headless route:

> "...files into a stone (server-side, via `GsFileIn`) via the extension's "Install Server Support" command, or headlessly with `npm run test:server:install-plugin` for a test stone."

docs/explanation/enhanced-inspector.md
MatiasFernandez: Good catch — that was a dropped subject. Applied your suggested wording:

`docs/explanation/enhanced-inspector.md:22`
```diff
-and files in that order over a single session, so a later file can compile against classes an earlier file just installed.
+and the installer files them in that order over a single session, so a later file can compile against classes an earlier file just installed.
```

CHANGELOG.md
MatiasFernandez: Agreed — removed the CHANGELOG entry. It never shipped in a released extension, so it doesn't belong in release notes read by the audience this file is written for.

PR #438, opened 2026-08-14
Migrate GCI e2e tests off the e2e suite (phase 5b): allowedCommits budget

MatiasFernandez:
## Summary

Continues the GCI e2e test migration by giving `useIntegrationTest` an **`allowedCommits` budget**, then using it to move tests that commit off the slow, never-in-CI `gci` e2e suite and into the regular integration suite (which runs in CI across the release matrix).

- **`useIntegrationTest(callback, { allowedCommits: N })`** — declares how many `System commitTransaction`s the suite's tests, and the production code they exercise, perform. `beforeEach` opens `N` levels of `System beginNestedTransaction` plus one level of headroom, so those commits land in a nested level instead of the harness's own transaction; `afterEach` unwinds every remaining level in one round-trip and fails the test if the level fell to the budget's floor, so a test that commits past its budget fails loudly instead of silently corrupting the next test's session. A `login()` from inside a test re-establishes both the test transaction and the budget on the new session. `allowedCommits` is validated as a non-negative whole number, and `0` (the default) is an explicit discontinuity: nothing is nested and the existing commit guard is the sole enforcement.
- Adds `oopToInteger`/`executeAndFetchInteger` to `gciLibrary`, decoding integers directly off the oop via `GciTsOopToI64` (bigint, exact) instead of a lossy `printString` + `Number()` round-trip. `debugQueries`' private `oopToInt` folds into it, which also fixes its ignoring of the wrapper's success flag.
- New `useIntegrationTest.integration.test.ts` covers the harness itself: the commit guard's existing behavior, the budget containing single and sequential commits, session cleanliness after nesting, re-login under a budget, the config validation, and the explicit-zero case.
- **Tests moved out of `gciInstVar.e2e.test.ts`**: the delete-history commit test → `refactoringInstVar.committing.integration.test.ts` (under a budget); the accessor-fail+migrate test → `refactoringInstVar.integration.test.ts` as a plain test (the accessor-compile failure gates `commitStructuralThenMigrate:`, so nothing ever commits). The e2e suite is down to migrate-instances and the two accessor-atomicity scenarios.
- Extracts shared Utf8-oop-cache assertions (`__tests__/support/utf8OopCache.ts`) and instVar-refactoring probes (`refactoring/__tests__/support/refactoring.ts`) so the split test files stay DRY.
- Documents the mechanism in `docs/reference/integration-test-harness.md` (options table, hook order, "aborts spend the budget too") and `docs/explanation/integration-test-isolation.md` (why nesting contains a commit, which committing tests it does *not* cover, the zero-vs-`N` discontinuity), and updates `.claude/rules` so a plain commit no longer automatically justifies writing a test in `gci/`.

Note for reviewers: an earlier iteration of this branch exposed the mechanism as a `commitStrategy: 'nested'` discriminated-union option with a `commitDepth`. That was collapsed into the single numeric `allowedCommits` in the last commit — no strategy union, no `useIntegrationTest` overloads, unified teardown, and the off-by-one in the public API fixed (the harness adds its own headroom rather than asking authors to budget for it).

## Test plan

- [x] `npm test` (client/server/mcp-server) against a live stone


client/src/__tests__/useIntegrationTest.ts
```diff
@@ -62,11 +89,28 @@ type LoginOptions = {
  * of each suite and restored afterward, so a local GemStone installation is
  * unaffected outside of test runs.
  */
-export function useIntegrationTest(callback: UseIntegrationTestCallback) {
+export function useIntegrationTest(callback: UseIntegrationTestCallback): void;
+export function useIntegrationTest(
+  callback: UseIntegrationTestCallback,
+  options: UseIntegrationTestOptions,
+): void;
```
npapagna: do we need overloading here? I believe that using options? is enough.

npapagna (commented):
Automated review pass: 6 findings, ranked most-severe first.

**Conventions (confirmed), `client/src/refactoring/__tests__/refactoringInstVar.integration.test.ts:31`** — not anchorable inline since it falls outside any changed hunk: the file's docstring still claims the committing paths (migrate instances, delete history) "cannot live here" and "stay in `__tests__/gci/gciInstVar.e2e.test.ts`", which this same diff makes false — it adds `refactoringInstVar.committing.integration.test.ts` using the new `commitStrategy: 'nested'` harness feature precisely to host one of them. Every other touched file's header comment was updated to reflect the split except this one.

client/src/__tests__/useIntegrationTest.ts
```diff
@@ -110,41 +154,80 @@ export function useIntegrationTest(callback: UseIntegrationTestCallback) {
...
+      } else {
+        gciLibrary.abortTransaction(session);
+      }
       gciLibrary.resetNonTransactionalSessionState(session);
     } catch (error) {
       sessionCleanupFailed = true;
       throw error;
     }
+
+    if (options?.commitStrategy === 'nested' && levelBeforeAbort! <= floorLevel!) {
```
npapagna: **Correctness (confirmed):** this budget check silently no-ops if `floorLevel` is `undefined` or stale, because `bigint <= undefined` is always `false` in JS (verified in Node: `5n <= undefined` -> `false`, no throw).

If `openNestedCommitBudget()` throws its own guard error (line ~271) during a test's `beforeEach`, `floorLevel` is never (re)assigned for that iteration — `undefined` on the first test in the file, or stale from an earlier test on any later one. Vitest still runs `afterEach` after a `beforeEach` throw, so this check runs anyway with a floor that no longer describes a real transaction. The budget-violation guard — the harness's whole "fail loudly instead of silently absorbing a regression" promise — gets silently skipped.

client/src/__tests__/useIntegrationTest.ts
```diff
@@ -154,6 +237,52 @@ export function useIntegrationTest(callback: UseIntegrationTestCallback) {
...
+
+    const commitDepth = options.commitDepth;
+
+    gciLibrary.executeDiscardingResult(
+      session,
+      `${commitDepth} timesRepeat: [System beginNestedTransaction]`,
+    );
+    const nestedLevel = gciLibrary.executeAndFetchInteger(session, 'System transactionLevel');
+
+    if (nestedLevel < commitDepth + 1) {
```
npapagna: **Correctness (plausible):** this only validates the nesting depth is at least `commitDepth + 1`, never that it's exactly that — a stray extra open level is silently absorbed into `floorLevel` instead of being caught.

If a session ever carries one extra open transaction level into `openTestTransaction()` (e.g. a doit under autoBegin implicitly opening one, as the surrounding comments note doits can do), `nestedLevel` comes back one higher than expected. The check `nestedLevel < commitDepth + 1` passes anyway, and `floorLevel = nestedLevel - commitDepth` is computed one level too high — silently shifting the whole budget window instead of failing loudly the way the sibling too-few-levels branch does.

client/src/__tests__/useIntegrationTest.ts
```diff
@@ -110,41 +154,80 @@ export function useIntegrationTest(callback: UseIntegrationTestCallback) {
...
       // committed and left the session dirty, so the cleanup below is not guaranteed to succeed. Bail out early
       assertCommitGuardIsStillArmed(session);
 
-      gciLibrary.abortTransaction(session);
+      // A single abortTransaction unwinds only one level, same as a commit --
+      // under the default strategy that one level is the whole transaction,
+      // but under 'nested' there can be several left open, so it takes one
+      // abort per remaining level to fully unwind rather than leaving the
+      // next test's beginTransaction to fail against a still-nested session.
+      if (options?.commitStrategy === 'nested') {
```
npapagna: **Simplification:** `options?.commitStrategy === 'nested'` is repeated as an inline guard 5 times across this file (lines 113, 180, 196, 229, 260) instead of being narrowed once.

This ties directly into the `floorLevel`/undefined finding above: because "we're in nested mode" isn't narrowed into a single typed value once, `floorLevel`/`levelBeforeAbort` stay optional everywhere and get non-null-asserted (`!`) at the point of use — the kind of duplication that let that gap go unnoticed. Worth folding into the same fix.

client/src/__tests__/useIntegrationTest.integration.test.ts
```diff
@@ -82,3 +83,96 @@ describe('integration test harness commit guard (integration)', () => {
...
+ * be a whole count of at least one level. Anything else describes a budget that
+ * cannot be opened, and the harness refuses it where the suite declares it --
+ * rather than accepting it and failing every test in that suite at teardown,
+ * blaming tests that did nothing wrong.
+ */
+describe('nested-transaction commit strategy configuration', () => {
+  const declaringASuiteWithCommitDepth = (commitDepth: number) => () =>
+    useIntegrationTest(() => {}, { commitStrategy: 'nested', commitDepth });
+
+  it.each([0, -1, 1.5, NaN, Infinity])('refuses a commitDepth of %s', (commitDepth) => {
```
npapagna: **Test naming (confirmed):** `'refuses a commitDepth of %s'` mirrors the internal `commitDepth` field name instead of describing the guarantee in plain language, per the repo's test-naming convention (`.claude/rules/tests.md`). Something like `'refuses a commit budget that isn't a whole count of at least one level'` would state the guarantee instead of echoing the option name.

client/src/refactoring/__tests__/support/refactoring.ts
```diff
@@ -30,3 +30,31 @@ const engineTestsPayloadPath = (): string =>
  */
 export const fileInEngineTestsExpr = (): string =>
   `GsFileIn fromServerPath: '${escapeString(toLocalGemPath(engineTestsPayloadPath()))}'.`;
+
+/**
+ * Small Smalltalk probes shared by the instance-variable and instance-variable-structure
+ * refactoring integration suites. Each takes the caller's `exec` so it works against any
```
npapagna: **Reuse (plausible, optional):** the new `userIndex` helper duplicates a Smalltalk snippet still copy-pasted verbatim in ~12 other files (`refactoringMoveMethod`, `addClassVariable`, `addAccessors`, `refactoringInlineTemporary`, `refactoringExtractMethod`, `refactoringExtractTemporary`, `refactoringTemporary`, `refactoringClassVariable`, `refactoringInlineMethod`, `classVarQueries`, `refactoring.integration.test.ts`, `explorerQueries.integration.test.ts`, and `gci/gciInstVar.e2e.test.ts`), only wired into 2 call sites here. Not blocking — flagging as a follow-up cleanup opportunity rather than scope for this PR.

client/src/__tests__/useIntegrationTest.ts
npapagna: Related to this one: how would you feel to reify the commit strategy? That might remove all the conditionals and clean up useIntegrationTest.

npapagna (approved):
🚀

PR #442, opened 2026-08-14
Explorer UX: filter icons and wording, filterable method categories, and honest class-comment affordances (#387)

ericwinger:
Closes #387.

A review pass over the reworked GemStone Explorer produced a 13-item checklist of UX tweaks. This
implements the ten that were worth doing; the other three were resolved on the issue, each with its
reasoning recorded there.

## Filter and find

- **Funnel means filter, magnifier means find** (items 1, 2, 3). The filter controls use `$(filter)`;
  the "filter is active" indicator carries the same funnel, so the indicator and the control no
  longer disagree about what concept they represent. The find action is retitled **Find Class in
  Explorer** — it does its work in the Explorer, and the old "Find Class in Browser Panels" was both
  inaccurate and confusable with the global **Find Class…**.
- **An active filter is easier to notice** (item 4). The indicator row is prefixed `Filter:`, so it
  reads as a statement about the tree's state rather than as another button.
- **The prompt no longer contradicts the behaviour** (item 5). The input said Enter confirms, which
  set up the expectation that nothing happens until you press it — but filtering is live. The live
  behaviour is unchanged; the message that misdescribed it is gone.
- **Method categories are filterable, not just selectors** (item 7). Typing `acc` surfaces the
  `accessing` category *with its methods* rather than an empty folder. Category names match on a
  prefix and honour the `*` wildcard, in both the grouped and flat views; a category reachable only
  by one of its selectors still works; an instance-variable-token filter deliberately does not pull
  in whole categories.

## Method list and categories

- **The flat-view toggle says what it does** (item 8). "Show Methods Without Categories" read as
  "the methods that have no category"; it is now **Don't Group Methods by Category**. The inverse
  label was already clear and is untouched.
- **The `** ALL METHODS **` pseudo-category is gone** (item 10). It duplicated what the real
  categories already showed and, being first and expanded by default, pushed them below the fold —
  so switching classes meant scrolling before any category was visible. No method is lost: every
  selector is still listed under its own category, uncategorized ones under "as yet unclassified",
  with the flat/grouped toggle as the see-everything view. The internal enumerate-everything lookup
  key is retained; only the visible pseudo-category was removed.

## Class metadata

- **A class comment opens as a preview, and only where there is one** (item 11). It used to open
  *pinned*, so a quick read cost two clicks to dismiss; it now opens as a preview tab, reused by the
  next comment and still promotable by double-click. The command is retitled **Open Comment**, since
  "Pin Comment" would describe behaviour it no longer has.

  The button also used to appear on every class — and this was worse than a button that sometimes
  does nothing. Since GS/64 3.1 a class comment lives under the `#comment` key of the class's extra
  dict, and `Class>>comment` *synthesises* `No class-specific documentation for X` plus a rendered
  hierarchy when that key is absent. Clicking it on an uncommented class opened an invented
  document, and no emptiness test on the returned string can detect the case. The class list now
  reports a `hasComment` flag read from the key itself, and the row offers the button only when it
  is set.

  Cost, since the issue raised it: **~0.12µs per class**, under 1ms cold for all 769 classes in
  `Globals`, and **no extra round trip** — the query listing a dictionary's classes already visits
  each one for its category, so the check rides along in that loop.

  The **Classes-pane toolbar** button is deliberately *not* gated. If every entry point were hidden
  on uncommented classes, a class's first comment could never be written from the Explorer.

- **An empty variable section is omitted, not grayed** (item 12). A class with instance variables
  but no class variables shows only the **instance variables** header, and vice versa; a class with
  neither shows nothing. This overrules the earlier convention of showing both headers and graying
  the empty one — a header for a definitively-empty section reads as something to open, and opening
  it is the only way to learn there was nothing there. Graying required a synthetic `resourceUri`
  plus a `FileDecorationProvider`; that is now unreachable and removed along with
  `explorerVarDecorations.ts`.

## Resolved on the issue rather than here

- **Item 6** (Escape does not cancel filtering) was the only outright defect; it was split out to
  #388 and shipped there.
- **Item 9** (rename "Collapse all" to "Collapse all categories") is declined. That label is VS
  Code's own built-in tree collapse-all action, enabled by `showCollapseAll: true` — there is no
  contributed command title to reword. Retitling would mean replacing the built-in with a custom
  command purely to change a hover, making this tree behave differently from every other tree in the
  editor.
- **Item 13** (method documentation on hover) is deferred. The method hover already carries
  Implementors and Senders as one-click links plus the override and session-method lines, and most
  methods have no comment to show, so a comment would push the useful part down without adding much.
  The stronger version of that want — an **Explain this method** action derived from the code and
  its senders, implementors and covering tests — is recorded on #441.

## A note for reviewers

The class row's `contextValue` gains a `.commented` variant, so the ten other class actions were
widened from `viewItem == explorerClass` to `viewItem =~ /^explorerClass(\.commented)?$/`. That
regex is anchored on purpose: a bare `^explorerClass` would also match `explorerClassVar` and put
class actions on class-variable rows. A manifest test evaluates every one of those clauses against
`explorerClass`, `explorerClass.commented` and `explorerClassVar`, and asserts the comment button is
the *only* difference between a commented and an uncommented row — a wrong clause would otherwise
silently drop an action from the context menu with nothing to catch it.

## Testing

Client-only TypeScript — no `gs-src`, `resources` or `server` changes, so no engine rebuild and
nothing for the RB SUnit layer to exercise.

- `npm run lint`, `npm run format:check`, `npm run compile` — clean.
- `npm test` — **378 client files / 5676 passed**, 322 server, 92 mcp-server, 0 failures, against a
  live 3.6.2 stone. Pre-commit and pre-push hooks ran and passed (no `--no-verify`).
- Roughly 380 lines of new tests: the filter/category behaviour, the `hasComment` query and its
  parse, the variable-side invariant, and a manifest test pinning the icons, titles and `when`
  clauses so a future edit cannot quietly undo items 1–4, 8 or 11.
- Items 11 and 12 were also exercised by hand in a running editor.


client/src/__tests__/getClassesWithCategory.test.ts
```diff
@@ -5,14 +5,45 @@ describe("listing a dictionary's classes with their categories", () => {
...
+    expect(code).toContain('_extraDictAt: #comment');
+    // `v comment` would always answer a non-empty string, making every class look
+    // commented — and would build a hierarchy report per class while doing it.
+    expect(code).not.toMatch(/\bv comment\b/);
+  });
+
+  it('keeps a tab in a category name from shifting the class name', () => {
+    // Category comes first and the class name is taken from the LAST tab, so the
+    // parse stays anchored even if a category ever contains one.
+    const execute = vi.fn().mockReturnValue('Odd\t1\tArray\n');
```
MatiasFernandez: 🚫 I think this test doesn't verify what its name and comment claim. The fixture `'Odd\t1\tArray\n'` has no tab inside the category name, so it's really just re-asserting the happy path — and the comment says the class name is taken from the *last* tab, but the parse takes the *second* one (`line.indexOf('\t', tab + 1)`).

With an actual tab in the category, e.g. `'Od\td\t1\tArray\n'`, you get `category: 'Od'`, `hasComment: false` (`'d' !== '1'`) and `className: '1\tArray'` — the shift the test is named after.

Parsing from the right would make the claim true: `className = line.slice(line.lastIndexOf('\t') + 1)`, then the flag from the character before it, then the category is everything left. Worth doing along with a fixture that actually contains the tab? Otherwise I'd drop the test and the comment rather than leave an invariant advertised that doesn't hold.

client/src/gemstoneExplorer.ts
```diff
@@ -1758,6 +1769,17 @@ export class ExplorerController {
...
 
+  // Whether a class carries a real comment — drives whether the row offers the
+  // comment button at all (#387 item 11), so the button never promises a document
+  // that turns out to be GemStone's synthesised "No class-specific documentation
+  // for …" placeholder. Answered from the class list already fetched for this
+  // dictionary, so asking costs no extra query. A class we have no entry for
+  // (a stale row, or one from another dictionary) is treated as uncommented: the
+  // Classes-pane toolbar button still reaches it, so nothing becomes unreachable.
+  classHasComment(className: string): boolean {
+    return this.classCategoryEntries.some((e) => e.className === className && e.hasComment);
```
MatiasFernandez: 💭 This is an O(n) scan, and `ClassProvider.getChildren` calls it once per class row — so the Classes pane goes quadratic in class count. For the 769 classes in `Globals` that's ~300k comparisons per render, which is a good deal more than the ~0.12µs/class the description measures on the engine side.

Its two siblings on the same row are O(1) map lookups — `classHasDefinedVars` reads `definedIvarCounts`/`definedClassVarCounts`, `classVersion` reads `classVersions`. Should this follow that pattern, e.g. a `commentedClasses: Set<string>` rebuilt wherever `classCategoryEntries` is assigned? (`classCategory` uses `.find` too, but that's a single lookup rather than one per row.)

client/src/__tests__/explorerPaneFilterChip.test.ts
```diff
@@ -20,8 +20,8 @@ function makeController(): ExplorerController {
   ctl.state.dictName = 'Globals';
   ctl.state.dictIndex = 2;
   (ctl as unknown as { classCategoryEntries: ClassCategoryEntry[] }).classCategoryEntries = [
-    { category: 'Collections', className: 'Array' },
-    { category: 'Kernel', className: 'Object' },
+    { category: 'Collections', className: 'Array', hasComment: true },
+    { category: 'Kernel', className: 'Object', hasComment: false },
```
MatiasFernandez: 💭 The `hasComment` flag looks well covered at both ends — the query test pins the Smalltalk and the parse, and the manifest test pins the `when` clauses — but I couldn't find anything covering the wiring in between: that an entry with `hasComment: true` actually ends up as a `ClassItem` with `contextValue === 'explorerClass.commented'`.

That path is a bit easy to break silently: `hasComment` is the 4th positional parameter of `ClassItem` and defaults to `false`, so a dropped argument or a flipped ternary degrades to "no comment button anywhere" with every test still green. This fixture already seeds one commented and one uncommented class — would asserting the two `contextValue`s off `classProvider.getChildren()` here be enough to close it?

client/src/queries/getClassesWithCategory.ts
```diff
@@ -22,16 +33,21 @@ dict ifNil: [^ ''].
 ws := WriteStream on: String new.
 dict keysAndValuesDo: [:k :v |
   v isBehavior ifTrue: [
-    | cat |
+    | cat cmt |
     cat := [v category] on: Error do: [:e | nil].
     (cat isNil or: [cat isEmpty]) ifTrue: [cat := 'as yet unclassified'].
-    ws nextPutAll: cat asString; tab; nextPutAll: k; lf]].
+    cmt := [(v _extraDictAt: #comment) notNil] on: Error do: [:e | false].
```
MatiasFernandez: 💭 `notNil` treats a stored-but-empty comment as a real one. If someone opens the comment editor, deletes the text and saves, the `#comment` key exists with `''` — so the row keeps offering the button and it opens an empty document, which is close to the "button that promises a document it can't deliver" this item set out to remove.

Would `[(v _extraDictAt: #comment) ifNil: [false] ifNotNil: [:c | c isEmpty not]]` (or similar) be worth it? The reasoning for avoiding `Class>>comment` still holds — this checks the stored value, not the synthesised one.

client/src/gemstoneExplorer.ts
```diff
@@ -3104,7 +3138,15 @@ export class ExplorerController {
...
-        (info) => filter === undefined || this.methodMatchesFilter(isMeta, info.selector, filter),
+        (info) =>
+          filter === undefined ||
+          this.methodMatchesFilter(isMeta, info.selector, filter) ||
+          // A category-name match keeps that category's methods here too (#387 item 7).
+          // Without this, filtering 'accessing' listed the category in grouped mode and
+          // then emptied the pane the moment the user turned grouping off, even though
+          // the filter had not changed. Ivar-token filters are excluded for free --
+          // methodCategoryMatchesFilter answers false for them.
+          this.methodCategoryMatchesFilter(info.category, filter),
```
MatiasFernandez: 💭 `methodCategoryMatchesFilter` re-runs `this.parseFilter(raw)` on every call, and here it's called once per method in the flat list — so the same pattern gets parsed N times for one render. Should the parse be hoisted out of the callback (parse once, pass the parsed filter in, or memoize on the raw string)?

Related: the comment at the `hasMatch` site above describes name-matching as "a string compare" to justify putting it first, which undersells it a little — it parses the filter before comparing. Might be worth rewording so the ordering rationale still reads accurately.

client/src/__tests__/explorerFilterUx387.test.ts
```diff
@@ -0,0 +1,301 @@
...
+      .methodCategories(false)
+      .flatMap((c) =>
+        (ctl.methodProvider.getChildren(c) as MethodItem[]).map((r) => r.info.selector),
+      )
+      .sort();
+
+    // Exactly what the pseudo-category used to enumerate.
+    expect(reachable).toEqual(
+      ctl
+        .selectorsFor(false, ALL_METHODS_CATEGORY)
```
MatiasFernandez: 💭 Both sides of this assertion come from `selectorsFor`, so the test passes vacuously if that ever returns `[]` — "nothing reachable" would equal "nothing enumerated" and the no-methods-lost claim wouldn't be tested at all. The very next test already hardcodes the full list; would asserting `reachable` against that same literal (or adding a non-empty precondition on it) make this one bite?

MatiasFernandez (approved):

client/src/gemstoneExplorer.ts
ericwinger: Eric: Manually verified with a log and ooh-boy that's a lot of checks.

ericwinger: Eric: Did some verification of the comment icon too. Icon stays & updates nicely (after fix)

client/src/queries/getClassesWithCategory.ts
ericwinger: Eric: Manually verified. Book icon sticks around

client/src/__tests__/getClassesWithCategory.test.ts
ericwinger: Both true, thanks — fixed in `837ad33a`.

The parse now reads from the right, as you suggested: a class name cannot hold a tab but a class category is free text, so the last tab ends the flag and the one before it ends the category. The fixture is `Od\td\t1\tArray` now, which under the old parse gave `category: 'Od'`, `hasComment: false`, `className: '1\tArray'` — every field wrong. So the test bites, and the comment above it is true.

client/src/gemstoneExplorer.ts
ericwinger: Confirmed and fixed in `837ad33a` — thanks, this was the real one.

`ClassProvider.getChildren` does call it once per row, so a 769-class `Globals` was doing ~300k comparisons per render while its two row siblings were map lookups. It is now a `commentedClasses: Set<string>`, as you suggested.

One wrinkle worth naming: `classCategoryEntries` is assigned in eight places (dict switch, refresh, class create, reveal, external compile, …), so a hand-maintained set had eight chances to go stale. The set is rebuilt by a private setter on `classCategoryEntries` instead, which leaves every existing assignment site untouched and means there is no site to forget — including the tests that poke the field through a cast. Five new cases in `explorerClassHasComment.test.ts` cover the staleness paths specifically (reload adds a comment, dictionary switch drops a class, entries cleared).

Manually F5-verified by Eric against a 769-class Globals.

client/src/queries/getClassesWithCategory.ts
ericwinger: Good catch — fixed in `837ad33a`, and it reproduced exactly as you described.

Confirmed on a live stone: `cls comment: ''` STORES the empty string rather than dropping the key, so `notNil` stayed true and the button survived its own comment being deleted.

I went slightly further than `isEmpty not`: the test is now "any non-whitespace character", because a save can leave a lone newline behind (VS Code's insert-final-newline) and a whitespace-only comment is no more readable than none:

```smalltalk
cmt := [(v _extraDictAt: #comment)
          ifNil: [false]
          ifNotNil: [:c | (c detect: [:ch | ch isSeparator not] ifNone: [nil]) notNil]]
        on: Error do: [:e | false].
```

Your reasoning for avoiding `Class>>comment` still holds and is unchanged. Since none of this — the synthesised placeholder, the stored `''` — is visible to a mocked unit test, there are four new cases in `explorerQueries.integration.test.ts` pinning it against a real stone: never commented, real text, emptied, whitespace-only. Manually F5-verified by Eric too.

client/src/gemstoneExplorer.ts
ericwinger: Taking the second half, pushing back on the first.

`parseFilter` already memoizes on the raw string, and has since `e5e9386d` (#259 2a):

```ts
private parsedFilter?: { raw: string; filter: MethodFilter };
private parseFilter(raw: string): MethodFilter {
  if (this.parsedFilter?.raw !== raw) this.parsedFilter = { raw, filter: parseMethodFilter(raw) };
  return this.parsedFilter.filter;
}
```

So a render parses once per distinct filter string; the per-method calls are a string compare and a field read. No change made there.

Your second point is fair though, and is fixed in `837ad33a`: the note now reads "a cached-parse lookup plus a string compare (parseFilter re-parses only when the raw filter string changes)", so the ordering rationale says what it actually is.

client/src/__tests__/explorerFilterUx387.test.ts
ericwinger: Right — fixed in `837ad33a`. It now asserts against the same written-out list the next test uses:

```ts
expect(reachable).toEqual(['name', 'name:', 'printOn:', 'printString', 'size']);
```

with a comment saying why, so nobody helpfully DRYs it back into a `selectorsFor` call later.

Issue #428, opened 2026-08-13
Omni Search (PR #423): manual-testing feedback — naming, shortcuts, scopes, stale-result bugs, UI polish


MatiasFernandez:
## Summary

Feedback from manually exercising the GemStone/Omni Search UI on the branch behind
#423 (issue #378). Grouped by area; each box is an independent item, so this can be
split out into separate issues if any one of them grows. A few are genuine defects
(stale results, highlighting, matcher) — the rest are naming, discoverability and
polish suggestions.

> **Round markers.** Each fixed item is ticked and carries its round's marker, linking to the comment
> that describes the fix. A 🚫 **Not fixing** item marks a decision not to pursue it, with the reasoning
> in the linked comment.
>
> **🔵 Round 1** — [PR #443](https://github.com/GemTalk/Jasper/pull/443), branch `eric/issue428-omni-highlight-and-literal-symbol`. The first batch: stale preview highlight, Literals symbol search, removal of the Quick Pick path, corpus staleness and the design/doc corrections.
>
> **🟣 Round 2** — [PR #446](https://github.com/GemTalk/Jasper/pull/446), branch `eric/issue428-round2`. The honest Methods scan cap with a configurable limit, and the skipped-scope hint. Both manually F5-verified by Eric.
>
> **🟢 Round 3** — [PR #445](https://github.com/GemTalk/Jasper/pull/445), branch `eric/issue428-round3`, commits `cc648f0a` + `fcc5eb90`. Explorer reveal robustness (category guard, dictionary-jump warning), Literals error surfacing, and the method-tree scroll. The method-tree scroll was manually F5-verified by Eric; see the [summary](https://github.com/GemTalk/Jasper/issues/428#issuecomment-5319384327) and the [record note](https://github.com/GemTalk/Jasper/issues/428#issuecomment-5320581299).
>
> **🟠 Round 4** — [PR #444](https://github.com/GemTalk/Jasper/pull/444), branch `eric/issue428-round4`. The reveal race, the panel width, and the settings/test nits, plus two items declined; see the [summary](https://github.com/GemTalk/Jasper/issues/428#issuecomment-5318862899).
>
[...]

PR #443, opened 2026-08-15
Omni Search #428 (Round 1): highlight/matcher/staleness fixes, Quick Pick removal, docs

ericwinger:
**Round 1 of [#428](https://github.com/GemTalk/Jasper/issues/428)** (Omni / GemStone Search follow-ups after #423). Seven items — three genuine defects, one showstopper-adjacent UI decision, and the docs the #423 review flagged. Each item's checkbox in #428 is now checked and links to a comment describing the fix.

### Fixes

| # | Item | Fix |
|---|------|-----|
| 🐛 | **Stale match highlighting** | `rehighlightSourcePreview()` recomputes the preview's blue match marks against the *current* query each run, so highlights track the search box, not the debounced host round-trip. `f08d64c6` |
| 🐛 | **Matcher bug (Literals symbol search)** | `sendersOf:` under-reports on 3.7.5, so `x not` senders leaked into `#not` results. Switched to `substringSearch ∩ referencesToLiteral`; new integration test (red on 3.7.5 pre-fix). `a02bd607` |
| 🐛 | **`prime()` corpus staleness** | Cached corpora refresh instead of going stale: a compiled class folds in live (targeted, Unicode7-safe lookup — no full re-enumerate); dictionaries refresh immediately on create/remove/rename; categories + all corpora rebuild on commit/abort. `b712d927` |
| 🔀 | **Quick Pick vs panel disagree / engine-controller duplication** | Removed the Quick Pick UI (`ui: "quickpick"`) + its controller (the duplicate search path). Panel + Spotter now share the single `omniEngine.ts` — they can't disagree. `3a2e2379` |
| 📄 | **DESIGN.md out of sync** | Rewritten to match the shipped architecture (two webview UIs on the shared engine, complete module map, all 8 settings, trimmed deferred list). `cdf62d35` |
| 📄 | **Provider header docs** | globals header (jumps to the value's class, not home dictionary) + categories header (precise reveal shipped). `4660dedc` / `b712d927` |

### Testing
- New/updated unit tests across the engine, all four provider shapes, the targeted class-lookup query, and the host-notification seams. Full gate green: lint / format / compile + client 5631 / server 322 / mcp 92 (pre-push, live 3.6.2 stone).
- **Manually F5-verified** (the `vscode` glue seams aren't unit-tested): a compiled class appears in search live; a class compiled into a new category appears in the Categories scope; a newly created dictionary appears immediately in the Dictionary scope.

### Notes
- The `quickpick` UI value is removed from `gemstone.omniSearch.ui` (now `panel` | `spotter`).
- Known by-design limits: an *empty* class category isn't an image entity (Explorer overlay only) so it can't appear until a class is filed in; globals created by compiling code (not a command) refresh on the next commit/abort.


client/src/omniSearch/DESIGN.md
```diff
@@ -3,167 +3,182 @@
...
 
-## Deferred / follow-ups (explicitly NOT in this issue)
+## Deferred / follow-ups (tracked in issue #428)
 
-- Phase-2 **webview Spotter** (labeled filter-button row, inspector preview, screenshots' look).
 - **Recents / empty-state** (IntelliJ shows recent files when the field is empty).
-- **Top-hit** promotion; result **highlight ranges** rendering (matcher already returns ranges).
-- Extra providers: **senders/implementors, commands, settings, globals/variables**.
+- **Top-hit** promotion.
+- Corpus refresh: `prime()` loads the class/global/dictionary corpus **once**, so a class created after
```
MatiasFernandez: 🚫 This bullet says the opposite of what the PR ships: the whole `prime()`-staleness fix (`applyChange` / `reprime` + the `notifyClassCompiled` / `notifySessionSynced` hooks) means a class created after the UI opened DOES appear now. Since the doc was rewritten here specifically to match the shipped architecture, can we drop this from Deferred and instead describe the refresh model — who produces a change (class compile, dictionary add/remove/rename, commit/abort), which providers fold vs. re-scan, and the known limits you list in the PR description (empty category, globals created by compiling code)? That's the part a future reader (or an agent) will most need and it isn't documented anywhere else.

client/src/omniSearch/omniSearchViewProvider.ts
```diff
@@ -69,6 +69,23 @@ export class OmniSearchViewProvider implements vscode.WebviewViewProvider {
...
+   *  unless we have an engine built for that same session. */
+  async onClassCompiled(sessionId: number, className: string, dictName?: string): Promise<void> {
+    if (!this.engine || this.builtForSession !== sessionId) return;
+    const view = await this.engine.applyChange({ kind: 'class', className, dictName });
+    if (view) this.postView(view);
+  }
+
+  /** The session synced (commit/abort): rebuild every cached corpus so changes from elsewhere —
+   *  including other sessions — appear, then redraw the current search. */
+  async onSessionSynced(sessionId: number): Promise<void> {
```
MatiasFernandez: 💭 Should this be gated on the view being visible, the way `onConfigChanged` does with `this.view?.visible`?

The engine isn't dropped when the panel is hidden, so once the panel view has been shown a single time in the window (it's the default UI), every commit and every abort — plus every dictionary add/remove/rename via `notifySessionSynced` — runs a full `resync`: re-enumerating all classes, globals and dictionaries. Those provider loads are synchronous GCI executes, so on a big image that's a UI-thread stall on an operation people run constantly, even with the panel collapsed and nothing typed in the box. Could we instead mark the corpora dirty and let the next `ready` / next search rebuild them when the view is actually visible?

client/src/queries/methodSearch.ts
```diff
@@ -161,20 +161,28 @@ ${methodSerialization(environmentId)}`;
...
   environmentId: number = 0,
 ): MethodSearchResult[] {
-  const code = `| symLit lit sends methods stream limit classDict sl |
+  const needle = escapeString(symbolExpr);
+  const code = `| symLit lit candidates methods stream limit classDict sl |
 symLit := ${symbolExpr}.
 lit := (ClassOrganizer new referencesToLiteral: symLit) at: 1.
-sends := (ClassOrganizer new sendersOf: symLit) at: 1.
-methods := lit reject: [:m | sends includes: m].
+candidates := (ClassOrganizer new substringSearch: '${needle}' ignoreCase: false) at: 1.
```
MatiasFernandez: 💭 Agreed the senders-subtraction was unsound, but the substring pre-filter buys precision at the cost of some genuine literal uses that the old query did find — a symbol whose source text isn't `#not`:

- inside a literal array: `#(not size)` / `#(at:put: foo)` — the elements are real symbols in the literal frame, but the source has no `#`
- a different-but-equivalent spelling: someone searching `#not` won't find a method that wrote `#'not'` (same interned symbol, different source text)

Are those acceptable? Either way, could the comment say so? Right now it explains only what we gain, so the next reader debugging a "missing" literal hit has no hint that the substring gate is the reason.

client/src/omniSearch/__tests__/corpusRefresh.test.ts
```diff
@@ -0,0 +1,246 @@
...
+      { kind: 'class', className: 'Foo' },
+      {
+        isCancelled: false,
+      },
+    );
+
+    expect(changed).toBe(false);
+  });
+
+  it('drops a removed class on the next fold', () => {
```
MatiasFernandez: 💭 This proves the provider handles a removal, but is anything in production producing that change? `ExplorerController.removeClass` doesn't call the new hooks, and the only `notifyClassCompiled` caller is the class-compile document event — so a class deleted from the Explorer keeps showing up (and stays clickable) in an open search until the next commit/abort resync. Since the dictionary add/remove/rename path did get wired here, should `removeClass` fire `onSymbolListChanged` too — or is the delete-then-stale window a deliberate follow-up worth noting in the by-design limits?

client/src/extension.ts
```diff
@@ -745,17 +748,24 @@ export function activate(context: vscode.ExtensionContext) {
...
+      // Distinguish "no selector at this position" (LSP answers null → rename the
+      // edited method) from "lookup unavailable" (throw → the command aborts rather
+      // than guess): a not-yet-started client or a request failure THROWS; only a
+      // live "no selector here" answers null.
+      if (!client) throw new Error('language server not started');
+      return client.sendRequest<string | null>('gemstone/selectorAtPosition', {
+        textDocument: { uri: document.uri.toString() },
+        position,
+      });
+      // A dictionary add/remove/rename changes the symbol list — refresh an open Omni Search so the
```
MatiasFernandez: 💭 This comment looks like it landed in the wrong place — it sits inside the `selectorAtPosition` callback, after its `return`, so it reads as dead text describing a lookup that has nothing to do with dictionaries. Should it move down next to the `(sid) => omniSearch?.notifySessionSynced(sid)` argument it actually explains?

client/src/extension.ts
```diff
@@ -745,17 +748,24 @@ export function activate(context: vscode.ExtensionContext) {
...
+      // new/renamed dictionary shows up without waiting for a commit.
+    },
+    (sid) => omniSearch?.notifySessionSynced(sid),
```
MatiasFernandez: 💭 Naming: a dictionary add/remove/rename isn't really a "session synced" event (the explorer side calls it `onSymbolListChanged`, which reads right), and routing it here means a rename triggers a full corpus rebuild of classes + globals + dictionaries. Would a narrower hook — say `notifySymbolListChanged`, refreshing just the dictionaries corpus — describe the trigger better and cost less? See also my note on `onSessionSynced`.

client/src/extension.ts
```diff
@@ -2512,8 +2530,7 @@ export function activate(context: vscode.ExtensionContext) {
       }
     }),
 
-    registerOmniSearch(sessionManager, context),
-
+    (omniSearch = registerOmniSearch(sessionManager, context)),
```
MatiasFernandez: nit: assigning inside the `subscriptions.push(...)` argument list is easy to miss when scanning for where `omniSearch` gets set (and it's the one entry in this long list that does two things). Would hoisting it read better?

```ts
omniSearch = registerOmniSearch(sessionManager, context);
// ...
context.subscriptions.push(..., omniSearch, ...)
```

client/src/omniSearch/omniSearchCommand.ts
```diff
@@ -38,28 +39,11 @@ import { createGlobalsProvider } from './providers/globalsProvider';
...
-
-/** Context key that's true only while the reference view is showing, so the Left-arrow "back"
- *  keybinding fires only in the pivot (elsewhere Left is normal cursor movement). */
-const OMNI_IN_PIVOT_CONTEXT = 'gemstone.omniSearchInPivot';
-
-/** The controller for the currently-open picker, so the (global) references command can act on its
- *  highlighted row. Cleared when the picker hides. */
-let activeController: OmniController | undefined;
-
 /** Where a result should open. The QuickPick passes nothing (open in the active group, the prior
```
MatiasFernandez: nit: this doc is stale now that the QuickPick is gone — it's the panel view that passes nothing. Maybe "the docked panel view passes nothing (open in the active group); the Spotter passes `Beside` + `preserveFocus`…"?

MatiasFernandez (approved):
Approved with comments. I used the following "emoji" styling so you can easily identify the intention of the comment, and you can decide what to do:
- 📝: Outdated doc; doc improvement
- 🧪: A test that can be improved
- 💭: question/comment regarding the code that may indicate changes worth analyzing

client/src/omniSearch/omniSearchViewProvider.ts
ericwinger: I was able to manually verify using F5 and setting appropriate breakpoints. commit, abort, and several dictionary actions do the full resync.

client/src/omniSearch/DESIGN.md
ericwinger: Manually verified with BP's

client/src/omniSearch/omniSearchViewProvider.ts
ericwinger: Good catch — fixed in 9e1b3727, and I reproduced it under the debugger first to be sure of the cost.

With the panel hidden and the box empty, a no-op commit really did reach `resync`, and `resync` maps over *all* `providers` (not `providersInScope`), so it fired 7 times: `classesProvider.reprime` → `getAllClassNames`, `dictionariesProvider.prime` → `getDictionaryNames`, `globalsProvider.prime` → `getAllGlobalNames`, plus `categoriesProvider.reprime` dropping its lazy cache and three awaited no-ops. Three image-wide synchronous executes per commit/abort, with nothing on screen.

Took your suggestion exactly: `onSessionSynced` now sets a `syncPending` flag and returns when `!this.view?.visible`, mirroring `onConfigChanged`. The rebuild is paid in `flushPendingSync()`, called on reveal (`onShown`), on `ready`, and before every webview-driven engine op — that last one matters so a search never reads a stale corpus. Many hidden syncs collapse into one rebuild, and the flag is cleared wherever the engine is dropped or rebuilt since a fresh engine primes anyway. A visible panel still refreshes live.

5 new tests in `omniSearchViewProvider.test.ts` cover: hidden sync doesn't touch the corpora, reveal pays it, next search pays it, N syncs → 1 rebuild, and a reveal with nothing pending stays quiet. Manually F5-verified by Eric.

I left `OmniSearchPanel.onSessionSynced` (the Spotter) ungated for now — same shape, but the window is much narrower and the Spotter's future is under discussion.

client/src/omniSearch/__tests__/corpusRefresh.test.ts
ericwinger: You're right, and it was worth chasing — nothing in production produced that change. Confirmed under the debugger: deleting a class hit `deleteClass` and ran to the end of `removeClass` without any notify, and the class stayed in an open search, still clickable. The dictionary hook fired normally for a dictionary add in the same session, which made the gap obvious.

Fixed in 9e1b3727 with the per-class fold rather than a session-wide resync, so a delete doesn't cost an image-wide re-enumeration:

- new `onClassRemoved?(sessionId, className)` on `ExplorerController` (also threaded through `registerGemStoneExplorer`), fired **once per class** — as you noted, the delete takes the whole subtree, so it loops the resolved `targets`
- only classes that **actually** deleted are reported: `removeClass` now tracks a `removed[]` beside `failures[]`, so a partial failure can't drop a class that is still in the image
- wired to a new `notifyClassRemoved` on the Omni API, which shares a `foldClassChange` helper with `notifyClassCompiled`

That routes into the same `applyChange` path `corpusRefresh.test.ts` was already testing: `getClassNameEntriesFor` re-looks-up the name, comes back empty (it uses `at: target otherwise: nil`, so a missing key is not an error), and the entry drops out.

I deliberately did **not** rename `notifyClassCompiled` → `notifyClassChanged` even though a removal now flows through it — `ExplorerHandle` has its own unrelated `onClassCompiled`, so the rename would collide across two hooks and churn 14 sites. The delegating `notifyClassRemoved` keeps the call sites honest instead.

4 new tests in `explorerRemoveClass.test.ts`: leaf reports once, a subtree reports every member in order, a failed delete reports nothing, a dismissed confirmation reports nothing. Manually F5-verified by Eric.

client/src/omniSearch/DESIGN.md
ericwinger: Agreed — that bullet described the exact bug this PR fixes. Fixed in 9e1b3727, split the way you suggested.

The positive model now lives in §3 (load-once providers), as a who-announces-what table, since that is where the corpora are described:

| Change in the image | Hook | Cost |
| --- | --- | --- |
| A class compiled locally | `notifyClassCompiled` → `applyChange` | re-fetch that ONE class name |
| A class removed (Explorer → Remove Class) | `notifyClassRemoved` → `applyChange`, fired once per class | re-fetch per name; the lookup comes back empty and the entry drops |
| Dictionary add / remove / rename | `onSymbolListChanged` → `notifySessionSynced` | full `resync` |
| Commit / abort / file-in | `notifySessionSynced` | full `resync`, deferred while hidden |

Same section also documents the hidden-panel deferral from your other comment, and why those three loads are worth deferring at all.

The Deferred entry keeps only what is still true: a global created by evaluating code (nothing announces it), a brand-new class category once the Categories scope is loaded, and anything changed by another session — all picked up by the next commit/abort resync.

client/src/queries/methodSearch.ts
ericwinger: Yes, both are acceptable — and you're right that the comment only sold the upside. Query unchanged; comment extended in 9e1b3727:

```
// What that costs — the gate is textual, so a genuine literal use whose SOURCE doesn't spell `#not`
// is missed even though the symbol really is in the literal frame:
//   - inside a literal array — `#(not size)`, `#(at:put: foo)`: real symbols, but no `#` per element
//   - an equivalent different spelling — a search for `#not` won't find a method that wrote `#'not'`
// Accepted deliberately (reviewed on PR #443): both are rare next to the bogus every-sender flood the
// old query produced. If you are chasing a "missing" literal hit, this filter is the reason.
```

That last line is aimed squarely at the reader you described — someone debugging a hit they expected to see.

client/src/extension.ts
ericwinger: Yes — it was stranded after the `return`, describing something the `selectorAtPosition` callback has nothing to do with. Moved in 9e1b3727 onto the argument it explains.

While there, the new `notifyClassRemoved` argument from your Remove Class comment landed next to it, so both hooks are now labelled at the call site:

```ts
    // A dictionary add/remove/rename changes the symbol list — refresh an open Omni Search so the
    // new/renamed dictionary shows up without waiting for a commit.
    (sid) => omniSearch?.notifySessionSynced(sid),
    // A removed class has to leave Omni Search's cached class corpus the same way, but one class at a
    // time — Remove Class takes the whole subtree with it.
    (sid, className) => omniSearch?.notifyClassRemoved(sid, className),
```

client/src/omniSearch/omniSearchCommand.ts
ericwinger: Stale indeed — took your wording, with one addition explaining *why* the panel passes nothing. Fixed in 9e1b3727:

```ts
/** Where a result should open. The docked panel view passes nothing (open in the active group — the
 *  editor area sits above the panel, so there is nothing to open beside); the Spotter passes `Beside`
 *  + a `preserveFocus` flag so a result opens beside the panel — optionally without stealing focus
 *  from the search field. */
```

PR #444, opened 2026-08-17
Omni Search #428 (Round 4): post-login reveal race, settings description, chrome width

ericwinger:
Fourth batch of follow-ups from issue #428 (the manual-testing + PR #423 review backlog for
GemStone Search). Marked **🟠 Round 4** on the issue, so it stays distinguishable from 🔵 Round 1
(PR #443, open) and 🟣 Round 2.

## What's in here

Every row links to the issue comment describing that fix, and each issue checkbox links back here.

| Fixed | Commit | Files |
|---|---|---|
| **Post-login panel reveal raced a fixed 400 ms wait** — [details](https://github.com/GemTalk/Jasper/issues/428#issuecomment-5318862899) | `1e1c6107` | `omniSearchCommand.ts`, `omniSearchViewProvider.ts` |
| **Settings description named a non-existent "open editors" scope** — [details](https://github.com/GemTalk/Jasper/issues/428#issuecomment-5318862899) | `967e5115` | `package.json` |
| **Matcher-test comment + range literal tidy-up** — [details](https://github.com/GemTalk/Jasper/issues/428#issuecomment-5318862899) | `470c12e4` | `__tests__/omniMatch.test.ts` |
| **Webview capped at a fixed 960px, wasting wide monitors** — [details](https://github.com/GemTalk/Jasper/issues/428#issuecomment-5319135727) | `c7a1ce93` | `omniSearchShared.ts` |

Two items were **declined** rather than fixed, both struck through on the issue with the reasoning:

| Declined | Why |
|---|---|
| **Spotter tab's odd preview semantics** — [details](https://github.com/GemTalk/Jasper/issues/428#issuecomment-5318961625) | Spotter-only; the default panel has no dispose path. VS Code also offers no API to make a webview panel a preview tab, so half the complaint was never fixable. |
| **One command per UI mode** — [details](https://github.com/GemTalk/Jasper/issues/428#issuecomment-5319167617) | With the Quick Pick gone there are only two modes, and the generic command already opens the default one — the remaining value is Spotter reachability. |

## The one behavioral change worth reviewing closely

`SessionManager.selectSession` fires `onDidChangeSelection` **before** it sets the
`gemstone.hasActiveSession` context key, and that key is the search view's `when` clause. The
post-login auto-reveal therefore ran while the view was not yet allowed to exist, and covered for it
by sleeping a flat 400 ms. On a slow or busy window that lost the race silently — no retry, no
message; on a fast one the panel arrived a visible beat late.

`revealPanelAfterLogin()` now asserts the context key and **awaits** it (idempotent — the session
manager sets the same value a tick later, but awaiting means it has landed), then reveals once and
waits for the view to announce itself. `resolveWebviewView` fires an event and
`OmniSearchViewProvider.focus()` waits on it with a give-up bound, returning `Promise<boolean>` that
means *this* reveal landed; a reveal that never lands is logged rather than swallowed.

Review changed the second half of this. The first version polled `focus()` up to 5 × 60 ms, and
`focus()` answered from a field set once in `resolveWebviewView` and never cleared — so past the first
login of a window it reported success unconditionally and the retry loop was dead code. Waiting on the
resolve event removes the loop, the attempt budget and the inter-attempt delay, and lets the remaining
timeout be a generous deadline for *reporting* failure rather than a guess at how long a reveal takes.

Still upstream, deliberately: `selectSession` fires its event before it writes the context key, which
is why the reveal has to assert that key itself. Fixing the ordering in `SessionManager` would help
every consumer of the key, not just this caller — left as a follow-up since it changes ordering for
code well outside the search feature.

Reviewer note: the reveal only fires on a 0 → active session transition, so exercising it by hand
means logging out of every session first. It is covered by unit tests rather than manual verification.

## Tests

- `revealPanelAfterLogin` in `omniSearchCommand.test.ts` (3) — the context key is set before the
  reveal, a reveal that lands says nothing, a reveal that never lands logs.
- `registerOmniSearch` in the same file (3) — the wiring around the helper: the panel is revealed on a
  login, not on a switch between existing sessions, and not at all when the Spotter is the chosen UI.
  Verified red by dropping each guard in turn.
- Reveal reporting in `omniSearchViewProvider.test.ts` (3) — reports failure when the view never
  resolves without hanging, returns as soon as it does resolve, answers at once for a view already
  built. Verified red by restoring the old `return this.view !== undefined`.
- **Dropped in review:** the chrome-width test. It asserted a stylesheet rule rather than a rendered
  width, which cannot catch the regression it named. The width fix itself is unchanged.
- **Gate at this tip (3.6.2 stone):** `npm run lint` / `format:check` / `compile` clean; full `npm test`
  green — **client 5876 passed / 12 skipped (391 files), server 322, mcp-server 92**. The pre-push hook
  re-ran the same suite in random order before pushing.
- **Second boundary (3.7.5 stone), at the pre-review tip:** full `npm test` run against a live 3.7.5 stone — server 322 and
  mcp-server 92 green, client **5736 passed / 8 skipped with 11 failures, all pre-existing and unrelated**:
  they are the in-stone RB engine SUnit suites (`src/refactoring/__tests__/*.integration.test.ts`), which
  that stone's loaded engine predates. Verified by running the same five files from a tree **without**
  these commits against the same stone: identical 11 failures. Round 4 changes no refactoring file — the
  diff is under `client/src/omniSearch/` plus one `package.json` string and two additions to the shared
  vscode test mock. The review changes since are client-only and touch no refactoring file either.
- **Not run:** the local CI-mirror matrix. It reprovisions fresh stones per version, which would have
  disrupted two sibling branches mid-work, and this diff has no server, RB-engine, GCI or Smalltalk
  surface for it to exercise. CI on this PR covers the same ground for a client-only change.

## Merge note

`main` has been merged in (`69e02123`). The earlier rounds that rewrote `omniSearchCommand.ts` have
landed, and both conflicts were code arriving beside this branch's edits rather than on top of them —
main's registration interface next to the reveal helper, and main's corpus-refresh methods next to
`focus()`. Both sides kept; full suite green afterwards.


client/src/omniSearch/omniSearchCommand.ts
```diff
@@ -303,6 +304,43 @@ export async function runOmniSearch(
...
+ * the same value a tick later, but awaiting means it has actually landed — then reveal and use
+ * `focus()`'s report of whether the view resolved, retrying briefly if the workbench hasn't caught up.
+ * A reveal that never lands is logged instead of swallowed; the shortcut and status-bar button still
+ * work, so this is a warning, not an error.
+ */
+export async function revealPanelAfterLogin(
+  viewProvider: Pick<OmniSearchViewProvider, 'focus'>,
+  sleep: (ms: number) => Promise<void> = (ms) => new Promise((resolve) => setTimeout(resolve, ms)),
+): Promise<boolean> {
+  await vscode.commands.executeCommand('setContext', 'gemstone.hasActiveSession', true);
```
MatiasFernandez: 💭 This works, but it leaves the actual root cause in place and moves ownership of a global context key into the search feature. `selectSession` (`client/src/sessionManager.ts:81-87`) fires `onDidChangeSelection` and *then* calls `setContext` (without awaiting it) — every other consumer of `gemstone.hasActiveSession` that reacts to that event has the same race, it just doesn't happen to reveal a view. Would flipping those two lines in `selectSession` (set the context first, then fire) be the cheaper fix? Then this function keeps only the retry + reporting part, and `omniSearch` doesn't need to know about a key it doesn't own.

If the ordering can't be changed for a reason I'm missing, worth a note here saying so — otherwise a future reader will fix `selectSession` and never notice this duplicate write.

client/src/omniSearch/omniSearchViewProvider.ts
```diff
@@ -69,13 +69,18 @@ export class OmniSearchViewProvider implements vscode.WebviewViewProvider {
...
+   *  Reports whether the view actually resolved. A reveal attempted while the view's `when` clause is
+   *  still false — i.e. before `gemstone.hasActiveSession` has propagated — silently shows nothing, so
+   *  a caller revealing on a login needs to know rather than assume (see `revealPanelAfterLogin`). */
+  async focus(): Promise<boolean> {
     this.focusPending = true;
     // Reveal (and, if needed, create) the view; then ask the field to take the cursor. If the webview
     // is still loading, the message is lost — the `ready` handler replays it (see onMessage).
     await vscode.commands.executeCommand(`${OMNI_VIEW_ID}.focus`);
     this.deliverFocus();
+    return this.view !== undefined;
```
MatiasFernandez: 💭 Is `this.view !== undefined` really "this reveal resolved the view", or "a view has resolved at some point"? `this.view` is set in `resolveWebviewView` and never cleared — there's no `view.onDidDispose` handler — so once the panel has been shown in this window it stays set.

The scenario I'm worried about is the second login: user logs in (view resolves), logs out of every session (the `when` clause goes false and the workbench disposes the view, but `this.view` still points at it), logs back in. `revealPanelAfterLogin` calls `focus()`, gets `true` on attempt 1, and never retries — exactly the case the retry exists for. Should `resolveWebviewView` register `view.onDidDispose(() => (this.view = undefined))` so the signal means what the TSdoc says?

client/src/omniSearch/omniSearchCommand.ts
```diff
@@ -303,6 +304,43 @@ export async function runOmniSearch(
 /** globalState flag: the one-time "Ctrl/Cmd+Shift+A opens GemStone Search" tip has been shown. */
 const SEARCH_TIP_SHOWN_KEY = 'gemstone.gemstoneSearchTipShown';
 
+/** Bounded retry budget for the post-login reveal (see `revealPanelAfterLogin`): the first attempt is
+ *  immediate, so this only costs anything on a window slow enough to actually lose the race. */
+const REVEAL_ATTEMPTS = 5;
+const REVEAL_RETRY_MS = 60;
```
MatiasFernandez: 💭 The budget here is 4 × 60 ms ≈ 240 ms of patience, which is *less* than the flat 400 ms it replaces. The failure this PR is fixing is "a slow or busy window (remote, WSL, a crowded activation) lost the race" — on that window, are 240 ms enough, or do we now give up sooner than the code we're replacing and just log about it? Since the first attempt is immediate the fast path costs nothing either way, so it seems cheap to be more generous: more attempts, or a small backoff (60/120/240…) so a genuinely slow window still lands instead of hitting the warning.

client/src/omniSearch/__tests__/omniSearchCommand.test.ts
```diff
@@ -70,3 +75,70 @@ describe('buildOmniHandlers', () => {
     );
   });
 });
+
+describe('revealPanelAfterLogin', () => {
```
MatiasFernandez: 💭 These four cover `revealPanelAfterLogin` itself nicely, but nothing covers the wiring — that `onSelection` calls it only on a 0 → active transition and only when `ui === 'panel'`. That call site is where the original bug lived, and it's also the part PR #443 rewrites in this same file, so it's the bit most likely to get lost in the reconciliation. Is it worth one test that drives `sessionManager.onDidChangeSelection` and asserts the reveal fires once on login and not on a session switch (and not at all for `ui: "spotter"`)?

client/src/omniSearch/__tests__/omniChromeWidth.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    ['docked panel', false],
+    ['Spotter tab', true],
+  ] as const) {
+    it(`does not cap the container at a fixed pixel width — ${host}`, () => {
+      const rule = omniRule(renderOmniHtml({ showPin }));
+
+      // Sanity: we matched the real container rule, not some other `#omni`-ish text.
+      expect(rule).toMatch(/height:\s*100vh/);
+
+      expect(rule).not.toMatch(/max-width:\s*\d+(\.\d+)?px/);
```
MatiasFernandez: 💭 The regex only catches a cap expressed in `px`. A reintroduced `max-width: 60rem` / `72ch` / `40em` — or the same cap moved onto `#results` — passes this guard while producing the exact bug from #428. Should it just assert there's no `max-width` at all in the `#omni` rule (`expect(rule).not.toMatch(/max-width:/)`)? That's stricter, still honest about what jsdom can check, and reads as "this container is not capped" rather than "not capped in pixels".

package.json
```diff
@@ -155,7 +155,7 @@
               "The result starts with the typed text."
             ],
             "default": "fuzzy",
-            "markdownDescription": "How Omni Search matches your query against classes, methods, dictionaries and open editors."
+            "markdownDescription": "How your typed query is matched against the names in each scope: classes, method selectors, dictionaries, globals and class categories. **Source** and **Literals** match on the stone instead (whole method source / literal frames), so this setting does not affect them."
```
MatiasFernandez: nit: this calls them "scopes" while `gemstone.omniSearch.categories` right below calls the same seven things "categories" (and its own description uses **Source** / **Literals** the same way). Since both descriptions sit in the same settings page, should this one say "category" too so a user reading top-to-bottom doesn't wonder whether scopes and categories are two different things?

client/src/omniSearch/__tests__/omniChromeWidth.test.ts
```diff
@@ -0,0 +1,46 @@
...
+import { renderOmniHtml } from '../omniSearchShared';
+
+/** The declarations inside the `#omni { … }` rule of the rendered chrome. */
+function omniRule(html: string): string {
+  const match = /#omni\s*\{([^}]*)\}/.exec(html);
+  if (!match) throw new Error('#omni rule not found in the rendered chrome');
+  return match[1];
+}
+
+describe('search chrome width (#428)', () => {
```
MatiasFernandez: ❗ This is yet another example of Claude working hard to invent a test for a scenario that is hard to test or not testable at all. I prefer missing a test for a CSS change than having a fragile or weird test attempting to test something very implementation-specific, which also provides a false sense of coverage (you can't ensure there's a CSS regression with this test). I would just remove this file

client/src/omniSearch/omniSearchCommand.ts
```diff
@@ -303,6 +304,43 @@ export async function runOmniSearch(
...
+  for (let attempt = 1; attempt <= REVEAL_ATTEMPTS; attempt++) {
+    if (await viewProvider.focus()) return true;
+    if (attempt < REVEAL_ATTEMPTS) await sleep(REVEAL_RETRY_MS);
+  }
+  logWarning(
+    'GemStone Search: the search panel did not appear after login — the view never resolved. ' +
+      `Open it with ${OMNI_OPEN_KEY_HINT} or the GemStone Search button in the status bar.`,
+  );
+  return false;
+}
```
MatiasFernandez: ❗ I'd like this solved a different way. What's here works, but it treats the symptom in three places at once — it re-asserts a global context key the search feature doesn't own, it polls for an outcome it can't actually observe, and it leaves the ordering bug that caused all of it untouched. There's an event to wait on for each half of the race, so I think the timers can go entirely.

**The root cause is upstream.** `selectSession` (`client/src/sessionManager.ts:81-87`) fires `onDidChangeSelection` and *then* calls `setContext` without awaiting it. Every consumer of `gemstone.hasActiveSession` that reacts to that event has this race; the search panel is just the one that happens to reveal a view and so notices. Fixing it there means starting the write before the fire and handing listeners something to await:

```ts
/** Resolves once the last `gemstone.hasActiveSession` write has landed in the workbench.
 *  Anything revealing UI gated on that key must await this first. */
private _contextSettled: Promise<void> = Promise.resolve();
get contextSettled(): Promise<void> {
  return this._contextSettled;
}

private setActiveSessionContext(value: boolean): void {
  this._contextSettled = Promise.resolve(
    vscode.commands.executeCommand('setContext', 'gemstone.hasActiveSession', value),
  ).then(() => undefined);
}

selectSession(id: number): void {
  // …
  this.setActiveSessionContext(true);   // start the write BEFORE listeners run
  this._onDidChangeSelection.fire(id);
}
```

The same helper replaces the duplicated write in `logout()`, and `omniSearch` stops needing to know about a key it doesn't own.

**The other half of the race isn't the context key at all**, which I think is why the retry loop feels necessary: `<viewId>.focus` can return before the workbench has instantiated the provider even when the `when` clause was true the whole time. That's provider-resolution latency, not context propagation — and `resolveWebviewView` is the definitive "the view exists now" signal for it, so it can be an event rather than something we poll:

```ts
private readonly _onDidResolve = new vscode.EventEmitter<void>();

resolveWebviewView(view: vscode.WebviewView): void {
  this.view = view;
  // …existing setup…
  this._onDidResolve.fire();
}

/** True when the workbench has actually instantiated the view. The deadline is a give-up bound,
 *  not a guess at how long the reveal takes — the happy path returns on the event. */
private whenResolved(deadlineMs: number): Promise<boolean> {
  if (this.view) return Promise.resolve(true);
  return new Promise((resolve) => {
    const sub = this._onDidResolve.event(() => {
      clearTimeout(timer);
      sub.dispose();
      resolve(true);
    });
    const timer = setTimeout(() => {
      sub.dispose();
      resolve(false);
    }, deadlineMs);
  });
}

async focus(): Promise<boolean> {
  this.focusPending = true;
  await vscode.commands.executeCommand(`${OMNI_VIEW_ID}.focus`);
  const resolved = await this.whenResolved(REVEAL_DEADLINE_MS);
  this.deliverFocus();
  return resolved;
}
```

With both, this function is two awaits — no `sleep` parameter, no attempt budget, no foreign `setContext`:

```ts
await sessionManager.contextSettled;   // the view's `when` clause is now true
if (await viewProvider.focus()) return true;
logWarning('GemStone Search: the search panel did not appear after login — …');
return false;
```

**Two things that motivated this rather than just tuning the numbers.**

The budget as written is 4 × 60 ms ≈ 240 ms of patience, *less* than the flat 400 ms it replaces. The failure being fixed is "a slow or busy window lost the race" — on that window we now give up sooner than the code we're replacing and log about it. Making the wait event-driven sidesteps the question: the one surviving `setTimeout` becomes a deadline for *reporting* failure rather than an estimate of how long the workbench needs, so it can be seconds without the fast path paying anything.

And `focus()`'s new return value doesn't mean what the loop reads it as. `this.view` is sticky — once `resolveWebviewView` has run it stays set for the life of the window, so `return this.view !== undefined` is unconditionally `true` from then on. The retry loop can therefore only ever detect a failed reveal on the *very first* login of a window; every login after that reports success whether or not the panel came forward. The new tests don't catch it because the stub provider returns the flag directly instead of modelling `this.view`. (Waiting on the resolve event short-circuits the same way for an already-resolved view — but there it's correct, because the view genuinely does exist.)

Fair warning that the first part touches `sessionManager.ts`, outside this PR's diff, and that PR #443 is already rewriting this file — so if you'd rather land Round 4 as-is and do this as a follow-up, that's reasonable. In that case the sticky-`this.view` behaviour is the piece I'd still want addressed here, at minimum as a comment saying the return value means "has ever resolved", not "this reveal landed".

MatiasFernandez (approved):

client/src/omniSearch/__tests__/omniSearchCommand.test.ts
ericwinger: Worth it — done in `a99e36cb`, as three tests that drive the real `registerOmniSearch` through a session-selection event rather than testing the helper again:

- reveals the panel when a session appears where there were none (a login);
- does not reveal when switching between sessions that already existed;
- does not reveal at all when the Spotter is the chosen UI.

They watch for `executeCommand('gemstoneOmniSearchView.focus')`, so they pin the call site itself, not a stub of it. I proved they bite: dropping the `!hadSession` guard fails the switch test, and dropping the `ui === 'panel'` guard fails the Spotter test — one each, no overlap.

Two notes. Driving the real registration needed two additions to the shared vscode test mock (`window.registerWebviewViewProvider` and a `Disposable.from` static); nothing else in the suite depended on their absence. And your reconciliation worry was well founded — `main` had moved under this branch, so it is now merged in (`69e02123`) and `ui` is `spotter | panel` only, which is why the third test names the Spotter rather than a Quick Pick.

package.json
ericwinger: Agreed — "category" it is, in `a99e36cb`:

> How your typed query is matched against the names in each **category**: classes, method selectors, dictionaries, globals and class categories. **Source** and **Literals** match on the stone instead (whole method source / literal frames), so this setting does not affect them.

That matches the setting right below it, and the setting key is `gemstone.omniSearch.categories`, so "category" is the word with something behind it.

The split is wider than this one string, and I deliberately left the rest alone: the `ui` setting's own text mentions "scope tabs" and "title-bar scope buttons", and the panel gets a `Scopes` button in PR #448, which is still in review. Renaming across an open branch seemed like the wrong way to settle vocabulary, so I have noted it for the naming cleanup in #447 instead.

client/src/omniSearch/__tests__/omniChromeWidth.test.ts
ericwinger: You are right, and the file is gone in `a99e36cb`. The width fix itself is untouched.

The tell was in the file's own header comment: it admitted jsdom never computes a box and then asserted the stylesheet text anyway. That is a test of how the rule happens to be written, not of anything a user would see — it would pass just as happily if the panel rendered one column wide, and it would fail on a harmless refactor of the stylesheet. Cheerfully deleted.

I also took the entry out of the PR description's Tests section so it does not advertise a file that no longer exists.

client/src/omniSearch/omniSearchCommand.ts
ericwinger: Taken, in `a99e36cb` — the view-provider half here, the session-manager half as a follow-up, which is the split you offered.

**Your sticky-`this.view` finding is exactly right, and it was worse than a mis-read return value: it made the retry loop dead code.** The field is set once in `resolveWebviewView` and never cleared, so `focus()` could only ever answer "no" before the view had been built at all — i.e. on the first login of a window. Every login after that answered "yes" regardless, so the loop never retried and the warning could never fire. Reproducing it needs no GemStone and no window: stub the reveal command to do nothing, call `focus()` (correctly `false`), call `resolveWebviewView` once, then call `focus()` again — still nothing revealed, now `true`, forever.

The fix is your event, not a tuned loop. `resolveWebviewView` fires `onDidResolveView`, and `focus()` waits on it:

```ts
private whenResolved(deadlineMs: number): Promise<boolean> {
  if (this.view) return Promise.resolve(true);
  return new Promise((resolve) => {
    const sub = this.onDidResolveView.event(() => {
      clearTimeout(timer);
      sub.dispose();
      resolve(true);
    });
    const timer = setTimeout(() => {
      sub.dispose();
      resolve(false);
    }, deadlineMs);
  });
}
```

So `revealPanelAfterLogin` is now what you sketched, minus the part that is still upstream:

```ts
await vscode.commands.executeCommand('setContext', 'gemstone.hasActiveSession', true);
if (await viewProvider.focus()) return true;
logWarning('GemStone Search: the search panel did not appear after login — …');
return false;
```

The retry loop, `REVEAL_ATTEMPTS`, `REVEAL_RETRY_MS` and the injected `sleep` are all gone. Your point about the budget landed too: the remaining `setTimeout` is a give-up bound for *reporting*, so it is 5 s rather than 240 ms, and the fast path never pays for it because the happy path returns on the event. The short-circuit for an already-resolved view is deliberate, for the reason you gave — the view genuinely exists, so the reveal had something to show.

Pinned by three tests on the provider: reports failure when the view never resolves (without hanging), returns as soon as it does resolve rather than sitting out the deadline, and answers at once for a view already built. The middle one is the regression guard — I restored the old `return this.view !== undefined` and watched only that one go red.

**Left for the follow-up:** `selectSession` still fires before it writes the context key, so the reveal asserts `gemstone.hasActiveSession` itself. That is a stopgap and is now labelled as one in the comment, with the reasoning you gave: the ordering belongs to the session manager, and starting the write before firing while handing listeners a `contextSettled` to await would fix every consumer of the key rather than this one caller. Not in this PR because it changes ordering for code well outside the search feature, and `main` had already moved under this branch (now merged in at `69e02123`).

ericwinger: The follow-up now has a number: #457 — the ordering fix in `SessionManager`, with your `contextSettled` sketch as the proposed direction and the provider-resolution half written up as already handled here.

PR #445, opened 2026-08-17
Omni Search #428 (Round 3): Explorer reveal robustness, Literals error surfacing, method-tree scroll

ericwinger:
## Round 3 of #428 follow-ups

Four independent, self-contained fixes from the #428 backlog. **Part of #428 — does not close it** (other rounds cover the rest).

Each item below links to its checkbox in #428; the matching #428 items link back here (bidirectional). This branch touches only `gemstoneExplorer.ts` and `literalsProvider.ts` (+ their tests and the one-line Literals log wiring in `omniSearchCommand.ts`), so it does not overlap the other in-flight #428 rounds.

### Items fixed

| # | Item (in #428) | What changed | Where |
|---|----------------|--------------|-------|
| **15** | Guard the category check before `selectClassCategory` ([review thread](https://github.com/GemTalk/Jasper/pull/423#discussion_r3776904477)) | `revealCategoryByPath` now checks the category exists in the loaded forest **before** `selectClassCategory` mutates the panes, so a missing category leaves the panes untouched (and warns) instead of scrolling to a node that isn't there | `gemstoneExplorer.ts` |
| **16** | `logWarning` on a dictionary-jump reveal failure ([review thread](https://github.com/GemTalk/Jasper/pull/423#discussion_r3776904483)) | `revealDictionaryByName` logs a failed `TreeView.reveal` to the GCI channel instead of swallowing it with a bare `catch`, mirroring the category-reveal path two methods down | `gemstoneExplorer.ts` |
| **18** | Literals `catch` swallows all errors ([review thread](https://github.com/GemTalk/Jasper/pull/423#discussion_r3776904511)) | The Literals provider distinguishes "still typing" (silent `[]`) from a real runner failure (GCI drop / aborted transaction). The latter is surfaced through an injected `onError` sink, wired to the GCI log in `buildProviders`, so it no longer masquerades as "no results". Still returns `[]` either way — throwing would abort sibling providers in the All scope | `literalsProvider.ts`, `omniSearchCommand.ts` (1-line wiring) |
| **43** | Selecting a method in the results doesn't scroll the method tree | Only `reveal({ focus: true })` scrolls in this VS Code build (`focus:false` selects but never scrolls). Editor-driven navigation now reveals with `focus:true` and hands focus straight back to the editor, so the method row scrolls into view without the tree keeping focus. Passive background resync stays a non-scrolling select. **F5-verified** (scrolls; cursor stays in the editor) | `gemstoneExplorer.ts` |

### Testing

New tests (all in new files, no shared test files touched):

- `explorerOmniReveal.test.ts` (6) — "# 15" reorder (fails on the old ordering), "# 16" log-on-failure / no-log-on-success.
- `literalsProviderErrors.test.ts` (4) — "# 18" real throw → `onError`; "still typing" stays silent; optional param safe.
- `buildProvidersLiteralsWiring.test.ts` (1) — "# 18" production wiring; fails if the `onError` arg is dropped.
- `explorerMethodReveal.test.ts` (2) — "# 43" navigation reveals with `focus:true` + returns editor focus; passive resync does not grab focus.

Gate: `compile` / `eslint` / `prettier` clean; explorer + omniSearch suites green; pre-push suite green (client + server 322 + mcp 92).

/cc @MatiasFernandez


client/src/gemstoneExplorer.ts
```diff
@@ -3270,14 +3270,25 @@ export class ExplorerController {
...
-        new MethodItem(isMeta, info, displayCategory, this.methodSourceUri(isMeta, info)),
-        { select: true, focus: false, expand: true },
-      );
+      // In this VS Code build focus:false selects the row but never scrolls it into view (#43); only
+      // focus:true scrolls. For editor-driven navigation we force the scroll with focus:true and hand
+      // focus straight back to the editor so the tree doesn't keep it. A passive background resync
+      // stays a plain (non-scrolling) select so it can't yank focus off whatever the user is doing.
+      if (opts.focusEditorAfter) {
+        await this.views?.method.reveal(item, { select: true, focus: true, expand: true });
+        await vscode.commands.executeCommand('workbench.action.focusActiveEditorGroup');
```
MatiasFernandez: 💭 #16 in this same PR is precisely about not swallowing a failed reveal — but `revealMethodRow` still ends in a bare `catch { /* ignore */ }`, and with the new `focus: true` branch the silent failure got a bit worse: if the `reveal` resolves and `executeCommand('workbench.action.focusActiveEditorGroup')` throws, the tree keeps the focus we deliberately took, the user's cursor is out of the editor, and nothing lands in the GCI log to explain it.

Should this path `logWarning` too, the way `revealDictionaryByName` now does?

client/src/__tests__/explorerMethodReveal.test.ts
```diff
@@ -0,0 +1,72 @@
...
+  const ctl = new ExplorerController(sessionManager);
+  const method = { reveal: vi.fn(async () => {}), selection: [], description: '', visible: true };
+  ctl.setViews({
+    dict: { reveal: vi.fn(), description: '' },
+    category: { reveal: vi.fn(), description: '' },
+    klass: { reveal: vi.fn(), description: '' },
+    hierarchy: { reveal: vi.fn(), description: '' },
+    method,
+  } as never);
+  // revealMethodRow is private; drive it directly for the two paths.
```
MatiasFernandez: 💭 Both cases drive the private `revealMethodRow` through a cast, so they pin the *mechanism* (focus:true + hand focus back) but not the *wiring*: if someone drops `{ focusEditorAfter: true }` from one of the three navigation call sites (3676, 3738, 4426), these two tests still pass and #43 silently regresses.

That's the same gap you covered for the Literals fix with `buildProvidersLiteralsWiring.test.ts` — is there a public entry point here (the `opts.revealMethod` path) that one of the cases could go through instead, so the option is asserted end-to-end?

client/src/gemstoneExplorer.ts
```diff
@@ -3533,12 +3548,12 @@ export class ExplorerController {
       return;
     }
     this.selectDict(new DictItem(dictName, idx + 1));
```
MatiasFernandez: nit: the comment below says a missing category "leaves the panes untouched", but `selectDict` has already run at this point, so the dictionary pane (and its class listing) does change even when we bail out. That's probably the behavior you want — jumping to the dictionary is still useful — just worth saying "leaves the category/classes panes untouched" so the next reader doesn't go looking for a bug when the Explorer visibly moves.

client/src/__tests__/explorerMethodReveal.test.ts
```diff
@@ -0,0 +1,72 @@
...
+          o?: { focusEditorAfter?: boolean },
+        ) => Promise<void>;
+      }
+    ).revealMethodRow(false, INFO, opts);
+  return { method, reveal };
+}
+
+beforeEach(() => vi.clearAllMocks());
+
+describe('#43 revealMethodRow scrolls to the method on navigation', () => {
```
MatiasFernandez: 💭 It's weird to have number references to past issues in test names. I would remove "#43" here

MatiasFernandez (approved):

client/src/gemstoneExplorer.ts
ericwinger: Good catch — fixed in 7e17fe58.

`revealMethodRow` no longer swallows anything. It is now two separate `try`s:

- a rejected `TreeView.reveal` logs `Explorer method reveal failed for <side> method <selector>: <message>`
- a failed `workbench.action.focusActiveEditorGroup` logs `Explorer could not return focus to the editor after revealing <side> method <selector>: <message>`

both via `logWarning` to the GCI channel, matching the wording of the dictionary and category reveal paths.

One extra change your comment prompted: the focus hand-back is now attempted **even when the reveal rejects**. The reveal may have taken focus before failing, and stranding the user's cursor in the tree is the worse outcome — so the `if (takesFocus)` block runs regardless of whether the reveal resolved.

Three tests added in `explorerMethodReveal.test.ts`: reveal rejects → logs and still hands focus back; the focus command throws → logs; both succeed → logs nothing (with an assertion that the reveal actually ran, so it can't pass vacuously). I confirmed both failure tests fail against the old `catch { /* ignore */ }` code, so they pin the fix rather than just describing it.

client/src/gemstoneExplorer.ts
ericwinger: You're right, the comment was wrong — fixed in 7e17fe58. The behaviour is what you guessed: landing on the home dictionary is useful, so that selection is meant to stick.

The comment now says so explicitly:

```
// Check the category actually exists in this dictionary's loaded forest BEFORE mutating the
// category/classes panes. selectDict has just (synchronously) loaded classCategoryEntries, so
// allCategoryPaths() is valid here. Note the dictionary selection above has ALREADY been applied
// and is meant to stick — landing on the home dictionary is still useful. Doing the check first
// means a missing category leaves only the category/classes panes untouched instead of scrolling
// them to a node that isn't there — the jump would otherwise silently appear to do nothing (the
// reported "strange spot").
```

I also swept the same wrong phrasing out of `explorerOmniReveal.test.ts` — both its file header and the inline comment on the missing-category test said "leaves the panes untouched".

client/src/__tests__/explorerMethodReveal.test.ts
ericwinger: Agreed, and thanks for flagging it — those numbers are items in a private checklist, not issue numbers, and read as issue numbers that point somewhere unrelated (`#15`, `#16`, `#18` and `#43` in this repo are all closed dependabot bumps). Removed in 7e17fe58, and applied across the whole PR rather than just this line:

| Was | Now |
|---|---|
| `describe('#43 revealMethodRow scrolls to the method on navigation')` | `describe('revealMethodRow scrolls to the method on navigation')` |
| `describe('#16 — revealDictionaryByName logs a failed reveal…')` | `describe('revealDictionaryByName logs a failed reveal…')` |
| `describe('#15 — revealCategoryByPath checks existence…')` | `describe('revealCategoryByPath checks existence…')` |
| `describe('buildProviders wires the Literals provider to the GCI log (#18)')` | `describe('buildProviders wires the Literals provider to the GCI log')` |

The same numbers are gone from all four test file headers, from the `revealMethodRow` source comment, and from the one in `omniSearchCommand.ts`. Each header now just describes the behaviour it covers.

While in there I also dropped two notes that named parallel in-flight work by round number — those would have gone stale as soon as that work landed.

PR #446, opened 2026-08-17
Omni Search #428 (Round 2): honest Methods scan cap + configurable limit, skipped-scope hint

ericwinger:
Round 2 of the [#428](https://github.com/GemTalk/Jasper/issues/428) follow-ups. Two items, one commit each. **Both manually F5-verified by Eric.**

## Which #428 items this closes

| #428 item (checkbox is ticked, marked 🟣 Round 2) | Commit | Fix write-up |
|---|---|---|
| **`MAX_SERVER_LIMIT` clamp is untested; footer reads "exact" at the fetch ceiling** — from the #423 review, [thread](https://github.com/GemTalk/Jasper/pull/423#discussion_r3776904491) | `e403ae5e` | [🟣 Round 2 fix](https://github.com/GemTalk/Jasper/issues/428#issuecomment-5318883645) |
| **Manual triggering of slow scopes is not discoverable** — "Search behavior & bugs" section | `c23124c0` | see below, linked from the issue |

Both checkboxes on #428 carry a 🟣 Round 2 marker and link back here, and the issue body has a 🟣 Round 2 legend beside the existing 🔵 Round 1 one, so each round's fixes stay attributable.

## 1 — The Methods scan cap was reported as an exact total (`e403ae5e`)

The Methods scan short-circuits once it has enough matches, so its rows are a **floor**, not a total. `exact` was derived from the display cap alone, so **Load all** printed a bare `200 results` over a slice that had been cut off, and the load controls vanished as though everything had been found. The clamp responsible had **no test at all** — that was the review's actual complaint.

There are two bounds, and which one binds changes the right answer:

```
serverLimit = min(maxResultsPerCategory × SERVER_OVERFETCH, maxServerScan)
```

So a provider now reports when its **own** fetch ceiling bound the results, via an optional `reportTruncated` argument to `OmniProvider.search`. The report carries the scope plus `ceiling` and `atCeiling` rather than a bare boolean:

- **`ceiling`** (the configured setting) is what the user sees. Reporting the slice actually scanned showed a limit nobody configured — with the display cap at 60 and the setting at 400 it read `capped at 240` — and it climbed on every Load-more.
- **`atCeiling`** says whether the *setting*, not the over-fetch, bound the scan. Only then can Load-more not get past it; while the over-fetch is the tighter bound, loading genuinely widens the scan, so warning there would be wrong.

The engine collects these into `OmniViewData.truncations`; `exact` now requires Load-all **and** nothing truncated; and the footer shows, beside the count:

> ⚠ Methods scan capped at 400 — narrow the search for the rest

with hover help covering why the cap exists, why the load buttons cannot reach past it, and the setting that raises it.

**The ceiling became a setting.** `gemstone.omniSearch.maxServerScan` (default 200, clamped 20–20 000) replaces the hard-coded constant — the honest answer to "I want more than 200" is "scan wider, and accept a slower search".

## 2 — An All-scope search silently skips three scopes (`c23124c0`)

Source, Literals and Categories are `explicitOnly`, so `providersInScope` drops them whenever the scope is All. Nothing said so, which makes an All-scope "no results" indistinguishable from "not in the image". Verified against a live stone: `no such element` finds **4** methods under Source and **0** under All.

While All is active and something is typed, the view now shows, under the field:

> Not searched here: Source · Literals · Categories — click one to search it

Each scope name is a button, so the hint names the problem *and* is the one-click way out; each carries that scope's own `searchHint` as its tooltip. It stays silent when a heavy scope is already active, when the field is empty, and when the heavy scopes are disabled via `categories`.

**Enter is deliberately unchanged.** The issue floated binding it to "run the search", but it activates the selected row today, and making it scope-dependent would trade one surprise for another.

## Tests — 28 new

| Layer | Count | Covers |
|---|---|---|
| `providers.test.ts` | 7 | the clamp itself (200, not 400 000), over-fetch below the ceiling, truncation reported from the **raw** row count rather than the post-filter count, and the cap-60/setting-400 case at three display caps proving the reported number stops drifting |
| `omniEngine.test.ts` | 6 | `exact` false after Load-all when truncated; the Load-more walk past the ceiling; a term that fits stays exact; one bounded provider among several taints the run; **Load All driven through the real `methodsProvider`** — every other capped-provider test uses a fake, so the actual clamp had never been exercised through `loadAll()` |
| `omniSearchShared.test.ts` (new) | 3 | the postMessage boundary, including a guard that fails if **any** `OmniViewData` field is not forwarded |
| `omniSearchView.test.ts` | 12 | footer text and the cap note's content, visibility and stable position; and all 8 scope-hint cases |
| `omniConfig.test.ts` | 4 | the new setting's default / floor / ceiling / coercion |
| `methodsCeiling.integration.test.ts` (new) | 2 | against a live stone: the generated selector scan really does stop at `limit`, which is what makes a full slice a sound truncation signal |

Each fix was verified **red** before green: reverting `exact` to the cap-only expression fails exactly its 3 engine tests, removing the clamp fails the Load-All test, and disabling the scope hint fails 5 of its 8.

## Notes for the reviewer

- **`LOAD_ALL_LIMIT` is now exported** from `omniEngine.ts` so tests bind to the engine's real Load-all cap instead of a hard-coded `100_000`.
- **`resultsMessage` lists the view's fields one by one instead of spreading the view**, so a new `OmniViewData` field reaches the engine but never the webview. That is how the #14 fix first shipped broken — the flag was computed and the footer still said `200 results`. The new guard test fails if a field is not forwarded; worth knowing if you add one.
- **`display` vs the stylesheet.** Both new elements start hidden via an inline style and the view sets explicit display values. A `display: none` rule in the webview stylesheet beats the view clearing the inline style — that is how the cap note first shipped invisible, and it is *still* how `#breadcrumb` and `#error` behave on `main`. Those two are unfiled defects found while doing this work, deliberately left out of scope here.
- Based on `main`, so this does **not** include Round 1 (#443). Both PRs touch `DESIGN.md`, so whichever merges second will need that conflict resolved; Eric has accepted that.
- The full suite is green in the worktree — client 5677 passed / 111 skipped, server 322, mcp 92 — and the pre-push hook ran it again on both pushes. The CI-mirror matrix (`test-ci-local.sh 3.6.2 3.7.5`) was **not** run, per Eric's call, matching how #443 was opened.


client/src/omniSearch/DESIGN.md
```diff
@@ -152,6 +165,56 @@ Task status against the pinned Phase-2 plan: **1 (webview)**, **2 (own highlight
...
+answer to "I want more than 200" is "raise the scan, and accept a slower search".
+
+The two bounds mean different things to the user, so a provider reports when its OWN ceiling was the
+one that bound it (`OmniTruncationSink`, an optional 4th argument to `OmniProvider.search`, carrying
+the category and the number it stopped at; providers that scan exhaustively and cap client-side never
+report). The engine collects those into `OmniViewData.truncations`:
+
+- **display cap reached** → `hasMore`; more rows are one Load-more away → `N+ shown` + the Load buttons.
+- **fetch ceiling reached** → a `truncations` entry; the count is a floor and no cap can reveal the
+  rest → `N+ shown`, no Load buttons (they cannot help), **and a visible note beside the count**
```
MatiasFernandez: 🚫 This bullet says that when the fetch ceiling is reached there are **no Load buttons (they cannot help)**, but I don't see that in the code: `updateFooter` computes `showLoad = !!view.hasMore && !view.exact`, with no reference to `truncations`. So with `maxServerScan` 200 and the display cap raised to 120 by one Load-more, the slice is `min(480, 200) = 200` → `atCeiling: true` → the note says "narrow the search for the rest", while `hasMore` is still true and both Load buttons are on screen inviting the user to do the thing the note just said cannot work.

And the view test `'never leaves the flex flow, so the Load buttons do not move when it appears'` asserts exactly that combination (`hasMore: true` + capped → `loadMore` visible), so I think the doc is the side that's wrong rather than the code. Could you either fix the bullet to describe what actually happens, or hide the buttons on `atCeiling`? Right now DESIGN.md promises a behavior the tests forbid.

client/src/omniSearch/providers/methodsProvider.ts
```diff
@@ -21,8 +28,17 @@ export type SelectorSearchRunner = (
...
+/** Ceiling on the server slice, bounding the scan cost regardless of maxResultsPerCategory. Comes from
+ *  the `gemstone.omniSearch.maxServerScan` setting (`OmniConfig.maxServerScan`, default 200) so a user
+ *  who wants a wider net can raise it.
+ *
+ *  It bounds the RESULTS as well as the cost: `searchSelectors` short-circuits the instant it has that
+ *  many matches, so a broad term can never return more no matter how far the display cap is raised —
+ *  "Load all" included. `search` therefore reports truncation to the engine, which would otherwise
+ *  present a cut-off count as an exact total, with nothing on screen saying the scan gave up (#14). */
+function scanCeiling(cfg: OmniConfig): number {
+  return cfg.maxServerScan > 0 ? cfg.maxServerScan : 200;
```
MatiasFernandez: 💭 Two small things about this fallback:

- the `200` is a second copy of `OMNI_DEFAULTS.maxServerScan` — shouldn't we use the constant, so raising the default doesn't silently leave a stale 200 here?
- `readOmniConfig` already clamps `maxServerScan` into 20–20 000, so the `> 0` guard can only ever fire for a hand-built `OmniConfig` (i.e. a test). Is that the case you had in mind, or would `cfg.maxServerScan` straight up be honest enough?

client/src/omniSearch/__tests__/omniSearchShared.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  // This payload lists its fields one at a time instead of spreading the view, so a NEW OmniViewData
+  // field reaches the engine but never the footer. That is precisely how the #14 fix first shipped
+  // broken: the engine computed the truncation flag, `resultsMessage` didn't forward it, the webview
+  // read `undefined`, and the count still said "200 results". This guard fails the moment another
+  // field is added to OmniViewData without being forwarded here.
+  it('forwards every OmniViewData field, so a new field cannot be silently dropped', () => {
+    const view = viewData({ pivotTitle: 'Senders of foo' });
+    const msg = resultsMessage(view, chrome);
+    for (const key of Object.keys(view)) {
```
MatiasFernandez: 💭 This guard iterates `Object.keys(view)` on the fixture, so it's only as exhaustive as `viewData()` is. A new **required** field on `OmniViewData` does get caught (the fixture stops compiling, you add it, the loop covers it), but a new **optional** one — the shape `pivotTitle` already has — never appears in `Object.keys` unless someone remembers to pass it via `over`, which is exactly the kind of remembering the #14 regression proved we can't count on.

Would typing the base fixture as `Required<OmniViewData>` (or building it from a `Record<keyof Required<OmniViewData>, unknown>` key map) close that? Then TS forces every future field — optional included — into the fixture, and the loop picks it up for free.

client/src/omniSearch/omniTypes.ts
```diff
@@ -126,10 +129,55 @@ export interface OmniCancel {
...
+ * user just hit (triage #14).
+ *
+ * It carries the scope and the number so the UI can name both ("Methods stopped after 200") rather
+ * than show an anonymous warning — and so any provider that gains a ceiling later is covered without
+ * touching the view.
+ */
+export interface OmniTruncation {
+  /** The category whose own fetch ceiling bound the results. */
+  categoryId: OmniCategoryId;
+  /** How many matches this run's scan actually collected before stopping. */
```
MatiasFernandez: nit: this says `scanned` is "how many matches this run's scan actually collected", but the provider reports `scanned: serverLimit` — the slice size it *asked* for, not `rows.length`. They coincide today because the report is gated on `rows.length >= serverLimit`, so maybe "the slice size that bound this run" describes it more accurately for the next reader.

package.json
```diff
@@ -19,8 +19,16 @@
     "node": ">=22.15.1"
   },
   "devEngines": {
```
MatiasFernandez: nit: these `devEngines` and `enum` reflows look like an editor's JSON formatter rather than part of the change — `format:check` only globs `**/*.{ts,mts,cts,js,mjs,cjs}`, so nothing gates package.json either way, and prettier wouldn't expand the `enum` arrays. Worth reverting the unrelated hunks so the diff is just the new `maxServerScan` setting?

MatiasFernandez (approved):
Approved. Worth reviewing the unrelated package.json format changes

client/src/omniSearch/DESIGN.md
ericwinger: Good catch — the doc was the wrong side, not the code. Fixed in `00278a73`: the bullet now says the Load buttons stay (they raise the display cap for the whole view and still widen any *other* in-scope provider that has more to show), and the ⚠ note — not a vanished button — is what says the capped scope can't grow. `updateFooter` and its view test are unchanged.

client/src/omniSearch/providers/methodsProvider.ts
ericwinger: Both addressed in `00278a73`: dropped the `scanCeiling` helper entirely and read `cfg.maxServerScan` straight. You're right on both counts — `readOmniConfig` already clamps it into 20–20 000 so the `> 0` guard was dead, and the second `200` was a stale duplicate of `OMNI_DEFAULTS.maxServerScan`. The ceiling semantics moved to a comment at the call site.

client/src/omniSearch/__tests__/omniSearchShared.test.ts
ericwinger: Done in `00278a73` — the base fixture is now typed `Required<OmniViewData>` (with `pivotTitle` populated), so TypeScript forces every future field, optional ones included, into it and the `Object.keys` loop covers it for free. Also dropped the now-redundant `pivotTitle` override in the guard test.

client/src/omniSearch/omniTypes.ts
ericwinger: Reworded in `00278a73` to "the slice size that bound this run — the number the scan was asked for and stopped at", noting it coincides with the collected row count only because the report is gated on `rows.length >= serverLimit`.

package.json
ericwinger: Reverted both reflow hunks (devEngines and the `matchMode` enum) in `00278a73` — the diff vs main is now just the new `maxServerScan` setting.

PR #448, opened 2026-08-17
Omni Search #428 (Round 6): panel controls for preview pane, All-scope filter and match algorithm

ericwinger:
Round 6 of the #428 follow-ups. Four items, all about giving the **bottom-docked panel** — wide but short, and the default UI — control over what a search spends, plus one investigation that ends with "change nothing".

Each fix links to its 🔴 Round 6 comment on #428.

| #428 item | What changed | Commit |
|---|---|---|
| [A toggle for the preview pane](https://github.com/GemTalk/Jasper/issues/428) | ◧ chip; hides the pane **and** stops the per-row source fetch | `9a4b0bc7` |
| [Scope exclusions in "All"](https://github.com/GemTalk/Jasper/issues/428) | `Scopes` menu; a scope leaves "All" but keeps its tab | `9a4b0bc7` |
| [Change the match algorithm from the UI](https://github.com/GemTalk/Jasper/issues/428) | `Fuzzy`/`Substring`/`Prefix` chip, cycles on click | `80fa1cc0` |
| [Explore `fuzzysort` vs the hand-rolled matcher](https://github.com/GemTalk/Jasper/issues/428) | Benchmarked; **not adopted**, with reasons in DESIGN.md | `80fa1cc0` |

## Why these, and why the panel

`#body` is a flex row: `#results` is `flex: 1 1 55%`, `#preview` `flex: 1 1 45%`. In a full editor tab that preview pays off; in the bottom panel it buys a source view a few lines tall while costing the labels — `Class>>selector` + dictionary + detail — nearly half their width. The host that pays most gets least, so the toggle hands that room back. The scope filter is the same argument for round-trips: Methods queries the stone on every keystroke, and per-keystroke fan-out is exactly where the panel hurts.

## One contract for all three controls

A setting supplies the **starting** value; the in-panel control owns it for the session; toggling never rewrites settings. That last part is load-bearing rather than stylistic — writing a setting fires `onDidChangeConfiguration`, which drops the cached engine and **re-primes the corpus**. A self-persisting toggle would cost a stone round-trip per click. It also matches how `caseSensitive` has always behaved.

New settings: `gemstone.omniSearch.previewPane`, `gemstone.omniSearch.excludeFromAll`.

## The distinction #41 turns on

`providersInScope` already held `explicitOnly` categories (Source/Literals/Categories) out of the "All" fan-out — *outside All, but still reachable by their own tab*. The filter simply lets the user put an ordinary category in that same state. That is what separates it from `gemstone.omniSearch.categories`, which removes the provider **and** the tab; conflating the two is the reported defect. Scoping directly to an excluded category always runs it, and there is a test pinning exactly that.

## Two defects found while building

- **`filterPivot` read `config.matchMode`** — the value baked in when the engine was constructed — so switching algorithms did nothing while a references list was open. Now reads the live value.
- **`resultsMessage` must not carry `previewPane`.** It enumerates fields by hand, and re-sending the toggle would silently undo it on the user's next keystroke. Guarded by a test.

## On #44 — the case for changing nothing

The real `omniMatch.ts` was compiled out of the worktree and benchmarked against `fuzzysort` 4.0.2 over the repo's vendored class names (2,063), selectors (8,230), and a 20,000-name corpus scaled from them (synthetic, authentic shapes).

fuzzysort is **3–24× faster** — and it does not matter: `omniMatch`'s worst case there is **16 ms for a one-character query**, inside a frame and behind a 120 ms debounce. **Recall was identical on every query.** Ranking is arguably better ours on camelCase identifiers (`oc` → `OCCURRENCE, Once, ONCE, OrderedCollection` vs `OCCURRENCE, SubOnlyCVar, OrderedCollection, Lock`).

What decides it: fuzzysort is **fuzzy-only** — no `substring`, no `prefix` — and has **no case-sensitive option**. Jasper ships both as settings, and this PR just made the mode switchable mid-search. Adopting it would remove two user-visible features to solve a problem nobody has.

**No dependency was added.** The benchmark ran in a scratch directory outside the checkout; `package.json`/`package-lock.json` are untouched and `omniMatch.ts` is unchanged.

## Verification

**All three UI items were manually F5-verified by Eric** — the preview toggle, the scope filter, and the match-algorithm chip including the pivot fix. #44 changed no behavior, so it is reproducible rather than verifiable: re-run the benchmark, or check `git diff package.json` is empty.

47 new tests across four new files (no existing test file touched). Two regression guards were verified **red** by breaking the source and re-running, not merely asserted: `previewPane` riding on a results message, and narrowing "All" failing to reset the page cap. The `filterPivot` fix was likewise verified red.

Gate: lint / format / compile clean; full `npm test` green — **client 5683 passed / 111 skipped, server 322, mcp 92**.

⚠️ The CI-mirror matrix (`test-ci-local.sh`) was **not** run, deliberately: it reprovisions the shared test stones, and two other #428 sessions are working against them right now. This diff is entirely `client/src/omniSearch/` plus two new settings — no server, RB-engine, GCI or Smalltalk surface for the matrix to exercise.

## Note for reviewers on merge order

Four #428 PRs are open and several touch the same webview files. This branch is off `main` `75af6306` and expects an ordinary 3-way merge with #446 in `omniSearchShared.ts` / `omniSearchView.js`. ⚠️ `resultsMessage`'s chrome parameter gained two required fields here (`excludedFromAll`, `matchMode`); a new caller must pass both.


client/src/omniSearch/omniSearchView.js
```diff
@@ -386,6 +557,10 @@
         clearTimeout(previewTimer);
         previewTimer = null;
       }
+      // The references list lives in this pane, so asking for references while the pane is hidden
+      // would be a gesture with no visible result. Treat the request as intent to see the pane and
+      // switch it back on (the toggle updates, so it doesn't look like a glitch).
+      if (!previewEnabled) setPreviewEnabled(true);
```
MatiasFernandez: 💭 `setPreviewEnabled(true)` ends by calling `requestPreview()`, which re-arms the exact timer the three lines above just cleared — so asking for references while the pane is hidden schedules a source fetch for the active row. The `previewMode === 'source'` guard on the `preview` reply throws the result away, so nothing renders wrong, but we pay a round-trip on the one gesture that was supposed to be about references, and it's the same round-trip the toggle exists to avoid.

Would flipping the order help — `setPreviewEnabled(true)` first, then the `clearTimeout` block, then `previewMode = 'refs'`? The clear would still win, and the sequence would read as "show the pane, cancel any pending source preview, render the refs".

package.json
```diff
@@ -212,6 +212,20 @@
...
+          "gemstone.omniSearch.previewPane": {
+            "type": "boolean",
+            "default": true,
+            "markdownDescription": "Show the source preview beside the results, filling in as you move down the list. The pane takes about **45%** of the width, which the bottom panel — wide but short — can least afford, so turning it off gives that room back to the result labels and also stops the per-row source fetches. This is the value the panel **starts** with; the ◧ button in the panel switches it for the current session without changing this setting."
+          },
+          "gemstone.omniSearch.excludeFromAll": {
+            "type": "array",
+            "items": {
+              "type": "string",
+              "enum": ["classes", "methods", "dictionaries", "globals"]
```
MatiasFernandez: 💭 This enum is a hand-maintained copy of the non-`explicitOnly` ids in `OMNI_CATEGORIES`. If a category is added later (or an existing one flips `explicitOnly`), the enum goes stale silently and settings.json rejects an id that `readOmniConfig` would happily honour — it already derives the same rule from `CATEGORY_BY_ID[id].explicitOnly`, so the source of truth is right there.

There's precedent in the repo for pinning package.json against code (`client/src/__tests__/databaseMenuOrder.test.ts`, `editorContextMenu.test.ts`). Worth a small test asserting this enum equals `OMNI_CATEGORIES.filter((c) => !c.explicitOnly).map((c) => c.id)`?

client/src/omniSearch/__tests__/omniPreviewToggle.test.ts
```diff
@@ -0,0 +1,206 @@
...
+
+// The real chrome from renderOmniHtml, trimmed to what this file touches — including the two
+// Round-6 controls, which the older view test's shell predates.
+const SHELL =
+  '<div id="omni">' +
+  '<div id="tabs"></div>' +
+  '<div id="field"><input id="query" type="text"><button id="clear" style="display:none">×</button></div>' +
+  '<button id="case">Aa</button>' +
+  '<button id="previewToggle" aria-pressed="true">◧</button>' +
+  '<span id="scopeFilterWrap"><button id="scopeFilter" aria-expanded="false">▽</button>' +
```
MatiasFernandez: 💭 Each of the three new jsdom files hand-copies a `SHELL` of `renderOmniHtml`'s chrome, and they have already drifted from it: this one (and `omniScopeFilterMenu`'s) renders the filter button as `▽` and omits `#matchMode` entirely, while the real chrome in `omniSearchShared.ts` says `Scopes` and does include the chip. The null guards in the view (`if (!matchModeEl) return;`) mean the tests stay green against chrome that can't exist in production, so the drift is invisible.

Since all three also repeat the same `beforeAll` script load, `api()` and `mount()`, is it worth one shared `__tests__/omniViewHarness.ts` — ideally deriving the shell from `renderOmniHtml({ showPin: false })` so it can't drift at all?

client/src/omniSearch/__tests__/omniScopeFilterMenu.test.ts
```diff
@@ -0,0 +1,217 @@
...
+    const { posted, view } = mount({ excludedFromAll: ['methods'] });
+    el('scopeFilter').click();
+
+    optionFor('methods').click();
+
+    expect(view.excludedFromAll()).toEqual([]);
+    expect(posted).toContainEqual({ command: 'setExcludedFromAll', excludedFromAll: [] });
+  });
+
+  it('stays open while several scopes are toggled', () => {
```
MatiasFernandez: 💭 In the panel these two clicks aren't adjacent: the first posts `setExcludedFromAll`, the host replies with a results message, and `renderScopeFilter()` tears down and rebuilds every option button before the second click lands. This test never sees that rebuild, so it proves the menu survives two synchronous clicks rather than the flow the name describes.

Worth interleaving a `results` echo between them, the way the match-mode test does with its `echo()` helper?

client/src/omniSearch/omniEngine.ts
```diff
@@ -411,6 +451,15 @@ export function createOmniEngine(deps: OmniEngineDeps): OmniEngine {
       pivot = null;
       return runSearch(lastRawValue);
     },
+    async setExcludedFromAll(ids) {
+      const cats = new Map(OMNI_CATEGORIES.map((c) => [c.id, c]));
```
MatiasFernandez: nit: `omniTypes` already exports `CATEGORY_BY_ID` (built once from `OMNI_CATEGORIES`), and `omniConfig.ts` in this same PR imports it for exactly this `explicitOnly` check. Could this use it rather than rebuilding the Map on every call?

```ts
excludedFromAll = new Set(ids.filter((id) => CATEGORY_BY_ID[id]?.explicitOnly !== true));
```

client/src/omniSearch/omniSearchShared.ts
```diff
@@ -92,8 +95,16 @@ export function resultsMessage(
...
     pinned: chrome.pinned,
     referencesInPreview: chrome.config.referencesInPreview,
+    // Which categories are currently held back from "All" — the scope filter renders its checkboxes
+    // from this, so the panel always shows what the engine actually did.
+    excludedFromAll: chrome.excludedFromAll,
+    // Engine-owned, like `caseSensitive` — so it IS re-sent every time, unlike `previewPane`.
+    matchMode: chrome.matchMode,
     placeholder: placeholderFor(view.pivot ? null : chrome.scopeId),
   };
+  // NOTE: `previewPane` is deliberately NOT here. Once the webview has it from `configMessage` the
```
MatiasFernandez: nit: This note is doing real work — the PR body calls the omission load-bearing — but it sits after the `return`, which is where the next person adding a field to `resultsMessage` is least likely to scroll. Would it be better in the TSdoc above the function, so it also shows on hover at the call site?

client/src/omniSearch/__tests__/omniMatchModeControl.test.ts
```diff
@@ -0,0 +1,246 @@
...
+    };
+    const engine = createOmniEngine({
+      providers: [provider],
+      config: cfg({ matchMode: 'fuzzy' }),
+      resolveReferences: () => references,
+    });
+    await engine.search('arr');
+    await engine.pivot(0);
+
+    // Fuzzy: "oc" matches OrderedCollection as a subsequence, and Collection too (C-o-...-c? no) —
```
MatiasFernandez: nit: `"Collection"` does match `oc` as a subsequence (`o` at index 1, `c` at index 5), so the `(C-o-...-c? no)` aside is backwards. The assertion itself is fine — it only checks `OrderedCollection` is present — but this comment is what a future reader will trust when working out what fuzzy does here.

client/src/omniSearch/DESIGN.md
```diff
@@ -152,6 +152,70 @@ Task status against the pinned Phase-2 plan: **1 (webview)**, **2 (own highlight
 indicator)** done; **3 (hover)** delivered as a preview pane; **5 (count)** shows an exact count once
 Load-all makes it exact, else `N+` (a true pre-count query stays a follow-up).
 
+## Controlling what a search costs (#428 items #40 / #41)
```
MatiasFernandez: nit: The heading says items #40 / #41, but the section's third bullet is the match algorithm, which is #42. Worth adding it so searching DESIGN.md for #42 lands here?

MatiasFernandez (approved):

package.json
ericwinger: Done in 70dd133e. Added a pin in `omniSettings.test.ts` that asserts the manifest enum *is* the derived list, so the two can't silently diverge:

```ts
const inAll = OMNI_CATEGORIES.filter((c) => !c.explicitOnly).map((c) => c.id);
expect(excludeFromAll.items.enum).toEqual(inAll);
```

Followed the `databaseMenuOrder.test.ts` precedent you pointed to. Checked it isn't vacuous — dropping `globals` from the manifest enum turns it red.

client/src/omniSearch/__tests__/omniPreviewToggle.test.ts
ericwinger: Done in 70dd133e. New `__tests__/omniViewHarness.ts` derives the mounted DOM straight from `renderOmniHtml({ showPin: false })` (stripping the inlined `<script>`, since jsdom won't run an innerHTML-injected script — the view code is loaded separately by `loadOmniView`). All three files now go through `loadOmniView` / `mountOmniView`; the hand-copied `SHELL` constants are gone (~160 lines removed).

You were right that the drift was live — two of the shells rendered the old `▽` button and omitted `#matchMode`, and the view's missing-element guards kept them green. To prove the harness now *can't* hide that, I temporarily dropped `#matchMode` from `renderOmniHtml` and the chip tests went red (they couldn't before). Also fixed a stale doc-comment in `omniScopeFilterMenu.test.ts` that still called it "the ▽ menu".

client/src/omniSearch/omniEngine.ts
ericwinger: Done in 70dd133e:

```ts
excludedFromAll = new Set(ids.filter((id) => CATEGORY_BY_ID[id]?.explicitOnly !== true));
```

Grepped the rest of the PR for `new Map(OMNI_CATEGORIES` — this was the only occurrence.

client/src/omniSearch/omniSearchShared.ts
ericwinger: Done in 70dd133e. Moved it into the TSdoc above `resultsMessage`, so it shows on hover at the call sites and sits where someone adding a field will actually read it — with a line reminding them to confirm a new field is engine-owned before forwarding it.

client/src/omniSearch/__tests__/omniMatchModeControl.test.ts
ericwinger: Fixed in 70dd133e. You're right — `o`@1, `c`@5, so `Collection` matches `oc` too. Reworded to say both names match under fuzzy and that the prefix assertion below is the real discriminator (neither starts with `oc`).

PR #449, opened 2026-08-17
Omni Search #428 (Round 7): references pivot honors the selected scope, and names its own exit

ericwinger:
Fixes item **#20** of issue #428: the references pivot ignored the selected scope while claiming to
apply it. **Manually F5-verified by Eric — both the defect and the fix.**

| #428 item | Fix comment | Commit |
|---|---|---|
| #20 — the `references` pivot ignores the selected scope / filter type | ⚪ [Round 7 fix](https://github.com/GemTalk/Jasper/issues/428#issuecomment-5320521262) | `cf5184ac` |
| #39 — should the Literals scope match a variable name? | 🚫 [Not doing](https://github.com/GemTalk/Jasper/issues/428#issuecomment-5320441641) | — (decision only, no code) |

## The bug

Three symptoms, one cause. `runSearch` opens with a pivot branch that returns before scope is ever
consulted:

```ts
if (pivot) { current = filterPivot(pivot.results, rawValue.trim()); return pivotView(); }
```

`setScope` sets `scopeId` and calls `runSearch(lastRawValue)`, so a scope picked during a pivot had
**zero** effect on what was displayed — while the reply's chrome still carried `scopeId`, which is what
lit the tab up. The scope then took effect **invisibly** on the next real search, i.e. when the pivot was
dismissed, narrowing a restored search the user never asked to narrow.

## The fix

`setScope` clears the pivot before re-running, so **picking a scope leaves the references view and
applies that scope to the restored search**. A scope belongs to the *search*, not to a references view:
every pivot row is a method (`methodRowsToResults`), so a Classes/Globals/Dictionaries filter over them
is meaningless. The tab's active state and the list now always agree.

Rejected, and recorded in `DESIGN.md` so it isn't relitigated:

- *Filter the pivot rows by scope* — meaningless with one category: every tab is either a no-op or
  empties the list.
- *Disable the tabs while pivoted* — honest, but it hides a control instead of giving it a meaning, and
  it lives in the webview rather than the engine.

## The pivot now names its own exit

The pivot's only ways out are `Esc` and `←` (the latter only with the caret at the start of the field),
neither visible, and clearing the box does **not** escape — an empty filter matches every reference row,
so clearing widens the list. Rather than overload the clear gesture, the new exported `PIVOT_EXIT_HINT`
is appended to `OmniViewData.pivotTitle`, so the breadcrumb reads `References to Foo — Esc to go back`.
Clearing deliberately stays a filter reset (Eric's call), and a test pins that so it isn't "fixed" later
by accident.

Deliberately **not** added to `ReferencePreview.title`: in `referencesInPreview` mode there is no pivot
and `Esc` closes the panel, so the same words there would be false. There is a test for that too.

⚠️ **Reviewer note:** the hint is only as visible as the breadcrumb, and the breadcrumb currently cannot
render at all — `omniSearchShared.ts` sets `#breadcrumb { display: none }` while the view reveals it by
clearing an *inline* style. That is a separate defect being handled with item #28 in another round; this
PR's behavioural fix does not depend on it, and this PR does not touch the webview files.

## Repro (needs `gemstone.omniSearch.referencesInPreview: false`)

At the default `true`, references fill the preview pane and the pivot is never entered, so the bug is
unreachable on the default path. Against a 3.7.5 stone:

1. `Ctrl+Shift+A`, type `Fraction` in the **All** scope — 25 rows (9 Class, 16 Method).
2. `Alt+Enter` on the `Fraction` class row — 14 reference rows, all methods.
3. Click the **Classes** tab — *before:* the tab lights up and the list stays the 14 method rows.
   *After:* you land back in the search, 9 class rows, tab and list agreeing.
4. `Esc` — *before:* the box still read `Fraction` but the list had silently dropped to 9 class rows.
   *After:* nothing to exit; the scope change already restored the list.

## Tests

8 tests in `omniEngine.test.ts`, **written before the fix — 6 failed red for the right reasons**, all
green after: a scope click leaves the pivot; the shown rows match the reported scope; no silent narrowing
on exit; the restored search is genuinely re-run rather than reused; a scope chosen *before* a pivot still
survives an `Esc` exit; clearing stays a filter reset; the breadcrumb carries the exit hint and keeps it
while filtering; the preview list's title stays clean.

Gate on `eric/issue428-round7`: lint, format, compile, omniSearch + keybindings **174/174**, full
`npm test` **client 5644 passed / 111 skipped, server 322, mcp 92** — and the pre-push hook re-ran the
whole suite green (no `--no-verify`).

⚠️ The CI-mirror matrix (`test-ci-local.sh`) was **not** run, deliberately: it reprovisions the shared
test stones that two other active #428 sessions are using, and this diff is 3 files inside
`client/src/omniSearch/` with no server, GCI, RB-engine or Smalltalk surface for the matrix to exercise.
Same reasoning Round 4 used for PR #444.

## Scope of the diff

3 files: `omniEngine.ts` (the pivot region + `PIVOT_EXIT_HINT`), `__tests__/omniEngine.test.ts`, and
`DESIGN.md` (one new section, *5c. A scope belongs to the search, not to a references view*). No webview
files — `omniSearchView.js` and `omniSearchShared.ts` belong to a concurrent round and are untouched.


client/src/omniSearch/omniEngine.ts
```diff
@@ -381,6 +389,13 @@ export function createOmniEngine(deps: OmniEngineDeps): OmniEngine {
     async setScope(newScope) {
       scopeId = newScope;
       scopeLimit = config.maxResultsPerCategory; // a fresh scope starts at the base cap
+      // A scope belongs to the SEARCH, not to a references view, so picking one LEAVES the pivot
+      // (#20). Every pivot row is a method (`methodRowsToResults`), so there is nothing meaningful for
+      // a Classes/Globals/Dictionaries filter to do to them. Without this the tab lit up as active
+      // while `runSearch`'s pivot branch returned the unchanged reference rows — the chrome claimed a
+      // filter that was never applied — and the scope then took effect invisibly on the way out,
+      // narrowing a restored search the user never asked to narrow.
+      pivot = null;
```
MatiasFernandez: 💭 The engine side of this is clean, but I think the webview can still end up in the "chrome says one thing, list shows another" state this PR is fixing — one step later.

`runSearch`'s pivot branch deliberately doesn't touch `lastRawValue`, so while pivoted the *box* holds the filter term and `lastRawValue` holds the search. In the webview the input is never rewritten from the host (`omniSearchView.js` only rewrites `inputEl.value` on the Clear click; `resultsMessage` carries no value), so: search `foo` → pivot on `Foo` → type `also` to filter the reference rows → click **Classes**. The list correctly becomes the `foo` search narrowed to classes, but the box still reads `also`, and the next keystroke turns it into a real search for `alsoX`, throwing the restored list away.

The QuickPick host handles the equivalent case explicitly (`omniSearchController.ts` `exitPivot` does `qp.value = lastRawValue` before refreshing) — is it worth having the webview do the same on a pivot→search transition? And either way, would it be worth a test that *filters inside the pivot first* and then picks a scope, so the `lastRawValue`-not-clobbered invariant is pinned? All eight new tests call `setScope` on an unfiltered pivot, which is the one case where the box and `lastRawValue` happen to agree.

client/src/omniSearch/omniEngine.ts
```diff
@@ -68,10 +68,18 @@ export interface OmniViewData {
...
+   *  see `PIVOT_EXIT_HINT`. */
   pivotTitle?: string;
 }
 
+/** Appended to the pivot breadcrumb so the way OUT of a references view is discoverable (#20). The
+ *  pivot replaces the whole result list and offers no visible exit affordance, so a user who does not
+ *  already know the gesture has to guess — clearing the box looks like it should work and doesn't (an
+ *  empty filter matches every reference row). Deliberately NOT added to `ReferencePreview.title`: in
+ *  `referencesInPreview` mode there is no pivot and Esc closes the panel, so the hint would be false. */
+export const PIVOT_EXIT_HINT = 'Esc to go back';
```
MatiasFernandez: 💭 Worth noting that this only reaches the webview host. `omniSearchController.ts` (the QuickPick path, still live — `omniSearchCommand.ts:281`) keeps its own copy of the pivot logic and sets `qp.title = decorateTitle(view.title)` on pivot, so no hint there. That one's fine in practice (it shows a real back button and a `← back to search` placeholder), so I'm not asking for a second implementation.

What did strike me is the other half: that controller's `setScope` doesn't clear `pivot` and its `refresh` early-returns while pivoted — i.e. the exact bug #20 describes. It's currently unreachable only because `pivotToReferences` replaces the title bar with `[backButton()]`, so there's no scope button to click while pivoted. That's a fairly invisible invariant holding the same bug shut in the other host. Should this comment (or a short one over there) say so, so nobody restores the scope buttons during a pivot for a good-looking reason and silently reopens #20?

client/src/omniSearch/omniEngine.ts
```diff
@@ -311,7 +319,7 @@ export function createOmniEngine(deps: OmniEngineDeps): OmniEngine {
         hasMore: false,
         exact: true,
         pivot: true,
-        pivotTitle: pivot?.title,
+        pivotTitle: pivot ? `${pivot.title} — ${PIVOT_EXIT_HINT}` : undefined,
```
MatiasFernandez: 💭 This bakes presentation into the data: the separator, the wording, and the decision to show a hint at all are now fixed inside the engine, and the only consumer that can undo it is one that string-splits on `—`.

Would a separate optional field (`pivotHint?: string`, or `pivotTitle` + `pivotExitHint`) be worth it instead? The view could then render the hint dimmed/smaller next to the breadcrumb rather than as part of the same run of text, and a host that already has a visible exit affordance could just drop it — which is roughly the situation the QuickPick controller is in today. It'd also make the tests assert the hint directly rather than through a concatenation.

client/src/omniSearch/__tests__/omniEngine.test.ts
```diff
@@ -392,6 +392,48 @@ describe('createOmniEngine', () => {
...
+    const engine = createOmniEngine({
+      providers: [fakeProvider('classes', [classResult('Foo')])],
+      config: cfg(),
+      resolveReferences: () => refView,
+    });
+    const search = await engine.search('foo');
+
+    const pivot = await engine.pivot(search!.rows[0].id);
+    expect(pivot!.pivotTitle).toBe(`References to Foo — ${PIVOT_EXIT_HINT}`);
+    expect(PIVOT_EXIT_HINT).toContain('Esc');
```
MatiasFernandez: 💭 Since the expectation interpolates `PIVOT_EXIT_HINT`, this assertion passes for *any* value of the constant — including an empty string or something that no longer names a key. `toContain('Esc')` covers part of that, but the ` — ` join and the final shape of the breadcrumb are still self-fulfilling.

Would it be worth hardcoding the expected string in at least this one test — `expect(pivot!.pivotTitle).toBe('References to Foo — Esc to go back')` — so a change to the user-facing wording has to be made deliberately and shows up in the diff? The interpolated form reads better in the other assertions where the title isn't what's under test.

client/src/omniSearch/DESIGN.md
```diff
@@ -73,6 +73,27 @@ icon, isEnabled, search(query, token) }`. Providers: **Classes, Methods, Diction
...
+   behaviour: the tab lit up as active while the list was left untouched, so the chrome advertised a
+   filter that was never applied — and the scope then took effect **invisibly** when the pivot was
+   dismissed, narrowing a search the user never asked to narrow.
+   **The pivot names its own exit.** Its only ways out are `Esc` and `←` (the latter only with the
+   caret at the start of the field), neither of them visible, and clearing the box does *not* escape —
+   an empty filter matches every reference row, so clearing widens the list instead. Rather than
+   overload the clear gesture, the breadcrumb carries the exit: `PIVOT_EXIT_HINT` ("Esc to go back") is
+   appended to `OmniViewData.pivotTitle`. It is deliberately **not** added to `ReferencePreview.title`
+   — in `referencesInPreview` mode there is no pivot and `Esc` closes the panel, so the same words
+   there would be false. ⚠️ The hint is only as visible as the breadcrumb: while `#breadcrumb` carries
```
MatiasFernandez: nit: this ⚠️ is accurate today but describes a defect that item #28 is about to fix, and nothing will prompt anyone to come back and delete it — DESIGN.md is the doc a future reader (or an agent) trusts about how the breadcrumb behaves. Could it carry the issue reference (e.g. "until #28 lands") so it's obvious when it has expired?

MatiasFernandez (approved):

client/src/omniSearch/omniEngine.ts
ericwinger: Agreed — done in `532fb68e`. The hint is its own field now, and the engine no longer decides how it looks.

```ts
  /** Breadcrumb title while pivoted (e.g. "Senders of printString"). Just the title: the way out
   *  travels separately, in `pivotHint`. */
  pivotTitle?: string;
  /** How to leave the pivot, offered ALONGSIDE `pivotTitle` rather than glued into it, so each host
   *  decides how to present it: the webview renders it as a quieter aside beside the breadcrumb, and a
   *  host whose own chrome already shows an exit can simply ignore the field. Set only while pivoted;
   *  the wording is `PIVOT_EXIT_HINT`. */
  pivotHint?: string;
```

```ts
        pivot: true,
        pivotTitle: pivot?.title,
        pivotHint: pivot ? PIVOT_EXIT_HINT : undefined,
```

The panel renders it as its own span rather than appending to the breadcrumb text, so it can be styled independently — `#breadcrumb .crumb-hint { margin-left: 0.6em; font-size: 0.85em; opacity: 0.75; }` — and the em-dash is gone; the gap and the lighter weight do the separating. `setBreadcrumb(title, hint)` skips the span entirely when a host sends no hint, so "drop it" is now just not setting the field.

The tests assert the hint directly (`expect(pivot!.pivotHint).toBe(PIVOT_EXIT_HINT)`) with a separate assertion that `pivotTitle` is the plain title, and there are new jsdom tests for: the hint rendering in its own element with the title text left plain, the title standing alone when no hint is offered, no stale hint left behind when the pivot is exited, and the stylesheet actually carrying the dimming rule (that last one because the stylesheet is a template literal, so a rule is easy to lose silently). I mutated each of those back to the old behaviour and confirmed the tests fail before restoring the fix.

One correction to your example, for the record: the QuickPick controller is gone — `omniSearchController.ts` was deleted on main, so the panel and Spotter are the only consumers today. Your point stands on its own for the panel (it wants to style the hint differently from the title), and the field shape is what makes a future host able to opt out.

Two things I also cleaned up while in these files:

- **Private tracking numbers.** Several comments, a doc-comment, a test doc-block and two `describe` names carried bare `#NN` markers from my own notes, which read as unrelated GitHub issue numbers to anyone else. Removed, with the prose saying what the behaviour is instead.
- **A stale comment.** `LOAD_ALL_LIMIT` claimed it "matches the QuickPick controller's constant so the two UIs behave identically" — that controller no longer exists, so the sentence pointed at nothing.

Eric manually F5-verified the rendered result: the breadcrumb reads `Senders of printString` with `Esc to go back` beside it, smaller and fainter, and it disappears cleanly on Esc.

This branch also has main merged in (it was 54 commits behind); the conflicts were the design doc's renumbered sections and the engine test's import list.

PR #450, opened 2026-08-17
Omni Search #428 (Round 5): heavy-scope icon, references affordance, scope cycling

ericwinger:
Round 5 of the #428 follow-ups — three Omni Search panel UI affordances (plus one companion fix), all **manually F5-verified by Eric**. Webview-only changes in `omniSearchView.js` / `omniSearchShared.ts` — no VS Code keybinding is involved (the panel is a webview, so `Tab` is captured inside it).

## What's in it

- **Distinct icon for heavy scopes.** The explicit/heavy scopes (Source, Literals, Class Categories) now carry a ⌛ hourglass marker instead of the plain 🔍 magnifier the whole field otherwise reads as, so an expensive full-image scan looks visibly different. (`omniSearchShared.ts` CSS)
- **References affordance.** A new `↗ References` chip in the field row (styled like the case-sensitivity toggle, per the review suggestion) lights up whenever references/senders are shown — both the classic list pivot **and** the default sticky preview-pane list — and click-to-exit. Also repairs the `#breadcrumb`: the stylesheet hid it with `display:none` while the view "revealed" it by clearing the inline style, so the "References to X" breadcrumb never actually appeared. Companion fix: the identical dead-affordance bug on the in-panel `#error` banner (errors already toast; this restores the in-context banner as a second layer).
- **Scope-switch shortcut.** `Tab` / `Shift+Tab` from the search field cycles the scope tabs (the JetBrains Search-Everywhere gesture), wrapping in render order, with a footer legend advertising the shortcut.

## Tests

New jsdom / guard tests: `omniHeavyScopeIcon`, `omniReferencesAffordance`, `omniScopeCycle`. Full omniSearch + keybindings suite green (186); full `npm test` green (client / server / mcp).

## Notes

- All three fixes were **manually F5-verified by Eric**; details in the [🟡 Round 5 summary](https://github.com/GemTalk/Jasper/issues/428#issuecomment-5320040568) on #428.
- `DESIGN.md` is intentionally untouched — this branch carries the pre-#443 design note that **#443** rewrites wholesale; the Round 5 reconcile is deferred to after #443 lands.

Follow-up to #428.


client/src/omniSearch/omniSearchView.js
```diff
@@ -636,6 +686,22 @@
...
+        ev.key === 'Tab' &&
+        ev.target === inputEl &&
+        previewMode !== 'refs' &&
+        !inPivot &&
+        !ev.ctrlKey &&
+        !ev.altKey &&
+        !ev.metaKey
+      ) {
+        ev.preventDefault();
+        cycleScope(ev.shiftKey ? -1 : 1);
```
MatiasFernandez: 💭 This is the one I'd most like to resolve before merging. With `preventDefault()` on both `Tab` and `Shift+Tab` in the ordinary (non-refs, non-pivot) state — which is the state the panel is in almost all the time — a keyboard-only user in the search field has no `Tab` egress at all: they can't reach `#clear`, `#case`, the new `#refindicator`, `#pin`, or `#loadMore`/`#loadAll` any more. The pre-existing comment on the handler just above even documents the old behaviour ("rather than tabbing to the clear/case buttons"), so we're knowingly trading that away. Is there an alternative egress we could keep — e.g. only cycling when the field is non-empty, or reserving the gesture for `Ctrl+Tab`, or leaving one un-hijacked direction? JetBrains gets away with it because its popup has no in-field buttons to reach.

Separately, the comment says the `inPivot` case is "handled just above / by the pivot", but the handler above only covers `previewMode === 'refs'` — in the classic list pivot `Tab` isn't intercepted at all, it just falls through to native focus movement. Could we reword so the next reader doesn't go looking for handling that isn't there?

client/src/omniSearch/omniSearchView.js
```diff
@@ -131,9 +136,35 @@
...
+        var c = lastCategories[i];
+        (c.explicitOnly ? searches : ids).push(c.id);
+      }
+      return ids.concat(searches);
+    }
+
+    function cycleScope(dir) {
+      var ids = orderedScopeIds();
+      if (ids.length < 2) return;
+      var cur = ids.indexOf(currentScopeId);
```
MatiasFernandez: 💭 `currentScopeId` is only refreshed when the host echoes results back through `renderTabs`, but `cycleScope` posts and returns immediately. So two quick `Tab` presses (easy to do — it's a repeat-friendly key) both compute `cur` from the same stale value and both post `setScope` for the *same* target scope, so the second press looks like it did nothing. Should we advance `currentScopeId` optimistically right after posting (`currentScopeId = ids[next]`) and let the host's `renderTabs` correct it if it disagrees?

client/src/omniSearch/omniSearchView.js
```diff
@@ -577,13 +612,28 @@
...
+      // the inline style falls BACK to that rule, so a bare '' left the breadcrumb permanently hidden.
+      breadcrumbEl.style.display = title ? 'block' : 'none';
+    }
+
+    // Show the references-mode chip (styled like the case toggle) whenever the panel is displaying
+    // references/senders — the classic list pivot OR the sticky preview-pane list (#428 item 28) — so
+    // it is obvious you are in a references view. Hidden the rest of the time.
+    function refreshRefIndicator() {
+      if (!refIndicatorEl) return;
+      var on = inPivot || previewMode === 'refs';
```
MatiasFernandez: 💭 Every other place that sets `previewMode = 'source'` is paired with a `refreshRefIndicator()` — except the `input` handler around line 665 (`previewMode = 'source'; // typing dismisses a sticky references list`). So if you're in preview-refs mode and start typing, the chip claims you're still in a references view until the next `results` message lands. Worth adding the call there too, or is the results round-trip fast enough that it never shows?

client/src/omniSearch/omniSearchView.js
```diff
@@ -721,6 +787,21 @@
...
+    if (refIndicatorEl) {
+      // Clicking the references chip exits the references view: for the classic list pivot that means
+      // `back` (restore the prior search); for the sticky preview-pane list it means re-selecting the
+      // active row so its source preview replaces the refs list. Either way the chip then hides.
+      refIndicatorEl.addEventListener('click', function () {
+        if (inPivot) {
+          scrollResetPending = true;
+          post('back');
+        } else if (previewMode === 'refs') {
+          setActive(activeIndex);
```
MatiasFernandez: 💭 `setActive` flips `previewMode` to `'source'` and hides the chip synchronously, but the actual source preview only arrives after `requestPreview`'s debounce plus a host round-trip. So for that window the refs list is still on screen with no chip explaining it — which is exactly the "it isn't obvious when you're looking at references" problem #428 #28 is about, just inverted. Should we `clearPreview()` (or show a placeholder) so the pane and the chip change state together?

client/src/omniSearch/omniSearchShared.ts
```diff
@@ -202,7 +202,11 @@ export function renderOmniHtml(opts: { showPin: boolean }): string {
       border-color: var(--vscode-button-background);
     }
     .tab.explicit { font-style: italic; border-style: dashed; }
-    .tab.explicit::before { content: '\\1F50D\\00A0'; font-style: normal; font-size: 0.9em; }
+    /* Heavy/slow scopes (Source/Literals/Categories) each run a full-image scan, so they carry a
+       distinct hourglass marker rather than the plain magnifier the whole field otherwise reads as
+       (#428 item 27). No backticks in this comment -- the stylesheet is a template literal.
+       Swap 231B (hourglass) for 1F422 (turtle) if the hourglass reads as "loading" instead of "slow". */
```
MatiasFernandez: nit: `Swap 231B (hourglass) for 1F422 (turtle) if the hourglass reads as "loading" instead of "slow"` reads like an authoring note to yourself, and it ships in the webview's stylesheet. Since you've already F5-verified the hourglass, can we drop that line and keep the first two (which explain the *why*)?

client/src/omniSearch/omniSearchShared.ts
```diff
@@ -414,6 +435,7 @@ export function renderOmniHtml(opts: { showPin: boolean }): string {
         <button id="clear" title="Clear search" aria-label="Clear search" style="display:none">×</button>
       </div>
       <button id="case" title="Case sensitivity" aria-pressed="false">Aa</button>
+      <button id="refindicator" title="Showing references — click to exit" aria-pressed="false" style="display:none">↗ References</button>
```
MatiasFernandez: 💭 `aria-pressed` here is `true` whenever the chip is visible and `false` only while it's `display:none`, so it never actually conveys anything to a screen reader — unlike `#case`, which is always on screen and genuinely toggles. A screen-reader user will hear "References, toggle button, pressed" with no way to observe the unpressed state. Would a plain button (no `aria-pressed`) with the label carrying the meaning be more honest here — e.g. `aria-label="Showing references — activate to exit"`? The two new tests assert the `aria-pressed` values, so they'd need to follow.

client/src/omniSearch/__tests__/omniScopeCycle.test.ts
```diff
@@ -0,0 +1,132 @@
...
+  it('does not cycle when a modifier other than Shift is held', () => {
+    const { handle, vscode, input } = mount();
+    handle.renderTabs(CATS, null);
+    pressTab(input, { ctrlKey: true });
+    pressTab(input, { metaKey: true });
+    pressTab(input, { altKey: true });
+    expect(lastSetScope(vscode)).toBeUndefined();
+  });
+
+  it('is a no-op when there is only the All scope', () => {
```
MatiasFernandez: 💭 The guard clause carries three conditions we deliberately added — `previewMode !== 'refs'`, `!inPivot`, and the modifier check — but only the modifier one is covered here. The other two are the interesting ones, since they're what keeps `Tab` working as "dive into the refs list" and as native focus movement in the pivot. Should we add two cases: `Tab` while a `refPreview` is showing, and `Tab` while `pivot: true`, both asserting no `setScope` is posted? Both states are already easy to reach from the helpers in `omniReferencesAffordance.test.ts`.

client/src/omniSearch/__tests__/omniReferencesAffordance.test.ts
```diff
@@ -0,0 +1,210 @@
...
+        shownCount: 1,
+        pivot: true,
+        pivotTitle: 'References to Foo',
+      }),
+    });
+    chip.click();
+    expect(vscode.postMessage).toHaveBeenCalledWith(expect.objectContaining({ command: 'back' }));
+  });
+
+  it('clicking the chip in preview-refs mode dismisses the refs list (chip hides)', () => {
```
MatiasFernandez: 💭 The name promises the refs list is dismissed, but the only assertion is that the chip hid — the test would still pass if the refs rows stayed on screen forever. Could we also assert the observable outcome, e.g. that a `preview` message was posted for the active row (or that `#preview` no longer holds the ref rows)? That's the behaviour a user would notice, and it's the part the chip's contract actually rests on.

MatiasFernandez (approved):

client/src/omniSearch/omniSearchShared.ts
ericwinger: Done in 01c453e3. Dropped the `Swap 231B (hourglass) for 1F422 (turtle)…` line and kept the two that explain why the marker exists:

```
/* Heavy/slow scopes (Source/Literals/Categories) each run a full-image scan, so they carry a
   distinct hourglass marker rather than the plain magnifier the whole field otherwise reads as.
   No backticks in this comment -- the stylesheet is a template literal. */
```

While there I also stripped a bare internal item number from that same comment so nothing but the rationale ships in the stylesheet.

client/src/omniSearch/__tests__/omniScopeCycle.test.ts
ericwinger: Added both cases in 01c453e3:

- **Tab while a references list fills the preview pane** — asserts no `setScope`. It presses plain **Tab** (which dives into the list) *and* **Shift+Tab**; the Shift+Tab press is the one that actually exercises the guard's `previewMode !== 'refs'` clause, since the earlier "dive into the list" handler only intercepts plain Tab. I confirmed this by mutation: deleting that clause fails the test only because of the Shift+Tab assertion.
- **Tab while `pivot: true`** — asserts no `setScope`; deleting `!inPivot` fails it.

Both states are reached through `onMessage` (a `results` message renders the tabs so cycling is genuinely possible, then a `refPreview` / `pivot` message enters the guarded state) — I found that `renderResults` alone doesn't populate `lastCategories`, so `cycleScope` would no-op regardless and the test would pass vacuously. Going through `onMessage` avoids that trap.

client/src/omniSearch/__tests__/omniReferencesAffordance.test.ts
ericwinger: Strengthened in 01c453e3. The `refPreview` now carries a real row, the test asserts that row is actually on screen (`#preview` contains `Baz>>qux`), and after clicking the chip it asserts the observable outcome you described — the active row's source `preview` is re-requested:

```ts
expect(preview.textContent).toContain('Baz>>qux'); // the refs list really is on screen
vscode.postMessage.mockClear();
chip.click();
expect(chip.style.display).toBe('none');
vi.advanceTimersByTime(300); // flush the preview debounce
expect(vscode.postMessage).toHaveBeenCalledWith(
  expect.objectContaining({ command: 'preview', id: 0 }),
);
```

Renamed it to "…dismisses the refs list and restores the source preview". Mutation-checked: a variant that hides the chip but skips the source re-request now fails (it passed under the old chip-only assertion).

PR #451, opened 2026-08-17
Move senders/implementors counts into the selector hover (#432)

### LINKED ISSUE #432: Senders/implementors counts shift the method source when it opens — give them a home outside the document flow
## Summary

Opening a method in the source editor makes the text jump up and down while the senders and
implementors counts are being worked out. The counts themselves are genuinely useful and should
stay — the problem is only where they live. Today they are CodeLens links above the method
(`client/src/gemstoneCodeLensProvider.ts`), so as each lens resolves and appears, the method body
shifts down; the effect reads as a jiggle right when you are trying to start reading the code.

Previous attempts to fix this in place have not worked. Rather than try again to smooth the
CodeLens behaviour, this asks for a different home for the counts: somewhere clearly visible, but
somewhere that cannot move the method source or change which line of the editor a given line of
code sits on.

## Acceptance criteria

- Opening a method never shifts the source text — the first line of the method stays where it was
  from the moment the editor opens, whether the counts are still loading, have arrived, or failed.
- Senders and implementors are still obvious without hunting for them, and still one click each to
  navigate.
- Both counts still resolve lazily (they are costly for a popular selector) and the existing
  caching keeps working.

## Implementation notes

Some possible directions, no preference settled:

- Move the counts out of the document flow entirely — a status-bar item, a view-title
  action, or a small entry in the editor's title area, updated as they resolve.
- Keep them as CodeLens but reserve the space up front: render both lenses immediately with a
  placeholder title, then fill in the numbers in place, so nothing is inserted after layout.
- Put them somewhere adjacent instead — the Explorer's method row, a hover, or a panel — where
  appearing late costs nothing visually.

Worth checking whether the debugger's inline-values CodeLens provider
(`client/src/inlineValuesCodeLens.ts`) contributes to the movement as well, since VS Code
re-resolves every provider's lenses on a document when any one of them fires a change.

ericwinger:
## What & why

Opening a method jiggled the source: the senders/implementors counts lived as CodeLens links *above* the method, so as each one resolved and appeared, the method body shifted down — right when you're trying to start reading. A CodeLens inherently occupies a line that only exists once the provider returns (after first paint), so that one shift is unavoidable while the counts live in the document flow. Previous in-place attempts couldn't remove it.

This moves the counts **out of the document entirely**, into the **selector hover** — where appearing late costs nothing visually and the source can never move.

## Changes

- **`gemstoneHoverProvider.ts`** — the selector hover now leads with clickable **senders · implementors** links (trusted markdown + codicons) with friendly tooltips (`Browse senders of #sel`), each firing the same command the old CodeLens did. The implementor preview list stays below. The senders count is cached per `selector|session|env` (`sendersOf` blocks the host and a hover fires on every mouse-rest). The hover now also appears for a selector that is *sent but not implemented*.
- **`gemstoneCodeLensProvider.ts`** — the `gemstone://` method editor no longer emits a senders/implementors CodeLens (topaz/file docs keep theirs), so the source never shifts on open.
- Tests updated/added (senders link, senders-only hover); the test `__mocks__` `MarkdownString` learns the `isTrusted` / `supportThemeIcons` flags the hover sets.

## Acceptance criteria (from #432)

- ✅ Opening a method never shifts the source text — the counts are no longer in the document flow.
- ✅ Senders and implementors are still obvious and one click each to navigate (via the hover links).
- ✅ Both counts still resolve lazily and stay cached.

## Testing

Client-only TypeScript change (no server / GemStone / RB surface). Full client gate green: `compile` / `lint` / `format:check`, and the client unit suite (**5637 passed, 111 skipped, 0 failed**). Manually exercised by @ericwinger during development (hover links + tooltips verified).

## Not in scope

A separate, pre-existing breadcrumb behavior — VS Code rebuilding the breadcrumb's symbol segment when a method opens from the bottom-panel Omni webview — can flash on a selector change. It's independent of this change (reproduces on `main`), is driven by the panel origin rather than anything the extension controls, and was confirmed unfixable from the extension via open-path/column tweaks. Left as-is here.

Closes #432

### DIFF
--- a/client/src/__mocks__/vscode.ts
+++ b/client/src/__mocks__/vscode.ts
@@ -714,6 +714,10 @@ export const InputBoxValidationSeverity = {
 // ── MarkdownString mock ──────────────────────────────────
 
 export class MarkdownString {
+  // Mirror the real MarkdownString flags the extension sets: `isTrusted` enables
+  // command: links, `supportThemeIcons` enables $(codicon) rendering.
+  isTrusted?: boolean;
+  supportThemeIcons?: boolean;
   constructor(public value: string = '') {}
   appendMarkdown(value: string): MarkdownString {
     this.value += value;
--- a/client/src/__tests__/gemstoneCodeLensProvider.test.ts
+++ b/client/src/__tests__/gemstoneCodeLensProvider.test.ts
@@ -90,10 +90,14 @@ name: aString
       expect(lenses).toHaveLength(4); // 2 methods × 2 lenses
     });
 
-    it('returns a senders+implementors lens pair for gemstone:// method URIs', () => {
+    it('returns no lenses for gemstone:// method URIs (#432: the selector hover owns those)', () => {
+      // The editable gemstone:// method editor no longer carries a
+      // senders/implementors CodeLens (it jiggled the source on open). Those counts
+      // now live in the selector hover (GemStoneHoverProvider) as clickable links,
+      // so the CodeLens provider emits nothing there.
       const doc = createMockDocument('name\n  ^ name', 'gemstone');
       const lenses = provider.provideCodeLenses(doc);
-      expect(lenses).toHaveLength(2); // 1 method × 2 lenses
+      expect(lenses).toHaveLength(0);
     });
 
     it('returns no lenses for an unsaved new-method template', () => {
--- a/client/src/__tests__/gemstoneHoverProvider.test.ts
+++ b/client/src/__tests__/gemstoneHoverProvider.test.ts
@@ -4,6 +4,7 @@ vi.mock('vscode', () => import('../__mocks__/vscode.js'));
 
 vi.mock('../browserQueries', () => ({
   implementorsOf: vi.fn(() => []),
+  sendersOf: vi.fn(() => []),
   getAllClassNames: vi.fn(() => []),
   getClassComment: vi.fn(() => ''),
 }));
@@ -13,9 +14,10 @@ import type * as vscode from 'vscode';
 import { GemStoneHoverProvider } from '../gemstoneHoverProvider';
 import { SelectorResolver } from '../gemstoneDefinitionProvider';
 import { SessionManager } from '../sessionManager';
-import { implementorsOf, getAllClassNames, getClassComment } from '../browserQueries';
+import { implementorsOf, sendersOf, getAllClassNames, getClassComment } from '../browserQueries';
 
 const mockImplementorsOf = vi.mocked(implementorsOf);
+const mockSendersOf = vi.mocked(sendersOf);
 const mockGetAllClassNames = vi.mocked(getAllClassNames);
 const mockGetClassComment = vi.mocked(getClassComment);
 
@@ -56,9 +58,11 @@ describe('GemStoneHoverProvider', () => {
   beforeEach(() => {
     __resetConfig();
     mockImplementorsOf.mockReset();
+    mockSendersOf.mockReset();
     mockGetAllClassNames.mockReset();
     mockGetClassComment.mockReset();
     mockImplementorsOf.mockReturnValue([]);
+    mockSendersOf.mockReturnValue([]);
     mockGetAllClassNames.mockReturnValue([]);
     mockGetClassComment.mockReturnValue('');
   });
@@ -72,7 +76,7 @@ describe('GemStoneHoverProvider', () => {
   });
 
   describe('selector hover', () => {
-    it('shows implementors with categories', async () => {
+    it('shows implementors with categories, plus clickable senders/implementors links', async () => {
       mockImplementorsOf.mockReturnValue([
         {
           dictName: 'Globals',
@@ -89,6 +93,7 @@ describe('GemStoneHoverProvider', () => {
           category: 'accessing',
         },
       ]);
+      mockSendersOf.mockReturnValue(Array.from({ length: 7 }, () => ({}) as never));
       const resolver: SelectorResolver = { getSelector: vi.fn(async () => 'size') };
       const provider = new GemStoneHoverProvider(makeSessionManager(true), resolver);
       const result = await provider.provideHover(makeDocument('self size'), pos(0, 5));
@@ -96,9 +101,25 @@ describe('GemStoneHoverProvider', () => {
       expect(result).not.toBeNull();
       const md = result!.contents as unknown as MarkdownString;
       expect(md.value).toContain('**#size**');
-      expect(md.value).toContain('*2* implementors');
+      // #432: counts are clickable command links, not plain text.
+      expect(md.value).toContain('7 senders](command:gemstone.sendersOfSelector?');
+      expect(md.value).toContain('2 implementors](command:gemstone.implementorsOfSelector?');
+      expect(md.isTrusted).toBe(true); // command: links only fire from a trusted hover
       expect(md.value).toContain('`Array` (accessing)');
       expect(md.value).toContain('`String` (accessing)');
+    });
+
+    it('shows a hover with just a senders link when the selector has no implementors', async () => {
+      mockImplementorsOf.mockReturnValue([]);
+      mockSendersOf.mockReturnValue(Array.from({ length: 3 }, () => ({}) as never));
+      const resolver: SelectorResolver = { getSelector: vi.fn(async () => 'onlySent') };
+      const provider = new GemStoneHoverProvider(makeSessionManager(true), resolver);
+      const result = await provider.provideHover(makeDocument('self onlySent'), pos(0, 5));
+
+      expect(result).not.toBeNull();
+      const md = result!.contents as unknown as MarkdownString;
+      expect(md.value).toContain('3 senders](command:gemstone.sendersOfSelector?');
+      expect(md.value).toContain('0 implementors](command:gemstone.implementorsOfSelector?');
     });
 
     it('shows singular "implementor" for one result', async () => {
@@ -116,7 +137,7 @@ describe('GemStoneHoverProvider', () => {
       const result = await provider.provideHover(makeDocument('self size'), pos(0, 5));
 
       const md = result!.contents as unknown as MarkdownString;
-      expect(md.value).toContain('*1* implementor\n');
+      expect(md.value).toContain('1 implementor](command:gemstone.implementorsOfSelector?');
     });
 
     it('shows "class" suffix for class-side implementors', async () => {
@@ -151,7 +172,7 @@ describe('GemStoneHoverProvider', () => {
       const result = await provider.provideHover(makeDocument('self size'), pos(0, 5));
 
       const md = result!.contents as unknown as MarkdownString;
-      expect(md.value).toContain('*15* implementors');
+      expect(md.value).toContain('15 implementors](command:gemstone.implementorsOfSelector?');
       expect(md.value).toContain('Class9');
       expect(md.value).not.toContain('Class10');
       expect(md.value).toContain('...and 5 more');
--- a/client/src/gemstoneCodeLensProvider.ts
+++ b/client/src/gemstoneCodeLensProvider.ts
@@ -2,7 +2,6 @@ import * as vscode from 'vscode';
 import { SessionManager, ActiveSession } from './sessionManager';
 import { parseTopazDocument } from './topazFileIn';
 import { extractSelector } from './systemBrowser';
-import { parseMethodUri } from './gemstoneFileSystemProvider';
 import * as queries from './browserQueries';
 
 interface CodeLensData {
@@ -90,12 +89,13 @@ export class GemStoneCodeLensProvider implements vscode.CodeLensProvider, vscode
     return lenses;
   }
 
-  private provideGemstoneCodeLenses(document: vscode.TextDocument): vscode.CodeLens[] {
-    const method = parseMethodUri(document.uri);
-    if (!method) return [];
-
-    const range = new vscode.Range(new vscode.Position(0, 0), new vscode.Position(0, 0));
-    return this.makeMethodLenses(range, method.selector, method.className, method.isMeta);
+  private provideGemstoneCodeLenses(_document: vscode.TextDocument): vscode.CodeLens[] {
+    // #432: the editable gemstone:// method editor no longer carries a
+    // senders/implementors CodeLens (it appeared after first paint and shoved the
+    // source down — the "jiggle"). Those counts now live in the selector hover
+    // (GemStoneHoverProvider) as clickable links, which is out of the document flow
+    // and can't move the source. The topaz/file paths above keep their lenses.
+    return [];
   }
 
   // Emit the senders + implementors pair on the same range, in that order.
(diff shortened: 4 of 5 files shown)


ericwinger:
Screen shot of new senders/implementors in hover help

image

npapagna (approved):
Thanks @ericwinger.
Gave this a try, it worked fine for me.
Left a couple of issues (that don't seem strictly related to this PR)

npapagna:
## Omni Search code review — 10 findings

⚠️ **Note:** these findings are not strictly related to this PR. They come from a `/code-review` pass over the full Omni Search feature (originally merged via #423), and none of the files below are touched by this PR's actual diff (which only changes `gemstoneCodeLensProvider.ts`/`gemstoneHoverProvider.ts`). Recording them here as a plain comment per request, rather than as inline line comments, since GitHub won't attach those to paths outside a PR's diff.

---

### 1. Stale references can overwrite fresh ones
**`client/src/omniSearch/omniEngine.ts:414`** (`referencesFor`)

No generation/ordering guard between the `await deps.resolveReferences(result)` and the write to the shared `referenceRows` array. Unlike `pivot()`, it never checks whether a newer call has superseded it.

*Failure:* Request references for row A, then quickly for row B. If B's (faster) query resolves first and A's slower one lands after, A's stale rows silently overwrite B's while the UI still shows B — clicking a visible B reference then opens the wrong method via `referenceResultFor`.

*Fix:* same generation-token pattern `runSearch` already uses: capture a token before the `await`, skip the write if superseded.

---

### 2. `pivot()`'s generation guard doesn't cover its own resolution
**`client/src/omniSearch/omniEngine.ts:398`** (`pivot`)

Bumps `generation` to invalidate in-flight work, but never captures its own token or re-checks it after `await deps.resolveReferences(result)` resolves.

*Failure:* Pivot to "senders of #foo" (slow, many hits), then type a new search term before it resolves. The new search legitimately updates the view; when the stale pivot call finally resolves, it unconditionally overwrites `pivot`/`current`, snapping the UI back to the discarded pivot.

*Fix:* same as `runSearch` — capture a token before the `await`, bail if invalidated.

---

### 3. `OmniSearchPanel.show()` discards new `deps` when a panel already exists
**`client/src/omniSearch/omniSearchPanel.ts:71`**

```ts
static show(deps: OmniPanelDeps): void {
  if (OmniSearchPanel.current) {
    OmniSearchPanel.current.panel.reveal(...);
    ...
    return; // deps is discarded
  }
  ...
}
```

*Failure:* Two-session window. Open Spotter while session A is active (bound to A's deps). Switch active session to B, reinvoke `gemstone.omniSearch`. The freshly-built deps for B are thrown away; the panel keeps searching/opening against session A with no indication anything's wrong.

*Fix:* on the "already exists" branch, reassign `OmniSearchPanel.current.deps = deps` (and anything derived from it) before revealing.

---

### 4. Reveal handlers drop `sessionId`, resolve against whatever session is currently selected
**`client/src/omniSearch/omniSearchCommand.ts:104-122`** (`revealDictionary`, `revealGlobal`, `revealCategory`)

Unlike `openClass`/`openMethod` in the same file, these three don't pass `a.sessionId` through to the Explorer commands — and the Explorer-side commands (`gemstoneExplorer.ts:5198-5212`) don't even accept a session parameter; they operate on whatever session the Explorer controller currently has selected.

*Failure:* Two-session window. Search while A is active, get a Dictionaries/Globals/Categories hit from A, switch active session to B, click the result — it resolves against B instead of A. Best case "not found," worst case a same-named dictionary/category in B and you land in the wrong session's data silently.

*Fix:* thread `a.sessionId` through these three commands and the Explorer's `revealDictionaryByName`/`revealCategoryByPath`/`findClass`, same as `openClass`/`openMethod` already do. Larger fix since it changes the Explorer's command signatures.

---

### 5. References pivot only ever queries environment 0
**`client/src/omniSearch/references.ts:24,36,46`** (`referenceRequestFor`) and **`references.ts:76`** (`methodRowsToResults`, hardcodes `environmentId: 0`)

The pre-existing `gemstone.sendersOfSelector`/`implementorsOfSelector` commands (`extension.ts:2156-2190`) loop every environment up to `gemstone.maxEnvironment`. Omni Search's references pivot never loops — every result's `environmentId` is hardcoded to 0 at creation, so `resolveReferencesUsing` (`omniSearchCommand.ts:128-139`) only ever searches environment 0.

*Failure:* On a stone with `maxEnvironment > 0`, pivoting to "Senders of #foo" from an Omni Search result misses senders in environment ≥ 1. The hover-based senders count (which does loop) reports a different, larger number for the same selector/session.

*Fix:* loop `req.environmentId` from 0 to `maxEnvironment` inside `resolveReferencesUsing`, matching the existing `extension.ts` pattern (dedup by class/selector).

---

### 6. `revealCategoryByPath` mutates Explorer state before checking the target exists
**`client/src/gemstoneExplorer.ts:3526-3549`**

`selectClassCategory(catItem)` — which clears the selected class/hierarchy, clears pane filters, and refreshes three tree providers — runs *before* the code checks `categoryPath` actually exists via `allCategoryPaths()`.

*Failure:* `gemstone.explorer.revealCategory` fires with a valid `dictName` but a stale/renamed `categoryPath` (e.g. Omni Search result predates a refactor). The Explorer gets mutated/filtered to nothing, *then* the check fails and shows a warning — leaving a warning toast plus a confusingly blank Explorer.

*Fix:* move the `allCategoryPaths()` existence check before calling `selectClassCategory` — it only needs `dictName`/`categoryPath`, not the constructed item.

> **Note:** open PR #445 ("Explorer reveal robustness") appears to already reorder this check ahead of `selectClassCategory` — worth confirming it's a complete fix before treating this as still open.

---

### 7. `literalsProvider.search()` swallows all errors, including real GCI failures
**`client/src/omniSearch/providers/literalsProvider.ts:61-74`**

The bare `catch { return []; }` is meant to cover "not yet a compilable expression" while typing a partial `#symbol`, but it also swallows genuine backend failures (dropped connection, timeout, torn-down session). Every sibling provider has no local catch at all — exceptions propagate to the controller's single `onError` chokepoint (`omniSearchController.ts:354` etc.).

*Failure:* A real GCI failure during a Literals search shows "no results" with zero indication anything went wrong, unlike every other search category which would surface the standard error toast for the same underlying failure.

*Fix:* narrow the catch to the case it's meant for (a compile/parse failure from `runSymbol` on a not-yet-valid expression), let anything else propagate.

> **Note:** open PR #445 ("Literals error surfacing") already adds an `onError` sink and restructures the bailouts vs. the try/catch — looks like it addresses this directly; worth confirming before treating this as still open.

---

### 8. Case-folding disables the camelCase match preference in default mode
**`client/src/omniSearch/omniMatch.ts:158-176`** (`match`) / **`115-126`** (`contiguousIndices`) / **`50-58`** (`isWordStart`)

`match()` lowercases `target` into `t` before calling `contiguousIndices(q, t)`, which relies on `isWordStart`'s camelCase-hump check (`/[A-Z]/.test(cur)`) — impossible to trigger once everything is lowercase. `scoreFrom` is later called with the original-case `target`, but by then `contiguousIndices` has already picked *which* occurrence to use, blind to case.

*Failure:* Default settings (`caseSensitive: false`). Target `"xcoreFooCore"`, query `"co"`: an early non-word-start "co" (inside "xcore") is chosen over the later, semantically preferred camelCase-boundary "Co" (in "FooCore"), degrading ranking for the majority (default) case-insensitive mode.

*Fix:* have the word-start check in `contiguousIndices` consult the original-case `target` even while matching the folded `q`/`t`.

---

### 9. Webview search UIs (Spotter panel, docked panel) have no debounce
**`client/src/omniSearch/omniSearchView.js:612-618`** (input handler) / **`omniSearchShared.ts:131-132`** (`dispatchEngineMessage`)

`inputEl.addEventListener('input', ...)` posts a `'query'` message on every keystroke straight into `engine.search()` — no debounce. The QuickPick controller (`omniSearchController.ts:224,365-370`) explicitly debounces via the user-configurable `gemstone.omniSearch.debounceMs` (default 120ms), but only the QuickPick path uses it.

*Failure:* Typing a 10-character term in the Spotter/docked panel fires 10 full provider fan-outs (one GCI round-trip per provider per keystroke) instead of ~1-2 after coalescing — visible lag, and more overlapping in-flight calls amplifying findings 1 and 2.

*Fix:* apply the same `config.debounceMs` timer around the `post('query', ...)` call (or the `case 'query'` dispatch), so all three UI surfaces share one debounce behavior.

---

### 10. Docked panel loses input focus on cold webview load
**`client/src/omniSearch/omniSearchViewProvider.ts:73-85,125-130`**

`resolveWebviewView` sets `this.view` as soon as VS Code creates the view — before the webview's page script has loaded and attached its message listener. `focus()` calls `deliverFocus()` right after, which sees `this.view` truthy, immediately clears `focusPending`, and posts `focusInput` into a page with no listener yet (lost). The `'ready'` handler's replay (`deliverFocus()` again) is meant to catch this race, but by the time `'ready'` fires, `focusPending` is already `false`, so the replay no-ops.

*Failure:* First-time (cold) open of the docked Omni Search panel in a session — the panel opens but the search input never gets keyboard focus, defeating the "open and start typing" shortcut. Subsequent opens (warm webview) work fine.

*Fix:* track webview readiness separately from `this.view`'s existence (e.g. a `ready` flag flipped by the `'ready'` message); have `deliverFocus()` only actually consume `focusPending`/post the message once ready, otherwise leave it for the `'ready'` handler to pick up.

> Note: open PR #444 changes `focus()`'s return type but doesn't appear to address this specific race — worth double-checking.

ericwinger:
Thanks @npapagna — much appreciated, and thanks for flagging that these are outside this PR's diff.

I checked all ten against current `main` (`15860c9c`). Your notes on **6** and **7** were right: both are already fixed, so nothing to do there.

**6 — `revealCategoryByPath` mutates before checking the target exists.** Fixed by #445. On `main` the existence check now runs *before* `selectClassCategory`, at `client/src/gemstoneExplorer.ts:3653`:

```ts
this.selectDict(new DictItem(dictName, idx + 1));

// Check the category actually exists in this dictionary's loaded forest BEFORE mutating the
// category/classes panes. ...
if (!this.allCategoryPaths().some((p) => p === categoryPath || p.startsWith(`${categoryPath}-`))) {
  void vscode.window.showWarningMessage(`Category "${categoryPath}" was not found in ${dictName}.`);
  return;
}

const segment = categoryPath.split('-').pop() ?? categoryPath;
const catItem = new ClassCategoryItem(segment, categoryPath, false);
this.selectClassCategory(catItem);
```

One deliberate difference from your suggested fix: the *dictionary* selection above the check is meant to stick even when the category is missing — landing on the home dictionary is still useful. It's only the category/classes panes that are now left untouched.

**7 — `literalsProvider.search()` swallows real GCI failures.** Also fixed by #445 (`client/src/omniSearch/providers/literalsProvider.ts:61-93`), along the lines you describe: the "still typing / not a literal" cases are now pure bailouts *before* the `try`, so anything reaching the runner is a complete, well-formed literal and a throw from it can only be a real backend failure. That failure goes to an `onError` sink rather than being silently dropped:

```ts
let rows: MethodSearchResult[];
try {
  rows = run();
} catch (e) {
  // A real failure from the server-side runner (GCI drop, aborted transaction, ...). Surface it
  // so it's diagnosable instead of masquerading as "no results". Still return [] — the sync
  // provider API has no error row, and throwing would abort sibling providers in the All scope.
  onError?.(`Literals search failed for ${term}: ${e instanceof Error ? e.message : String(e)}`);
  return [];
}
```

The remaining eight I confirmed as still live on `main`, and I'm filing them as their own issues rather than folding them into this PR (its diff is only the hover/CodeLens pair, and it's already approved). I'll post the issue links back here once they're up. **10** looks like a third race in the same story as #457 — that issue already names `OmniSearchViewProvider.focus()` and provider-resolution latency — so I'll add it there rather than open a separate one, and **3**/**4** will be cross-linked to #437.

ericwinger:
@npapagna — all ten findings from your review are now resolved or tracked. Thanks again; several of these
were real and none of them were things the #451 diff would have surfaced.

Everything was re-verified against `main` first, because your review ran against this PR's branch rather
than `main` — that branch was 64 commits behind, which matters for a couple of the findings (noted below).

| # | Finding | Outcome |
|---|---|---|
| 1 | Stale references can overwrite fresh ones | Fixed — #460, PR #465 (`963d806e`) |
| 2 | `pivot()`'s guard doesn't cover its own resolution | Fixed — #460, PR #465 (`963d806e`) |
| 3 | `OmniSearchPanel.show()` discards new `deps` | Fixed — #459, PR #464 |
| 4 | Reveal handlers drop `sessionId` | Fixed — #459, PR #464 |
| 5 | References pivot only queries environment 0 | Fixed — #461, PR #464 |
| 6 | `revealCategoryByPath` mutates before the existence check | Already fixed by #445 — [details above](https://github.com/GemTalk/Jasper/pull/451#issuecomment-5345940048) |
| 7 | `literalsProvider.search()` swallows real GCI failures | Already fixed by #445 — [details above](https://github.com/GemTalk/Jasper/pull/451#issuecomment-5345940048) |
| 8 | Case-folding disables the camelCase match preference | Fixed — #462, PR #465 (`94bd5f56`) |
| 9 | Webview search UIs have no debounce | Fixed — #463, PR #465 (`5e45ecf3`) |
| 10 | Docked panel loses input focus on cold webview load | Tracked in #457 — [added there](https://github.com/GemTalk/Jasper/issues/457#issuecomment-5345989300) as an acceptance criterion plus implementation notes |

Two PRs, split by which files they touch so they share none and can merge in either order:
**#464** (findings 3, 4, 5) and **#465** (findings 1, 2, 8, 9).

Three places where the fix or the framing came out differently from the report — worth flagging since you
may look at the PRs:

**Finding 1 — `referencesFor` needed its own counter, not the shared `generation`.** Your suggestion was
the `runSearch` token pattern, which is right in shape, but reusing `generation` doesn't work either way:
capturing it read-only doesn't fix the bug, because two successive reference loads capture the same value
and neither invalidates the other — which is exactly the reported race. Bumping it does fix the ordering
but breaks `referencesFor`'s contract of leaving the search untouched: an in-flight search would be
cancelled by a preview load and the result list would sit stale until the next keystroke. So there's now a
separate `referenceGeneration`, with a test that fails if anyone collapses the two back together.

**Finding 9 is worse on `main` than on the branch you read.** You saw the Quick Pick debouncing while the
newer surfaces didn't, which reads as an inconsistency fixable by copying the working path. On `main` the
Quick Pick is gone (#443), so nothing debounces anywhere, `gemstone.omniSearch.debounceMs` has no consumer
at all, and there's nothing to copy from — a user who feels the lag and turns the setting up gets nothing.
Wired it up rather than deleting the setting, because DESIGN.md asserts debounced behaviour in three
places, including a matcher benchmark whose conclusion is argued *from* a 120 ms debounce.

**Finding 8's effect is narrower than "degrades ranking", and that's why it's easy to miss.** `scoreFrom`
is called with the original-case target, so scoring always could see the humps; the damage is done earlier,
in `contiguousIndices` picking *which* occurrence to score. So it shows up as the wrong characters
highlighted and a row in the wrong place, not as an obviously wrong score. Fixed by judging boundaries on
the original case while still matching on the folded string.

Also, for anything you review here in future: findings 7 and 9 cite `omniSearchController.ts`, which no
longer exists on `main` — it went with the Quick Pick removal. Worth pointing `/code-review` at `main`
rather than a PR branch when the target is the feature rather than the diff.

PR #454, opened 2026-08-18
GemStone Manager: one screen for the GemStone environment

srbaker:
Running GemStone means keeping four things in order: an operating system
configured to hand out shared memory, a version installed from it, a database
made from that version, and a login that reaches the database. Each has its own
sidebar view, which is fine once everything is up — and four separate lists to
walk when it isn't.

This adds **GemStone Admin: GemStone Manager**, an editor tab that puts all four
on one screen in the order they matter. A machine that cannot hand out enough
shared memory says so before anything else; a machine with nothing installed
leads with the versions; once neither is true the screen leads with connecting,
because that is what it is usually opened to do.

## Reading it

Eight commits. The first two stand on their own without the panel as their
excuse, and are worth taking first:

- **Put the 1 GB shared-memory rule in one place.** Three places rederived
  `shmmax / 2^30`, `shmall / 2^18` and `>= 1` from scratch — the Start Stone
  preflight, Quick Setup, and the OS view's own status node, which also formats
  the reading for display. One helper now answers it, and the status node loses
  its if/else entirely.
- **Answer what is installed without asking the network.** `fetchCatalog` does
  the scrape; `versionsFrom` turns a catalog plus the disk into rows;
  `fetchAvailableVersions` is the two together and is unchanged for its callers.
  `getInstalledVersions` becomes `versionsFrom([])`, which is what makes the
  offline list filtered and ordered like the online one — it had not been, so the
  panel's first paint showed releases Jasper cannot drive, which then vanished.

Then the panel itself, the two commits that complete it, the acceptance chapter,
a test-hardening pass, and the changelog note.

## What the screen offers

- **Operating System** — a checklist, one row per prerequisite this machine
  actually has (shared memory with how much is still free rather than just the
  limit, RemoveIPC, the WSL version, mirrored networking, both `gs64ldi` services
  entries, the hosts entry), each saying what the machine reports and carrying
  its own remedy only while it is failing.
- **Connect** — every login with its user and stone and whether that stone is up;
  log in, start-and-log-in, edit, delete, duplicate, select a session, plus the
  session actions including Export Classes and Full Logical Backup.
- **Versions** — download, extract, install, uninstall, register and unregister
  local builds, open the folder or a terminal, the Windows client trio, the
  walkthrough.
- **Databases** — per row: version, stone and NetLDI state, processes with PIDs
  and ports, logins, extent chooser, log/conf/backup files; create, delete,
  start/stop, replace extent, back up, restore, install server support, terminal,
  reveal, create and edit logins, clear a stale lock, copy host.

Everything the four sidebar views offered, the manager offers. The panel accepts
47 kinds of message, two of which fan out through an allow-list — 8 session
actions and 9 operating-system remedies — and each was checked against the
contributions of the view that used to own it.

## How it is built

It dispatches the existing `gemstone.*` commands rather than reimplementing them,
so confirmations, progress and tree refreshes come from the code that already
owns them. It does not trust the webview: session actions and OS remedies are
matched against allow-lists, and a file path is re-checked against its database
before anything is opened or restored from. It keeps itself current from the
admin views' own change events — coalesced, quiet, one scan at a time, and idle
while its tab is hidden.

## Two things worth calling out

**`@vscode/codicons` is a production dependency.** The panel draws every icon
with the codicon font, which a webview inherits nothing of, and `vsce` prunes
devDependencies — so a `.vscodeignore` whitelist over one would find nothing left
to package. Two files ship (`codicon.css`, `codicon.ttf`), the icons are
CC BY 4.0, and THIRD-PARTY.md carries the attribution. Both files and the
dependency are pinned by a guard, because losing either costs nothing at compile
time and nothing in the dev host: the panel renders with every glyph a blank box,
and only once packaged.

**The manager is palette-only for now.** An icon beside the GEMSTONE title needs
`viewContainer/title`, which is a proposed API the workbench only honours for
`workbench.view.debug`. A gear on the container becomes available once the
sidebar collapses toward a single view.

## Verified

`lint`, `format:check`, `compile`, `lint:supply-chain`, `lint:lockfile` clean.
`npm test` green — 5,421 client, 322 server, 92 mcp, against a live test stone.

A new acceptance chapter runs green headless in the container
(`npm run test:acceptance:docker`): the four sections on opening, the
operating-system checklist read from the machine's real `sysctl` and `ipcs`, and
logging in to a database from Connect. Writing it found three bugs the unit tests
could not — a section snapping shut on redraw, a live session offered "Start &
log in", and a stone this machine did not create always reading as stopped.

The tests were then audited by mutation rather than by reading: break a line of
production code, see whether anything fails. Fourteen changes broke nothing, the
worst being that neither allow-list had a test that the list is ever consulted —
the fixture held no sessions, so nothing reached that path, and deleting both
checks left the suite green. All fourteen now fail a test.

## Known gaps

Four `project-management` acceptance scenarios are red, and were red before this
branch — verified by stashing its work and re-running. Diagnosed separately: the
shared `Given I am logged in to a database` asserts on a sidebar that logging in
has replaced, since the Explorer is its own view container.

The Connect scenario reaches the database through "Start & log in" rather than
"Log in": the container's GemStone lives outside the folder Jasper scans, so it
finds no installed version to run `gslist` from and cannot see the stone at all.
That is an honest picture of a login to a stone made elsewhere, but it leaves the
plain "Log in" path without acceptance coverage.


client/src/gemstoneManager.ts
```diff
@@ -0,0 +1,1509 @@
...
+   * same string `buildDatabases` puts on the wire, so the panel never has to ship
+   * credentials to the webview. Delegates to `gemstone.login`, inheriting its
+   * keychain lookup, password prompt and session wiring.
+   */
+  /**
+   * Bring a stopped database up and then log in — the two steps a stopped login
+   * would otherwise need, in the order that works. The stone must answer before
+   * the login is attempted, so the start is awaited rather than fired alongside.
+   */
+  /** Run one of the allowed session commands against a live session. */
```
MatiasFernandez: 🚫 Three doc blocks have stacked up on this one method, and the two above it belong to methods further down: "Connect as a specific login..." is `connectLogin`'s, and "Bring a stopped database up and then log in..." is `startAndConnect`'s -- both of which now sit below with no doc of their own. Someone hovering `runSessionAction` is told it connects a login.

Same shape at line 759: `installNewVersion` carries `installVersion`'s "Installing a chosen release is a single action: fetch the archive, then extract it" ahead of its own block.

Looks like a reorder that left the comments behind -- can you move each block back onto the method it describes? It matters a bit more than usual here since `runSessionAction` is one of the two allow-list gates, so it's a method people will read before touching.

client/src/__tests__/gemstoneManagerPanel.test.ts
```diff
@@ -0,0 +1,644 @@
...
+    changeSetting('gemstone.rootPath');
+
+    await vi.waitFor(() => expect(postedStates(panel)).toHaveLength(1));
+  });
+
+  it('ignores a setting it does not read', async () => {
+    const { panel } = openPanel();
+
+    changeSetting('editor.fontSize');
+    await new Promise((resolve) => setTimeout(resolve, 0));
```
MatiasFernandez: 🚫 I think this one passes whether the code is right or wrong. `markStale` schedules the rebuild on the 200 ms coalescing timer, so at `setTimeout(..., 0)` nothing has been posted yet *even if* `editor.fontSize` had wrongly triggered a rebuild -- the assertion is satisfied by the timer not having fired, not by the setting being ignored.

`await settle()` (400 ms, as the neighbouring tests use) would make it fail if the filter regressed. Worth calling out given the mutation audit in the description: this is exactly the shape that survives one -- delete both `affectsConfiguration` checks and this test stays green.

client/src/sharedMemoryTreeProvider.ts
```diff
@@ -31,7 +31,8 @@ export type OsConfigNode =
 /** Does `/etc/services` inside the default WSL distro have a
  *  `gs64ldi <port>/tcp` entry? Routed through wslExecSync so the check
  *  inspects the Linux distro, not any Windows file of the same name. */
-function wslServicesHasGs64ldi(): boolean {
+/** Does the WSL distro's /etc/services carry the gs64ldi entry? */
```
MatiasFernandez: 💭 The original block comment is still sitting above this one, so the function now carries two docs -- and TS surfaces only the nearest, which means the tooltip loses "Routed through wslExecSync so the check inspects the Linux distro, not any Windows file of the same name". That was the part explaining why the function exists rather than just reading the file. Same at line 48 for `windowsServicesHasGs64ldi`.

Was the intent to replace them? If so, should the routing rationale fold into the new one-liner rather than being dropped?

Related, while you're in here: `isSharedMemoryConfigured` (line 149) lost its doc entirely in this pass. "Treats an unreadable sysctl as not configured" is still true and still not obvious from the body.

client/src/__tests__/gemstoneManagerPanel.test.ts
```diff
@@ -0,0 +1,644 @@
...
+
+/** Tell the panel its tab was shown or hidden, as VS Code would. */
+function changeVisibility(panel: MockPanel, visible: boolean): void {
+  panel.visible = visible;
+  const [handler] = panel.onDidChangeViewState.mock.calls[0] as [() => void];
+  handler();
+}
+
+/** Long enough for the panel's coalescing window to close and a rebuild to land. */
+function settle(): Promise<void> {
```
MatiasFernandez: 💭 `settle()` is a flat 400 ms sleep and it's awaited around 20 times in this file, so roughly eight seconds of the suite is spent waiting out a 200 ms timer -- and it's the kind of margin that gets thin on a loaded CI box.

Would `vi.useFakeTimers()` plus `await vi.advanceTimersByTimeAsync(COALESCE_MS + 1)` work here? It would make the "exactly one more pass" and "rebuilds once for a burst" tests deterministic rather than merely likely, which is where the real value is -- those are the ones asserting an exact count. The tests that just wait for something to arrive already use `vi.waitFor`.

client/src/gemstoneManagerView.js
```diff
@@ -0,0 +1,879 @@
...
+      (a.dir ? ` data-dir="${esc(a.dir)}"` : '') +
+      (a.folder ? ` data-folder="${esc(a.folder)}"` : '') +
+      (a.login ? ` data-login="${esc(a.login)}"` : '') +
+      (a.name ? ` data-name="${esc(a.name)}"` : '');
+    const title = ` title="${esc(a.title || label)}"`;
+
+    if (a.iconOnly) {
+      return `<button type="button" class="icon-btn" data-action="${action}"${data}${title} aria-label="${esc(label)}">${icon(iconKey)}</button>`;
+    }
+    const variant = cls === 'btn-primary' ? ' btn-primary' : '';
```
MatiasFernandez: 💭 `btn()` only honours `'btn-primary'` -- every other value of `cls` is silently discarded. `'btn-ghost'` is passed at 24 call sites and `'btn-secondary'` at one, and neither class appears in the CSS block in gemstoneManager.ts, so today they read as styling intent that never lands.

Should the argument just be dropped where it isn't `btn-primary`, or was `.btn-ghost` a rule meant to ship with the stylesheet? Either way it would stop the next reader assuming those buttons already render differently from the default secondary treatment.

client/src/gemstoneManagerView.js
```diff
@@ -0,0 +1,879 @@
...
+      action: el.dataset.cmd,
+    });
+  }
+
+  // Changing the extent select is destructive, so it opens the guarded replace
+  // flow through the host rather than swapping the file silently.
+  function onChange(e) {
+    const el = e.target.closest('[data-select]');
+    if (!el) return;
+    post({ command: el.dataset.select, dirName: el.dataset.dir });
```
MatiasFernandez: 💭 The extent the user picked never leaves the webview -- only `dirName` goes on the wire. `gemstone.replaceExtent` then opens its own QuickPick (databaseManager.ts:346) which pre-picks `currentExtent`, so choosing "extent1" here drops you into a second chooser sitting on "extent0".

I'm with you on not swapping silently -- replacing an extent destroys the database and its tranlogs, so routing through the guarded flow is right. But should the selection ride along so the confirmation starts on the extent they actually chose? As it stands the select is a control that forgets which option you picked, which is a bit odd right next to a destructive action.

client/src/gemstoneManager.ts
```diff
@@ -0,0 +1,1509 @@
...
+
+  private async buildVersions(source: VersionSource): Promise<VersionRow[]> {
+    let list: GemStoneVersion[];
+    if (source === 'local') {
+      list = this.deps.versionManager.getInstalledVersions();
+    } else {
+      try {
+        this.catalog ??= await this.deps.versionManager.fetchCatalog();
+        list = this.deps.versionManager.versionsFrom(this.catalog);
+      } catch {
```
MatiasFernandez: 💭 This catch reads every failure as "offline", but it also covers `versionsFrom`, which is pure disk work -- a scrape whose regex stopped matching, or a `readVersionTxt` throwing on a malformed install, comes out as a version list that quietly shrinks to what's on disk with nothing anywhere saying why.

`appendSysadmin` is already imported for `failed()`. Worth logging the error before falling back, so a "my versions disappeared" report has something to go on?

client/src/gemstoneManager.ts
```diff
@@ -0,0 +1,1509 @@
...
+    gbLabel: string,
+    inUse: number | undefined,
+  ): OsCheck[] {
+    // shmall is a ceiling for the machine, not for one stone: caches other stones
+    // already hold count against it. A database Jasper creates asks for a 100 MB
+    // cache (SHR_PAGE_CACHE_SIZE_KB), so less than that free means the next start
+    // fails however comfortably the limit itself clears 1 GB.
+    const shmallBytes = mem ? mem.shmall * 4096 : undefined;
+    const free = shmallBytes !== undefined && inUse !== undefined ? shmallBytes - inUse : undefined;
+    const roomForACache = free === undefined || free >= 100 * 1024 * 1024;
```
MatiasFernandez: 💭 This 100 MB is the cache size `databaseManager.ts:252` writes into every generated conf as `SHR_PAGE_CACHE_SIZE_KB = 100000`, but the two are independent literals -- and they don't quite agree, since 100000 KB is 102,400,000 bytes and this is 104,857,600.

Should the conf value become a shared constant both can read? The comment just above already names the coupling, which is what makes it worth pinning: if someone bumps the cache in databaseManager, this check keeps warning against the old figure and nothing fails.

client/src/gemstoneManager.ts
```diff
@@ -0,0 +1,1509 @@
+// GemStone Manager — a single, consolidated editor-tab webview that manages the
```
MatiasFernandez: 💭 I see that all the logic for the gemstoneManager is sitting at the root of the client workspace. We've been working with Eric on reorganizing the code a bit to keep the code for big features grouped together. Could you create a `gemstoneManager` folder in `client/src` to group the logic and the related tests there?

THIRD-PARTY.md
```diff
@@ -69,3 +69,35 @@ SOFTWARE.
 >   not the Refactoring Browser AST vendored here (a distinct Brant/Roberts
 >   lineage), so only the Pharo Project copyright above is reproduced for this
 >   component. See the linked `LICENSE` for the complete picture.
+
+---
+
+## Microsoft codicons (icon font)
```
MatiasFernandez: 💭 Why are we including this here if we are not vendoring codicons?

client/src/__tests__/vscodeignoreShippedAssets.test.ts
```diff
@@ -80,6 +80,40 @@ describe('runtime-injected webview assets are shipped in the .vsix', () => {
...
+
+  it.each([
+    'node_modules/@vscode/codicons/dist/codicon.css',
+    'node_modules/@vscode/codicons/dist/codicon.ttf',
+  ])('ships %s (exists on disk and is whitelisted in .vscodeignore)', (file) => {
+    expect(fs.existsSync(path.join(repoRoot, file))).toBe(true);
+    expect(whitelist.has(`!${file}`)).toBe(true);
+  });
+
+  // vsce prunes devDependencies before it packages, so a whitelist over one
```
MatiasFernandez: 💭 I don't understand the purpose of this test; parsing the `package.json` feels weird. And the comment mentions `devDependencies` but the package is imported as a normal dependency. What do we lose if we remove this?

.vscodeignore
```diff
@@ -58,6 +58,12 @@ mcp-server/node_modules/**
 !node_modules/koffi/index.d.ts
 !node_modules/koffi/LICENSE.txt
 !node_modules/koffi/build/**
+# The GemStone Manager webview draws every icon with the codicon font, which a
```
MatiasFernandez: 💭 Eric has been working on multiple webviews for different features, but this new one seems to be using a different styling/UI design. Are there specific limitations you found with the current established font/icons/styling, or is this just a preference? I'm worried about the visual consistency across the different webviews of Jasper.

MatiasFernandez (commented):
Left a few comments, I want to spend some time testing the new view. We discussed with Eric in the past that the sidebar would require some improvements but we haven't settle on a path forward, so I want to play around with this proposal to see how it feels

ericwinger:
> Left a few comments, I want to spend some time testing the new view. We discussed with Eric in the past that the sidebar would require some improvements but we haven't settle on a path forward, so I want to play around with this proposal to see how it feels

Yes, and I'd like to give it a try too.

THIRD-PARTY.md
ericwinger: The codicon font ships *inside* the .vsix — the `.ttf`/`.css` come from `node_modules/@vscode/codicons`, whitelisted by two `!` lines in `.vscodeignore`, and it's a production dependency so vsce doesn't prune it. We attribute it because those bytes are redistributed in the package, even though we don't check the source into the repo — the entry says "not vendored" to mark exactly that distinction. Happy to reword if "vendored" reads as the wrong header for a shipped-but-not-checked-in dependency.

client/src/__tests__/vscodeignoreShippedAssets.test.ts
ericwinger: It guards that the icon font actually ships. It's whitelisted in `.vscodeignore`, but vsce prunes devDependencies before packaging — so if `@vscode/codicons` were ever a devDependency, the whitelist would point at files that got pruned and the icons would silently vanish from the released extension. Parsing package.json is how it checks the package stays a regular dependency. Remove it and nothing catches a future move to devDependencies. I'll make the comment say that more clearly.

client/src/gemstoneManager.ts
ericwinger: Agreed — grouping the Manager's code and tests into a `client/src/gemstoneManager/` folder is tracked in #482, alongside extracting the inline `activate()` handlers into their own tested functions (the two are the same reorg). Keeping it out of this PR so the move doesn't tangle with the feature diff.

ericwinger:
**Update — integration work + review fixes now on this branch**

Since the last review, this branch has grown well beyond the original Manager panel. Summary of what's new, for re-review:

**Features**
- **One activity-bar icon.** The Explorer panes, Rowan, Inspector, and MCP Server view now share a single *GemStone Explorer* container; the administrative views moved out of the sidebar into the Manager.
- **Guided setup for a fresh machine** — a next-step header, a "Show me how" coach-mark tour, Quick Setup surfaced up front, and two zero-question commands (Create Database with defaults; Create Login).
- **Configuration viewer/editor — this completes and closes #232.** Once a session is selected, the Manager shows that session's stone and gem configuration: read-only config-file parameters vs runtime ones, inline editing with the stone as the authority on what a change does, a filter, and tooltips drawn from the product's `system.conf`.
- **The webview now scales with the workbench font size** instead of fixed pixels; kept codicons as the standard iconography.

**Correctness fixes from two review passes** (each with a regression test)
- Deleting a local database no longer deletes a *remote* login of the same name (was silent credential loss).
- "Start & log in" no longer starts a local stone for a remote login.
- The OS checklist no longer shows a "run sudo script" button on a healthy row.
- A downloaded-but-unextracted version no longer disappears offline; "no room for another cache" now counts as a warning; the raw `shmall` readout is clamped.
- Removed Open System Browser from session actions (it was retired from session rows).

**Also:** substantial test coverage added across the config layer and the panel; CHANGELOG updated. A follow-up refactor (extract the `activate()` handlers into a `gemstoneManager/` folder; remove the now-dead sidebar-views switch and its stale docs) is tracked in #482.

client/src/__tests__/vscodeignoreShippedAssets.test.ts
MatiasFernandez: I still think it's a little weird to have a test like that. I don't have a blocker that it feels weird that we are guarding configuration using a test. It's similar to the kind of tests Claude wrote in the past to verify that certain CSS was defined in a certain way.

THIRD-PARTY.md
MatiasFernandez: Agreed on the entry — but I'd tighten the justification, because as written it implies a rule we don't follow.

The stated reason is "the bytes ship in the `.vsix`." Sound rule, but `esbuild.mjs` bundles with `external: ['vscode', 'koffi']` only, so ~100 production packages are inlined verbatim into `extension.js` / `server.js` / `index.js` — 59 MIT, plus BSD and Apache-2.0, all of which want their notice on binary distribution too. `grep -c Copyright` on `extension.js` returns 0. So by the criterion as stated we're attributing 1 of ~100, and a reader can't tell why codicons made the cut.

I think it's the right call for a different reason, and I'd rather write that one down: CC BY 4.0 isn't a software license (affirmative, medium-aware attribution clause), it auto-terminates on breach with a 30-day cure while MIT has no termination clause, and it's a non-code asset shipping as two identifiable standalone files. That also covers the AST-Core entry, and gives us an answer to "shouldn't we attribute a lot of other stuff?" that holds up.

Three small things for this PR:

1. **Scope sentence in the preamble.** It currently says the file lists "vendored third-party source," and the new section's second bullet opens with "not vendored" — the header no longer describes the file. Something like: *"…material shipping in the `.vsix` as identifiable standalone files, or vendored as source into this repo. Permissively-licensed npm dependencies compiled into the bundles are covered separately."*
2. **Ship codicons' `LICENSE` and `LICENSE-CODE`** — two more `!` lines. Makes it match the koffi handling four lines above, where the license travels with the bytes. Cheap insurance on the one dependency with a termination clause.
3. **One line under `## License` in the README** pointing at it.

I'm going to file a separate issue to track improvements we can include for the other packages I mentioned we are not including the notice for

MatiasFernandez: Issue created here: https://github.com/GemTalk/Jasper/issues/495

PR #456, opened 2026-08-18
Silence libgcits' per-login stdout banner (87% less test output)

MatiasFernandez:
## Why

Every GemStone login and logout makes `libgcits` `printf` a line straight to the process's stdout:

```
 gcits login: session 0x88140c000 lgc 0x88140c008 rpc gem processId 24917
 gcits logout: nbExecInProg 0x0 session 0x88140c000 rpc gem processId 24917
```

It bypasses the JS `console` entirely — it comes from native code, not from JS — so **no vitest reporter, `--silent`, or `onConsoleLog` hook can suppress it**. Every test run, in CI and locally, carries one pair of these lines per session. Nobody reads them, and nothing in Jasper depends on them.

The cost is real and it is paid three times over: CI log noise, humans scrolling past it, and — the motivating case — AI coding agents, which pay for the whole test transcript in context tokens on every single run. Wasted context is wasted budget and wasted attention.

The fix is already in our hands: bit 4 of the `loginFlags` argument we pass on every login — `GCI_LOGIN_QUIET = 0x10` — gates every one of those stdout writes. Jasper hardcoded `loginFlags = 0` at all six production call sites, so we were opting into the noise by omission.

## What changed

- **`gciConstants.ts`** — new `GCI_LOGIN_QUIET = 0x10`, documented alongside the existing `GCI_PERFORM_FLAG_*` block.
- **`gciLibrary.ts`** — a private `quietedLoginFlags()` helper ORs the bit over whatever the caller passed, applied in all four raw login wrappers (`GciTsLogin`, `GciTsLogin_`, `GciTsNbLogin`, `GciTsNbLogin_`). They all share `doLogin` inside the library, so blocking, netldi and non-blocking paths behave identically.
- **No call site changes.** `sessionManager.ts`, `installHelpers.ts`, `mcp-server/src/mcpSession.ts`, `GciLibrary.login()` and every test keep passing `0`; the OR happens one level below them. Doing it in the wrapper rather than at each call site means future callers get it for free and cannot forget it.
- **Docs** — the class-level "thin 1:1 wrappers" claim now records this as its single deliberate exception, and `login()`'s doc no longer says flags are left at zero.
- **New test** — `client/src/gciLibrary/__tests__/gciLoginQuiet.test.ts` mocks koffi with a distinct stub per native function and asserts against the recorded native arguments: the quiet bit reaches all four wrappers, and a caller passing `GCI_LOGIN_PW_ENCRYPTED` (`1`) arrives as `0x11` — the bit is **added**, never substituted.

## The wins (measured, not estimated)

Full `npm test`, before vs after, `VITEST_SEED` pinned so runs are comparable:

| run | lines | bytes | tokens | `gcits` lines |
|---|---|---|---|---|
| all passing, before | 159 | 9,567 | **3,491** | 110 |
| all passing, after | 49 | 1,427 | **461** | 0 |
| one failing test, before | 147 | 9,785 | **3,590** | 110 |
| one failing test, after | 37 | 1,643 | **538** | 0 |

**~87% fewer tokens per test run** (85% when a test fails). 110 of the 159 baseline lines were the banner — it was the *majority* of our test output. The on-demand `npm run test:gci` suite, which performs ~20 real logins, went from 45 banner lines to 0.

Two things worth calling out:

- **Output filtering cannot fix this.** Running the same suite through a token-optimizing output filter saved 0.5% before this change (3,491 → 3,472 tokens), and on a failing run came out 3 tokens *larger*. Interleaved native stdout is invisible to line-based filters. This had to be fixed at the source.
- **Nothing diagnostic is lost.** `GCI_LOGIN_QUIET` does not affect the client trace file, so `GS_LGC_DEBUG` tracing still works if anyone ever needs the session details. Verified by deliberately breaking a live-login integration test: the `FAIL` header, assertion diff, `file:line:col` and code frame all still come through intact, and the received value was real data fetched from the stone — proof the login itself succeeded with the flag set.

## Compatibility

`GCI_LOGIN_QUIET` carries the value `0x10` in every version from 3.6.2 through 3.7.5, so **no version gating is needed** and `gciVersionGated.test.ts` is unaffected. Verified empirically against a live stone: login still succeeds, and the logout banner goes quiet too.

## Verification

- `npm run lint && npm run format:check && npm run compile` — clean.
- `npm test` — 377 client files / 5,642 tests, server 322, mcp-server 92, all passing; zero `gcits` lines.
- `npm run test:gci` — comes back quiet (45 → 0 banner lines)
- Existing tests that assert `GciTsLogin` was called with `..., 0, 0` (`mcpSession.test.ts`, `sessionManager.test.ts`) mock the *wrapper*, which still receives `0` — they stay green and keep asserting what the caller intends.

### DIFF
--- a/client/src/gciConstants.ts
+++ b/client/src/gciConstants.ts
@@ -30,6 +30,17 @@ export const GCI_PERFORM_FLAG_SINGLE_STEP = 4;
 // No effect where native code is unavailable anyway (e.g. Darwin/ARM builds).
 export const GCI_PERFORM_FLAG_INTERPRETED = 0x20;
 
+// Login flags (from gci.ht)
+// Tells GciTsLogin the `gemstonePassword` argument is already ciphertext, as produced by
+// GciTsEncrypt, rather than plaintext.
+export const GCI_LOGIN_PW_ENCRYPTED = 0x1;
+
+// Suppresses libgcits' unsolicited ` gcits login:` / ` gcits logout:` printf()s
+// to stdout, which are the majority of the test suite's output and bypass the
+// JS console entirely. Present with this value since 3.6.2, so no version
+// gating is needed.
+export const GCI_LOGIN_QUIET = 0x10;
+
 // Class OOPs (from gcioop.ht)
 export const OOP_CLASS_STRING = 74753n;
 export const OOP_CLASS_UTF8 = 154113n;
--- a/client/src/gciLibrary.ts
+++ b/client/src/gciLibrary.ts
@@ -1,6 +1,6 @@
 import koffi from 'koffi';
 import * as path from 'path';
-import { OOP_FALSE, OOP_ILLEGAL, OOP_NIL, OOP_TRUE } from './gciConstants';
+import { GCI_LOGIN_QUIET, OOP_FALSE, OOP_ILLEGAL, OOP_NIL, OOP_TRUE } from './gciConstants';
 import { GciLibraryError } from './gciLibraryError';
 import { escapeString } from './queries/util';
 
@@ -151,6 +151,15 @@ function toBigInt(value: number | bigint): bigint {
   return typeof value === 'bigint' ? value : BigInt(value);
 }
 
+// Forces GCI_LOGIN_QUIET on top of a caller's loginFlags. Jasper never wants
+// libgcits' ` gcits login:` / ` gcits logout:` banner on stdout, so every login
+// wrapper ORs the bit in rather than making all ~6 call sites remember it. This
+// is the one deliberate deviation from the raw wrappers' 1:1 contract; all
+// other caller-supplied bits pass through untouched.
+function quietedLoginFlags(loginFlags: number): number {
+  return loginFlags | GCI_LOGIN_QUIET;
+}
+
 // Rejects a Promise-returning (async) callback at the type level -- see
 // GciLibrary.executeAndRelease's doc comment for why that matters.
 type NotPromise<T> = T extends Promise<unknown> ? never : T;
@@ -163,6 +172,8 @@ type NotPromise<T> = T extends Promise<unknown> ? never : T;
  * - Raw `GciTsXxx` methods (e.g. `GciTsExecute`, `GciTsLogin`) are thin 1:1
  *   wrappers around the C functions of the same name — one per exported
  *   symbol, following the struct/pointer patterns already established below.
+ *   The one deliberate exception: the login wrappers force `GCI_LOGIN_QUIET`
+ *   into `loginFlags` — see the `quietedLoginFlags` helper above for why.
  * - The remaining methods (grouped into labeled sections further down: Session
  *   lifecycle, Code execution, Symbol resolution & Utf8 caching, Object
  *   lifecycle & PureExportSet, UserGlobals management, SessionTemps
@@ -409,10 +420,11 @@ export class GciLibrary {
     this._GciTsNbLogout = this.lib.func(`int GciTsNbLogout(GciSessionPtr, _Out_ GciErrSType *)`);
     this._GciTsSessionIsRemote = this.lib.func(`int GciTsSessionIsRemote(GciSessionPtr)`);
     // Optional: not exported by some libraries (e.g. GemStone 4.0). Jasper's
-    // login path passes the password in the clear (GciTsLogin with loginFlags
-    // 0), so GciTsEncrypt is never called during connect — binding it eagerly
-    // would abort library load for a library that lacks it. Only the ergonomic
-    // GciTsEncrypt() wrapper (used by tests) would throw if it were absent.
+    // login path passes the password in the clear (GciTsLogin without
+    // GCI_LOGIN_PW_ENCRYPTED), so GciTsEncrypt is never called during connect —
+    // binding it eagerly would abort library load for a library that lacks it.
+    // Only the ergonomic GciTsEncrypt() wrapper (used by tests) would throw if
+    // it were absent.
     this._GciTsEncrypt = this.optionalFunc(
       'GciTsEncrypt',
       `char* GciTsEncrypt(const char *, _Out_ char *, size_t)`,
@@ -769,7 +781,7 @@ export class GciLibrary {
       gemServiceNrs,
       gemstoneUsername,
       gemstonePassword,
-      loginFlags,
+      quietedLoginFlags(loginFlags),
       haltOnErrNum,
       executedSessionInit,
       err,
@@ -804,7 +816,7 @@ export class GciLibrary {
       gemstoneUsername,
       gemstonePassword,
       netldiName,
-      loginFlags,
+      quietedLoginFlags(loginFlags),
       haltOnErrNum,
       executedSessionInit,
       err,
@@ -837,7 +849,7 @@ export class GciLibrary {
       gemServiceNrs,
       gemstoneUsername,
       gemstonePassword,
-      loginFlags,
+      quietedLoginFlags(loginFlags),
       haltOnErrNum,
       loginPollSocket,
     );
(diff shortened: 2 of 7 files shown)


client/src/gciLibrary/__tests__/gciLoginQuiet.test.ts
```diff
@@ -0,0 +1,115 @@
+import { describe, it, expect, beforeEach, vi } from 'vitest';
```
ericwinger: 🧪 Placement: this is a koffi-mocked unit test, but `client/src/gciLibrary/README.md` says this folder's `__tests__/` "holds integration tests for the **raw bindings** only" — and every sibling here is named `*.integration.test.ts` and needs a live stone. The repo's only other koffi-mocked, library-free test is `client/src/__tests__/gciOptionalFunctions.test.ts`. Either move this file next to that one, or update the README: as it stands the folder no longer matches its own description.

client/src/gciLibrary.ts
```diff
@@ -163,6 +172,8 @@ type NotPromise<T> = T extends Promise<unknown> ? never : T;
  * - Raw `GciTsXxx` methods (e.g. `GciTsExecute`, `GciTsLogin`) are thin 1:1
  *   wrappers around the C functions of the same name — one per exported
  *   symbol, following the struct/pointer patterns already established below.
+ *   The one deliberate exception: the login wrappers force `GCI_LOGIN_QUIET`
```
ericwinger: 📝 The exception is recorded here, but the same "1:1" claim also lives in `client/src/gciLibrary/README.md`, which defines that folder's raw bindings as "the wrappers that are 1:1 with a GCI C entry point" — the very contract this PR carves an exception out of. That README went stale with this change and isn't in the diff.

client/src/gciLibrary/__tests__/gciLoginQuiet.test.ts
```diff
@@ -0,0 +1,115 @@
...
+import { GCI_LOGIN_PW_ENCRYPTED, GCI_LOGIN_QUIET } from '../../gciConstants';
+
+// One stub per native function, so each login binding's arguments can be
+// inspected independently. Keyed by the function name parsed out of the koffi
+// signature string. vi.hoisted because the vi.mock factory below is hoisted
+// above this module's imports.
+const nativeStubs = vi.hoisted(() => new Map<string, ReturnType<typeof vi.fn>>());
+
+vi.mock('koffi', () => {
```
ericwinger: 🧪 This factory is near-identical to the one in `client/src/__tests__/gciOptionalFunctions.test.ts:31-54` — same `mockLib` shape, same `struct`/`array`/`opaque`/`pointer`/`union`/`load` stubs; only `func`'s behaviour differs. If the two files end up in the same folder, a shared helper parameterized by the `func` implementation would keep the koffi surface in one place; today the next `koffi.*` call added to `GciLibrary`'s constructor has to be mirrored into both.

client/src/gciLibrary/__tests__/gciLoginQuiet.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      pointer: vi.fn(() => 'MockPointer'),
+      union: vi.fn(() => 'MockUnion'),
+      load: vi.fn(() => mockLib),
+    },
+  };
+});
+
+import { GciLibrary } from '../../gciLibrary';
+
+/** The `loginFlags` argument the named native function was last called with. */
```
ericwinger: 📝 The doc says "last called with", but the body reads `mock.calls[0]` — the *first* call — after asserting there is exactly one. Harmless as written (first == last), but "the `loginFlags` argument the named native function was called with, asserting it was called exactly once" is what it actually does.

client/src/gciLibrary/__tests__/gciLoginQuiet.test.ts
```diff
@@ -0,0 +1,115 @@
...
+  let gci: GciLibrary;
+
+  beforeEach(() => {
+    nativeStubs.clear();
+    gci = new GciLibrary('/fake/libgcits.dylib');
+  });
+
+  // `loginFlags` sits at index 7 in GciTsLogin/GciTsNbLogin and at index 8 in
+  // the `_` variants, which take an extra netldiName before it.
+  const LOGIN_FLAGS_INDEX = 7;
```
ericwinger: 🧪 Optional strengthening: every assertion reads one positional index, so if a wrapper's argument order ever shifted these would quietly compare a *different* argument instead of failing. Asserting the whole recorded call for one wrapper — `expect(stub.mock.calls[0]).toEqual([...])` — would pin the index and also prove `quietedLoginFlags`' claim that "all other caller-supplied bits pass through untouched".

Related small gap: `GciTsNbLogin_` is the one wrapper never exercised with caller-supplied flags (it's covered for the no-flags case at line 77; the other three get the `GCI_LOGIN_PW_ENCRYPTED` treatment at lines 85 and 93).

client/src/gciConstants.ts
```diff
@@ -30,6 +30,20 @@ export const GCI_PERFORM_FLAG_SINGLE_STEP = 4;
...
+// Login flags (from gci.ht)
+// Tells GciTsLogin the password argument is already ciphertext, as produced by
+// GciTsEncrypt, rather than plaintext. Jasper's own login path passes the
+// password in the clear and never sets this; it exists here so callers and
+// tests name the bit instead of writing a bare 1.
+export const GCI_LOGIN_PW_ENCRYPTED = 0x1;
+
+// Suppresses libgcits' unsolicited ` gcits login:` / ` gcits logout:` printf()s
+// to stdout, which are the majority of the test suite's output and bypass the
+// JS console entirely. Does not affect the client trace file, so GS_LGC_DEBUG
```
ericwinger: 💭 "so `GS_LGC_DEBUG` tracing still works" is the one claim in this PR that nothing covers, and there is no existing coverage to lean on: `GS_LGC_DEBUG` appears nowhere else in the repo — no code, test, or doc references it. The nearest thing is `client/src/__tests__/gci/gciAsync.test.ts:100`, and it is a different mechanism (the `GciTsGemTrace` GCI call writing `gci<pid>trace.log`, not the env var); it asserts only `err.number` and the previous-level return values, never file contents, and it lives in the on-demand `gci` project that `npm test` excludes.

If this was checked by hand against a live login with the flag set, saying so here (or narrowing the sentence to what was actually observed) would stop the next reader from reading it as covered behaviour.

client/src/gciConstants.ts
```diff
@@ -30,6 +30,20 @@ export const GCI_PERFORM_FLAG_SINGLE_STEP = 4;
 // No effect where native code is unavailable anyway (e.g. Darwin/ARM builds).
 export const GCI_PERFORM_FLAG_INTERPRETED = 0x20;
 
+// Login flags (from gci.ht)
+// Tells GciTsLogin the password argument is already ciphertext, as produced by
```
ericwinger: 📝 `GciTsLogin` takes two passwords — `HostPassword` and `gemstonePassword` — so "the password argument" is ambiguous. `gcits.hf` is specific: "GCI_LOGIN_PW_ENCRYPTED bit in loginFlags applies to gemstonePassword argument"; the host password has its own `hostPwIsEncrypted` boolean parameter instead. Naming `gemstonePassword` here settles it.

client/src/gciConstants.ts
ericwinger: Eric: I noticed this and asked Claude to confirm. Might be worth a simple test if possible.

ericwinger (approved):
LGTM

client/src/gciLibrary/__tests__/gciLoginQuiet.test.ts
MatiasFernandez: This is ok for now. I'm in the process of reorganizing this, so it's hard to have a single way to describe this folder. I'll improve the README in a separate PR.

client/src/gciConstants.ts
MatiasFernandez: I missed this `GS_LGC_DEBUG` reference. This is AI slop; that flag has nothing to do with the change I did, so I'm going to clean this comment up.

npapagna (approved):
This is great! Thanks @MatiasFernandez!

PR #458, opened 2026-08-19
GemStone Search cleanup: one name, Class Categories, dead-code removal, hint-in-pivot fix

### LINKED ISSUE #447: Omni Search: naming, DESIGN.md, and dead-code cleanup (deferred until #443 lands)
## Summary
Four small Omni Search follow-ups from #428 are all entangled with **#443** (which removes the Quick Pick, rewrites `DESIGN.md` wholesale, and reworks `methodSearch.ts`). Grouping them here as one small batch to pick up **after #443 merges**, so they don't churn against that PR's changes.

## Where
`client/src/omniSearch/` — `methodSearch.ts`, `omniTypes.ts`, `omniSearchShared.ts`, `DESIGN.md`, and the `gemstone.omniSearch.*` settings / command titles.

## Items (all deferred until #443 lands)
1. **`referencesToLiteral` looks dead — document or remove it.** The only references are its own test and the DESIGN note; confirm reachability or drop it. (from #428; [PR #423 review](https://github.com/GemTalk/Jasper/pull/423#discussion_r3779019885))
2. **Pick one name for the feature.** The code and settings say "Omni Search" (`client/src/omniSearch/`, `gemstone.omniSearch.*`), while the user-facing command title, the settings-page title and the panel say "GemStone Search". Suggestion: standardize on **Omni Search** — more explicit than the generic "search", and it makes the command palette easier to filter. (from #428)
3. **The "Categories" scope is ambiguous — say "Class Categories".** Class categories or method categories? It looks like class categories (`providers/categoriesProvider.ts` builds its corpus from `getAllClassCategories`). Worth making that explicit in the UI label, in the `gemstone.omniSearch.categories` enum description / `markdownDescription`, and in the code naming. Already prototyped (the tab label in `omniTypes.ts` and `SCOPE_LABEL` in `omniSearchShared.ts` renamed to "Class Categories", plus a guard test) but held uncommitted so it ships together with the naming decision (item 2) and the `DESIGN.md` sync, after #443. (from #428)
4. **Decide `DESIGN.md`'s home — in-repo vs external support doc.** Whether the design note stays in the repo or becomes an external support doc. ([PR #423 review](https://github.com/GemTalk/Jasper/pull/423#discussion_r3779182761); from #428)

## Why deferred
#443 rewrites `DESIGN.md` (~253 lines) and reworks `methodSearch.ts`, so doing these now creates conflicts and churn. Tracking them here as a single small batch to clear once #443 merges.

Follow-up to #428.

ericwinger:
Resolves the small GemStone Search follow-up batch that was deferred until the Quick Pick removal / `DESIGN.md` rewrite landed.

## Changes

- **One name.** The user-facing name is **GemStone Search** everywhere it is rendered — command, settings page, panel/Spotter title, the bottom-panel tab, the status-bar button, error toasts, and tooltips — and in comments, docs, and tests. A few surfaces still said "Omni Search"; they no longer do. The `omniSearch` prefix is kept **only** as an internal identifier (`client/src/omniSearch/`, the `OmniSearch*` classes, the `gemstone.omniSearch.*` setting keys), because renaming the setting keys would break users' existing `settings.json`. Rationale recorded on the issue: https://github.com/GemTalk/Jasper/issues/447#issuecomment-5345377117
- **"Categories" → "Class Categories".** The scope is relabeled in the tab, the placeholder, the setting `markdownDescription`, and new per-value `enumDescriptions`, so it is explicit that it searches **class** categories, not method categories. A guard test pins the label.
- **Dead-code removal.** The unused `referencesToLiteral` TypeScript wrapper is gone (only its own test used it). The Smalltalk selector `ClassOrganizer>>referencesToLiteral:` is still used by `literalSymbolReferences` and is untouched.
- **`DESIGN.md` stays in-repo**, next to the code it documents; it records the naming decision and the `OMNI_CATEGORIES` / `categories`-id naming collision (left as-is deliberately).

## Fixed

- **The "not searched here" hint no longer appears while a references pivot is showing.** A pivot's rows are a fixed list of senders already fetched from the stone, not a search, so nothing was being skipped — and each scope name in the hint is a button that would have started a fresh search and silently discarded the pivot. Two jsdom tests cover it (hidden during a pivot; restored on the way out).

## Testing

`npm run lint` / `format:check` / `compile` clean; full `npm test` green — client **5939 pass / 12 skip**, server **322**, mcp **92** (against a running 3.6.2 stone). No server / RB / GCI runtime surface changed.

Closes #447.


client/src/omniSearch/__tests__/omniSearchView.test.ts
```diff
@@ -902,7 +902,7 @@ describe('Omni Search view — references in the preview pane', () => {
 // `providersInScope` whenever the scope is All, so an All-scope search never runs them and nothing said
 // so. Verified live: `no such element` finds 4 methods under Source and 0 under All. The hint names the
 // skipped scopes and doubles as the one-click way into them.
-describe('Omni Search view — scopes skipped under All', () => {
+describe('GemStone Search view — scopes skipped under All', () => {
```
MatiasFernandez: 💭 The rename pass seems to have stopped at the `describe` titles in this file — three leftovers around this block still carry the old name:

- **L77**: the `CATEGORIES` fixture still has `{ id: 'categories', label: 'Categories', explicitOnly: true }`. Since every test in this file renders the tab row from that fixture, no jsdom test actually renders `Class Categories` — the new label is only exercised by the `omniSettings.test.ts` guard.
- **L922**: `expect(hint().textContent).toContain('Categories')` passes for *both* labels (it's a substring of `Class Categories`), so the hint — arguably the surface where the class-vs-method-category ambiguity bites hardest — has no assertion that can tell the two apart.
- **L901**: the block comment still reads `(Source / Literals / Categories)`.

Would you mind updating the fixture to `'Class Categories'` and tightening L922 to `toContain('Class Categories')`? That way a regression back to the ambiguous label fails here too, not only in the settings guard.

client/src/omniSearch/omniSearchView.js
```diff
@@ -891,16 +891,23 @@
...
       for (var i = 0; i < lastCategories.length; i++) {
         if (lastCategories[i].explicitOnly) heavy.push(lastCategories[i]);
       }
-      // Only under All (no scope), only with a term to search, and only if any heavy scope is enabled
-      // at all — a user who disabled them via `categories` is not missing anything.
-      var show = lastScopeId === null && inputEl.value.trim().length > 0 && heavy.length > 0;
+      // Only under All (no scope), never during a pivot, only with a term to search, and only if any
+      // heavy scope is enabled at all — a user who disabled them via `categories` is not missing anything.
+      var show =
+        !inPivot && lastScopeId === null && inputEl.value.trim().length > 0 && heavy.length > 0;
```
MatiasFernandez: 💭 The two reasons in the doc comment above pull apart a bit for the *other* references mode, and I'm not sure which way you intended it.

With `referencesInPreview` on (the default), Alt+Enter goes through `post('referencesInline')`: the left list stays as the All-scope search results and only `previewMode` becomes `'refs'` — `inPivot` stays false, so the hint keeps showing. The first reason still holds there (the left rows really are a search, so scopes *are* being skipped), but the second one holds too: clicking a scope name starts a fresh search, which throws away the sticky references preview just as silently as it throws away a pivot. `refreshRefIndicator` treats the two states alike (`inPivot || previewMode === 'refs'`), which is what made me look.

Is the asymmetry deliberate — hint stays because the list is still a search — with the discarded-preview cost accepted as smaller than losing the "nothing was searched here" warning? If so it'd be worth a clause in the comment, since the default path is the un-guarded one.

package.json
```diff
@@ -173,6 +173,15 @@
                 "source",
                 "literals",
                 "categories"
+              ],
+              "enumDescriptions": [
```
MatiasFernandez: 💭 Nice addition — the per-value help is exactly what was missing from this setting.

One thing about `enumDescriptions`: it's positional, so inserting a value into `enum` without inserting at the matching index silently shifts every description after it. VS Code doesn't validate the pairing, so the settings page just shows the wrong help text and nothing fails. Given `omniSettings.test.ts` already reads this block out of `properties`, would it be worth pinning it there — the length matching `enum.length`, and maybe the `categories` entry mentioning "not method categories" — the same way the keybindings manifest test pins the `gemstone.omniSearch` binding?

client/src/omniSearch/omniSearchShared.ts
```diff
@@ -41,7 +41,7 @@ const SCOPE_LABEL: Record<string, string> = {
   globals: 'globals',
   source: 'source',
   literals: 'literals',
-  categories: 'categories',
+  categories: 'class categories',
```
MatiasFernandez: 💭 I think this one is unreachable: `placeholderFor` returns `CATEGORY_BY_ID[scopeId].searchHint` first when there is one, and `categories` has `'Type a class-category name, e.g. Kernel-Objects'` — so it never falls through to `SCOPE_LABEL`. (Same for `all`, which `placeholderFor(null)` short-circuits with a hardcoded string, so this predates the PR.)

Happy either way — keeping it consistent is cheap insurance if a scope ever loses its `searchHint`. Just flagging it so nobody reads this as the string the Class Categories placeholder shows.

client/src/omniSearch/__tests__/omniSettings.test.ts
```diff
@@ -29,6 +29,15 @@ describe('gemstone.omniSearch.ui setting', () => {
   });
 });
 
+describe('the class-categories scope', () => {
+  it('labels itself "Class Categories", not the ambiguous "Categories"', () => {
+    // The scope searches CLASS categories (getAllClassCategories), never method categories — the tab
+    // and every user-facing surface say so explicitly so the distinction is not left to the reader.
+    const categories = OMNI_CATEGORIES.find((c) => c.id === 'categories');
+    expect(categories?.label).toBe('Class Categories');
```
MatiasFernandez: nit: with the optional chaining, a rename of the `categories` **id** fails this as `undefined` vs `'Class Categories'`, which reads like the label regressed rather than the lookup missing. An `expect(categories).toBeDefined()` first (or dropping the `?.` so TS makes you assert it) would point at the real cause.

client/src/omniSearch/DESIGN.md
```diff
@@ -10,6 +10,18 @@ Global "search anything browsable" for the GemStone IDE — the Jasper answer to
...
+> settings-page section, the panel/Spotter title, the bottom-panel tab, the status-bar button, error
+> toasts, tooltips, and throughout this document. The name keeps the GemStone brand on the panel strip
+> next to Terminal / Output, where a generic name would say nothing about which extension owns it. The
+> code still uses the `omniSearch` identifier prefix — the `client/src/omniSearch/` module, the
+> `OmniSearch*` classes, the `gemstone.omniSearch` command, and the `gemstone.omniSearch.*` setting
+> keys — because those keys live in users' `settings.json` and renaming them would break existing
+> configuration for no user-visible gain. Treat `omniSearch` as an internal identifier, not a second
+> name. This design note **stays in the repo** next to the code it describes rather than moving to an
+> external support doc, so it is reviewed and kept in sync in the same change as the code (both
+> decisions resolve #428 / [#447](https://github.com/GemTalk/Jasper/issues/447)).
```
MatiasFernandez: 💭 Thanks for writing the rationale down here rather than leaving it on the issue thread ❤️ — this is the note I'd want to find in six months when someone asks why the directory says `omniSearch`.

Since it says both decisions resolve #428 as well, should the PR description close that one too? Right now it only says `Closes #447`, so #428 would stay open with its answer sitting in a merged design doc.

client/src/__tests__/keybindings.test.ts
```diff
@@ -102,7 +102,7 @@ describe('keybindings', () => {
     }
   });
 
-  it('Omni Search opens on Ctrl/Cmd+Shift+A, not a notebook run-cell gesture', () => {
+  it('GemStone Search opens on Ctrl/Cmd+Shift+A, not a notebook run-cell gesture', () => {
     const omni = keybindings.filter((kb) => kb.command === 'gemstone.omniSearch');
```
MatiasFernandez: Your PR description says renaming the command keys would affect users, but we've been treating features as experimental recently, so I think that if we want, we should rename the commands for now. I'm not convinced that using "gemstone search
 as an identifier is descriptive enough, but just wanted to mention that if you want to rename the commands, go for it. I think Claude by default avoids that, which is a good practice when you already have an established user base. But given this is a very recent feature and we still don't have an established user base, the risks of conflict are very, very low.

client/src/omniSearch/__tests__/omniSettings.test.ts
```diff
@@ -29,6 +29,15 @@ describe('gemstone.omniSearch.ui setting', () => {
   });
 });
 
+describe('the class-categories scope', () => {
+  it('labels itself "Class Categories", not the ambiguous "Categories"', () => {
```
MatiasFernandez: 🧪 This test is weird; it's the same pattern Claude follows when it forces a test creation to assert _something_. In this case, we are just verifying that a constant has the correct string for the label; it's like "testing the configuration" instead of behavior. In this case, this is just a label for humans, so just renaming the value in the original place should be enough; we don't need a test to ensure it won't change

MatiasFernandez (approved):

client/src/omniSearch/__tests__/omniSearchView.test.ts
ericwinger: Good catch — all three were leftovers of the rename pass. Fixed in `1d35c408`:

- **L77**: the fixture is now `{ id: 'categories', label: 'Class Categories', explicitOnly: true }`, so every jsdom test in this file renders the new label rather than the old one.
- **L922**: tightened to `toContain('Class Categories')`, which the old label can no longer satisfy.
- **L901**: the block comment now reads `(Source / Literals / Class Categories)`.

So a regression back to the ambiguous label fails right here — which matters more than it did when you wrote this, since the `omniSettings.test.ts` guard you flagged further down is now deleted. This file is the assertion that keeps the label honest.

client/src/omniSearch/omniSearchView.js
ericwinger: The asymmetry is deliberate, and you read the intent correctly — so I've added the clause, in `1d35c408`.

The reasoning now spelled out above `updateScopeHint`: in references-*preview* mode `inPivot` is false because the left list really is the live All-scope search, so the first reason no longer applies — those rows genuinely did skip the explicit-only scopes, and the warning is true. Clicking a scope does discard the sticky preview just as silently as it would drop a pivot, but that ephemeral preview is the smaller loss compared with suppressing an accurate "nothing was searched here" warning on a live search. Only the full-list pivot suppresses the hint.

You're right that `refreshRefIndicator` treats the two states alike (`inPivot || previewMode === 'refs'`) — that one is about whether to *show the indicator*, where both modes do have a references pane, so the two really are alike there. It's the hint that needs to tell them apart. The comment now says so, since the default path is the un-guarded one.

client/src/omniSearch/omniSearchShared.ts
ericwinger: You're right, it is unreachable — `placeholderFor` returns `CATEGORY_BY_ID[scopeId].searchHint` first, and `categories` has one, so the `SCOPE_LABEL` fall-through never fires. And as you note, `all` short-circuits earlier still, so that half predates this PR.

Leaving it as-is, on your "happy either way": it's the cheap insurance case — if a scope ever loses its `searchHint` the fall-through is the only thing standing between the user and an empty placeholder. No code change here, but the thread is worth having on record so nobody later reads that string as what the Class Categories placeholder actually shows.

client/src/__tests__/keybindings.test.ts
ericwinger: Renamed, in `1d35c408` — the command is now **`gemstone.search`**.

You're right about the risk calculus, and it's the reason I raised it rather than just doing it: the reflex to preserve command ids is worth having, but it's aimed at an installed base that this feature doesn't have yet. Better to pay the rename now than to carry `omniSearch` forever.

On "descriptive enough" — I went with `gemstone.search` over `gemstone.omniSearch` precisely because "omni" was jargon that only made sense from the inside; the palette title is still **GemStone Search…**, so what the user reads is unchanged. It sits beside the two narrower ones (`gemstone.searchRowanClasses`, `gemstone.searchMethods`), and being the unqualified one of the three reads about right for the search that spans everything. Happy to take a better name if you have one.

One deliberate exception: the **`gemstone.omniSearch.*` settings namespace stays.** Those keys live in users' own `settings.json`, where a rename silently drops their configuration rather than erroring — a different and worse failure than a command id going missing. The note above the namespace in `omniConfig.ts` now says that's why, so it doesn't read as a spot the rename missed.

Renamed across 6 call sites plus `DESIGN.md` (×2) and the `omniSearchShared` / `omniSearchCommand` doc comments.

client/src/omniSearch/__tests__/omniSettings.test.ts
ericwinger: Agreed — deleted in `1d35c408`.

You've named the pattern exactly: it asserts that a constant equals its own literal, which can only ever fail if someone edits the constant, and a human-facing label is allowed to change. It was also load-bearing for the wrong reason — it was the *only* place rendering the new label, because the view-test fixture still said `'Categories'`. With that fixture fixed (your other comment), the label is exercised where it's actually used, and this test has nothing left to add.

PR #464, opened 2026-08-19
GemStone Search: bind reveals/searches to the result's session, sweep all method environments

### LINKED ISSUE #459: GemStone Search searches and reveals against the wrong session when two sessions are open
## Summary

In a window with two logged-in GemStone sessions, GemStone Search can search and reveal results
against the wrong session. Two related defects: the Spotter panel keeps using the session it was first
opened with, and the Explorer "reveal" actions (dictionary / global / class-category) resolve against
whatever session is currently selected instead of the session the result came from.

## Environment

- **Jasper:** 1.8.11
- **GemStone server:** N/A (client-side session routing; independent of stone version)

## Steps to reproduce

Both scenarios need two logged-in sessions, A and B.

**Spotter panel keeps the old session:**
1. Log into session A and session B.
2. With A active, open GemStone Search (the Spotter panel).
3. Switch the active session to B.
4. Invoke GemStone Search again.

**Reveal resolves against the wrong session:**
1. Log into session A and session B.
2. With A active, search and get a Dictionaries / Globals / Class Categories result.
3. Switch the active session to B.
4. Click the result.

## Expected behavior

- Re-invoking GemStone Search after switching sessions searches and opens against the now-active
  session.
- Clicking a result reveals it in the session the result came from, regardless of which session is
  currently selected. If that session is no longer active, warn rather than silently resolving against
  a different one.

## Actual behavior

- The Spotter panel is revealed but stays bound to the session it was first opened with, so searches
  and result-opens still run against that session with no indication anything is wrong
  (`OmniSearchPanel.show()` reveals the existing panel and discards the incoming deps).
- The reveal actions resolve against the currently-selected session: best case a "not found" in B;
  worst case a same-named dictionary or category exists in B and you land in the wrong session's data
  silently. The reveal handlers (`revealDictionary` / `revealGlobal` / `revealCategory` in
  `client/src/omniSearch/omniSearchCommand.ts`) drop the result's session id, and the Explorer commands
  (`revealDictionaryByName` / `findClass` / `revealCategoryByPath` in `client/src/gemstoneExplorer.ts`)
  don't accept one — they operate on whatever session the Explorer currently has selected.

## Potential fix

Both surfaced in a code review of the GemStone Search feature (on PR #451, though outside that PR's
diff). A fix is in progress on branch `eric/omni-search-stale-session`:

- `OmniSearchPanel.show()` compares the incoming session id — same session → reveal + refocus;
  different session → replace the panel. (Its engine, providers, activation and preview are built once
  from the original session's deps in the constructor, so it can't be re-pointed in place.)
- Thread the result's session id through the three reveal handlers and the Explorer commands, selecting
  that session first (and warning if it's gone) — matching how `openClass` / `openMethod` already carry
  the session.

### LINKED ISSUE #461: GemStone Search references pivot misses senders/references in method environments other than 0
## Summary

On a stone configured with more than one method environment (`gemstone.maxEnvironment > 0`), pivoting a
GemStone Search result to its senders/references only searches environment 0, so senders or references
that live in environment ≥ 1 are silently missed.

## Environment

- **Jasper:** 1.8.12
- **GemStone server:** any — the bug only manifests when `gemstone.maxEnvironment > 0` (multiple method
  environments in use).

## Steps to reproduce

1. Use a stone with `gemstone.maxEnvironment > 0`, with at least one sender of a selector compiled in a
   method environment ≥ 1.
2. In GemStone Search, find a method (or class) result.
3. Pivot to its senders / references (the per-row ↗ button, or Alt+Enter on the row).

## Expected behavior

The pivot lists senders/references across all method environments (`0..maxEnvironment`), deduped by
class/selector — the same coverage the existing Senders/Implementors commands already provide — and each
result opens in the environment it was found in.

## Actual behavior

Only environment 0 is queried, so any sender/reference that exists solely in environment ≥ 1 is missing
from the pivot list. The hover-based senders count (which does loop environments) can report a larger
number than the pivot shows for the same selector.

## Potential fix

Surfaced in a code review of the GemStone Search feature (on PR #451, though outside that PR's diff). A
fix is in progress on branch `eric/omni-search-fixes`: `resolveReferencesUsing` sweeps
`0..maxEnvironment` (matching `gemstone.sendersOfSelector` / `implementorsOfSelector`), dedups by
class/selector keeping the lowest-environment copy, and carries each surviving row's environment into its
open action so it opens where it was found. `maxEnvironment` defaults to 0, so single-environment stones
are unaffected (one query, as today).

ericwinger:
## What & why

Three correctness fixes for GemStone Search, each a state bug that only surfaces with more than one
session or method environment. They were surfaced by a `/code-review` pass over the GemStone Search
feature (posted on #451, but outside that PR's diff), and are handled here as their own change.

### Fixes

- **The Spotter panel stayed bound to the session it was opened with.** `OmniSearchPanel.show()` used to
  reveal the existing panel and discard the incoming `deps`, so after the active session changed it kept
  searching and opening against the old session with no sign anything was wrong. It now compares the
  session: same session → reveal + refocus (unchanged); different session → replace the panel (the
  engine, providers, activation and preview are built once from `deps` in the constructor, so it can't be
  re-pointed in place). (#459)

- **The dictionary / global / class-category reveal actions ignored the result's session.** The three
  reveal handlers dropped the result's `sessionId`, and the Explorer commands resolved against whatever
  session was selected *now* — so searching in one session, switching active to another, then clicking a
  result revealed it in the wrong session (best case "not found", worst case a same-named
  dictionary/category in the other session, silently). The handlers now thread `a.sessionId` through, and
  `findClass` / `revealDictionaryByName` / `revealCategoryByPath` select that session first (warning if
  it's gone) — matching how `openClass` / `openMethod` already carry the session. (#459)

- **The references pivot only queried method environment 0.** On a stone with `maxEnvironment > 0`,
  pivoting to senders/references missed every hit in environment ≥ 1. `resolveReferencesUsing` now sweeps
  `0..maxEnvironment` and dedups by class/selector — matching the existing
  `sendersOfSelector` / `implementorsOfSelector` commands — and carries each surviving row's environment
  into its open action so it opens where it was found. `maxEnvironment` defaults to 0, so the common
  single-environment case is unchanged (one round trip). (#461)

## Testing

Client-only TypeScript change (no server / GemStone / RB surface). Fast gate green — `lint`,
`format:check`, `compile`, and the full `npm test` (client **5949 passed / 12 skipped** incl. the live
integration test, server **322**, mcp **92**), re-run after merging current `main`.

Each fix is pinned by unit tests that **fail on the old code and pass on the new** (verified by reverting
only the source): new `omniSearchPanel.test.ts` and `resolveReferences.test.ts`, plus additions to
`omniSearchCommand.test.ts`, `explorerOmniReveal.test.ts`, and `references.test.ts`.

**No F5 verification:** all three require the active session to change between a search and the action on
its result (or a `maxEnvironment > 0` stone), and there is no quick session switch to stage that by hand
— so the unit tests are the verification of record here.

Closes #459
Closes #461

### DIFF
--- a/client/src/gemstoneExplorer.ts
+++ b/client/src/gemstoneExplorer.ts
@@ -764,7 +764,14 @@ export class ExplorerController {
       );
       return undefined;
     }
-    this.sessionManager.selectSession(sessionId);
+    // Only switch when the result came from a DIFFERENT session than the one already selected. The
+    // common case — searching and clicking a result in the session you are already in — must not fire a
+    // needless `selectSession`, which unconditionally emits onDidChangeSelection (sessionManager.ts) and
+    // drives a full downstream refresh it doesn't need: symbol-cache invalidation, admin-panel restale,
+    // sunit resync, tree refreshes. Mirrors OmniSearchPanel.show()'s session-id compare before switching.
+    if (this.sessionManager.getSelectedSession()?.id !== sessionId) {
+      this.sessionManager.selectSession(sessionId);
+    }
     return session;
   }
 
--- a/client/src/gemstoneExplorer.ts
+++ b/client/src/gemstoneExplorer.ts
@@ -750,6 +750,24 @@ export class ExplorerController {
     return this.sessionManager.getSelectedSession();
   }
 
+  // Resolve the session a reveal should run against. When a caller (Omni Search) names the session its
+  // result came from, switch the Explorer to that session first so the reveal targets the right data —
+  // otherwise, after the user switches the active session, a click would resolve against the new one
+  // and land in the wrong session (or a same-named dictionary/category). With no id, fall back to the
+  // normal "resolve the selected session (prompting if ambiguous)" path.
+  private async resolveSessionFor(sessionId?: number): Promise<ActiveSession | undefined> {
+    if (sessionId === undefined) return this.sessionManager.resolveSession();
+    const session = this.sessionManager.getSession(sessionId);
+    if (!session) {
+      void vscode.window.showWarningMessage(
+        `That result's GemStone session (${sessionId}) is no longer active.`,
+      );
+      return undefined;
+    }
+    this.sessionManager.selectSession(sessionId);
+    return session;
+  }
+
   setViews(views: ExplorerViews): void {
     this.views = views;
     this.syncTitles();
@@ -3559,10 +3577,11 @@ export class ExplorerController {
   // Browser "Find Class…"), then cascade the new panes to the chosen class:
   // select its dictionary and class-category, reveal the class row, and open its
   // definition. An explicit `name` arg (programmatic callers) skips the picker.
-  async findClass(name?: string): Promise<void> {
+  async findClass(name?: string, sessionId?: number): Promise<void> {
     // Resolve rather than require a pre-selected session: if one session is
     // logged in it's chosen automatically (a bare getSelectedSession() no-ops).
-    const session = await this.sessionManager.resolveSession();
+    // An explicit sessionId (Omni Search) pins the reveal to the result's own session.
+    const session = await this.resolveSessionFor(sessionId);
     if (!session) return;
 
     let entries: queries.ClassNameEntry[];
@@ -3607,8 +3626,8 @@ export class ExplorerController {
   // Reveal+select a dictionary row by name in the Dictionaries pane (used by Omni
   // Search). Resolves the 1-based symbol-list index from the live list, cascades
   // the panes to that dictionary, and highlights its row. Warns on an unknown name.
-  async revealDictionaryByName(name: string): Promise<void> {
-    const session = await this.sessionManager.resolveSession();
+  async revealDictionaryByName(name: string, sessionId?: number): Promise<void> {
+    const session = await this.resolveSessionFor(sessionId);
     if (!session) return;
     const names = queries.getDictionaryNames(session);
     const idx = names.indexOf(name);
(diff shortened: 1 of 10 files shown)


client/src/gemstoneExplorer.ts
```diff
@@ -750,6 +750,24 @@ export class ExplorerController {
...
+  private async resolveSessionFor(sessionId?: number): Promise<ActiveSession | undefined> {
+    if (sessionId === undefined) return this.sessionManager.resolveSession();
+    const session = this.sessionManager.getSession(sessionId);
+    if (!session) {
+      void vscode.window.showWarningMessage(
+        `That result's GemStone session (${sessionId}) is no longer active.`,
+      );
+      return undefined;
+    }
+    this.sessionManager.selectSession(sessionId);
```
MatiasFernandez: 💭 `resolveSessionFor` calls `this.sessionManager.selectSession(sessionId)` unconditionally whenever a `sessionId` is passed, even when it already matches the currently-selected session — which is the common case (searching and clicking a result in the session you're already in). `selectSession` always fires `onDidChangeSelection` (sessionManager.ts:81-87) regardless of whether the id actually changed, and that event drives real work downstream: `gemstoneSymbolProvider` invalidates its cache, `gemstoneManager.markStale()` flags the admin panel stale, `sunitTestController` resyncs, several tree views refresh, etc.

So every Omni Search reveal in the single-session case (by far the most common) would now trigger a full cache-invalidation/refresh cycle it didn't before, even though nothing changed. Should this short-circuit when `sessionId === this.sessionManager.getSelectedSession()?.id`, similar to how `OmniSearchPanel.show()` in this same PR compares session ids before deciding whether to switch?

MatiasFernandez (approved):

client/src/gemstoneExplorer.ts
ericwinger: Fixed in `4efffc36` — good catch, and the downstream cost you traced is the reason.

`resolveSessionFor` now compares first and only switches on a genuine change:

```ts
if (this.sessionManager.getSelectedSession()?.id !== sessionId) {
  this.sessionManager.selectSession(sessionId);
}
```

So the common single-session reveal no longer fires `onDidChangeSelection` and no longer drags the symbol-cache invalidation, `markStale()`, the sunit resync and the tree refreshes behind it. And yes — this now mirrors `OmniSearchPanel.show()`'s id compare in this same PR, which was the inconsistency worth removing on its own.

Covered by a new test in `explorerOmniReveal.test.ts` — *"does not re-select when the result already came from the active session"* asserts `selectSession` is never called, alongside the existing case that asserts it *is* called for a cross-session reveal, so neither direction can regress silently.

PR #465, opened 2026-08-19
GemStone Search: reference/pivot ordering, camelCase matching, and keystroke debounce

### LINKED ISSUE #460: GemStone Search: a slow reference load or pivot can overwrite a newer one, so a visible reference can open the wrong method
## Summary

GemStone Search asks the stone for a result's references (senders of a selector, references to a class)
and writes the answer into shared state when it comes back. Neither of the two paths that do this checks
whether a newer request has superseded it while it was waiting, so answers that arrive out of order can
overwrite each other. The visible consequences are a reference row that opens a method other than the one
it names, and a search view that jumps back to a reference list the user already left.

Found by @npapagna in a `/code-review` pass over the whole GemStone Search feature, reported as findings 1
and 2 of https://github.com/GemTalk/Jasper/pull/451#issuecomment-5345768636. They are unrelated to that
PR's diff. Confirmed still present on `main`.

## Environment

- **Jasper:** 1.8.12
- **GemStone server:** N/A — client-side ordering of asynchronous replies; no server behavior involved.

## Steps to reproduce

**A — the preview list can open the wrong method** (`referencesFor`, `client/src/omniSearch/omniEngine.ts`)

1. Open GemStone Search and search for something that returns several rows.
2. With the references-in-preview pane active, move the active row from one result to another quickly —
   fast enough that the first row's reference query is still running when the second row's starts.
   The effect needs the first (slower) query to resolve *after* the second, so it is easiest to see when
   the first row is a widely-sent selector and the second is a rare one.
3. Click a row in the reference list now on screen.

**B — the view jumps back to an abandoned reference list** (`pivot`, same file)

1. Pivot a result to its references — a common selector, so the query takes a noticeable moment.
2. Before it comes back, type a new search term. The new search draws its results correctly.
3. Wait for the original pivot to finish.

## Expected behavior

- **A:** the row that is clicked opens the method it names.
- **B:** the view stays on the search the user asked for. A reference view that was abandoned while it was
  loading never appears.

## Actual behavior

- **A:** `referencesFor` writes into a single shared `referenceRows` array with no ordering check, so the
  slower first query lands last and replaces the second row's rows. The list on screen is still the second
  row's, so the row ids the user can see no longer match the array `referenceResultFor` looks them up in —
  clicking opens one of the *first* row's references instead. There is no error or warning.
- **B:** `pivot` bumps the shared `generation` counter to invalidate older in-flight work, but never
  captures or re-checks a token of its own, so on resolving it unconditionally overwrites the pivot and
  current-results state. The view snaps to the reference list the user had moved on from.

## Potential fix

`runSearch` in the same file already has the pattern this needs: capture a generation token before the
`await`, and discard the result if the token no longer matches when it resolves. One approach would be to
apply the same shape to both paths.

Worth noting one wrinkle if that route is taken: `referencesFor` probably should not share the `generation`
counter with the search. Capturing it read-only would not fix case A, since two successive reference loads
would capture the same value and neither would invalidate the other. Bumping it would fix the ordering but
break `referencesFor`'s documented contract of leaving the search untouched — an in-flight search would
then be cancelled by a preview load and the result list would sit stale until the next keystroke. A
counter of its own for reference loads would avoid both, at the cost of one more piece of state to explain.

A regression test for each case is straightforward at the engine level: the engine takes an injected
reference resolver, so a test can hold the first call open, let the second complete, then release the first
and assert the stale one is discarded and the visible rows still resolve to themselves. A third test
asserting that a reference load and an in-flight search do not cancel each other would pin the counter
decision above, whichever way it goes.

### LINKED ISSUE #462: GemStone Search: case-insensitive matching (the default) can never prefer a camelCase boundary, so ranking picks the wrong occurrence
## Summary

`match()` lowercases both the query and the target before choosing which occurrence of the query to
match. The occurrence-picking code then asks whether a candidate position begins a word, and part of
that test looks for a camelCase hump — a lowercase or digit followed by an uppercase letter. Once the
target has been lowercased there are no uppercase letters left, so that branch can never fire. With
the default `gemstone.omniSearch.caseSensitive: false`, the camelCase half of the word-boundary
preference is dead code, and search results are ranked off a worse choice of occurrence than intended.

Found by @npapagna in a `/code-review` pass over the whole GemStone Search feature, reported as
finding 8 of https://github.com/GemTalk/Jasper/pull/451#issuecomment-5345768636. Confirmed on `main`.

## Environment

- **Jasper:** 1.8.12
- **GemStone server:** N/A — pure client-side ranking; no server behavior involved.

## Steps to reproduce

1. Leave `gemstone.omniSearch.caseSensitive` at its default (`false`).
2. Search GemStone Search for `co` against a corpus containing a name shaped like `xcoreFooCore` —
   i.e. an early lowercase "co" that is *not* at a word boundary, and a later "Co" that starts a
   camelCase hump.
3. Look at which characters are highlighted as the match, and where the row ranks against its
   neighbours.

## Expected behavior

The later `Co` — the one that begins a camelCase word — is preferred as the match, the same way it
would be with `caseSensitive: true`. Turning case sensitivity off should change what *counts as equal*,
not silently switch off the word-boundary preference.

## Actual behavior

The early, mid-word `co` is chosen instead, and the row is ranked and highlighted accordingly.

The mechanism, all in `client/src/omniSearch/omniMatch.ts`:

- `match()` folds the target with `fold(target, opts.caseSensitive)` into `t`, then hands `t` to the
  matchers.
- `contiguousIndices(q, t)` walks the occurrences of `q` in `t` and returns the first one where
  `isWordStart(t, at)` is true, falling back to the leftmost occurrence otherwise.
- `isWordStart` has two ways to say yes: the previous character is a separator, or `/[a-z0-9]/` on the
  previous character and `/[A-Z]/` on the current one. On a lowercased `t` the second test can never
  pass, so only separator boundaries survive.

`scoreFrom` is later called with the ORIGINAL-case `target`, so scoring itself still sees the humps —
but by then `contiguousIndices` has already decided *which* occurrence is being scored, blind to case.
That is why the effect shows up as a wrong choice among several matches rather than as an obviously
wrong score.

## Potential fix

One approach would be to have the word-boundary check consult the original-case target even while the
matching itself runs on the folded strings — e.g. passing `target` alongside `t` into
`contiguousIndices` and using it for the `isWordStart` call only. The greedy subsequence fallback in
`matchFuzzy` has the same blindness and may be worth the same treatment.

Worth expecting some churn in the existing `omniMatch` / `rank` test expectations: this changes which
occurrence wins in the default mode, which is the point, but it does mean the ranking of real result
lists shifts. Probably worth eyeballing a few live searches, not just re-greening the unit tests.

### LINKED ISSUE #463: GemStone Search fires a full search on every keystroke, and the debounceMs setting meant to prevent it is not read by anything
## Summary

Typing in GemStone Search posts a query to the search engine on every single `input` event, with no
coalescing, so a ten-character term fans out ten complete provider searches — each one a set of GCI
round-trips to the stone — instead of roughly one after the typing settles.

There is a setting for exactly this, `gemstone.omniSearch.debounceMs` (default 120 ms, documented in
the manifest as "Delay after a keystroke before searching"), and it is still read and range-clamped
into the config object. But **no production code consumes it.** The only consumer was the Quick Pick
controller, which was removed along with that UI, so today the setting is inert: changing it has no
effect whatsoever.

The two together mean there is **no way to mitigate this from the user side**: someone who notices the
lag and turns `debounceMs` up gets no change, because nothing is listening.

Found by @npapagna in a `/code-review` pass over the whole GemStone Search feature, reported as
finding 9 of https://github.com/GemTalk/Jasper/pull/451#issuecomment-5345768636. **That report
understates it**, because it was written against an older branch that still had the Quick Pick: there
it reads as an inconsistency between UI surfaces, where one debounces and the others do not, fixable by
copying the working one. On `main` the Quick Pick is gone, so no surface debounces at all, the setting
that documents the behavior has no consumer left, and there is nothing to copy from.

## Environment

- **Jasper:** 1.8.12
- **GemStone server:** N/A for the defect itself, though the cost lands on the stone — each dropped
  keystroke is real provider queries against a live session.

## Steps to reproduce

1. Open GemStone Search (either the docked panel or the tab).
2. Type a ten-character term at a normal speed.
3. Observe the searching: ten separate fan-outs rather than one settled result. Most visible on a large
   image, or with a heavier scope enabled, where each fan-out takes long enough to see.
4. Separately: set `gemstone.omniSearch.debounceMs` to something extreme, e.g. 2000, and repeat.
   Nothing changes.

## Expected behavior

- Keystrokes are coalesced, so a burst of typing produces roughly one search rather than one per
  character.
- `gemstone.omniSearch.debounceMs` actually governs that delay, matching what the setting's description
  promises.

## Actual behavior

- `client/src/omniSearch/omniSearchView.js` posts `query` directly from the `input` listener, with no
  timer around it. The host's dispatch calls `engine.search()` immediately.
- `gemstone.omniSearch.debounceMs` is defined in the manifest, typed in `omniTypes.ts`, and parsed and
  clamped in `omniConfig.ts` — and referenced nowhere else outside tests. The clamping tests still pass,
  so the suite reports the setting as working; nothing asserts that anything acts on the value.
- A comment right next to the `input` handler still refers to "the debounced reply", describing
  behavior that no longer exists on this path. Worth fixing in the same change so the next reader is
  not misled.

## Potential fix

One direction would be a timer around the `query` post in the webview (or around the `query` case in
the host's dispatch), keyed off the `debounceMs` value that is already delivered to the webview in the
config message — which would need adding to that payload, since it enumerates its fields explicitly
rather than forwarding the whole config.

The alternative, if per-keystroke searching is considered acceptable, would be to **remove** the
setting rather than leave it advertising behavior it does not deliver. Either resolution is better than
the current state; a setting that silently does nothing is its own bug.

Two things worth checking alongside: the busy indicator is currently set on every keystroke, so its
timing may want to move with the debounce; and reducing overlapping in-flight searches also reduces
exposure to the out-of-order result races tracked separately.

ericwinger:
## What & why

Three independent defects in GemStone Search, all from @npapagna's `/code-review` pass over the
feature, recorded as a plain comment on #451 (findings 1, 2, 8 and 9 of
https://github.com/GemTalk/Jasper/pull/451#issuecomment-5345768636). None of them were in that PR's
diff, so they were filed separately and are fixed here. Findings 3, 4 and 5 from the same review are
being handled in parallel; findings 6 and 7 were already fixed by #445 and are answered on #451.

One commit per issue, so each can be read or reverted on its own.

## Changes

**`963d806e` — a stale reference load or pivot can no longer overwrite a newer one (#460)**

Two paths asked the stone for a result's references and wrote the answer into shared state without
checking whether a newer request had superseded them. `runSearch` has always had this guard; these two
never took a token.

- `referencesFor` gets its **own** counter, `referenceGeneration`. It deliberately does not share
  `generation`: capturing that read-only would not fix the bug (two successive reference loads capture
  the same value, so neither invalidates the other), and bumping it would break this call's contract of
  leaving the search intact. Before: arrowing from row A to row B while A was still resolving left B's
  list on screen over A's rows, so clicking a visible reference opened one of A's — silently.
- `pivot` already bumped `generation` but never re-checked its own token, so a slow pivot resolving
  after the user typed a new term snapped the view back to the abandoned pivot.

**`94bd5f56` — case-insensitive matching can prefer a camelCase boundary again (#462)**

`match` folds the target before picking which occurrence to match, and the picker asked the *folded*
string whether a position begins a word. Half of that test is the camelCase-hump check, which cannot
fire once every capital is gone — so with the default `caseSensitive: false` the preference silently
collapsed to separators only. `contiguousIndices`/`matchFuzzy` now judge boundaries on the original-case
target while still matching on the folded one. Searching `co` in `xcoreFooCore` now picks the `Co` of
`Core`, agreeing with what `caseSensitive: true` always did.

**`5e45ecf3` — keystrokes are coalesced, and `debounceMs` does something again (#463)**

Typing posted a query on every `input` event, so a ten-character term fanned out ten full provider
searches. `gemstone.omniSearch.debounceMs` was still parsed and clamped but read by nothing — its only
consumer was the Quick Pick controller, removed with that UI — so turning the knob up did nothing and
there was no user-side mitigation. Wired up rather than removed, because DESIGN.md asserts debounced
behaviour in three places, including the matcher benchmark whose conclusion is argued *from* a 120 ms
debounce.

- `configMessage` now carries `debounceMs` (that payload enumerates its fields, so it had to be added).
- The webview gained `sendQueryDebounced` / `cancelPendingQuery`. A delay of 0 — including before the
  first config message arrives — searches immediately, so the field is never unresponsive.
- The clear (×) button stays immediate **and** cancels a pending keystroke search, so a keystroke queued
  a moment earlier cannot land after the clear and refill the list.
- A scope change deliberately does **not** cancel a pending query: the pending value is newer than the
  host's, so letting it fire is what lands the right end state.

## Testing

Client-only TypeScript/JS change — no server, GemStone or refactoring-engine surface. Client gate green:
`lint` / `format:check` / `compile`, and the full suite (**client 5952 passed / 12 skipped, server 322,
mcp-server 92**), with the live-stone integration tests running rather than skipping.

**14 new tests.** Every one was checked against reverted source, so none of them is decorative:

| Tests | Reproduce the defect | Pin behaviour that must not change |
|---|---|---|
| #460 — 4 | 3 (stale reference load; pivot superseded by a search; pivot superseded by a second pivot) | 1 — a reference load and an in-flight search do not cancel each other, which fails if the two counters are ever collapsed |
| #462 — 5 | 2 (the camelCase occurrence is chosen; folded result agrees with the case-sensitive one) | 3 — separator boundaries, the leftmost fallback, and relative ranking |
| #463 — 5 | 4 (burst coalesces to one search; a keystroke restarts the window; clear supersedes a pending search; a nonsense value is not adopted) | 1 — the immediate/uncoalesced path for a 0 delay and a config-less field |

## DESIGN.md

No change needed, which is worth stating explicitly because it looks like it should need one. DESIGN.md
already describes debounced search in three places — the Spotlight comparison, the Methods provider, and
the matcher benchmark's "behind a 120 ms debounce" — and describes the word-start weights as tuned for
camelCase identifiers. All four were written as the intended design; it was the CODE that had drifted away
from them. These commits restore the behaviour the doc already documents, so every one of those lines is
accurate again rather than needing a rewrite.

Closes #460
Closes #462
Closes #463

### DIFF
--- a/client/src/omniSearch/__tests__/omniEngine.test.ts
+++ b/client/src/omniSearch/__tests__/omniEngine.test.ts
@@ -817,6 +817,32 @@ describe('createOmniEngine — out-of-order reference and pivot results', () => 
     expect(engine.referenceResultFor(fastB!.rows[0].id)!.label).toBe('Y>>usesBeta');
   });
 
+  it('a new search supersedes a reference load the previous term left in flight', async () => {
+    let releaseRefs: (() => void) | undefined;
+    const gate = new Promise<void>((res) => (releaseRefs = res));
+    const engine = createOmniEngine({
+      providers: [fakeProvider('classes', [classResult('Alpha'), classResult('Beta')])],
+      config: cfg(),
+      resolveReferences: async () => {
+        await gate; // still resolving when the next search lands
+        return {
+          title: 'References to Alpha',
+          target: 'Alpha',
+          results: [methodResult('X>>usesAlpha', 'usesAlpha')],
+        };
+      },
+    });
+    const search = await engine.search('a');
+    const idA = search!.rows.find((r) => r.label === 'Alpha')!.id;
+
+    const slowRefs = engine.referencesFor(idA);
+    const next = await engine.search('al'); // a new term starts while the reference load is mid-flight
+    releaseRefs!();
+
+    expect(await slowRefs).toBeNull(); // superseded by the search — its rows never overwrite the list
+    expect(next!.rows.map((r) => r.label).sort()).toEqual(['Alpha', 'Beta']); // the search landed intact
+  });
+
   it('a reference load neither cancels nor is cancelled by an in-flight search', async () => {
     let releaseSearch: (() => void) | undefined;
     const gate = new Promise<void>((res) => (releaseSearch = res));
@@ -846,7 +872,9 @@ describe('createOmniEngine — out-of-order reference and pivot results', () => 
     releaseSearch!();
 
     expect(preview!.rows.map((r) => r.label)).toEqual(['A>>useFoo']); // not cancelled by the search
-    expect(await searchP).not.toBeNull(); // and it did not cancel the search either
+    // ...and it did not cancel the search either: the search's own rows land intact, not clobbered by
+    // the reference load's row.
+    expect((await searchP)!.rows.map((r) => r.label)).toEqual(['Foo']);
   });
 
   it('supersedes an earlier pivot when a second pivot is started', async () => {
--- a/client/src/omniSearch/__tests__/omniEngine.test.ts
+++ b/client/src/omniSearch/__tests__/omniEngine.test.ts
@@ -774,3 +774,134 @@ describe('createOmniEngine — scope vs. the references pivot', () => {
     expect(back!.rows.map((r) => r.label).sort()).toEqual(['Bar>>foo', 'Foo']);
   });
 });
+
+describe('createOmniEngine — out-of-order reference and pivot results', () => {
+  it("supersedes a slow reference load so its rows cannot overwrite a newer row's", async () => {
+    const rowA = classResult('Alpha');
+    const rowB = classResult('Beta');
+    const refsForA: ReferenceView = {
+      title: 'References to Alpha',
+      target: 'Alpha',
+      results: [methodResult('X>>usesAlpha', 'usesAlpha')],
+    };
+    const refsForB: ReferenceView = {
+      title: 'References to Beta',
+      target: 'Beta',
+      results: [methodResult('Y>>usesBeta', 'usesBeta')],
+    };
+    let releaseA: (() => void) | undefined;
+    const gate = new Promise<void>((res) => (releaseA = res));
+    const engine = createOmniEngine({
+      providers: [fakeProvider('classes', [rowA, rowB])],
+      config: cfg(),
+      // Alpha is the slow one and resolves LAST, exactly the order that used to corrupt the list.
+      resolveReferences: async (r) => {
+        if (r.label === 'Alpha') {
+          await gate;
+          return refsForA;
+        }
+        return refsForB;
+      },
+    });
+    const search = await engine.search('a');
+    const idA = search!.rows.find((r) => r.label === 'Alpha')!.id;
+    const idB = search!.rows.find((r) => r.label === 'Beta')!.id;
+
+    const slowA = engine.referencesFor(idA);
+    const fastB = await engine.referencesFor(idB); // arrowed on to Beta while Alpha still resolves
+    releaseA!();
+
+    expect(await slowA).toBeNull(); // stale load discarded rather than written
+    expect(fastB!.rows.map((r) => r.label)).toEqual(['Y>>usesBeta']);
+    // The rows behind the ids the UI is showing are still Beta's — opening one cannot land in Alpha's.
+    expect(engine.referenceResultFor(fastB!.rows[0].id)!.label).toBe('Y>>usesBeta');
+  });
+
+  it('a reference load neither cancels nor is cancelled by an in-flight search', async () => {
+    let releaseSearch: (() => void) | undefined;
+    const gate = new Promise<void>((res) => (releaseSearch = res));
+    let call = 0;
+    const slow: OmniProvider = {
+      category: CATEGORY_BY_ID.classes,
+      async search() {
+        call++;
+        if (call === 2) await gate; // the SECOND search blocks, with a reference load landing meanwhile
+        return [classResult('Foo')];
+      },
+    };
+    const refView: ReferenceView = {
+      title: 'References to Foo',
+      target: 'Foo',
+      results: [methodResult('A>>useFoo', 'useFoo')],
+    };
+    const engine = createOmniEngine({
+      providers: [slow],
+      config: cfg(),
+      resolveReferences: () => refView,
+    });
+    const first = await engine.search('fo');
+
+    const searchP = engine.search('foo');
+    const preview = await engine.referencesFor(first!.rows[0].id);
+    releaseSearch!();
+
+    expect(preview!.rows.map((r) => r.label)).toEqual(['A>>useFoo']); // not cancelled by the search
+    expect(await searchP).not.toBeNull(); // and it did not cancel the search either
+  });
+
+  it('supersedes an earlier pivot when a second pivot is started', async () => {
+    let releaseAlpha: (() => void) | undefined;
+    const gate = new Promise<void>((res) => (releaseAlpha = res));
+    const engine = createOmniEngine({
+      providers: [fakeProvider('classes', [classResult('Alpha'), classResult('Beta')])],
+      config: cfg(),
+      // Alpha resolves LAST — the ordering that used to let the abandoned pivot win.
+      resolveReferences: async (r) => {
+        if (r.label === 'Alpha') {
+          await gate;
+          return {
+            title: 'References to Alpha',
+            results: [methodResult('X>>usesAlpha', 'usesAlpha')],
+          };
+        }
+        return { title: 'References to Beta', results: [methodResult('Y>>usesBeta', 'usesBeta')] };
+      },
+    });
+    const search = await engine.search('a');
+    const idAlpha = search!.rows.find((r) => r.label === 'Alpha')!.id;
+    const idBeta = search!.rows.find((r) => r.label === 'Beta')!.id;
+
+    const slowPivot = engine.pivot(idAlpha);
+    const fastPivot = await engine.pivot(idBeta); // pivoted somewhere else before Alpha came back
+    releaseAlpha!();
+
+    expect(await slowPivot).toBeNull(); // stale pivot discarded
+    expect(fastPivot!.pivotTitle).toBe('References to Beta');
+    // The live list is still Beta's, so activating a row cannot open one of Alpha's references.
+    expect(engine.resultFor(fastPivot!.rows[0].id)!.label).toBe('Y>>usesBeta');
+    expect(engine.state().pivot).toBe(true);
+  });
+
+  it('supersedes a slow pivot so it cannot snap the view back to an abandoned reference view', async () => {
+    let releasePivot: (() => void) | undefined;
+    const gate = new Promise<void>((res) => (releasePivot = res));
+    const engine = createOmniEngine({
+      providers: [fakeProvider('classes', [classResult('Foo')])],
+      config: cfg(),
+      resolveReferences: async () => {
+        await gate;
+        return { title: 'References to Foo', results: [methodResult('A>>useFoo', 'useFoo')] };
+      },
+    });
+    const search = await engine.search('foo');
+
+    const pivotP = engine.pivot(search!.rows[0].id);
+    const retyped = await engine.search('fo'); // a new term while the pivot is still resolving
+    releasePivot!();
+
+    expect(await pivotP).toBeNull(); // stale pivot discarded
+    expect(retyped!.pivot).toBe(false);
+    expect(engine.state().pivot).toBe(false); // the view stayed on the search the user asked for
+    expect(engine.resultFor(retyped!.rows[0].id)!.label).toBe('Foo');
+  });
+});
--- a/client/src/omniSearch/__tests__/omniSearchView.test.ts
+++ b/client/src/omniSearch/__tests__/omniSearchView.test.ts
@@ -987,3 +987,106 @@ describe('Omni Search view — scopes skipped under All', () => {
     expect(hint().textContent).toBe('');
   });
 });
+
+describe('search field debounce', () => {
+  /** Mount, then push a host config carrying `debounceMs` (the only way the webview learns it). */
+  function mountWithDebounce(debounceMs: number) {
+    const m = mount();
+    m.handle.onMessage({
+      data: { command: 'config', categories: [], scopeId: null, debounceMs },
+    });
+    m.vscode.postMessage.mockClear();
+    return m;
+  }
+
+  function type(text: string) {
+    const input = document.getElementById('query') as HTMLInputElement;
+    input.value = text;
+    input.dispatchEvent(new Event('input', { bubbles: true }));
+  }
+
+  function queries(vscode: { postMessage: ReturnType<typeof vi.fn> }) {
+    return vscode.postMessage.mock.calls
+      .map((c) => c[0] as { command: string; value?: string })
+      .filter((m) => m.command === 'query')
+      .map((m) => m.value);
+  }
+
+  it('coalesces a burst of keystrokes into ONE search, carrying the final text', () => {
+    vi.useFakeTimers();
+    try {
+      const { vscode } = mountWithDebounce(120);
+      type('f');
+      type('fo');
+      type('foo');
+      expect(queries(vscode)).toEqual([]); // nothing sent while the user is still typing
+
+      vi.advanceTimersByTime(120);
+      expect(queries(vscode)).toEqual(['foo']); // one search, the finished term
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
+  it('waits for quiet — a keystroke inside the window restarts the delay', () => {
+    vi.useFakeTimers();
+    try {
+      const { vscode } = mountWithDebounce(120);
+      type('f');
+      vi.advanceTimersByTime(100); // not yet
+      type('fo');
+      vi.advanceTimersByTime(100); // still inside the restarted window
+      expect(queries(vscode)).toEqual([]);
+
+      vi.advanceTimersByTime(20);
+      expect(queries(vscode)).toEqual(['fo']);
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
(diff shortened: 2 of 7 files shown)


ericwinger:
Correction to my own PR body: the "Known gap" section about `DESIGN.md` was wrong and has been replaced. I had it backwards — those four lines (debounced search in three places, camelCase-tuned weights) were never stale prose describing removed behaviour. They describe the *intended* design, and it was the code that had drifted off them. These commits restore the documented behaviour, so the lines are accurate as they stand and `DESIGN.md` needs no edit here. Nothing outstanding, and no follow-up issue.

client/src/omniSearch/omniEngine.ts
```diff
@@ -573,7 +586,12 @@ export function createOmniEngine(deps: OmniEngineDeps): OmniEngine {
       if (!deps.resolveReferences) return null;
       const result = current[rowId];
       if (!result) return null;
+      // `referenceRows` is the single array every preview row id indexes into, so a slow load must not
+      // land after a newer one: arrowing from row A to row B while A is still resolving would leave B's
+      // list on screen over A's rows, and `referenceResultFor` would then open the wrong method.
+      const gen = ++referenceGeneration;
       const view = await deps.resolveReferences(result);
+      if (gen !== referenceGeneration) return null; // a newer reference load superseded this one
```
MatiasFernandez: 💭 This closes reference-load-vs-reference-load, but a search still never supersedes an in-flight reference load — only `referencesFor` touches `referenceGeneration`. Scenario: arrow onto row 3, hit the references key, the resolve is slow; the user types a new term and the new search lands and re-renders the list. Row ids are array indices, so the late reply comes back with `forId: 3` and the webview guard (`msg.forId === activeRowId()`, omniSearchView.js:640) matches whatever row 3 is *now* — the preview pane flips to a references list belonging to a row from the previous search, and `referenceRows` is overwritten underneath it, so `referenceResultFor` opens from that stale list.

Would bumping `referenceGeneration` at the top of `runSearch` cover this without giving up the contract you're deliberately pinning? A reference load started *after* the search still takes the newer token, so "a reference load neither cancels nor is cancelled by an in-flight search" keeps passing — what becomes invalid is only a load that was already in flight when a new search started, which is exactly the case above. What do you think?

client/src/omniSearch/omniSearchShared.ts
```diff
@@ -131,6 +131,10 @@ export function configMessage(config: OmniConfig, pinned: boolean): Record<strin
     previewPane: config.previewPane,
     excludedFromAll: config.excludedFromAll,
     matchMode: config.matchMode,
+    // How long the field waits after a keystroke before searching. The webview owns the timer (it is
+    // the only side that sees keystrokes); the host just forwards the configured value, so a settings
+    // edit takes effect on the next config push like every other field here.
+    debounceMs: config.debounceMs,
```
MatiasFernandez: 💭 This is the one production line that closes the loop, and I don't think anything covers it: the webview tests hand-roll the payload (`{ command: 'config', categories: [], scopeId: null, debounceMs }`), and I couldn't find any test asserting what `configMessage` actually returns. If this line were dropped, `debounceMs` would stay 0 in the webview — every keystroke fanning out again, the exact bug this commit fixes — and all 14 new tests would still be green.

Since the whole point of the commit is that the setting has a consumer again, is it worth a small assertion that `configMessage(config, false).debounceMs` is the configured value?

client/src/omniSearch/__tests__/omniEngine.test.ts
```diff
@@ -774,3 +774,134 @@ describe('createOmniEngine — scope vs. the references pivot', () => {
...
+      resolveReferences: () => refView,
+    });
+    const first = await engine.search('fo');
+
+    const searchP = engine.search('foo');
+    const preview = await engine.referencesFor(first!.rows[0].id);
+    releaseSearch!();
+
+    expect(preview!.rows.map((r) => r.label)).toEqual(['A>>useFoo']); // not cancelled by the search
+    expect(await searchP).not.toBeNull(); // and it did not cancel the search either
```
MatiasFernandez: 💭 `not.toBeNull()` is a fairly loose way to say "the search survived" — it passes as long as *something* came back, even if the reference load had clobbered the search's rows (which is the failure this test exists to catch). Should this assert what the search actually returned, the way the sibling tests do — e.g. the row labels, or `engine.resultFor(view!.rows[0].id)!.label`?

client/src/omniSearch/omniMatch.ts
```diff
@@ -111,12 +111,18 @@ function matchSubstring(q: string, t: string): number[] | null {
...
  * scattered match that happens to start at index 0 outrank an exact mid-string one.
+ *
+ * `t` is the case-folded target (what we search in), `cased` the original (what we judge boundaries in).
+ * They must be separate: half of `isWordStart` is the camelCase hump test, which needs a real uppercase
+ * letter to fire, and in the DEFAULT case-insensitive mode `t` has none left. Asking `t` about word
+ * starts silently reduced the preference to separators only and picked a mid-word occurrence over a
+ * later camelCase one — the majority of searches, since case-insensitive is the default.
  */
-function contiguousIndices(q: string, t: string): number[] | null {
+function contiguousIndices(q: string, t: string, cased: string): number[] | null {
```
MatiasFernandez: 💭 Two adjacent same-typed string params where swapping them is silent in the default case-sensitive tests feels like the shape of the bug you just fixed. Would `folded` / `original` (or `folded` / `cased`) read more clearly than `t` / `cased`? `t` is the one that no longer says what it holds, and it's the one whose misuse caused #462 — a name that says "folded" makes `isWordStart(folded, at)` look wrong on sight at the call site.

client/src/omniSearch/__tests__/omniSearchView.test.ts
```diff
@@ -987,3 +987,106 @@ describe('Omni Search view — scopes skipped under All', () => {
...
+      type('foo');
+      (document.getElementById('clear') as HTMLButtonElement).click();
+      expect(queries(vscode)).toEqual(['']); // immediate, not waiting out the timer
+
+      vi.advanceTimersByTime(500);
+      expect(queries(vscode)).toEqual(['']); // and 'foo' never lands after the clear
+    } finally {
+      vi.useRealTimers();
+    }
+  });
```
MatiasFernandez: 💭 The PR description calls out that a scope change deliberately does NOT cancel a pending query (the pending value is newer than the host's, so letting it fire lands the right end state) — that's a subtle decision sitting right next to the clear button, which does the opposite. Worth pinning it in a test here? Otherwise the next person reading `cancelPendingQuery` is quite likely to "fix" the scope path for symmetry, and nothing would fail.

client/src/omniSearch/omniSearchView.js
```diff
@@ -989,7 +1022,7 @@
       // debounced reply would leave it a keystroke behind.
       updateScopeHint();
       rehighlightSourcePreview(); // clear/refresh the stale blue match marks now, not on the fetch
-      post('query', { value: inputEl.value });
+      sendQueryDebounced(inputEl.value);
```
MatiasFernandez: 💭 `setBusy(true)` a few lines up still fires on every keystroke, but now nothing is in flight until the timer expires — with the setting near its 2000 ms ceiling the panel reads as busy for two seconds with no stone work happening. Is that the behaviour you want (busy = "the list is stale, results are coming"), or should the busy state move into `sendQueryDebounced` alongside the `post`?

MatiasFernandez (approved):

client/src/omniSearch/omniEngine.ts
ericwinger: You're right, and the fix is the one you proposed — `11f8360d` bumps `referenceGeneration` at the top of `runSearch`, just past the pivot early-return.

Your scenario was the real hole: array-index row ids mean the late reply's `forId: 3` matches whatever row 3 became, so the webview guard can't catch it, the preview flips to references for a row from the previous search, and `referenceRows` gets overwritten so `referenceResultFor` then opens from that stale list.

And the contract survives, exactly as you predicted: a reference load started *after* the search takes a newer token, so "a reference load neither cancels nor is cancelled by an in-flight search" still passes untouched. What's now invalid is only a load already in flight when a new search starts.

New test — *"a new search supersedes a reference load the previous term left in flight"* — gates the superseded load returning `null` **and** the new search's rows landing intact, plus a comment on `referencesFor` pointing at the `runSearch` bump so the two generation sites read as one mechanism.

client/src/omniSearch/__tests__/omniEngine.test.ts
ericwinger: Fair — `not.toBeNull()` couldn't fail for the reason the test exists. Tightened in `11f8360d` to assert what the search actually returned:

```ts
expect((await searchP)!.rows.map((r) => r.label)).toEqual(['Foo']);
```

Now it catches the clobbering case: if the reference load's row had overwritten the search's, this compares `['A>>useFoo']` against `['Foo']` and fails, where before it passed on any non-null result. Matches how the sibling tests assert, as you suggested.

client/src/omniSearch/omniMatch.ts
ericwinger: Renamed in `11f8360d` — `contiguousIndices` and `matchFuzzy` now take `(q, folded, original)`.

Your read of *why* is the part I'd have missed: `t` had stopped saying what it holds, and it's the argument whose misuse caused the bug. With the new names, `isWordStart(folded, at)` looks wrong on sight at the call site, which is the only real defence against a silent swap between two same-typed strings that the default case-insensitive tests can't distinguish. The doc comment above `contiguousIndices` now spells out which string is which and why word starts must be judged on `original`.

client/src/omniSearch/omniSearchView.js
ericwinger: Moved, in `11f8360d` — `setBusy(true)` now lives inside `sendQueryDebounced` next to each `post`, on both the immediate `debounceMs === 0` path and inside the timer callback.

Your second reading is the right one: busy should mean "a query is out to the host," not "you touched the keyboard." The old placement was visibly wrong near the 2000 ms ceiling — two seconds of spinner with nothing in flight. The doc comment now states the invariant, so a future edit doesn't drift it back into the keystroke handler.

PR #473, opened 2026-08-20
Add a Windows leg to the Health Check matrix

npapagna:
## Summary
- Adds a `windows-latest` leg to the `health-check` job matrix, alongside the existing `ubuntu-latest` one, so the GemStone integration suite also runs against the native Windows GCI client.
- Windows has no native GemStone server build, so the new leg provisions WSL via a new composite action (`setup-gemstone-in-windows`), installs and starts the server inside the guest, downloads the native Windows GCI client, and points the native Windows test run at the WSL-hosted server remotely.
- `gs-create-test-env-file.sh` gains `--remote`, `--gci-library-path`, and `--global-dir` so the same script can write a `.env.test` that's reachable across that WSL/native-Windows boundary; `gs-test-setup.sh` just forwards extra args through to it.
- The server download cache is keyed on `gemstone-server-download-linux-<version>` (no `runner.os`), since the Windows leg's server install also runs as Linux under WSL and can share the same cached archive; a future macOS leg would need its own key, since `gs-config.sh` resolves a different archive there.
- The job's `timeout-minutes` is scoped per platform (5 for Linux, 10 for the Windows leg's added WSL/GCI-download overhead) instead of raised for everyone, so a Linux-side regression that doubles runtime still trips the guard.

## Test plan
- [x] `npm run lint && npm run format:check && npm run compile && npm test` locally
- [x] CI matrix green on both `ubuntu-latest` and `windows-latest` legs across the GemStone version matrix
- [x] Windows leg's WSL setup, GCI client download, and remote `.env.test` connection succeed end-to-end in CI

### DIFF
--- a/.github/workflows/health-check.yml
+++ b/.github/workflows/health-check.yml
@@ -72,6 +72,7 @@ jobs:
         with:
           sparse-checkout: |
             .github/workflows
+            .github/actions
           sparse-checkout-cone-mode: false
           # zizmor(artipacked): checkout must not persist the token past
           # this job
@@ -121,12 +122,42 @@ jobs:
         run: npm audit signatures
 
   health-check:
-    timeout-minutes: 5
-    name: GemStone S/64 ${{ matrix.gemstone-version }}${{ matrix.node-version && format(' (Node {0})', matrix.node-version) || '' }}
+    # The Windows leg needs extra headroom (WSL provisioning + GCI client
+    # download); the Linux legs still fit in 5. Scoped per-platform so a
+    # Linux-side regression that doubles the runtime still trips the guard
+    # instead of quietly passing under the Windows leg's larger budget.
+    timeout-minutes: ${{ matrix.platform == 'windows-latest' && 10 || 5 }}
+    name: ${{ matrix.platform }} ${{ matrix.gemstone-version }}${{ matrix.node-version && format(' (Node {0})', matrix.node-version) || '' }}
     needs: prepare
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.platform }}
+    # zizmor(template-injection): bind the untrusted matrix values to an env
+    # var instead of interpolating ${{ }} directly into a script
+    env:
+      # One name (for upload-artifact's `name`) and one path (for
+      # VITEST_JSON_OUTPUT and upload-artifact's `path`) per skip-report
+      # file. Both include matrix.platform, gemstone-version and
+      # node-version so the Linux and Windows legs of the same
+      # version/node combo don't produce same-named files — upload-artifact
+      # rejects duplicate names within a run, and report-skips'
+      # merge-multiple download would clobber same-named files. Paths are
+      # absolute (github.workspace-rooted): VITEST_JSON_OUTPUT is read by
+      # each npm workspace's vitest run, which cwds into client/, server/,
+      # or mcp-server/, so a relative path would land inside whichever
+      # workspace ran last instead of one shared test-results/ at the repo
+      # root.
+      #
+      # Each var spells out the matrix expression on its own rather than
+      # building on a shared base var: GitHub Actions' env context can't
+      # see sibling keys from the same env mapping (they'd silently
+      # resolve empty), so there's no legal way to factor this out further
+      # short of an extra step writing to $GITHUB_ENV.
+      SKIP_REPORT_BARE_NAME: skips-${{ matrix.platform }}-${{ matrix.gemstone-version }}-${{ matrix.node-version || 'nvmrc' }}-bare
+      SKIP_REPORT_BARE_PATH: ${{ github.workspace }}/test-results/skips-${{ matrix.platform }}-${{ matrix.gemstone-version }}-${{ matrix.node-version || 'nvmrc' }}-bare.json
+      SKIP_REPORT_PLUGIN_NAME: skips-${{ matrix.platform }}-${{ matrix.gemstone-version }}-${{ matrix.node-version || 'nvmrc' }}-plugin
+      SKIP_REPORT_PLUGIN_PATH: ${{ github.workspace }}/test-results/skips-${{ matrix.platform }}-${{ matrix.gemstone-version }}-${{ matrix.node-version || 'nvmrc' }}-plugin.json
     strategy:
       matrix:
+        platform: [ubuntu-latest, windows-latest]
         gemstone-version: ${{ fromJson(needs.prepare.outputs.versions) }}
         node-version: ['']            # empty ⇒ dev jobs resolve Node from .nvmrc (node-version-file); floor include overrides
         # node-version's declared axis value is only '' (which falls back to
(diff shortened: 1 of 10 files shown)


.github/workflows/health-check.yml
```diff
@@ -215,14 +218,31 @@ jobs:
...
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
         with:
           path: client/tmp/downloads
-          key: gemstone-download-${{ runner.os }}-${{ matrix.gemstone-version }}
-      - name: Setup test environment
+          # The Windows leg's server install also runs as Linux (inside WSL,
+          # see setup-gemstone-in-windows), so every matrix platform
+          # downloads the same archive for a given version: one shared key,
+          # no runner.os, lets either leg warm the cache for the other.
+          key: gemstone-download-${{ matrix.gemstone-version }}
```
MatiasFernandez: 💭 The comment justifies dropping `runner.os` on the premise that "every matrix platform downloads the same archive for a given version" — true today for `ubuntu-latest` + `windows-latest` (both install the Linux zip), but `gs-config.sh` also has a Darwin branch that resolves a different `FILENAME` (`GemStone64Bit<v>-arm64.Darwin.dmg`). If a `macos-latest` leg ever joins the `platform` axis, the first leg to save wins the key, the other platform re-downloads on every run, and nothing signals that caching stopped working.

Should we key on the archive's platform rather than on the assumption — e.g. `gemstone-download-linux-${{ matrix.gemstone-version }}`? That keeps the ubuntu↔WSL sharing explicit (and self-documenting) while leaving a future macOS leg on its own key.

.github/workflows/health-check.yml
```diff
@@ -121,12 +122,13 @@ jobs:
         run: npm audit signatures
 
   health-check:
-    timeout-minutes: 5
-    name: GemStone S/64 ${{ matrix.gemstone-version }}${{ matrix.node-version && format(' (Node {0})', matrix.node-version) || '' }}
+    timeout-minutes: 10
```
MatiasFernandez: 💭 This doubles the budget for the `ubuntu-latest` legs too, which were fitting in 5. Only the Windows leg needs the extra headroom (WSL provisioning + GCI client download), so should we scope it — something like `timeout-minutes: ${{ matrix.platform == 'windows-latest' && 10 || 5 }}` — so a Linux-side regression that doubles the runtime still trips the guard instead of passing quietly? `scripts/lint-workflow-timeouts.sh` only asserts the key is present, so an expression value wouldn't upset it.

client/src/__tests__/rowanLoad.test.ts
```diff
@@ -27,7 +27,7 @@ function write(rel: string, content: string): void {
   fs.writeFileSync(full, content);
 }
 
-describe('findRowanLoadSpecs', () => {
+describe('findRowanLoadSpecs', { timeout: 10000 }, () => {
```
MatiasFernandez: 💭 These two bumps (10s here, 20s on `updateGitRepo`) aren't mentioned in the PR description and don't say why they're needed. I'm guessing it's the Windows leg — git operations against the runner's filesystem being much slower than on Linux — but someone hitting a flake here later has no way to tell whether the numbers are measured or arbitrary. Could we add a one-liner explaining the origin? The rest of this file is commented exactly that way (the `GIT_*`-stripping comment right below is a good example).

Also worth a sanity check while you're here: are these slow on Windows because of real filesystem cost, or because something inside `updateGitRepo` is retrying/waiting on something that could be avoided?

client/bin/gs-test-setup-wsl.sh
```diff
@@ -0,0 +1,63 @@
...
+#                        its /mnt/... path; the server download is symlinked
+#                        in from here so this leg shares the same
+#                        actions/cache entry as the Linux legs instead of
+#                        re-downloading every run
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+USAGE_MESSAGE="Usage: $0 <version> <name> <gci-library-path> <global-dir> <shared-download-dir>"
+VERSION="${1:?$USAGE_MESSAGE}"
+NAME="${2:?$USAGE_MESSAGE}"
+GCI_LIBRARY_PATH="${3:?$USAGE_MESSAGE}"
```
MatiasFernandez: 💭 These two are named exactly like the variables `gs-config.sh` exports — and `GEMSTONE_GLOBAL_DIR` in particular is read by GemStone's own tools for locks/logs — but here they hold *native Windows* paths. It's safe today because they're never exported and `su gsuser -c` starts from a fresh environment, but if anyone later adds an `export` (or switches to `su -p` / `env ...`), the WSL-side GemStone would try to use a Windows path for its lock/log directory and the failure would be pretty baffling to debug.

Would renaming them to say whose paths they are — `WINDOWS_GCI_LIBRARY_PATH` / `WINDOWS_GLOBAL_DIR` — make that collision impossible to trip over?

client/bin/gs-info.py
```diff
@@ -0,0 +1,77 @@
...
+# on macOS, where a dev testing this locally would need something else
+# entirely (e.g. `ipconfig getifaddr en0`). Python's socket module gives one
+# implementation that works the same everywhere.
+
+import json
+import socket
+import subprocess
+import sys
+
+def netldi_ip():
```
MatiasFernandez: 💭 Two small things on this helper:

1. The name says `netldi_ip`, but nothing here is NetLDI-specific — it's this machine's own outbound address (the comment says as much). Something like `local_outbound_ip` would describe what it actually returns, and would contrast usefully with `netldi_port`, which really is NetLDI-specific.
2. `connect()` can raise `OSError` when there's no default route, and the socket then leaks with a bare traceback — noticeably worse than the friendly stderr message `netldi_port` prints for its own failure. Wrapping it in `with socket.socket(...) as probe_socket:` would at least guarantee the close; worth a matching error message too?

client/bin/gs-info.py
```diff
@@ -0,0 +1,77 @@
...
+    if not entries:
+        print(f"Could not find a running Netldi named {name} in gslist output:", file=sys.stderr)
+        print(result.stdout, file=sys.stderr)
+        print(result.stderr, file=sys.stderr)
+        sys.exit(1)
+
+    return entries[0]["Port"]
+
+def main():
+    name = sys.argv[1]
```
MatiasFernandez: 💭 Every sibling script guards its arguments (`VERSION="${1:?$USAGE_MESSAGE}"`), but this one indexes `sys.argv` directly — so invoking it with no arguments produces an `IndexError` traceback instead of the usage line the header documents. Worth a small guard for parity, e.g. `if len(sys.argv) != 2: sys.exit(f"Usage: {sys.argv[0]} <netldi-name>")`?

client/bin/gs-create-test-env-file.sh
```diff
@@ -1,38 +1,98 @@
...
-VERSION="${1:?Usage: $0 <version> <name>}"
+VERSION="${1:?$USAGE_MESSAGE}"
 # shellcheck disable=SC2034
 # This variable is read by gs-config.sh when sourced below
-NAME="${2:?Usage: $0 <version> <name>}"
+NAME="${2:?$USAGE_MESSAGE}"
+shift 2
+
+remote_flag=""
+gci_library_path_to_use=""
```
MatiasFernandez: nit: `gci_library_path_to_use` / `gemstone_global_dir_to_use` read a little awkwardly — the `_to_use` suffix doesn't say what makes them different from `gs-config.sh`'s values. Since they exist purely to override what `gs-config.sh` computed, would `gci_library_path_override` / `gemstone_global_dir_override` be clearer? The defaulting line below then reads exactly as the fallback it is: `${gci_library_path_override:-$GCI_LIBRARY_PATH}`.

MatiasFernandez (approved):

PR #474, opened 2026-08-20
Run SUnit tests from the GemStone Explorer, and let a run be stopped

### LINKED ISSUE #427: No way to run SUnit tests from the GemStone Explorer
## Summary

There is no way to run SUnit tests from the GemStone Explorer. The existing commands —
`gemstone.runSunitClass`, `runSunitClasses`, `runSunitMethods` — are dispatched only from
`systemBrowser.ts`; `gemstoneExplorer.ts` never mentions SUnit and contributes no such action.

That became a real gap with [#421](https://github.com/GemTalk/Jasper/issues/421), which removed
the buttons advertising the System Browser and pointed the documentation at the Explorer. Running
tests on a class is now the one place where "prefer the Explorer" costs a user a feature, and the
only reason left to go back to the browser for everyday work.

## Acceptance criteria

- Tests can be run from the Explorer, without opening the System Browser.
- Results are visible and readable — which tests ran, which failed, and enough of the failure to
  act on it.

## Implementation notes

Two sizes of answer, and they are not exclusive — the first could ship now and the second later.

**Smaller: a run action on a test class.** A button or context-menu action on a class row in the
Explorer that runs its tests and shows the outcome in a webview, in the spirit of the other
result panels. The server-side commands already exist and are registered in `extension.ts`, so
this is mostly a matter of contributing the menu entries to the Explorer's class (and possibly
method) rows and rendering the result. Deciding which rows offer it is worth some thought —
always-on for every class, or only where the class actually looks like a `TestCase` subclass.

**Larger: a proper SUnit Runner.** A dedicated view that discovers test classes, lets you pick a
scope to run, keeps a run history, and offers navigation from a failure to the failing method —
closer to what a test-runner UI usually provides. Whether it belongs in the Explorer's container
or its own is part of the question.

Related: `docs/explanation/system-browser-and-explorer.md` records this as the known
browser-only gap, and should be updated when it closes.

ericwinger:
Running SUnit tests was the one place where "prefer the Explorer" still cost a real
feature, and the documentation had already stopped advertising the System Browser.
This closes that, and fixes several things found while making it work.

## Running tests from the Explorer

- A test class row — in **Classes** or **Hierarchy** — and a test method row in
  **Methods** carry an inline ▶. Both dispatch the same commands the System Browser
  uses, so a run started anywhere reports in the Testing view.
- Each of those rows shows the outcome of its last run. A result that has gone stale
  (the code was recompiled since) keeps its shape but takes the queued colour, so a
  green check never quietly outlives the code it described.
- A `TestCase` subclass with no tests of its own is not offered as runnable — the run
  would do nothing. One whose count the stone could not report stays runnable, since
  "no tests found" beats a silently missing button.
- Navigation between the three places you might be looking from, so none of them is a
  dead end you have to leave by finding the class again by hand:
  - **Shift+Enter** on a GemStone Search result opens that class or method **in the
    Testing view** — selected and scrolled to, with the view brought up if it wasn't
    showing. A method result reveals down to its own test row, not just its class.
  - **Reveal in Testing View** does the same from an Explorer class or method row.
  - **Reveal in GemStone Explorer** goes the other way, from a Testing view row.
- **Clear Test Results** wipes both stores: ours, which paints the Explorer rows, and
  VS Code's run history, which paints the Testing view. Clearing only ours would leave
  the tester still showing the old verdicts. It is offered in four places, none of
  which take room from the rows they are about:
  - right-click a test class or test method row in the **Explorer** (Classes,
    Hierarchy, or Methods), under Run Tests;
  - right-click a row in the **Testing view**;
  - the **`…` overflow** on the Classes / Hierarchy / Methods pane headers;
  - **GemStone: Clear Test Results** in the Command Palette.

## Reading a run in the Test Results pane

- **A run is named after what ran** — `AiDemoBankAccountTest`,
  `AiDemoBankAccountTest>>testOverdrawIsRejected`, `2 test classes` — instead of VS
  Code's fallback `Test run at 8/20/2026, 11:52:18 AM`, which says nothing about which
  tests these were. With several entries in the history they all read alike otherwise.
  Names come from the items' own labels, so a class whose name exists in two
  dictionaries keeps its `{Dictionary}` qualifier there.
- **The class no longer reports into the run.** It used to appear in the results list
  as a sibling of its own methods, marked "Some tests failed." and timed at 0.0ms —
  reading like a fourth test that had failed. Every method already reports, and VS Code
  rolls the tree row's state up from them. The class-level roll-up is still kept in our
  own store, which is what paints the Explorer rows.
- **A re-run blanks the previous outcome first.** Re-running a failing test that fails
  again would otherwise repaint the mark it was already showing, leaving no sign the
  run happened at all.

The pane's own wording is VS Code's and not ours to change — "Test Results", "1 older
result", "The test run did not record any output" all come from the editor.

## Stopping a run

A test can run for minutes, and the blocking GCI call held the extension host for its
whole duration. That is why nothing could interrupt one — not even VS Code's own stop
button, whose cancellation event could not be delivered until the run had finished.

Runs now go through the pollable path. One press is enough: a soft break goes first,
because it lets the gem stop at a safe point and leaves the session healthy, and it
escalates to a hard break on its own rather than making you press again to discover
the test was in a tight loop. A stopped test is reported skipped, with no verdict left
behind — calling it an error would blame the test for the user's decision.

While a run is in flight the Explorer row's ▶ becomes a ■, and the Testing view's rows
gain one beside VS Code's own run icon (which an extension cannot replace).

## Fixes found along the way

These are in `nbRunner` and `gemstoneFileSystemProvider`, and affect every cancellable
call in Jasper — Execute It and a Rowan load as much as a test run:

- **A hard break sent immediately after a soft one crashes the client process.** Not an
  exception — a native fault in the GCI library. A second break arriving within 300ms
  of the first is now deferred to that mark. Found by an integration test that pressed
  both at once and took the vitest worker down with it.
- **A break left the session unusable for every later call.** Abandoning a call without
  collecting its result leaves the session reporting one in progress, so the next call
  was refused outright — which reads as "I pressed stop and now nothing runs". The
  abandoned result is drained, and a call that follows a stop waits for that to finish.
  Both the hard-break and the poll-error path (which is how a soft break usually ends)
  were affected.
- **A logout mid-call left the poll running forever**, with the progress notification
  claiming work was in flight and the caller never hearing back.
- **`initialize/release` broke the Methods pane.** `buildMethodUri` asserted a category
  was slash-free, so building a row for a method in a stock GemStone category threw,
  surfacing as a toast and taking the pane's render down with it. Categories now ride
  in the path through the same sentinel selectors have always used.

## Tests

63 tests across the branch, including 7 integration tests against a live stone. The
integration tests earned their place immediately: all three `nbRunner` bugs above were
found by them and by no unit test — the crash included. Nothing they do is committed;
the probe fixture is installed inside each test's own transaction, which the harness
aborts.

The larger "dedicated SUnit runner view" option in the issue is deliberately not built:
VS Code's Testing view covers it. Gaps we find in that view are worth their own issues.

Closes #427.


client/src/nbRunner.ts
```diff
@@ -97,6 +98,78 @@ export interface NbRunOptions {
...
  * and rejects with `NbCancelledError`.
  */
+/**
+ * The least time that may pass between a soft break and the hard break that
+ * follows it. Sending both back-to-back crashes the client process outright — a
+ * native fault in the GCI library, not an error that can be caught — so a second
+ * cancel arriving too soon is deferred rather than obeyed. Found by an
+ * integration test that pressed both at once and took the worker down with it.
+ */
+const MIN_HARD_BREAK_GAP_MS = 300;
```
MatiasFernandez: 🚫 Inserting these consts here detached `pollNbToCompletion`'s doc comment from the function: the block ending in "a second sends a hard break and rejects with `NbCancelledError`" now sits above `MIN_HARD_BREAK_GAP_MS`, so the exported function ends up with no TSdoc and the const looks like it has two. The text is also no longer accurate — a second cancel may now be *deferred* up to 300ms instead of sending the hard break straight away. Should we move that block back down onto `pollNbToCompletion` and mention the deferral in it?

client/src/nbRunner.ts
```diff
@@ -97,6 +98,78 @@ export interface NbRunOptions {
...
+  }).finally(() => {
+    draining.delete(session.id);
+  });
+
+  draining.set(session.id, done);
+  return done;
+}
+
+/** Wait for any abandoned call on this session to be collected. */
+export function awaitSessionDrain(session: ActiveSession): Promise<void> {
```
MatiasFernandez: 💭 `awaitSessionDrain` doesn't seem to have any caller — `runNbCall` reads the `draining` map directly. Is it meant to be used from somewhere (the SUnit controller before a re-run?), or can it go? An exported helper with no callers tends to get copied by the next reader as if it were the sanctioned way in.

client/src/queries/debugTestMethod.ts
```diff
@@ -0,0 +1,31 @@
...
+export function debugTestMethodCode(
+  className: string,
+  selector: string,
+  dictName?: string,
+): string {
+  const sel = escapeString(selector);
+  return `| cls tc |
+${classLookupOrRaiseExpr(className, dictName)}
+tc := cls selector: #'${sel}'.
+tc setUp.
```
MatiasFernandez: 💭 `tc setUp` is outside the `ensure:`, so a `setUp` that raises never runs `tearDown` — which is exactly the case a debug run is for. The TSdoc above promises "tearDown runs through `ensure:` so a suspended — then terminated — test still tears its fixture down", and that isn't true for a raise in setUp. SUnit's own `TestCase>>runCase` protects both (`[self setUp. self performTest] ensure: [self tearDown]`). Should we do the same here: `[tc setUp. tc perform: #'${sel}'] ensure: [tc tearDown]`?

client/src/gemstoneExplorer.ts
```diff
@@ -744,8 +828,77 @@ export class ExplorerController {
...
+    dictName: string | undefined,
+    className: string,
+    selector?: string,
+    isMeta = false,
+  ): void {
+    if (dictName === undefined || !this.sunit?.isTestClass(dictName, className)) return;
+    // A test class's non-test methods (setUp, helpers) are not runnable rows.
+    if (
+      selector !== undefined &&
+      !(!isMeta && selector.startsWith('test') && !selector.includes(':'))
```
MatiasFernandez: 💭 This inline predicate is the same rule `isTestSelector` implements a few lines up (instance-side, `test*`, unary), just with the class/dict passed in rather than read from `state`. And `isTestSelector` has no production caller any more — only the tests reach it — so the rule now lives in two places and only one of them is the one that runs. Should `decorateTestRow` call it (or should `isTestSelector` take the same explicit params and become the single home for the rule)? Otherwise the next change to what SUnit considers a test has two spots to find.

client/src/gemstoneExplorer.ts
```diff
@@ -5226,8 +5408,27 @@ export function registerGemStoneExplorer(
   // Called once per class that Remove Class actually deleted, so GemStone Search can drop it from its
   // cached corpus instead of showing (and offering to open) a class that no longer exists.
   onClassRemoved?: (sessionId: number, className: string) => void,
+  // True when a URI is the document a SUnit test item points at; see
```
MatiasFernandez: nit: these two comment paragraphs look like a leftover from an earlier shape of this parameter — the first one describes an `isTestItemUri` predicate that is now a field of `ExplorerSunitHooks`, and the "Late-bound, because…" sentence is repeated. Worth trimming to just the second paragraph?

client/src/sunitTestController.ts
```diff
@@ -361,50 +931,152 @@ export class SunitTestController implements vscode.Disposable {
...
+      passedCount,
+      totalCount: children.length,
+    });
+  }
+
+  /**
+   * Show the tests as running before the (blocking) stone call starts, having
+   * first shown them as never-run.
+   *
+   * Both yields matter. The queries are synchronous, so without handing the
```
MatiasFernandez: 💭 "The queries are synchronous" is no longer true after this PR — `runTestClassNb`/`runTestMethodNb` are exactly what stopped being blocking. The two yields still earn their place (the blank frame is what makes a re-run visible), so it's just the *reason* given that has gone stale. Reword to lead with the blank-frame argument?

client/src/sunitTestController.ts
```diff
@@ -361,50 +931,152 @@ export class SunitTestController implements vscode.Disposable {
...
-        run.passed(classItem);
-      } else {
-        run.failed(classItem, new vscode.TestMessage('Some tests failed.'));
-      }
+      // Deliberately no run.passed/failed on the class itself. Every method
+      // already reported, and VS Code rolls a parent's state up from its
+      // children in the tree — reporting the class too only adds a row to the
+      // Test Results list that duplicates the run's own name. The class-level
+      // roll-up below is our own store, which the tree rows read.
+      const allPassed = passedCount === totalCount;
```
MatiasFernandez: 💭 When no child matches a result (`resultMap` misses every selector — e.g. the stone answered nothing, or the selectors moved under us), `totalCount` stays 0 and `passedCount === totalCount` holds, so the class row is stored as `passed` even though every method was reported skipped. A green check on a class where nothing actually ran is the one indicator we most want not to lie. Should this be `totalCount > 0 && passedCount === totalCount`, or should the class result be left out entirely when `totalCount === 0`?

client/src/codeExecutor.ts
```diff
@@ -468,6 +468,65 @@ export class CodeExecutor {
...
+      if (!success) {
+        throw new Error(startErr.message || `GemStone error ${startErr.number}`);
+      }
+
+      await this.pollForResultOop(session);
+      return { raised: false };
+    } catch (e: unknown) {
+      // A cancelled (hard-break) run tells us nothing about the test, so it is
+      // reported as "raised" rather than passed off as a pass.
+      if (e instanceof NbCancelledError) return { raised: true, message: 'Execution cancelled.' };
```
MatiasFernandez: 💭 A debug run that gets cancelled (the progress notification's Cancel is still offered, even though our ■ isn't) comes back as `{ raised: true, message: 'Execution cancelled.' }`, and `debugSingleTest` turns that into `run.errored` plus a stored `error` outcome. That contradicts the rule the run path follows — and that the PR description states — that a stopped test is reported skipped with no verdict left behind, because calling it an error blames the test for the user's decision. Should the cancelled case be distinguishable here (a third state, or rethrowing `NbCancelledError`) so `debugSingleTest` can `reportSkipped` it the way `runSingleTest` does?

client/src/extension.ts
```diff
@@ -1260,6 +1314,9 @@ export function activate(context: vscode.ExtensionContext) {
         uri: vscode.Uri,
         opts?: { viewColumn?: vscode.ViewColumn; preserveFocus?: boolean; preview?: boolean },
       ) => {
+        // Claim the open: without this the Explorer reads an open of a test method's
+        // document as a Testing-view row click and leaves its panes where they were.
+        explorer.markAttributedOpen(uri);
```
MatiasFernandez: 💭 The claim is added before the open, but it is only consumed when `syncToEditor` runs — and if the document is already the active editor, or `openTextDocument` throws, no editor-change event fires and the entry stays in `attributedOpens` for the rest of the session. The next genuine Testing-view click on that same method would then be treated as a deliberate navigation and move the panes. Is it worth clearing the claim if the open didn't happen (or expiring it), rather than leaving a set that only ever grows?

client/src/sunitQueries.ts
```diff
@@ -38,6 +50,46 @@ export function runTestClass(session: ActiveSession, className: string, dictName
...
+  className: string,
+  dictName?: string,
+  onStart?: (cancel: () => void) => void,
+) {
+  return executeFetchStringNb(
+    session,
+    `Run tests: ${className}`,
+    runTestClassCode(className, dictName),
+    `GemStone: running ${className} tests…`,
+    false,
```
MatiasFernandez: nit: this bare `false` is `suppressNotification` — hard to read at the call site (I had to open `executeFetchStringNb` to be sure it wasn't the cancel-related flag). Worth a named constant, or naming it in a trailing comment the way the `onStart` hook is documented above?

MatiasFernandez (approved):
Looks good, left some non-blocking comments

Issue #475, opened 2026-08-21
Two method-search queries ignore the environment they are given, so results outside environment 0 are invisible


ericwinger:
## Summary

Two method-search queries accept an `environmentId` and then ignore it, so they always search
method environment 0. Every caller loops `0..gemstone.maxEnvironment` over them, which means a
method that lives in a higher environment is invisible to the search — and the loop does N
identical full-image scans to produce the same environment-0 answer each time.

Both are the same species of bug and the same shape of fix, so they belong together.

## 1. `referencesToObject` builds its organizer without an environment

`client/src/queries/methodSearch.ts` — `referencesToObject`

```smalltalk
methods := (ClassOrganizer new referencesToObject:
  (System myUserProfile symbolList objectNamed: #'...')).
```

The `environmentId` argument reaches `methodSerialization(environmentId)` only. The organizer is
built bare, so the scan itself always runs in environment 0. Compare `sendersOf` a few lines
above, which correctly does `ClassOrganizer new environmentId: N; yourself`.

Affected callers, all of which sweep environments:

- **Browse References** (`extension.ts`, `gemstone.browseReferences`)
- **GemStone Search**, references pivot (`omniSearch/omniSearchCommand.ts`)
- the **MCP** references tool (`mcpTools.ts`)

Minor, same function: the result is handed to `methodSerialization` without `asArray`, which then
indexes it with `at: i`.

## 2. `hierarchyImplementorsOf` selects with environment-0-only reflection

`client/src/queries/methodSearch.ts` — `hierarchyImplementorsOf`

```smalltalk
(cur includesSelector: #'sel') ifTrue: [methods add: (cur compiledMethodAt: #'sel')].
```

`includesSelector:` and the bare `compiledMethodAt:` are both environment-0 only, so an
implementor that exists only in a higher environment is never collected — in either direction of
the walk. Its caller (`gemstone.hierarchyImplementorsOf` in `extension.ts`) loops environments
over it, so superclass implementors and subclass overrides outside environment 0 are invisible
where that command is used.

## Confirmed on a live stone

`PrbUser>>usesInEnv0` compiled into environment 0 and `PrbUser>>usesInEnv1` into environment 1,
both referencing `PrbTarget`:

| expression | answer |
| --- | --- |
| `ClassOrganizer new referencesToObject: PrbTarget` | `PrbUser>>usesInEnv0` |
[...]

Issue #470, opened 2026-08-20
Multi-environment support is incomplete in GemStone Search and the GemStone Explorer


MatiasFernandez:
## Summary

With `gemstone.maxEnvironment` raised above 0, method environments other than 0 are only
partially supported by the IDE. GemStone Search never finds them, and the GemStone Explorer
lists them but can't open them and gives no indication of which environment a method belongs to.

Three related gaps:

1. **GemStone Search doesn't search non-zero environments.** Searching for a selector or for
   method source never returns Env 1 methods, even plain Smalltalk ones. There's no way to widen
   the search across environments.
2. **The Explorer lists Env 1 methods but opening them fails.** The rows show up under the class,
   but selecting/opening one does not bring up the method. The classic System Browser handled
   this correctly.
3. **The Explorer doesn't say which environment a method belongs to.** Env 0 and Env 1 methods
   with the same selector are visually indistinguishable in the method list.

## Acceptance criteria

- GemStone Search returns methods from environments `0..gemstone.maxEnvironment` — for both the
  selector search and the method-source search — either by default or via an explicit opt-in.
- When more than one environment is in scope, search results indicate which environment each hit
  came from, and opening a hit opens the method in that environment.
- Opening an Env 1 method from the GemStone Explorer opens the correct method source, matching
  what the classic System Browser did.
- Explorer method rows carry visual feedback for their environment when `gemstone.maxEnvironment`
  is greater than 0 (env 0 can stay unadorned so the common single-environment setup is unchanged).

## Implementation notes

The URI layer already models this — `buildMethodUri`/`parseUri` in
`client/src/gemstoneFileSystemProvider.ts` carry an `environmentId` and serialize it as `?env=N`.
The gaps look like places where that dimension is dropped:

- `client/src/gemstoneExplorer.ts` builds method URIs with a hardcoded `environmentId: 0` in both
  `openMethod` and `methodSourceUri`, which would explain an Env 1 row opening the wrong (or no)
  method. Upstream of that, the `SelectorInfo` record the tree is built from has no env field, even
  though `getClassEnvironments` (`client/src/queries/getClassEnvironments.ts`) does return an
  `envId` per category line — so carrying `envId` through into `SelectorInfo` would potentially
  serve both the open fix and the visual-feedback item (icon/description/tooltip on the row).
- On the search side, `searchMethodSource` in `client/src/queries/methodSearch.ts` passes a fixed
[...]

PR #476, opened 2026-08-21
Safe delete: check what still references a method, class or variable before removing it

ericwinger:
## What & why

Fixes #433.

Deleting a method, class, instance variable or class variable was unguarded. The same modal
appeared whatever the target, nothing was checked, and you found out afterwards — from a
doesNotUnderstand or a failed recompile — that something still needed it. Safe delete was the one
refactoring named in the "Done when" of #295 that had not been built.

Each of the four now scans first:

- **nothing references it** → the deletion happens with no question and a notification saying so,
  making the click one step instead of two;
- **something does** → the confirmation says how many and which, one line per referencing class
  with the class named once however many of its methods are involved. **Show References…** lists
  them for browsing; opening one abandons the deletion, because asking to read a method is not an
  answer to the question and a modal whose default action is destructive should not follow you
  there. Closing the list without opening anything brings the question back.

References that go away *with* the target don't count — a recursive method's send of its own
selector, a method inside the subtree of the class being removed, or an override whose senders
will resolve to the inherited implementation. A scan that cannot answer falls back to asking:
"could not check" must never resolve to "delete without asking".

The scans match how the refactoring engine decides rather than reading source text — bytecode
access for an instance variable, literal-frame binding identity for a class variable,
dictionary-scoped object identity for a class — so a mention in a comment, a same-named global,
and a same-named class in another dictionary are all correctly excluded. They ask the base image
directly rather than going through the engine, so the guard behaves the same with or without the
server plugin installed; deleting a method and deleting a class work without it today, and gating
them behind an engine install would have been a regression.

Two commits, so the feature and the follow-up fixes can be read separately.

## Changes

**`1eb9a0d5` — the guard, plus Remove Class Variable**

- The four guarded deletes, and the shared decision in `refactoring/safeDelete.ts` — kept free of
  any query so the whole decision table unit-tests without a stone.
- **Remove Class Variable**, which the Explorer never had: it could add and rename one but not
  remove it. Lightweight like adding one — a class variable is not part of instance layout, so
  nothing is reshaped, no class version is created, and no engine is involved.
- Removing an instance variable no method accesses now applies without opening the preview panel,
  since a preview of a change that breaks nothing is a question with one answer. The engine keeps
  the veto: if it reports methods that will not recompile, the panel opens after all.
- The results picker moves out of a closure in `activate()` into its own module so the
  confirmation can share it, and reports whether the user opened anything.
- Every Explorer removal now shows the trash can. The two variable removals used a minus sign,
  which reads as "take out of this list" rather than "delete this".
- Hand-testing fixtures under `gs-src/fixtures`, with each guarded kind present both referenced
  and unreferenced, plus a companion building the same class name in two dictionaries.

**`02c9595f` — environments, and the scans' honesty about what they saw**

The feature looped over method environments from the start, but three of the four scans underneath
it did not, so the loop was close to a no-op — and the failure mode was exactly the one the guard
exists to prevent.

- **A class-reference scan built its `ClassOrganizer` without an environment**, so it answered
  environment 0 whatever it was asked for. **The two variable scans enumerated `selectors`**,
  which lists environment 0 only, so a method living anywhere else was never offered to the
  accessed-variables test. Confirmed on a live stone before and after; both fixes have an
  integration test verified to fail against the old code.
- **Found methods now carry the environment they were found in.** A selector implemented in two
  environments counts as two methods instead of collapsing into one row, and each result opens the
  document it actually lives in — which also fixes the same latent bug in Senders, Implementors and
  References, all of which have always opened results as environment 0.
- **Removing an override no longer asks.** Previously the worst case in the feature: deleting
  something as ordinary as `printOn:` meant a whole-image walk followed by a dialog listing
  hundreds of methods that were never at risk. Asking the hierarchy is bounded by its depth, and
  the check is one-directional — finding an implementor above skips the scan, failing to find one
  only means we fall through and ask.
- **The scans run under a progress notification.** They walk the image and block the extension
  host, so without one the editor simply froze where the old delete popped a modal instantly.
- **A count at the server-side row cap is reported as a floor** ("At least 500 methods still
  reference it", list marked incomplete) instead of stated as fact when the truth may be thousands.
- **An instance variable could not be removed from a class with no subclasses at all** — found
  while checking the icon sweep's own claim. The action was gated on the row's context value being
  `explorerIvar`, but a leaf class builds its rows as `explorerIvarNoSubs`, so no delete affordance
  appeared, inline or in the menu. It now matches both, as rename and push-up already did.

## Testing

- **6059** client tests on 3.6.2 and **6062** on 3.7.5 (the difference is version gating), plus
  server and mcp-server suites; lint, format and compile clean on both.
- Integration tests run against live stones at **both** boundaries — the issue's stated criterion.
- Manual verification against a dev stone using the committed fixtures: the silent deletes, the
  confirming deletes and their exact reference lists, the three discriminations that must *not*
  appear (a name in a comment, a same-named global, a reference from inside the doomed subtree),
  and the two-dictionary shadow case, where deleting each `SdDemoShadow` named only its own caller.
- No GS SUnit test: no engine class changed. The instance-variable apply path this rides on is
  already covered by `GsInstVarRefactoringTest`.

## Notes for the reviewer

- **References are a confirmation, not a refusal.** #433 floats declining outright. You can delete
  a referenced method today, so refusing would remove a capability — the guard's job is that you
  are told first.
- **A leaf class still deletes silently.** Worth a second opinion: a misclicked class is the
  expensive one to get wrong, and recovery is `abort`, which discards every other uncommitted edit
  in the session. Left consistent with the other three after manual testing; easy to make classes
  always confirm if you'd rather.
- **A method scan asks about the selector, image-wide**, because that is what the image can
  answer — a dispatch through `perform:` is invisible to it. Documented rather than papered over.
- Two pre-existing environment bugs in adjacent queries, and the "sweep environments then dedupe"
  block that is written out nine times, are deliberately **not** in this diff — they affect
  commands outside this feature and are filed as #475.


npapagna (commented):
Reviewed the safe-delete changes and want to flag a few things before this merges. Grouped by priority below; inline comments carry the details on the ones GitHub lets me anchor to a diff line.

**Blockers**
1. `gemstoneExplorer.ts` (removeClass) - doomed-subtree exclusion filters by class name, not identity (inline comment).
2. `safeDelete.ts` - the "list not complete" cap hedge is computed after exclusion filters run, so it can go missing when it shouldn't (inline comment).
3. `gemstoneExplorer.ts` (removeMethod) - self-send exclusion ignores environmentId (inline comment).
4. `extension.ts` - the six pre-existing dedup blocks (`sendersOfSelector`, `implementorsOfSelector`, `hierarchyImplementorsOf`, `browseReferences`, `sendersOf`, `implementorsOf`, roughly lines 2167-2394) still dedupe on `className|isMeta|selector` only, not reusing the new environment-aware `dedupeMethodResults`. I could not attach this inline since those lines are not part of this PR's diff, but the CHANGELOG claims this PR fixes the same latent bug in Senders/Implementors/References too, and as far as I can tell it doesn't yet.
5. `methodResultsPicker.ts` - the open-browser path still ignores the row's environment (inline comment).

**Important, but fine as a follow-up PR**
6. `gemstoneExplorer.ts` (`scanReferences`) - one blocking round trip per environment instead of one server-side loop (inline comment).

**Non-blockers, just a friendly ask**
Nice work closing the environment-blindness bug across the scan paths. A few spots ended up with the same logic duplicated in two or more places; would be great to fold these together when convenient, no rush and not holding up this PR (inline comments on each).

client/src/gemstoneExplorer.ts
```diff
@@ -4214,29 +4429,28 @@ export class ExplorerController {
...
+    // (Doomed class >> new naming Doomed, a subclass's own method), so it is no reason to
+    // ask. What is left is the references that will still be there afterwards.
+    const doomed = new Set(targets.map((t) => t.className));
+    const scan = await this.scanReferences(`Finding references to ${className}…`, (env) =>
+      queries.referencesToClassInDict(session, className, dictIndex, env),
+    );
+    const target: SafeDeleteTarget = {
+      kind: 'class',
+      label: `${className} from ${dictName}`,
+      references: scan.references.filter((r) => !doomed.has(r.className)),
```
npapagna: **Blocker.** `doomed` is built from class names only (`targets.map(t => t.className)`), but `referencesToClassInDict` deliberately resolves by object identity through the dictionary so a shadowed class name elsewhere does not collide. Filtering `scan.references` with `!doomed.has(r.className)` reintroduces the identity problem this scan was built to avoid: a real reference from an unrelated same-named class in another dictionary gets silently dropped, and the delete can go through as "nothing references it" when something outside the subtree still does.

client/src/refactoring/safeDelete.ts
```diff
@@ -0,0 +1,244 @@
...
+  if (target.scanFailed) {
+    lines.push(`Could not check what references it: ${target.scanFailed}`);
+  }
+  if (target.references.length > 0) {
+    const n = target.references.length;
+    // The scan is capped server-side and the client cannot tell a full page from an exact
+    // answer, so at the cap the count is stated as a floor rather than as fact. Saying
+    // "500 methods reference it" when the truth may be thousands is a worse failure than
+    // hedging, because the whole point of the dialog is that its numbers can be trusted.
+    const atCap = n >= METHOD_SEARCH_RESULT_LIMIT;
```
npapagna: **Blocker.** `n` here is `target.references.length` after the caller's own exclusion filters (self-send, doomed subclasses) have already run, so `atCap` can read false even when the underlying scan hit the 500-row cap. Example: a recursive method with exactly 500 senders hits the cap, the self-send exclusion drops it to 499, and the confirmation reports "499 methods still reference it" with no "list not complete" hedge, understating risk in exactly the way this branch's commit message says it is fixing.

client/src/gemstoneExplorer.ts
```diff
@@ -3545,12 +3727,42 @@ export class ExplorerController {
...
+      ? { references: [] as queries.MethodSearchResult[], scanFailed: undefined }
+      : await this.scanReferences(`Finding senders of #${selector}…`, (env) =>
+          queries.sendersOf(session, selector, env),
+        );
+    const target: SafeDeleteTarget = {
+      kind: 'method',
+      label: `#${selector} from ${side}`,
+      // The method's own send of its own selector goes away with it, so a recursive method
+      // is not a method with a surviving sender.
+      references: scan.references.filter(
```
npapagna: **Blocker.** This exclusion matches on `className`/`isMeta`/`selector` only, not `environmentId`, even though `dedupeMethodResults` treats those four together as a method's identity. If the class has two methods for the same selector/side in different environments, deleting one wrongly excludes the other's self-send as if it were the deleted method's own recursion, understating real references.

client/src/methodResultsPicker.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const picked = await vscode.window.showQuickPick(items, {
+    placeHolder: `${results.length} method${results.length === 1 ? '' : 's'} found`,
+    matchOnDescription: true,
+    matchOnDetail: true,
+  });
+  if (!picked) return false;
+
+  const r = picked.result;
+  if (!SystemBrowser.navigateTo(sessionId, r)) {
```
npapagna: **Blocker.** This fix covers the no-browser-open fallback, but `SystemBrowser.navigateTo` (the branch taken whenever a browser is already open) still never reads `result.environmentId` and opens whatever environment the browser currently has selected. That is the more common path, and it still has the bug this comment says was fixed.

client/src/gemstoneExplorer.ts
```diff
@@ -3526,16 +3639,85 @@ export class ExplorerController {
...
+  ): Promise<{ references: queries.MethodSearchResult[]; scanFailed?: string }> {
+    return vscode.window.withProgress(
+      {
+        location: vscode.ProgressLocation.Notification,
+        title,
+        cancellable: false,
+      },
+      () => {
+        try {
+          const found = this.environmentsToScan().flatMap((env) => scan(env));
```
npapagna: **Important, but okay as a follow-up PR.** `scanReferences` does one blocking GCI round trip per configured environment (`environmentsToScan().flatMap(...)`) rather than a single server-side loop. With `gemstone.maxEnvironment` above 0, every delete does N sequential blocking whole-image scans, and UI freeze time scales linearly with the number of environments. Worth fixing, but not a correctness issue, so landing it separately is fine.

client/src/gemstoneExplorer.ts
```diff
@@ -3526,16 +3639,85 @@ export class ExplorerController {
...
+    const envs: number[] = [];
+    for (let env = 0; env <= this.maxEnv(); env++) envs.push(env);
+    return envs;
+  }
+
+  // The nearest ancestor that also implements `selector`, or undefined. Answers the class a
+  // send would reach once this implementation is gone, which is what makes removing an
+  // override harmless. Best-effort: a failure here answers undefined, which only means the
+  // caller falls back to the full sender scan.
+  private superclassImplementorOf(
```
npapagna: **Non-blocker, just a friendly ask.** `superclassImplementorOf` (here) and the `doomed` set in `removeClass` are two independent ways of saying "this reference resolves elsewhere, so it does not count." Since `safeDelete.ts` already aims to be the pure decision module for this feature, would it make sense to fold both into a shared concept there instead of hand-rolling each in the controller? Not blocking this PR, just flagging it so the next safe-delete kind does not end up growing a third bespoke version.

client/src/refactoring/queries/methodsAccessingInstVar.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ *  The class is resolved through `dict` (a 1-based SymbolList index, canonical for Jasper,
+ *  or a name); a class the dictionary does not bind answers nothing rather than an error.
+ *
+ *  Selectors are enumerated with `selectorsForEnvironment:`, not `selectors`: the latter
+ *  lists environment 0 only, so a method that exists solely in another environment would
+ *  never be offered to the accessed-variables test at all and the scan would wrongly report
+ *  the variable unused. Verified on a live stone — `selectors` answers the environment-0
+ *  selector and `selectorsForEnvironment: 1` the environment-1 one. */
+export function methodsAccessingInstVar(
```
npapagna: **Non-blocker, just a friendly ask.** This, `methodsAccessingClassVar`, `referencesToClassInDict`, and `sendersOf` each hand-write a very similar Smalltalk loop over `selectorsForEnvironment:`/`compiledMethodAt:environmentId:` with just the predicate differing. Worth considering a shared parameterized scan primitive at some point, mostly so a future fix to per-environment enumeration (like the `selectorsForEnvironment:` change in this PR) only needs to happen once instead of four times. No rush, just leaving it here for later.

client/src/gemstoneExplorer.ts
```diff
@@ -2147,19 +2153,126 @@ export class ExplorerController {
     return choice === undefined ? undefined : choice === YES;
   }
 
-  // "−" inline on an instance-variable row: remove it (with the will-not-recompile
-  // preview).
+  // 🗑 inline on an instance-variable row: remove it. Guarded like every other delete —
+  // a variable no method reads or writes goes straight through (no preview to show, since
+  // nothing can break) and is announced afterwards, while one that IS accessed raises a
+  // confirmation naming the accessors and then opens the will-not-recompile preview.
   async removeInstVar(item: IvarItem): Promise<void> {
```
npapagna: **Non-blocker, just a friendly ask.** `removeInstVar`, `removeClassVar`, `removeMethod`, and `removeClass` all follow the same scan, build a `SafeDeleteTarget`, call `decideSafeDelete`, bail on cancelled, delete, announce-on-silent shape, with only the scan query and the delete RPC really varying between them. A shared `runSafeDelete` helper could make those four call sites mostly just the parts that differ. Not blocking, just something to consider when there is time.

npapagna:
## Testing gaps

While cross-checking the blockers above against the test suite, I noticed each one has a related test that passes but stops just short of the scenario that would actually catch the bug. Flagging in case it's useful when fixing them.

1. **removeClass doomed-by-name** (blocker 1) - `explorerRemoveClass.test.ts:105` tests plain doomed-subclass exclusion, and the shadow test at `:201` only checks that `deleteClass` targets the right dictionary. Neither gives an unrelated same-named class a real reference to see if it gets wrongly swallowed by the `doomed.has(r.className)` filter.
2. **Cap hedge lost after filtering** (blocker 2) - `safeDelete.test.ts:496-526` builds a `SafeDeleteTarget` directly with 500 references, bypassing the real controller path entirely. No test builds a scan of exactly 500 combined with a self-send or doomed exclusion that drops it back under 500, which is exactly how the bug hides.
3. **Self-send exclusion ignoring environment** (blocker 3) - `explorerRemoveMethod.test.ts:118` tests self-send exclusion at environment 0 only; the multi-environment sweep tests at `:304-347` never combine with a same-selector self-send in a different environment.
4. **extension.ts dedup blocks** (blocker 4) - no test in this PR touches any of the six dedup sites for multi-environment behavior at all.
5. **navigateTo ignoring environment on the open-browser path** (blocker 5) - `methodResultsPicker.test.ts:131` covers only the no-browser fallback. `systemBrowser.test.ts`'s `navigateTo` tests all use `environmentId: 0`; none passes a non-zero environment and checks what actually opens.

A couple of gaps beyond the blockers themselves:
- No test drives the real scan -> filter -> target pipeline all the way to the 500-row cap for any of the four delete commands (method/class/ivar/classvar); only the decision-logic layer is exercised at the cap.
- No test checks that rows found in two different environments both survive into the same confirmation dialog's grouped display, except for removeMethod's sweep.

Worth adding coverage for these alongside the fixes so the blockers stay fixed.

client/src/gemstoneExplorer.ts
ericwinger: @/tmp/claude-8554/-home-ewinger/c9e4c0b7-413c-4e42-8116-dfd93efee146/scratchpad/r1.md

client/src/refactoring/safeDelete.ts
ericwinger: @/tmp/claude-8554/-home-ewinger/c9e4c0b7-413c-4e42-8116-dfd93efee146/scratchpad/r2.md

client/src/gemstoneExplorer.ts
ericwinger: @/tmp/claude-8554/-home-ewinger/c9e4c0b7-413c-4e42-8116-dfd93efee146/scratchpad/r3.md

client/src/methodResultsPicker.ts
ericwinger: @/tmp/claude-8554/-home-ewinger/c9e4c0b7-413c-4e42-8116-dfd93efee146/scratchpad/r4.md

client/src/gemstoneExplorer.ts
ericwinger: @/tmp/claude-8554/-home-ewinger/c9e4c0b7-413c-4e42-8116-dfd93efee146/scratchpad/r5.md

client/src/gemstoneExplorer.ts
ericwinger: @/tmp/claude-8554/-home-ewinger/c9e4c0b7-413c-4e42-8116-dfd93efee146/scratchpad/r6.md

client/src/refactoring/queries/methodsAccessingInstVar.ts
ericwinger: @/tmp/claude-8554/-home-ewinger/c9e4c0b7-413c-4e42-8116-dfd93efee146/scratchpad/r7.md

client/src/gemstoneExplorer.ts
ericwinger: @/tmp/claude-8554/-home-ewinger/c9e4c0b7-413c-4e42-8116-dfd93efee146/scratchpad/r8.md

ericwinger:
## All five blockers fixed, and the branch is up to date with `main`

Thanks — every one of these was real, and four of the five failed in the same direction: telling the user that *less* depends on the target than really does. That's the one direction this guard can't afford, because a delete that looks safe goes through with no question at all.

Replies are inline on each thread. Summary:

| Blocker | Fix | Commit |
|:--|:--|:--|
| 1 · doomed subtree matched by name | keys on name **and** home dictionary | `236ba0a4` |
| 2 · cap hedge computed after exclusions | truncation read off raw rows, per environment | `236ba0a4` |
| 3 · self-send exclusion ignores environment | matches the environment actually being removed | `236ba0a4` |
| 4 · six dedup blocks in `extension.ts` | all route through the shared fold | `236ba0a4` |
| 5 · open-browser path ignores row environment | browser switches to the row's environment | `236ba0a4` |

### On blocker 4, which you couldn't anchor inline

You were right that the CHANGELOG claimed something the code didn't do. All six sites now call `dedupeMethodResults`, which has moved to `client/src/queries/methodSearch.ts` — beside `MethodSearchResult` and the row cap, where every caller that sweeps environments already looks. Importing the refactoring module into `extension.ts`'s search commands would have been the wrong direction.

The CHANGELOG entry is corrected and now also names the half that was genuinely missing: results were folded environment-blind *and* opened as environment 0, and both are fixed.

Those six handlers can't be reached from a unit test without standing up the whole extension, so a source scan pins the absence of a second private copy of the rule (`methodResultDedupe.manifest.test.ts`), alongside the helper's own tests. It found a **seventh** site you hadn't listed — `omniSearch/omniSearchCommand.ts`. That one is deliberate and documented at the call site: GemStone Search shows one row per selector and carries the environment into the open action, which is a different contract from a reference list. It's allow-listed with that reason rather than silently skipped, so please push back if you disagree.

### Testing gaps

Your second comment was the more useful of the two — every item was accurate, and the two extras at the end were the ones that mattered most. All closed.

Rather than trust line coverage, I mutated each behaviour this branch changed and required a test to fail. Sixteen mutations. The first pass left **three survivors**, all your second extra generalised: remove instance variable, remove class variable and remove class each wire the scan's cap through to their own confirmation, and nothing checked that any of them did — only `removeMethod` had cap tests, and the wiring is per call site. Seven tests added in `ae8101aa`; all sixteen mutations are now caught.

### Also changed, from hand-testing

The dialogs never said which method environment anything was in, so two environments' methods collapsed into one line and the count read "1 method" for what was really two — the same collapse the scans here were just fixed to stop making. References outside environment 0 are now labelled `Account [env 1] >> #balance` and get their own line, environment 0 stays unlabelled, and removing a method whose selector is also implemented elsewhere says so and says only the environment-0 one is going (`8682372b`).

### Follow-ups filed

- **#483** — the per-environment round trips from your item 6, with the two constraints a server-side loop has to keep.
- **#484** — the three non-blocking cleanups, consolidated. Your scan-primitive point is flagged there as the highest-value of them.

The Explorer's own multi-environment gaps (lists environment-1 methods it can't open, no per-row environment indication) are already #470; the two queries that ignore a given environment are #475.

### Verification

Branch merged up to `main` (`51e84001`, one trivial import conflict). Gate: 411 test files, 6192 tests, lint and typecheck clean. Hand-tested against a live 3.7.5 stone with `gemstone.maxEnvironment` raised, using a fixture committed on the branch (`gs-src/fixtures/safeDeleteScanDemo.gs`) that builds each of these cases and self-checks that it did — all scenarios confirmed by hand except the already-open-System-Browser path, which was skipped because that file is frozen.

Ready for another look.

npapagna (approved):
Thanks @ericwinger !

PR #478, opened 2026-08-21
Address first-use friction: login clarity, navigation discoverability, and new-user onboarding (#468)

### LINKED ISSUE #432: Senders/implementors counts shift the method source when it opens — give them a home outside the document flow

### LINKED ISSUE #468: First-use feedback: remote login fields, invisible selection in High Contrast, and Methods pane discoverability
Observations from the second of three screen-capture sessions in which a developer used Jasper for the first time and narrated as he went. This issue covers **logging in to a remote stone and the first attempt to browse code**. Timestamps refer to video 2.

He did get connected this time, and his overall verdict on browsing was *"so far this looks pretty intuitive"* — the items below are the friction he hit on the way there.

### 1. "Host user" / "host password" have no explanation (1:00–1:13)

*"I'm not sure what to do about host user and host password, if I'm supposed to fill in my user and password because that's how I SSH to the host. Or not."* He guessed, left both blank, and it happened to work.

- [ ] Say in the field description what these are for and when they are required. If they're only needed for certain topologies, mark them optional.

### 2. Documentation order doesn't match task order (0:27, 1:19–2:06)

*"If I'm reading the instructions correctly, I need to skip down to logins and sessions"*, then later he went back and re-read to find *"find the login entry you just saved and click the plug connect."* He was jumping around the document to assemble the sequence.

- [ ] Provide a linear "connect to a remote stone" walkthrough that matches the order the UI presents the steps.

### 3. A successful connection isn't clearly announced (2:18–3:00)

*"It looks like the connection may have worked. So now what do I do?"* … and after clicking around, *"so it looks like I am successfully connected, I guess."* He confirmed it only by opening a dictionary and seeing contents.

- [ ] Give unambiguous connected/disconnected feedback, and a suggested next action after a successful login.

### 4. Browsing a dictionary offers to rename it (2:48)

*"Oh, I don't want to rename the dictionary."* A rename was triggered by ordinary exploratory clicking.

### 5. Same problem with instance variables (7:10–7:14)

*"Move instance variable up. No, I do not want to change anything. I just want to look. Just trying to look around and see what's going on."*

Together these are one theme: **mutating and refactoring actions are on a hair trigger for someone whose only intent is to read.** A first-time user should not be able to rename a dictionary or reorder instance variables by clicking around.

- [ ] Require a deliberate gesture (context menu, explicit edit affordance) for rename and for instance-variable reordering — not a single click or an accidental drag on the tree item.
- [ ] Consider an explicit browse/read-only posture for exploration.

### 6. Selection is nearly invisible under the High Contrast theme (4:12–4:46, 8:15–8:34)

He runs VS Code in High Contrast because of low vision. The consequences:

- Selected method: *"it's kind of subtle… I have to look for this light little amber outline around the thing that I have selected."*
- Selected class: *"I can't tell that I've clicked on it… I can kind of tell, because it's white and all the others are amber. So my current class selection is very subtly indicated."*

This is an accessibility defect, not a preference. Selection state is a primary signal in a class browser.

- [ ] Use theme-aware selection colors that hold up in High Contrast (and in the standard light/dark themes), rather than a low-contrast outline.

### 7. The current selection is left scrolled out of view when panes rearrange (3:55–4:46)

*"What happened was when the windows rearranged themselves, my current selection just jumped out of view."* He believed the browser had lost his selection and started re-verifying it.

- [ ] Scroll the selected item back into view whenever a pane is resized or the layout changes.

### 8. No obvious way to reach a second method of the same class (5:25–8:41)

Having opened one method, he was stuck for three minutes: *"Now what if I want to look at a different method on file exists, and I don't see an obvious way to do it."* He tried the breadcrumb (*"What happens if I click on this? Nothing much"*), doubted whether the class simply had one method, tried other classes, and only eventually found it: *"I am forgetting that when I click on the class itself… I'm noticing the methods are down here in the Methods pane. OK. Getting used to this, figuring it out."*

- [ ] Make the Methods pane discoverable — it is the primary navigation surface and he did not see it.
- [ ] Make the breadcrumb do something useful when clicked, or stop presenting it as clickable.

### 9. Selecting a class doesn't update the editor (7:18–8:07)

*"I noticed that my view has not changed from file exists. I kind of expected that it would once I click on that."* … *"But still, the view up here is not changing. So I'm not sure why."* The editor only updated once he clicked a *method* — which he hadn't found yet (see above), so the two problems compounded.

- [ ] Decide what selecting a class should show in the editor, and make it show something, so class selection produces visible feedback.

### 10. Nothing worth looking at on first connect (7:29–7:47, 8:54)

He was browsing raw kernel classes (`FileExists`, `FileLocator`, `FileSystemHandle`) and said *"I'm also not entirely sure if I'm even looking at classes that are good to look at."* His own fix was to plan to load a familiar test package to get *"a richer experience of navigating around classes and code."*

- [ ] Suggest a good starting point after login — a sample package, a getting-started tour, or a curated set of classes — rather than dropping a new user into the kernel.

ericwinger:
## Summary

Addresses the actionable items from #468 — observations from the second new-developer first-use session (remote login, then a first attempt to browse code). The work groups into six areas; each is one focused commit (plus one merge of `main`), described below with **what to expect** and **how to test** it in a real editor window.

Where an item turned out not to need code, it's listed under "Items closed without code" with the reasoning.

## What changed, and what to test

Log in to a stone first (any 3.7.x). Each area can be checked independently.

### 1. Login field help + connection feedback (items 1, 2, 3)
**Expect:** The login editor has a "Help me login" toggle (open by default for a new login) with per-field help; **Host User** / **Host Password** are marked optional. A left status-bar item shows a spinner while connecting, and on failure turns red and is click-to-explain. `README` / the walkthrough's connect page are rewritten to follow the UI order.
**Test:** Add a login and connect — watch the status bar go from connecting → (banner on success, see #2) or red on a bad password (click it to see why). Open the login editor and confirm the field help and the optional markers read clearly.

### 2. Green "connected" banner (item 3)
**Expect:** On a successful connect, a green `✅ Connected to <stone>` banner flashes at the top of the GemStone Explorer's Dictionaries view for a few seconds.
**Test:** Connect and confirm the banner appears and then clears on its own. (VS Code status bars can't render green, so this is a `TreeView.message` banner — please eyeball placement.)

### 3. High Contrast selection visibility (item 6)
**Expect:** In the class browser, the selected row draws a full outline in the theme's focus-border color — visible in every theme and bold under High Contrast.
**Test:** Switch VS Code to a High Contrast theme, click a class/method, and confirm the selection is unmistakable. (Accessibility fix — please verify in a real HC window.)

### 4. Explorer selection stability + Methods-pane discoverability (items 7, 8)
**Expect:** The Explorer's **Open Editors** pane is gone (it reshuffled the sidebar and scrolled your selection out of view when you opened your first editor); a left status-bar **Close N GemStone editor(s)** button replaces its one-click close-all. The Methods pane opens each method in one reusable preview tab; a one-time hint explains this at the moment a click first replaces a previewed method, the pin command reads **Keep Method Open (Pin)**, and the method-row tooltip spells out preview-vs-keep.
**Test:** Open a class, single-click several methods — confirm the selection no longer jumps away, and the one-time hint fires the first time a click replaces a method. Confirm the status-bar close-all button tallies and closes open GemStone editors.

### 5. Breadcrumb as a live class navigator + Back/Forward (items 8, plus #432)
**Expect:** The editor breadcrumb over an open method now drills into the stone at every level (dictionary → class → side → category → selector), so you can navigate sideways to sibling classes/methods without leaving the editor. Two title-bar **GemStone Go Back / Forward** buttons retrace the methods you've viewed (VS Code's own history can't, because methods share one preview tab). Also fixes #432: the senders/implementors hover no longer disappears when one of its two queries fails.
**Test:** Open a method, click through the breadcrumb dropdowns to reach a sibling method and the class definition. Use Go Back / Forward to retrace. Rest on a selector to confirm the senders/implementors hover still shows even on a stone without the RB plugin.

### 6. "Start Here" onboarding for a newly-connected user (item 10)
**Expect:** On connect, a purple **GemStone - Start Here** button appears in the status bar. Clicking it opens a quick pick — *Browse a class*, *Search your code*, *Open a workspace*, *Take the tour*. It stays put (it's unobtrusive) and is removed only via its own **Hide the Start Here button** entry; **GemStone: Reset Getting Started** brings it back. The same quick pick is always available from the Command Palette as **GemStone: Start Here**. The Get Started walkthrough also gained **Browse classes** and **Search your code** steps.
**Test:** Connect and confirm the button appears; click it and try each quick-pick action. Confirm dismissing the quick pick (clicking away) leaves the button in place, and that **Hide the Start Here button** removes it. Run **GemStone: Reset Getting Started** and reconnect to confirm it returns. Open Help → Welcome → Get Started with GemStone and confirm the two new steps.

## Items closed without code
- **Items 4 & 5 (accidental rename / ivar reorder):** not reproducible — rename requires an explicit icon/menu plus a confirmation dialog, and ivar moves require an explicit arrow plus a destination quick pick. Dropped; reopen with a real accidental-repro case.
- **Item 8, breadcrumb tooltip separators:** the visible breadcrumb bar already uses `›`. The remaining `/`-separated string is the editor tab's hover *tooltip*, which VS Code derives from the resource URI — no extension API can override it. Not fixable from the extension.
- **Item 9 (selecting a class should update the editor):** effectively resolved by items 6 + 8 — clicking a class now visibly populates the (larger, discoverable) Methods pane and the selection is clearly highlighted. We deliberately did **not** auto-open a class-definition tab on every click (visually noisy for read-oriented browsing). Left for the original tester to confirm.

## Testing done
Fast gate green: ESLint, Prettier, `tsc`, and the full unit + integration suites — **client 5912 · server 322 · mcp-server 92** (integration tests log into a live 3.7.x stone).

Closes #468


greghess-gemtalk:
This dialog is confusing: it looks like it offers to install the optional Gemstone support, always install, never install or cancel logging in. The actual behavior appears to be: install, always install, never install and don't install; but continue with login. Suggest rewording the "cancel" button to something like "don't install." ;-) And rearrange them to be "install," "don't install," "always install" and "never install." Unless you had a really specific reason for the odd order they're currently in. ;-)
image

ericwinger:
@greghess-gemtalk — addressing your note on the install dialog (https://github.com/GemTalk/Jasper/pull/478#issuecomment-5403816634). This is now on the branch (c53b433e):

- **"Cancel" → "Not Now."** It's the modal's close/Escape affordance (`isCloseAffordance`), so declining is now a labeled choice instead of a generic Cancel that read as "cancel the login." It declines *this* connect only — the setting stays `ask`, so you're offered again next time — which is why it's "Not Now" rather than "Never."
- **Reordered to Install / Not Now / Always / Never** — the two one-time actions first, then the two sticky "remember my choice" options. One caveat: VS Code positions the close-affordance button by OS convention, so "Not Now" may still render in the trailing (old Cancel) slot regardless of the order we pass — that's the native placement, not something we can force.
- While in there, also reworded the user-facing **"optional" server support → "recommended"** across the prompt, the uninstall confirmation, the `gemstone.serverSupport.autoInstall` setting, and the Get Started walkthrough.

Thanks for the catch. 🙂

greghess-gemtalk (approved):
Filed one issue. No obvious problems noted with this round of changes. Looks good!

PR #480, opened 2026-08-22
Detect servers started outside Jasper's environment and offer to restart them under Jasper

### LINKED ISSUE #472: Detect stones/netldis started outside Jasper's environment and offer to restart them under Jasper and connect
## Reproduction (what the user hits)

Setup: a stone and netldi that Jasper manages get started **outside** Jasper — e.g. by hand from a terminal after a machine reboot — from a shell whose GemStone environment differs from Jasper's own (a different `GEMSTONE_GLOBAL_DIR`, or an inherited `GEMSTONE_NRS_ALL` that points their registration at a different directory). The external stone and netldi use the **same names** as a database Jasper manages.

1. Outside Jasper, start `<stone>` and its `<netldi>` from a terminal (same names as a Jasper-managed database), in a shell whose GemStone environment differs from Jasper's.
2. In Jasper, click the login for that database and try to connect.
3. The connect does not just work — instead:
   - The Databases tree shows a **contradictory state** — e.g. `<stone>` *Running* but `<netldi>` *Stopped* — even though **both** processes are alive on the host.
   - The login **fails as though the stone isn't running** (Jasper can't traverse a NetLDI its own `gslist` doesn't see), and Jasper offers to start the database.
   - Accepting the offer **collides** with the already-running external servers (name already in use / registered elsewhere), and in some states a raw `GEMSTONE environment variable is not defined` error surfaces.
4. Seeing "not running" and "GEMSTONE not defined," the user goes off to debug their **shell profile and GemStone install** — when the real cause is simply that the servers were started outside Jasper's environment and registered where Jasper's own `gslist` doesn't look.

## Summary

Jasper runs its own `gslist` in its own environment — it sets `GEMSTONE_GLOBAL_DIR` to its configured root (`sysadminStorage.getRootPath()`) and discovers servers by their `.LCK` files in that root's `locks/` directory. A stone or netldi started **outside** Jasper — by hand, or from a shell carrying different GemStone environment (e.g. an inherited `GEMSTONE_NRS_ALL` or a different `GEMSTONE_GLOBAL_DIR`) — registers in a **different** `locks/` directory, so Jasper's `gslist` can't see it even though the OS process is alive. Servers started outside Jasper can't be prevented, so Jasper should recognize this situation and offer to reconcile it instead of showing a misleading state.

This issue also folds in two related state-accuracy fixes (see **Related state-accuracy fixes** below) that came out of the same investigation: status should reflect *reachability* rather than mere process presence, and Jasper's discovery environment should be deterministic rather than perturbed by the launching shell's leftover GemStone variables.

## Motivation / what goes wrong today

- The tree can show a **contradictory, partial state** — e.g. "Stone Running / NetLDI Stopped" while **both** processes are alive — because one registered where Jasper looks and the other didn't.
- A login **fails** (it must traverse a NetLDI that Jasper's `gslist` can't see), and the failure reads as "the stone isn't running" when it plainly is.
- In some states a raw GemStone shell-script error (`GEMSTONE environment variable is not defined`) is **relayed verbatim**, sending the user off to debug their shell profile for a problem that isn't in their shell.

## Acceptance criteria

- [ ] For each managed database, Jasper detects when a `stoned`/`netldid` process matching that database's stone/ldi **name and version** is running but is **not** visible in Jasper's own `gslist` (its `locks/` dir), and classifies it as "started outside Jasper's environment."
- [ ] On connect (and reflected on tree refresh), when such a server is detected, Jasper offers to reconcile with a dialog worded as a **mismatch**, not a missing install. Example wording:

  > "\<stone> and \<netldi> are running, but were started outside Jasper's environment, so they're registered where Jasper's own `gslist` doesn't look. (Jasper runs its own `gslist` in its own environment, which can differ from `gslist` on the host — so a server can look *Stopped* to Jasper while it's alive on the host.) Restart them under Jasper's environment and connect?"
  >
  > **[Restart & Connect]  [Connect as-is]  [Cancel]**

- [ ] **Restart & Connect** stops the external stone/netldi (clean `stopstone`/`stopnetldi`, with the existing force-kill fallback), then does Jasper's normal start + connect, so everything lands in Jasper's own `locks/` dir.
- [ ] Before **Restart & Connect** stops anything, confirm the running same-named server is actually the managed database. If it can't be confirmed, warn that it may be a **different** stone that merely shares the name, and do **not** auto-stop it. (Identity by name alone is ambiguous — see Implementation notes.)
- [ ] The reconcile dialog warns that restarting the stone drops any **uncommitted sessions** on it.
- [ ] If Jasper **cannot** stop or kill the external server, it does **not** dead-end: it shows enough to resolve by hand — **stone name, netldi name, process IDs, and the directory the server is registered in** — plus how to inspect it (e.g. `gslist` pointed at that directory).
- [ ] Jasper never surfaces the child script's bare `GEMSTONE environment variable is not defined`; that path is replaced by the situation-specific message above.
- [ ] The failure message is **not** phrased as "GemStone \<version> install path isn't configured" — a user who already installed that version would be baffled by that.

## Related state-accuracy fixes (folded in)

Both came out of the same investigation and share the root cause above — Jasper's *view* of server state must be accurate and self-consistent regardless of how or where servers were started. Filed here rather than separately.

### Status should reflect reachability, not just process presence

Today the tree shows *Running* whenever a matching process appears in Jasper's own `gslist`, even if the stone has no reachable NetLDI and a login cannot actually connect — so the tree can directly contradict a failing login.

- [ ] A stone that is running but **not connectable** (its NetLDI is unreachable, or it's registered outside Jasper) is shown in a distinct "running but not connectable" state — never a plain, healthy *Running*.
- [ ] The Running/Stopped state the tree shows must not contradict what a connect attempt will actually do. One approach: give the `isStoneRunning`-style checks a reachability notion instead of trusting `gslist` presence alone.

### Discovery environment should be deterministic

The start path (`getEnvironment`) sets `GEMSTONE_NRS_ALL`, but `versionEnvironment` — used for `gslist` discovery **and** the version terminal — sets no `GEMSTONE_NRS_ALL` at all. So a stray inherited `GEMSTONE_NRS_ALL` / `GEMSTONE_GLOBAL_DIR` from the shell that launched the editor can sway what Jasper discovers and where a version-terminal command lands.

- [ ] The environment used for discovery (and the version terminal) normalizes these variables the same way the start path does, so Jasper's view matches what it actually manages and isn't perturbed by the launching shell's leftovers.

## Implementation notes

*(Possibilities, not prescriptions.)*

- The building blocks exist: each database's stone/ldi name and version are known; `processManager` already does ps-based process inspection (`classifyPidOwnership`, `forceKillStone`, `inspectStaleLock`) and has a clean-stop-with-force-fallback path; version matching has a `versionsMatch` guard to reuse so two installed versions sharing a stone name don't collide.
- The missing piece is the **cross-check**: "a process matching my stone/ldi name+version exists, but it's absent from my own `gslist`." One approach is a broad process scan (for `stoned <stone>` / `netldid <ldi>`) compared against the `gslist` Jasper already runs against its configured root.
- A running stone's `GEMSTONE_GLOBAL_DIR` is fixed at start time — a live stone can't be re-homed into Jasper's directory — so reconcile must be **stop + start**, not re-registration.
- **Deciding "is this really my stone?" is the hard part, and name alone can't answer it** — the same name can belong to a different stone. Matching the extent/data files the running `stoned` has open (e.g. from its command line or open files) against the managed database's path is a *candidate* signal, but not foolproof: a stone can be started against a different extent. So treat "can't confirm" as "not confirmed" (warn-only, don't stop it), and finding a more robust identity check is an open problem worth solving while implementing this.

## Background (why the guard is part of this, not a separate issue)

The `GEMSTONE environment variable is not defined` errors observed in practice were the GemStone shell script's own stderr, relayed by Jasper, and occurred only while a mismatched/external server was in the picture. Jasper's own start path sets `GEMSTONE` correctly — a clean Jasper start+connect works even with `GEMSTONE` unset in the shell. So the hardening bundled here is (a) the situation-specific messaging above, and (b) resolving/verifying the environment up front and failing with an actionable message rather than relaying a raw shell-script error.

ericwinger:
Closes #472.

## The problem

Jasper runs its own `gslist` against the root it manages, and a GemStone server registers in the `locks/` directory of whatever `GEMSTONE_GLOBAL_DIR` it was started with. A stone or NetLDI started by hand — from a shell whose GemStone environment differs from Jasper's — therefore registers somewhere Jasper does not look, and `gslist` alone reported it as *Stopped* while the process was plainly alive.

What that looked like in practice:

- The Databases tree showed **Stone Running / NetLDI Stopped** with *both* processes up, because one had registered where Jasper looks and the other had not.
- The login failed as though nothing were running, and Jasper offered to start the database.
- Accepting that offer collided with the servers already holding those names.
- In some states a raw `The environment variable 'GEMSTONE' is not defined` was relayed verbatim, sending the user off to debug a shell profile and an install that were both fine.

## What this does

Cross-checks the host process table against Jasper's own `gslist` and marks such a server **Running outside Jasper**, with its PID and registration directory in the row's tooltip. On connect — or from the row's **Restart Under Jasper's Environment** action — Jasper offers to stop it and start it again under its own environment, worded as the environment mismatch it is rather than as a missing install.

The clean `stopstone`/`stopnetldi` runs against the directory the server actually registered in, which is what lets it stop cleanly instead of being killed; a force-kill sits behind that, and clears the lock it orphans.

**Identity is the careful part.** Two databases can share a name, so the rule splits by consequence:

- An **unconfirmed stone** is never touched. Stopping a stranger's stone loses whatever its sessions had not committed, and no warning makes that recoverable.
- An **unconfirmed NetLDI** may be restarted, with the warning kept. A NetLDI started without `-l` carries nothing that could ever identify it, so demanding proof there would make the restart permanently unreachable in its commonest case — while the worst outcome of being wrong is dropped connections, not lost work.

If a stop or kill fails, Jasper reports both names, the PIDs, the registration directory, the `gslist` invocation that will show the servers where they actually are, and the lock files a kill leaves behind — rather than dead-ending.

## Folded in from the same investigation

- **The tree no longer shows a healthy *Running* for a database a login cannot reach.** Rows distinguish Running, Stopped, *not responding*, *not connectable* (the stone is fine but nothing can traverse its NetLDI), and *Running outside Jasper*. `classifyStartNeed` leads with the same `isConnectable` predicate the tree derives its statuses from, so the two agree by construction rather than by coincidence.
- **Discovery no longer inherits `GEMSTONE_NRS_ALL`** from the shell that launched the editor.
- **`deleteDatabase` and `replaceExtent` could act under a live stone.** Both gated on "appears in Jasper's `gslist`", which is false for an externally started stone even though that process has the extent open. Both now refuse for a stone alive anywhere on the host, and say what to do about it.
- **"GemStone \<version> not found. Please extract it first."** now names the root Jasper searched and the directory it looked for.

## Real situations this ran into

The manual pass was run on a working developer machine rather than a clean one, and every one of these came from that rather than from a contrived test:

**A shell profile containing `unset GEMSTONE`.** With no interactivity guard, it ran for non-interactive shells too. Jasper built a correct environment, handed it to `startstone`, and the shell in between deleted the one variable that mattered — so GemStone reported `GEMSTONE` undefined while Jasper had demonstrably set it, and the error read as a broken install. It reached both the spawned command and the database terminal.

This is the same failure as the one this PR is about, pointed the other way: that one is the launching shell's leftovers leaking into what Jasper *discovers*, this is the shell overwriting what Jasper *sets*. Both are now closed. The `ulimit` wrapper runs with `--noprofile --norc` and a blanked `BASH_ENV`; the database terminal re-exports Jasper's values after the user's startup files have run, the way the WSL terminal already did.

**A shell exporting `GEMSTONE_NRS_ALL=#dir:…`,** which `startnetldi` folded into its own NRS string — a live instance of the stray-`NRS_ALL` hazard the discovery fix above addresses, found by accident while setting up an unrelated test.

**Several worktrees on one machine, each running a test stone under the same name.** Exactly the same-name ambiguity the identity check exists to resolve. It also caught a test in this PR picking another checkout's process by name alone.

**A NetLDI started by hand without `-l`,** which is the ordinary way anyone starts one — and which is why the identity gate had to be relaxed for NetLDIs.

## Testing

`npm run lint && npm run format:check && npm run compile && npm test` clean — **6,600 tests**.

Mostly unit-level, plus an integration test that runs the scan against a live stone and reproduces the bug end to end: the same servers and the same `gslist` binary, pointed at a different `GEMSTONE_GLOBAL_DIR`, report "no GemStone servers" while `ps` still sees them.

**A full manual UI pass was run on Linux — all 8 scenarios pass.** `EXTERNAL-SERVERS-UI-TEST-PLAN.md` records it, ordered by risk rather than setup convenience, with `external-server-test-env.sh` to set up a shell for the ones that need servers started outside Jasper.

That pass found five real bugs, all fixed here:

1. The shell-clobbered `GEMSTONE` above.
2. The matcher meant to replace GemStone's bare complaint was written from the wording in the issue report — a paraphrase GemStone never emits — so it matched nothing while its tests passed against the same paraphrase. It is now built from the message templates in the `startstone` binary.
3. Its replacement message then asserted "this usually means a server was started outside Jasper's environment" in a case where nothing external was running — the same confident misdirection in new clothes. It now states only what is certain and leads with a diagnostic.
4. The identity gate refused to restart a NetLDI it could not identify, which is a NetLDI's normal state.
5. The row's action offered **Restart & Connect** but had no login to connect to. It now says **Restart** / **Leave as-is** from that path.

The dialog also named the wrong server in its warning (the database's stone rather than whichever server was actually external), and had verb-agreement bugs in the single-server case. Both fixed.

## Not verified

**macOS.** Two things: `ps -Ao` selecting detached daemons is reasoned from flag semantics rather than tested — `-e` is *not* a selection flag on BSD, which would have made the feature silently inert on every Mac had it shipped as first written. And if `ps eww` will not expose another process's environment there, the registration directory is unknown, identity stays unconfirmed, and **Restart & Connect** is never offered for a stone — only *Connect as-is*. Detection itself still works. The integration test asserts this loudly so a macOS runner reports it rather than degrading quietly.

**Windows / WSL.** The path conversion between a Windows database path and the WSL paths a server reports is unit-tested in both directions but has never been run. The live-stone integration test skips on Windows with a stated reason: it calls `ps` directly rather than through `wslExecSync`, and Windows has no `ps` — Jasper reaches the process table through WSL there, which that test does not drive. The rest of the Windows matrix runs as normal.

## Suggested review order

1. `client/src/externalServerScan.ts` — parsing and identity rules, the load-bearing part
2. `client/src/processManager.ts` — `getExternalServers`, `stopExternalServer`, `killHostServer`
3. `client/src/externalServerReconcile.ts` — the decision flow and the wording
4. `client/src/wslBridge.ts` — small, but it changes how every Jasper-spawned command gets its environment
5. Everything else is wiring


HANDOFF.md
```diff
@@ -0,0 +1,102 @@
+# Handoff — issue #472, servers started outside Jasper's environment
```
MatiasFernandez: 🚫 Is this meant to be merged? It reads as a session-parking note rather than project documentation — it names your worktree path (`/export/uffda1/users/ewinger/worktrees/...`), calls the PR a draft, and lists intermediate SHAs that will not survive a squash. Everything durable in it is already in the PR description and the CHANGELOG. Can we drop it from the branch?

client/src/processManager.ts
```diff
@@ -458,6 +853,14 @@ export class ProcessManager {
...
       });
       terminal.show();
+      // Re-export after the shell has started, the way the WSL branch above
+      // already does. `env` is applied *before* the interactive shell reads the
+      // user's startup files, so an `unset GEMSTONE` or a competing
+      // `export GEMSTONE=…` in .bashrc silently wins — and this terminal is
+      // supposed to be the one place a user can trust to have the database's
+      // environment. Sending the exports afterwards makes Jasper's values
+      // authoritative without taking the user's shell customizations away.
+      terminal.sendText(exportCommand(env), true);
```
MatiasFernandez: 💭 This re-export includes `PATH`, which `versionEnvironment` sets to the fixed `${gsPath}/bin:/usr/local/bin:/usr/bin:/bin`. Since it is sent *after* the startup files have run, anything the user's `.bashrc`/`.zshrc` prepended — nvm, homebrew, pyenv, cargo — is gone from that terminal, so the comment's "without taking the user's shell customizations away" isn't quite true for the one variable most likely to be customized.

Note the WSL branch above doesn't have this problem: it exports and *then* `exec bash`, so the interactive shell's rc files still win.

Should we re-export only the GemStone variables here, and leave `PATH` to the `env` option on `createTerminal` (which the shell can then extend)? Something like `exportCommand(omit(env, 'PATH'))`, or an explicit `export PATH="${gsPath}/bin:$PATH"` so the prefix is asserted without discarding the rest.

client/src/databaseManager.ts
```diff
@@ -299,19 +299,45 @@ export class DatabaseManager {
...
   async deleteDatabase(db: GemStoneDatabase): Promise<boolean> {
-    // Check if processes are running
-    if (this.processManager.isStoneRunning(db.config.stoneName, db.config.version)) {
-      vscode.window.showErrorMessage(
-        `Stone "${db.config.stoneName}" is still running. Stop it before deleting.`,
-      );
+    // isServerAlive, not isStoneRunning: a server started outside Jasper's
+    // environment is absent from Jasper's gslist but has the extent open, and
+    // deleting the directory under it would corrupt a running database.
+    if (this.processManager.isServerAlive(db, 'stone')) {
```
MatiasFernandez: 💭 `isServerAlive` reads `cachedProcesses` and — through `getExternalServers` — the memoized per-refresh verdict, and neither `gemstone.deleteDatabase` nor `gemstone.replaceExtent` calls `refreshProcesses()` first. So if the user starts a stone by hand after the last tree refresh and then hits Delete, `gslistReadable`/`cachedExternal` can answer from a reading that predates the stone, and we're back to deleting under a live database — the exact failure this guard was widened to prevent.

`gemstone.restartExternalServers` in [extension.ts](client/src/extension.ts) does re-read for this reason ("the user may have dealt with the server by hand since"). Should the destructive guards do the same — a `refreshProcesses()` at the top of `deleteDatabase`/`replaceExtent`?

client/src/wslBridge.ts
```diff
@@ -487,11 +487,21 @@ export function wslSpawn(cmd: string, args: string[], env?: Record<string, strin
...
     return spawn('wsl.exe', wslArgs, { env: process.env });
   }
-  const mergedEnv = { ...process.env, ...env };
+  // BASH_ENV is the one way a non-interactive `bash -c` can still be made to
+  // read a startup file, and a startup file is free to clobber the environment
+  // Jasper just built. Users really do have `unset GEMSTONE` in their .bashrc —
+  // and a GemStone command whose GEMSTONE has been unset out from under it
+  // reports the variable as undefined, which reads as a broken install rather
+  // than as someone else's shell rewriting Jasper's environment.
+  const mergedEnv = { ...process.env, ...env, BASH_ENV: '' };
```
MatiasFernandez: 💭 `wslSpawn` is now protected from a startup file rewriting the environment, but `wslExecSync` right below still does a plain `execSync(cmd, { env: { ...process.env, ...env } })` with no `BASH_ENV: ''`. `execSync` runs through `/bin/sh`, which is bash on a fair number of distros, so the same `unset GEMSTONE` in a `BASH_ENV` file can reach it — and `gslist -cvl` (via `refreshProcesses`) goes through that path.

That would land as an unreadable gslist, which the new code correctly refuses to interpret, so it fails safe rather than wrong. Still — is there a reason to leave the two asymmetric, or should `wslExecSync` blank `BASH_ENV` too?

client/src/externalServerReconcile.ts
```diff
@@ -0,0 +1,334 @@
...
+  if (different.length > 0) {
+    const subject = different.map((s) => `${s.kind} "${s.name}"`).join(' and ');
+    const verb = different.length > 1 ? 'were' : 'was';
+    return (
+      `Warning: the paths ${subject} ${verb} started with point outside this database's ` +
+      `directory, so ${different.length > 1 ? 'these are' : 'this is'} probably a different ` +
+      `database that happens to share the name.`
+    );
+  }
+  const unknown = report.servers.filter((s) => s.identity === 'unknown');
```
MatiasFernandez: 💭 When `report.servers` is empty, this falls through to the `unknown` branch with an empty `subject` and `unknown.every(...)` vacuously true, producing "Warning: Jasper could not confirm this is  belonging to this database rather than another of the same name — the running process gave nothing away about which database it serves."

`reconcileTitle` explicitly defends against the zero-server case and `describeExternalServers` has a test for it ("has nothing to confirm when nothing was found"), so the shape is clearly considered reachable — but `reconcileMessage` takes the `else` branch there and emits that sentence. Both current callers guard with `hasExternalServer`/`need.kind === 'external'`, so I don't think it can happen today; it's more that the two functions disagree about whether it can. Worth an early return in `reconcileMessage` for `servers.length === 0`?

external-server-test-env.sh
```diff
@@ -0,0 +1,83 @@
...
+#
+# Source this in EVERY terminal you use for these tests. The variables it sets
+# do not survive a new terminal, and the failure mode when they are missing is
+# quiet rather than loud: $DB expands to nothing, so `-l "$DB/log/x.log"`
+# becomes `-l /log/x.log` and GemStone reports a missing directory instead of a
+# missing variable.
+#
+# Safe to source more than once.
+
+export ROOT=/uffda1/users/ewinger/jasperStones
```
MatiasFernandez: 💭 These are your machine's paths (`/uffda1/users/ewinger/...`), so anyone else sourcing this gets a shell pointed at directories that don't exist — and per the header comment the failure mode is quiet. Should these default from the environment instead, e.g. `ROOT=${ROOT:-$HOME/Documents/GemStone}` / `GS=${GS:-$ROOT/GemStone64Bit$VERSION-...}`, with a check that `$GS/bin/startstone` exists before exporting anything?

Separately: this plus `EXTERNAL-SERVERS-UI-TEST-PLAN.md` are two new files at the repo root, next to `README`/`CHANGELOG`/`ROADMAP`. Would `docs/` (or `scripts/`) be a better home, given they're for one feature's manual pass?

package.json
```diff
@@ -2826,6 +2832,16 @@
...
         },
+        {
+          "command": "gemstone.restartExternalServers",
+          "when": "view == gemstoneDatabases && viewItem =~ /^gemstoneDb(Stone|Netldi)External$/",
+          "group": "inline@1"
+        },
+        {
+          "command": "gemstone.restartExternalServers",
+          "when": "view == gemstoneDatabases && viewItem =~ /^gemstoneDb(Stone|Netldi)External$/",
+          "group": "navigation@1"
```
MatiasFernandez: nit: This registers the command twice on the same `when` — once in `inline@1` and once in `navigation@1`. Every other `gemstoneDatabases` item entry in this section (`startStone`, `stopStone`, `startNetldi`, `stopNetldi`, `replaceExtent`, …) uses `inline` only. Is the second one deliberate, to also get a right-click entry that the siblings don't have?

MatiasFernandez (approved):

client/src/processManager.ts
ericwinger: Good catch — fixed in 21f87b6a. Re-exporting the fixed PATH did wipe the user's nvm/homebrew/pyenv/cargo additions. The re-export now asserts GemStone's bin as a *prefix* (`export PATH='<gsBin>':"$PATH"`) and leaves the rest of the user's PATH intact; only the non-PATH GemStone vars are re-exported verbatim. Comment updated to match, and I added a test that the native re-export prepends rather than replaces.

client/src/databaseManager.ts
ericwinger: Fixed in 21f87b6a. `deleteDatabase` and `replaceExtent` now call `refreshProcesses()` before the `isServerAlive` guard, the same way `restartExternalServers` does, so a stone the user started by hand after the last tree refresh is seen rather than deleted/replaced under. Added a test that the guard re-reads before it checks.

client/src/wslBridge.ts
ericwinger: Fixed in 21f87b6a — `wslExecSync` now sets `BASH_ENV: ''` as well, for parity with `wslSpawn`. As you noted it already failed safe (an unreadable gslist is refused, not misread), but there's no reason to leave the two asymmetric. Added a test that the Linux branch blanks BASH_ENV.

package.json
ericwinger: Deliberate. The `navigation@1` entry is what gives the external-server rows a right-click "Restart Under Jasper's Environment" item; the `inline@1` one is the hover button. The ordinary start/stop siblings don't need the extra entry, but external reconciliation is the one action here worth making more discoverable — a user who spots the anomalous row is likely to right-click it — so it gets both. Happy to drop the right-click one if you'd rather keep the section inline-only.

PR #481, opened 2026-08-22
Breakpoints: aim at a step point, enable/disable, and a manager view

### LINKED ISSUE #262: Breakpoint support & step point support
We need gemstone breakpoint support in jasper. Looking at Jadeite for Pharo (and probably Jade) would be a good starting point for this.

1. Be able to set, enable, disable, clear breakpoints from method source.
2. Provide visibility of the step point that breaks are set at easily.
3. Visible in a source pane as well as debugger source pane
4. Breakpoint editor should allow group breakpoint settings including variable breakpoints (if possible). Jadeite for Dolphin had a decent one that could be reviewed.
5. We should also display step points when the cursor is hovering over a location in the source (or as near to the step point as possible).  Should look at Jade/JfP for inspiration. 
6. Other features?

---
_Migrated from GitLab: [GemStone/grail#74](https://code.gemtalksystems.com/GemStone/grail/-/work_items/74)_

ericwinger:
Closes #262.

Setting a breakpoint meant a gutter click, which lands on the leftmost step point of the line. A Smalltalk line routinely holds several, so the one you wanted was often not the one you got — and there was no way to disable a breakpoint, act on all of them at once, or see what the session actually had.

## The rule everything follows

A GemStone breakpoint lives in the **gem**. It is not in the repository — no `commit` persists it, verified on 3.6.2 and 3.7.5 — so it cannot outlive the thing it was set in. VS Code's breakpoint list is the *working* model (which is what makes the gutter, the per-breakpoint checkbox and the built-in Enable/Disable/Remove All controls drive GemStone for free), but deliberately **not** a durable one. A breakpoint is dropped from that list, not just from the gem, when:

- the **session** logs out — and anything VS Code's own cross-restart persistence brings back on startup is pruned;
- the **method** is recompiled — the new method is a different object, and after an edit the same step point may be different code, so re-applying would move the breakpoint somewhere the developer never put it.

Both of those were the opposite way round in the first draft of this branch. They changed because a marker that promises to stop execution it cannot stop is worse than no marker.

## What you get

**Aiming at a step point.** Precision rides on the breakpoint's column: a gutter click carries none and means the leftmost step point on the line, while an inline breakpoint or **Toggle Breakpoint at Cursor** (`Shift+F9`) carries the exact column and picks the step point nearest it. The token that will actually break is outlined — grey and faded when disabled, matching the grey the gutter dot goes.

**Enable / disable.** Per breakpoint from the Breakpoints view checkbox or the cursor commands; **Enable All** / **Disable All** / **Remove All** cover every GemStone breakpoint, including ones Jasper never set. A disabled breakpoint is applied as set-then-disable, because `disableBreakAtStepPoint:` is a no-op on a step point with no breakpoint.

**Step points, numbered but out of the way.** Numbers are inlay hints — VS Code's own dim style, suppressed or restyled by its `editor.inlayHints.*` settings like any other — each clickable to toggle a breakpoint there. `gemstone.stepPoints.display` defaults to `debugging`, so they appear while a debug session runs and stay out of the way while you are reading or writing code. Hovering a step point always reports its number and breakpoint state whatever the setting. Numbers are withheld while a buffer has unsaved edits, since the stone's offsets no longer describe the text on screen.

**A breakpoint manager.** A **Breakpoints** view in the GemStone sidebar lists what the gem actually holds, grouped by class and method with the step point each breakpoint resolved to. It is the gem's truth rather than a copy of VS Code's list, so it also surfaces breakpoints Jasper never set — from topaz, or a `halt` left in the code — which were invisible until execution stopped on one.

**Break on entry by naming a method.** The `+` button in VS Code's Breakpoints panel takes a name rather than a location. Type a bare selector and Jasper finds the implementors, asking which class you meant when there is more than one; a qualified `Account>>balance` or `Account class>>new` skips the prompt but is still checked against the image, since a typo would otherwise set a breakpoint that silently never fires. The name is then **replaced** by an ordinary breakpoint on the method's first step point, rather than carried as a second kind of breakpoint — so it gains a real location and red dot, enable/disable, and a row in the manager view, and cannot fight a source breakpoint over which of them owns a method's breaks.

Everything works in the debugger's source pane too, since the debugger opens ordinary editors.

## Unifying two managers that fought each other

The old cursor-toggle command kept its breakpoints in a private map, so they never appeared in the Breakpoints view **and were silently wiped whenever a gutter breakpoint on the same method was applied** (that path calls `clearAllBreaks` first). It now adds a normal VS Code breakpoint, so the two kinds are one kind.

## Bugs found along the way

- **A gutter breakpoint could land one step point late.** Line-to-step-point mapping compared GemStone's 1-based source offsets against 0-based line offsets, so a step point sitting exactly on a line boundary was read as belonging to the previous line. Run to Cursor already corrected for this; the gutter path did not.
- **`_allMethodBreakpoints`' tuple stride is version-dependent** — 3 fields on 3.6.2, 4 on 3.7.5, which gained `breakpointLevel`. Decoding it by hand walks off the end of the array on the older release. `getAllBreakpoints` reads the kernel's own `_breakReport:` instead, which knows its own stride. It also skips breakpoints stranded on a **superseded** method version, which a recompile leaves behind and the gem goes on reporting.
- **Cursor commands failed silently** in four cases — a non-method editor, an unsaved buffer, a diff view, a method whose step points can't be read. Invoked from a keystroke, a silent no-op is indistinguishable from a dead keybinding. Each now says why.
- **A named breakpoint's implementor search skipped environment 0.** `gemstone.maxEnvironment` is a ceiling, not a selection, and was passed straight through as *the* environment to search — so on a stone configured above 0 an ordinary selector came back as "nothing implements it". It now sweeps 0..max and remembers which environment each hit came from.
- **VS Code's `+` creates the breakpoint blank and only then opens it for editing**, and the typed name arrives as a *change*, not an addition. Rejecting the blank deleted the row before it could be typed into; watching additions alone never saw the name.

## Known limitation

VS Code's own **Deactivate Breakpoints** button greys the breakpoints out but cannot reach GemStone: `vscode.debug` exposes no activation state and no event for it, so an extension cannot observe the flag and the breakpoints stay armed in the gem. **Disable All Breakpoints** is the one that disarms them. Documented in the README.

## Testing

`npm run lint`, `npm run format:check`, `npm run compile` and `npm test` all clean: **6152 client, 322 server, 92 mcp-server**.

`breakpoints.integration.test.ts` pins the GemStone semantics the design rests on against a live stone, since none of them is documented anywhere we control — the disable-is-a-no-op rule, set-as-enable, out-of-range being silently ignored, breakpoints surviving a transaction abort, and the version-independence of the breakpoint report. **Run against both 3.6.2 and 3.7.5.**

Manually verified end to end in the Extension Development Host across 20 checks: gutter and cursor breakpoints, the panel and the manager view, enable/disable individually and globally, step point numbering and its dirty-buffer suppression, hover, reveal-from-list, the four global toolbar actions, all six named-breakpoint cases, actually stopping on a breakpoint, recompile, logout, and class-side methods.


client/src/breakpointManager.ts
```diff
@@ -49,138 +166,646 @@ export class BreakpointManager {
...
-            verified: true,
-          });
-          tracked.push({ stepPoint: result.stepPoint, actualLine: result.actualLine });
-        } catch {
-          results.push({ stepPoint: 0, actualLine: line, verified: false });
         }
-      } else {
-        results.push({ stepPoint: 0, actualLine: line, verified: false });
+        applied.push(bp);
+      } catch {
```
npapagna: **Correctness:** this catch block silently swallows every GCI failure from `setBreakAtStepPoint`/`disableBreakAtStepPoint` — no `showErrorMessage`/logging, unlike the deleted `selectorBreakpointManager.ts` which surfaced this. Worse: if `disableBreakAtStepPoint` throws *after* `setBreakAtStepPoint` already succeeded, the step point is left armed in the gem, but the entry is excluded from `applied` (only `verified = false` is set) — so the gutter shows an unverified/hollow marker while the gem will actually stop execution there, with zero user-visible feedback.

client/src/breakpointManager.ts
```diff
@@ -49,138 +166,646 @@ export class BreakpointManager {
...
-      sourceOffsets = queries.getSourceOffsets(
-        session,
-        method.className,
-        method.isMeta,
-        method.selector,
-        method.environmentId,
-      );
-    } catch {
-      return lines.map(() => ({ stepPoint: 0, actualLine: 0, verified: false }));
+    const info = this.stepPoints.fetch(session, uri, method);
```
npapagna: **Correctness:** if `stepPoints.fetch` fails here, right after `clearAllBreaks` already wiped the gem's breakpoints for this method, the function returns early without clearing `this.applied`, refreshing decorations, or firing `onDidApply`. A method that had breakpoints from a prior successful apply is left with stale decorations/hover/tree-view entries pointing at breakpoints that no longer exist in the gem, and nothing else cleans this up.

client/src/breakpointManager.ts
```diff
@@ -49,138 +166,646 @@ export class BreakpointManager {
...
+    this.byOop(bp, 'clearBreakAtStepPoint:');
+  }
+
+  /**
+   * The VS Code breakpoint behind a gem-reported one, or undefined when Jasper
+   * didn't set it. Matched on the method coordinates we applied plus the step
+   * point, rather than on the gem's dictionary/category strings, so it still
+   * matches when the same class name is bound in more than one dictionary.
+   */
+  private ownedBreakpoint(bp: GemStoneBreakpoint): vscode.SourceBreakpoint | undefined {
```
npapagna: **Correctness (latent — needs multiple concurrent sessions):** `ownedBreakpoint()` matches purely on class/selector/isMeta/environmentId/stepPoint, never checking that the candidate belongs to the currently selected session, even though sibling methods (`pruneOrphans`, `clearAllForSession`) already filter on the `gemstone://<sessionId>/` prefix available here too. With two sessions sharing a class/selector/step-point, toggling a breakpoint in one session's tree row could silently mutate a `SourceBreakpoint` belonging to a different session. Not exploitable today since Jasper only supports one open session, but worth fixing before multi-session support lands rather than rediscovering it then.

client/src/breakpointManager.ts
```diff
@@ -49,138 +166,646 @@ export class BreakpointManager {
...
-        const uri = bp.location.uri;
-        if (uri.scheme === 'gemstone') {
-          affectedUris.add(uri.toString());
-        }
+      if (bp instanceof vscode.SourceBreakpoint && bp.location.uri.scheme === 'gemstone') {
+        affected.add(bp.location.uri.toString());
       }
     }
+    for (const uriStr of affected) {
+      this.applyToUri(session, vscode.Uri.parse(uriStr));
```
npapagna: **Correctness (latent — needs multiple concurrent sessions):** `onBreakpointsChanged` applies breakpoint changes to whichever session is currently *selected*, not the session the changed breakpoint's own URI belongs to — `applyToUri` never checks `method.sessionId` against the passed-in session. With two sessions open, toggling a breakpoint in a method opened from session A while session B is selected would clear/set breakpoints against session B's gem using session A's coordinates. Same caveat as the `ownedBreakpoint` finding above: currently unreachable, worth tracking.

client/src/breakpointManager.ts
```diff
@@ -49,138 +166,646 @@ export class BreakpointManager {
...
+
+  /**
+   * Enable or disable every GemStone breakpoint.
+   *
+   * Flips Jasper's own breakpoints in VS Code's model first — that's the durable
+   * record, and it re-applies them through `onDidChangeBreakpoints` — then
+   * sweeps the gem, which also catches breakpoints Jasper never set (from topaz,
+   * another tool, or a `halt` in the code). "All" has to mean all of them.
+   */
+  setAllEnabled(enabled: boolean): void {
```
npapagna: **Correctness (latent — needs multiple concurrent sessions):** `setAllEnabled`/`removeAll` flip or remove VS Code breakpoint rows across *every* live session, but only sweep the *currently selected* session's gem (`queries.disableAllBreakpoints(session)` etc. only target one session). With two live sessions, "Disable All Breakpoints" would show every row as disabled while the non-selected session's gem breakpoints stay armed. Same latent-until-multi-session caveat as above.

client/src/functionBreakpoints.ts
```diff
@@ -0,0 +1,325 @@
...
+    });
+
+    logInfo(
+      `[breakpoints] ${qualifiedName(target)} entry is line ${entry.line} col ${entry.character}; ` +
+        `converting to ${uri.toString()}`,
+    );
+
+    // Replace, don't add alongside: the function breakpoint has done its job as
+    // a way of naming a method, and leaving it would show two rows for one break.
+    vscode.debug.removeBreakpoints([bp]);
```
npapagna: **Correctness:** `convert()` never re-checks that the original `FunctionBreakpoint` is still present after awaiting the class-picker prompt before calling `addBreakpoints`. If a bare selector resolves to multiple classes (QuickPick opens) and the developer deletes that function breakpoint from the Breakpoints panel while the picker is open, `convert()` proceeds regardless and resurrects the breakpoint the developer just removed. `inFlight` only guards against re-resolving the same pending name, not against mid-await deletion, and `event.removed` is never forwarded to `handle()`.

client/src/stepPointSelectors.ts
```diff
@@ -0,0 +1,134 @@
...
+        pos++;
+        continue;
+      }
+      if (ch === ')' || ch === ']' || ch === '}') {
+        depth--;
+        if (depth < 0) break;
+        pos++;
+        continue;
+      }
+      if (ch === '.' || ch === ';') break;
```
npapagna: **Correctness:** the statement-separator check (`.`/`;`) here fires regardless of bracket depth, unlike the depth-gated identifier scan elsewhere in this function. For a keyword send like `self assert: (x foo; bar) equals: true.`, the scan enters the parens (depth 1) but the `;` still breaks the loop immediately, so `equals:` is never found as a continuation keyword. No existing test combines paren-nesting with an internal separator, so this is an untested gap rather than an intentional rule.

client/src/stepPointHints.ts
```diff
@@ -0,0 +1,126 @@
...
+        if (e.affectsConfiguration('gemstone.stepPoints.display')) {
+          this.display = readDisplaySetting();
+          this.refresh();
+        }
+      }),
+    );
+  }
+
+  /** Flip the numbers on or off, and remember it in the user's settings. */
+  async toggle(): Promise<void> {
```
npapagna: **Correctness:** `toggle()` reads `this.visible()`/computes `next` synchronously before `await`ing the config update, and only assigns `this.display = next` after the await resolves. A rapid double-invocation (double-click on the editor-title icon, fast repeated keybinding) computes the same stale `next` both times, so two toggles collapse into one net change instead of flipping back and forth as the user would expect.

client/src/breakpointManager.ts
```diff
@@ -49,138 +166,646 @@ export class BreakpointManager {
...
+  );
+  vscode.debug.removeBreakpoints(breakpoints);
+  vscode.debug.addBreakpoints(replacements);
+}
+
+function positionOf(document: vscode.TextDocument, offset: number): vscode.Position {
+  return document.positionAt(offset);
+}
+
+function message(e: unknown): string {
```
npapagna: **Reuse:** `message(e: unknown)` is duplicated verbatim here and in `functionBreakpoints.ts:323`, when an equivalent, already-exported helper (`messageOf` in `serverPlugin/installHelpers.ts`) exists and is already imported by 4 other files. Worth importing the shared helper instead of a 5th copy.

client/src/functionBreakpoints.ts
```diff
@@ -0,0 +1,325 @@
...
+  if (inst) return { className: inst[1], isMeta: false, selector: inst[2].trim() };
+
+  // A bare selector. Reject anything holding '>>', which was a qualified name
+  // the patterns above failed on — a malformed class half, most likely.
+  if (name.includes('>>')) return null;
+  return { isMeta: false, selector: name.replace(/^#\s*/, '') };
+}
+
+/** How a name should be shown once it has been pinned to one class. */
+export function qualifiedName(target: MethodSearchResult): string {
```
npapagna: **Reuse:** `qualifiedName()` reimplements `methodResultsPicker.ts`'s `describeMethodResult()` for the same `MethodSearchResult` type, and the format has already diverged (`Account class>>reset` here vs. `Account class >> #reset` from the shared helper). Log lines and reject messages here read inconsistently with every other method-list feature (Senders/Implementors/References, safe-delete) that already uses the shared helper.

client/src/queries/getAllBreakpoints.ts
```diff
@@ -0,0 +1,121 @@
...
+ * Dictionary and category come back too, so a caller can open the method in an
+ * editor without a second round trip per breakpoint. The dictionary is matched
+ * by class *identity*, not by name, so a class name shadowed in two
+ * dictionaries resolves to the one actually holding this class.
+ */
+export function getAllBreakpoints(execute: QueryExecutor): GemStoneBreakpoint[] {
+  const code = `| ws sl dictOf isCurrent |
+ws := WriteStream on: String new.
+sl := System myUserProfile symbolList.
+dictOf := [:aCls | | base found |
```
npapagna: **Reuse:** this `dictOf` block hand-rolls the "walk symbolList to find which slot holds this class object by identity" algorithm instead of calling the existing `symbolListIndexOfClassExpr` helper in `queries/util.ts`, which four other query files already reuse for exactly this purpose (that file documents itself as "the single home for this idiom"). A future fix to the identity-lookup idiom is likely to miss this call site.

client/src/breakpointManager.ts
```diff
@@ -201,7 +826,7 @@ export function buildLineOffsets(source: string): number[] {
```
npapagna: **Reuse:** `buildLineOffsets`/`mapOffsetToStepPoint`/`mapLineToStepPoint` in this file duplicate `stepPointModel.ts`'s canonical `buildLineStarts`/`resolveStepPoint` algorithm (that file is explicitly documented as "the single home for where are this method's step points"). These are also the very helpers the still-unconverted `sourceReference` branch in `gemstoneDebugSession.ts` relies on instead of the shared resolver (see comment there). Two copies of the same "leftmost on line, else nearest by column, else fall forward" logic risk drifting apart on a future fix.

client/src/gemstoneDebugSession.ts
```diff
@@ -156,6 +156,10 @@ export class GemStoneDebugSession extends DebugSession {
     }
 
     const requestedLines = args.breakpoints.map((bp) => bp.line);
```
npapagna: **Correctness / consistency:** this diff only threads `requestedColumns` into the `args.source.path` branch below. The `sourceReference` branch further down (doit/decompiled frames) is untouched — it still uses the old line-only `mapLineToStepPoint` path in `breakpointManager.ts` and calls `queries` directly, bypassing `BreakpointManager` entirely. Setting a breakpoint on a doit/'Executed Code' frame (exactly the case this PR calls out as needing first-class support) ignores the DAP column, never populates `BreakpointManager.applied`, and is invisible to `pruneOrphans`/`clearAllForSession`/Enable All/Disable All.

client/src/stepPointModel.ts
```diff
@@ -0,0 +1,325 @@
...
+      return { problem: 'This method has no step points to break at.' };
+    }
+    return { info };
+  }
+
+  /**
+   * Step points for a method identified by coordinates rather than an open
+   * document — the path the debug adapter and the breakpoint applier take.
+   */
+  fetch(session: ActiveSession, uri: vscode.Uri, method: MethodUriRef): StepPointInfo | null {
```
npapagna: **Efficiency:** `fetch()` issues three sequential GCI round trips per method (`getMethodSource`, `getSourceOffsets`, `getStepPointSelectorRanges`) instead of one combined query. `provideInlayHints`/`provideHover` call into this synchronously on the extension host, so the first hover or inlay-hint render on a freshly opened method blocks the extension host for 3 round trips instead of 1 — visible stutter exactly when a developer opens a method to set a breakpoint.

client/src/breakpointManager.ts
npapagna: 🚫 **BLOCKER — Correctness:** this catch block silently swallows every GCI failure from `setBreakAtStepPoint`/`disableBreakAtStepPoint` — no `showErrorMessage`/logging, unlike the deleted `selectorBreakpointManager.ts` which surfaced this. Worse: if `disableBreakAtStepPoint` throws *after* `setBreakAtStepPoint` already succeeded, the step point is left armed in the gem, but the entry is excluded from `applied` (only `verified = false` is set) — so the gutter shows an unverified/hollow marker while the gem will actually stop execution there, with zero user-visible feedback.

client/src/breakpointManager.ts
npapagna: **Correctness:** if `stepPoints.fetch` fails here, right after `clearAllBreaks` already wiped the gem's breakpoints for this method, the function returns early without clearing `this.applied`, refreshing decorations, or firing `onDidApply`. A method that had breakpoints from a prior successful apply is left with stale decorations/hover/tree-view entries pointing at breakpoints that no longer exist in the gem, and nothing else cleans this up.

client/src/breakpointManager.ts
npapagna: **Correctness (latent — needs multiple concurrent sessions):** `ownedBreakpoint()` matches purely on class/selector/isMeta/environmentId/stepPoint, never checking that the candidate belongs to the currently selected session, even though sibling methods (`pruneOrphans`, `clearAllForSession`) already filter on the `gemstone://<sessionId>/` prefix available here too. With two sessions sharing a class/selector/step-point, toggling a breakpoint in one session's tree row could silently mutate a `SourceBreakpoint` belonging to a different session. Not exploitable today since Jasper only supports one open session, but worth fixing before multi-session support lands rather than rediscovering it then.

client/src/breakpointManager.ts
npapagna: **Correctness (latent — needs multiple concurrent sessions):** `onBreakpointsChanged` applies breakpoint changes to whichever session is currently *selected*, not the session the changed breakpoint's own URI belongs to — `applyToUri` never checks `method.sessionId` against the passed-in session. With two sessions open, toggling a breakpoint in a method opened from session A while session B is selected would clear/set breakpoints against session B's gem using session A's coordinates. Same caveat as the `ownedBreakpoint` finding above: currently unreachable, worth tracking.

client/src/breakpointManager.ts
npapagna: **Correctness (latent — needs multiple concurrent sessions):** `setAllEnabled`/`removeAll` flip or remove VS Code breakpoint rows across *every* live session, but only sweep the *currently selected* session's gem (`queries.disableAllBreakpoints(session)` etc. only target one session). With two live sessions, "Disable All Breakpoints" would show every row as disabled while the non-selected session's gem breakpoints stay armed. Same latent-until-multi-session caveat as above.

client/src/functionBreakpoints.ts
npapagna: **Correctness:** `convert()` never re-checks that the original `FunctionBreakpoint` is still present after awaiting the class-picker prompt before calling `addBreakpoints`. If a bare selector resolves to multiple classes (QuickPick opens) and the developer deletes that function breakpoint from the Breakpoints panel while the picker is open, `convert()` proceeds regardless and resurrects the breakpoint the developer just removed. `inFlight` only guards against re-resolving the same pending name, not against mid-await deletion, and `event.removed` is never forwarded to `handle()`.

client/src/stepPointSelectors.ts
npapagna: **Correctness:** the statement-separator check (`.`/`;`) here fires regardless of bracket depth, unlike the depth-gated identifier scan elsewhere in this function. For a keyword send like `self assert: (x foo; bar) equals: true.`, the scan enters the parens (depth 1) but the `;` still breaks the loop immediately, so `equals:` is never found as a continuation keyword. No existing test combines paren-nesting with an internal separator, so this is an untested gap rather than an intentional rule.

client/src/stepPointHints.ts
npapagna: **Correctness:** `toggle()` reads `this.visible()`/computes `next` synchronously before `await`ing the config update, and only assigns `this.display = next` after the await resolves. A rapid double-invocation (double-click on the editor-title icon, fast repeated keybinding) computes the same stale `next` both times, so two toggles collapse into one net change instead of flipping back and forth as the user would expect.

client/src/breakpointManager.ts
npapagna: **Reuse:** `message(e: unknown)` is duplicated verbatim here and in `functionBreakpoints.ts:323`, when an equivalent, already-exported helper (`messageOf` in `serverPlugin/installHelpers.ts`) exists and is already imported by 4 other files. Worth importing the shared helper instead of a 5th copy.

client/src/functionBreakpoints.ts
npapagna: **Reuse:** `qualifiedName()` reimplements `methodResultsPicker.ts`'s `describeMethodResult()` for the same `MethodSearchResult` type, and the format has already diverged (`Account class>>reset` here vs. `Account class >> #reset` from the shared helper). Log lines and reject messages here read inconsistently with every other method-list feature (Senders/Implementors/References, safe-delete) that already uses the shared helper.

client/src/queries/getAllBreakpoints.ts
npapagna: **Reuse:** this `dictOf` block hand-rolls the "walk symbolList to find which slot holds this class object by identity" algorithm instead of calling the existing `symbolListIndexOfClassExpr` helper in `queries/util.ts`, which four other query files already reuse for exactly this purpose (that file documents itself as "the single home for this idiom"). A future fix to the identity-lookup idiom is likely to miss this call site.

client/src/breakpointManager.ts
npapagna: **Reuse:** `buildLineOffsets`/`mapOffsetToStepPoint`/`mapLineToStepPoint` in this file duplicate `stepPointModel.ts`'s canonical `buildLineStarts`/`resolveStepPoint` algorithm (that file is explicitly documented as "the single home for where are this method's step points"). These are also the very helpers the still-unconverted `sourceReference` branch in `gemstoneDebugSession.ts` relies on instead of the shared resolver (see comment there). Two copies of the same "leftmost on line, else nearest by column, else fall forward" logic risk drifting apart on a future fix.

client/src/gemstoneDebugSession.ts
npapagna: 🚫 **BLOCKER — Correctness / consistency:** this diff only threads `requestedColumns` into the `args.source.path` branch below. The `sourceReference` branch further down (doit/decompiled frames) is untouched — it still uses the old line-only `mapLineToStepPoint` path in `breakpointManager.ts` and calls `queries` directly, bypassing `BreakpointManager` entirely. Setting a breakpoint on a doit/'Executed Code' frame (exactly the case this PR calls out as needing first-class support) ignores the DAP column, never populates `BreakpointManager.applied`, and is invisible to `pruneOrphans`/`clearAllForSession`/Enable All/Disable All.

client/src/stepPointModel.ts
npapagna: **Efficiency:** `fetch()` issues three sequential GCI round trips per method (`getMethodSource`, `getSourceOffsets`, `getStepPointSelectorRanges`) instead of one combined query. `provideInlayHints`/`provideHover` call into this synchronously on the extension host, so the first hover or inlay-hint render on a freshly opened method blocks the extension host for 3 round trips instead of 1 — visible stutter exactly when a developer opens a method to set a breakpoint.

npapagna (changes requested):
Thanks @ericwinger! Left a couple of comments.

client/src/breakpointManager.ts
ericwinger: Fixed in `5e3d0c75`.

The bare `catch` is gone. The arming loop now handles the two calls separately:

- a failed `setBreakAtStepPoint` refuses that step point with the stone's own message, shown with `showErrorMessage` **and** carried back in `VerifiedBreakpoint.message` so the debug adapter relays it to the gutter — an unverified marker on its own is unreadable, since it looks exactly like a breakpoint on a line with no step point;
- the case you spotted — a `disableBreakAtStepPoint` that fails *after* the set succeeded — now calls `clearBreakAtStepPoint` to take the break back out, rather than leaving a step point armed behind a marker that reads "disabled". If that removal also fails, the message says the step point is still armed and will stop execution.

Failures are collected and reported once per apply rather than one toast per step point.

Three tests cover it: the set-fails path, the armed-but-could-not-disable path asserting `clearBreakAtStepPoint` runs, and the both-calls-failed path asserting the "still armed" wording. Each was mutation-checked — restoring the silent `catch` turns them red.

client/src/breakpointManager.ts
ericwinger: Fixed in `5e3d0c75`.

You were right that the early return leaves a lie behind: `clearAllBreaks` has already emptied the method, so anything the manager still remembers for it points at breakpoints that exist in no gem — and the decorations, hover text and tree rows keep being drawn from that record.

The path now does what the "nothing wanted" branch above it already did: deletes the URI from `applied`, refreshes the editors, fires `onDidApply`, and warns the developer that the method's breakpoints were cleared because its step points could not be read. The unverified results also carry that reason for the adapter.

Tested by dropping the step point cache (which is what a session re-select does) and making the next fetch throw: the applied record empties, `onDidApply` fires, and the warning is shown. Mutation-checked by restoring the early return.

client/src/breakpointManager.ts
ericwinger: Fixed in `5e3d0c75`, and it turned out to be reachable today rather than latent — see the note on the `onBreakpointsChanged` thread: concurrent sessions ship behind `gemstone.sessionMode: "multiple"`.

`ownedBreakpoint()` now skips any URI outside `gemstone://<selected session>/` before matching. The rows it matches against were read out of the selected session's gem, so only a method from that session can be behind one — the same prefix rule `pruneOrphans` and `clearAllForSession` already use, as you noted.

Tested with two sessions holding the same class, selector and step point: the row's disable action leaves the other session's VS Code breakpoint untouched and falls through to the by-OOP path instead. Worth saying that my first attempt at this test passed with the guard removed — it had no VS Code breakpoint in the list to flip — so I rewrote it until the mutation failed.

**Manually verified against a live stone by Eric**, with two sessions open on one method: disabling the row greys only its own session's marker.

client/src/breakpointManager.ts
ericwinger: Fixed in `5e3d0c75`.

One correction to the premise, because it changes the severity rather than the fix: this is **not** latent. Concurrent sessions ship today behind a setting — `gemstone.sessionMode: "multiple"` (`package.json`), with `evaluateLoginPolicy` in `sessionManager.ts` allowing the second login and the single-mode refusal message naming the setting. The README documents the mode, and its "Single vs. multiple sessions" note documents the exact state this mishandles: *"an open workspace/editor stays bound to the session that opened it even after you switch the active session."* So a developer who flips one documented setting hits it.

The fix is a new `sessionForUri()`: `onBreakpointsChanged` resolves each affected URI to the session in its own authority and applies there, skipping a URI whose session has logged out (pruning has already removed its row). `thawIfClean` and `frozenResults` go through it too. The early "no selected session, return" is gone, so a breakpoint in a live session's method is applied even when a different session is selected.

Three tests with two sessions live and the other one selected: the break is armed in the gem the method was opened from, the wrong gem's breakpoints on that method are not cleared, and a URI with no live session arms nothing.

**Manually verified against a live stone by Eric**: with the old code, a breakpoint set in one session's method editor while another session was active does not stop that session's execution.

client/src/breakpointManager.ts
ericwinger: Fixed in `5e3d0c75`.

`setAllEnabled` and `removeAll` now sweep **every logged-in session's** gem through a shared `sweepEveryGem` helper, since the rows they flip are one VS Code list spanning all sessions — a gem left un-swept keeps stopping execution behind a row that reads "disabled". One session failing no longer abandons the sweep: the rest are still swept, the failures are collected, and the message names the sessions that refused. `onDidApply` fires either way so the views redraw for the gems that did succeed.

Same note as the sibling threads: reachable today via `gemstone.sessionMode: "multiple"`, not latent.

Three tests: both gems swept for disable, both for enable, and one gem throwing while the other is still swept and named in the error.

**Manually verified against a live stone by Eric** across two sessions.

client/src/functionBreakpoints.ts
ericwinger: Fixed in `5e3d0c75`.

`convert()` now re-checks `vscode.debug.breakpoints.includes(bp)` after the class picker resolves, and returns quietly when the developer deleted the row instead of answering — they got what they asked for, so there is nothing to report. The check is deliberately only on that path: the single-candidate path never yields, so the list cannot move under it.

One thing worth recording for whoever reads this later: this really is unreachable in the shipped UI, and for a reason nobody wrote down. `showQuickPick` sets no `ignoreFocusOut`, so clicking the Breakpoints panel to delete the row dismisses the picker first and takes the reject path. Nothing else can remove a `FunctionBreakpoint` mid-await, since `gemstoneBreakpoints()` filters to `SourceBreakpoint` with the `gemstone` scheme. So it is a real gap held shut by an untested default that a future `ignoreFocusOut: true` would reopen — which is why it is fixed with a unit test rather than left alone. No manual repro exists.

client/src/stepPointSelectors.ts
ericwinger: Fixed in `5e3d0c75`.

The separator now ends the keyword send only at depth 0, matching the depth-gated identifier scan below it. A `.` or `;` inside parentheses or a block belongs to an inner expression, so `self foo: (s add: 1; yourself) bar: 2` and `self foo: [a bar. c baz] qux: 2` keep their trailing keywords — each had been losing everything after the bracket, which meant no step point number, no hover and nowhere to aim a breakpoint on that part of the send.

Three tests, written red first and mutation-checked both ways: reverting the depth gate fails three, and removing the stop entirely fails two.

**Manually verified against a live stone by Eric.**

client/src/stepPointHints.ts
ericwinger: Fixed in `5e3d0c75`.

`toggle()` now assigns `this.display = next` **before** awaiting the configuration write, so a second invocation arriving mid-await reads the first one's answer and flips back instead of computing the same stale `next`. The configuration listener still redraws when the write lands; the eager assignment only decides what the next call sees.

Tested, and **manually verified by Eric** — a fast double-click of the editor-title button now leaves the numbers where they started rather than flipped.

client/src/breakpointManager.ts
ericwinger: Fixed in `5e3d0c75`. The local `message()` is deleted and `messageOf` from `serverPlugin/installHelpers` is imported instead, so there is no fifth copy. A grep confirms no `message(` helper is left anywhere in the tree.

The same finding caught a second copy I had just written: the new failure messages needed to name a method, and I had added a local `describeMethod` — which is the duplication your `qualifiedName` comment objects to. That is gone too; see the reply on that thread.

client/src/functionBreakpoints.ts
ericwinger: Fixed in `5e3d0c75`. `qualifiedName()` is deleted and every call site — the candidate list, the reject messages and the log lines — uses `describeMethodResult`, so the function-breakpoint flow now reads `Account class >> #reset` like Senders, Implementors, References and safe delete.

One extension, since the shared helper could not serve every caller as written: its parameter is now the three parts of a name (`{className, isMeta, selector}`) rather than a whole `MethodSearchResult`. Body and output are unchanged and every existing caller still fits — but it let the breakpoint manager's new messages use it instead of growing yet another copy of the format, which is the failure mode this finding is about.

client/src/queries/getAllBreakpoints.ts
ericwinger: Fixed in `5e3d0c75`. The hand-rolled walk is gone; `dictOf` now calls `symbolListIndexOfClassExpr` from `queries/util.ts` and reads the name off that index, so the identity-lookup idiom has one home again and a future fix to it cannot miss this call site.

client/src/breakpointManager.ts
ericwinger: Fixed across `746162fc` and `5e3d0c75` — all three copies are gone.

`mapLineToStepPoint` went with the `sourceReference` work. `buildLineOffsets` and `mapOffsetToStepPoint` are now deleted too (59 lines, plus 58 lines of their tests, which duplicated coverage `stepPointModel.test.ts` already has). Their one remaining caller, the debugger panel's Run to Cursor, builds a `StepPointInfo` and calls the canonical `stepPointAtOffset`, so Run to Cursor and a breakpoint set in an editor now resolve step points through the same code and cannot drift apart. The stone's offsets are 1-based and the shared model is 0-based, so the conversion happens where the info is built.

A stale comment in `debuggerView.js` claiming Run to Cursor is unavailable on an **Executed Code** frame is corrected while I was there: such a frame is breakable — `breakable` is `homeMethodOop !== 0n`, and a doit is broken in by method OOP.

**Manually verified against a live stone by Eric**: Run to Cursor on an Executed Code frame stops at the send nearest the cursor (`sqrt` in `a factorial + a sqrt truncated`), not at the leftmost one on the line.

client/src/gemstoneDebugSession.ts
ericwinger: Addressed in `746162fc` and `30c86eb6`, but by **refusing the case rather than threading it through `BreakpointManager`** — so this is a decision to check rather than the fix you asked for.

The reasoning: a doit's method is compiled for one execution and is gone afterwards, so a breakpoint on it could never be hit again. Giving it first-class support would mean maintaining a breakpoint that can only ever fire during the execution that is already stopped in front of you. Eric's call was that offering it is worse than refusing it.

So the frame source view no longer claims a language, which is what made VS Code offer its gutter there (that also cost its syntax highlighting; the debugger panel's own source pane still highlights), and the adapter refuses a `sourceReference` request that arrives another way — `debug.allowBreakpointsEverywhere`, or a non-VS-Code client — with a message saying to set the breakpoint in the method instead. The column-ignoring `mapLineToStepPoint` path is deleted rather than fixed.

Two things this does not resolve, both filed: the debugger panel's companion tab still shows a gutter and refuses the click ([#515](https://github.com/GemTalk/Jasper/issues/515)), and Run to Cursor on such a frame does still work — it breaks by method OOP, which is legitimate because the break is temporary and belongs to the execution in front of you.

Since the `args.source.path` branch is now the only one that carries a refusal reason out to the developer, it has tests it did not have before: a refused breakpoint arrives as `reason: 'failed'` plus the message, an accepted one carries neither, and the DAP column is forwarded.

client/src/stepPointModel.ts
ericwinger: Fixed in `5e3d0c75`. `fetch()` now makes one round trip: a new `queries/getStepPointBundle.ts` asks for source, step point offsets and selector ranges together, framed so the variable-length parts come first and the source is carried verbatim at the end — no escaping. It is also less work for the stone, which computed the first two to answer the third anyway.

Worth flagging what that traded away and how it is now covered. Replacing three proven queries with one new doit put an untested query on the hover and inlay-hint hot path: every unit test mocked it, so a syntax error or any difference in `sourceString` / `_sourceOffsets` across releases would have stopped step point numbers entirely with the suite still green. `20f30fdb` closes that with integration cases against a live stone, ungated so they run on both boundaries in CI:

- the bundle's three parts equal what `getMethodSource`, `getSourceOffsets` and `getStepPointSelectorRanges` return for the same method;
- the reply is not merely empty — a returned selector range lands on real source text (equality alone would pass on two empty answers);
- a fixture whose source carries a tab-indented line, blank lines, a `'1,2,3'` literal shaped like the offsets header and a tab-separated literal shaped like a selector row round-trips verbatim, which is what pins the framing.

ericwinger:
Thanks @npapagna — all fourteen are addressed, each with a reply in its own thread and the thread resolved. Branch is up to date with `main`.

**Both blockers.** The swallowed GCI failure now reports the stone's message and, in the case you spotted, removes a break that armed but could not be disabled rather than leaving it stopping execution behind a marker that reads "disabled". The doit/decompiled blocker went a different way than you asked: the case is now **refused** rather than threaded through `BreakpointManager`, because a doit's method is compiled for one execution and gone afterwards, so a breakpoint on it could never be hit again. That is a decision worth your eye — the reasoning is in that thread, along with the residual filed as #515.

**One premise correction, because it changes severity rather than the fix.** The three session-scoping findings were called latent on the grounds that Jasper supports only one open session. Concurrent sessions ship today behind `gemstone.sessionMode: "multiple"`, documented in the README and in the CHANGELOG entry that introduced Logins & Sessions — and the README's own note describes the exact state the code mishandled: *"an open workspace/editor stays bound to the session that opened it even after you switch the active session."* All three were reachable by a developer flipping one documented setting, and all three are fixed and tested.

**Verified by hand against a live stone, by Eric** — as distinct from the rest, which rest on automated tests:

- a breakpoint is applied to the gem its method was opened from, not the selected session;
- a Breakpoints view row disables only its own session's breakpoint;
- Disable All reaches every logged-in session's gem;
- the keyword scan carries past a separator nested inside brackets;
- the step point numbers toggle survives a fast double-click;
- Run to Cursor through the shared resolver stops at the send nearest the cursor on an Executed Code frame.

**Also in this round**, from a second pass over the branch after your review:

- the one-round-trip step point query had never executed against a stone — every test mocked it, so a syntax error or a release difference in `sourceString` / `_sourceOffsets` would have stopped step point numbers entirely with the suite green. It now has ungated integration cases comparing its three parts against the three queries it replaced, proving the reply is not merely empty, and pinning the framing with source containing tabs, blank lines and header-shaped text;
- the `args.source.path` branch of `setBreakpointsRequest` — the only path that now carries a refusal reason to the developer — has tests for the reason arriving, for an accepted breakpoint carrying none, and for the column being forwarded;
- a method held still while its editor was dirty was only ever released by a revert, never by a save, so the URI leaked for the life of the window; it is released on recompile and swept on logout;
- CHANGELOG entries for four fixes that had none, and the README now states that a breakpoint belongs to its own session's gem and that Enable/Disable/Remove All sweep every logged-in session.

Every source change was mutation-checked: reverted in turn, with the suite required to go red. Gate is green — lint, formatting, compile, 6,818 client tests, 322 server, 92 mcp, including the live-stone integration tests.

npapagna (approved):
Thanks @ericwinger! 🚀

Issue #492, opened 2026-08-25
RB engine compiles through a selector marked for deprecation that hardwires method environment 0


ericwinger:
## Summary

Every method the refactoring engines compile goes into method environment 0, because the
selector they compile through hardwires it. GemStone also marks that selector for possible
deprecation. Both point at the same migration.

`Behavior>>compileMethod:dictionaries:category:` is defined in the base image as:

```smalltalk
compileMethod: sourceString
dictionaries: aSymbolList
category: aCategoryString

"May be deprecated in the future.
 Returns nil, or an Array of error descriptors"

^ [ self compileMethod: sourceString
       dictionaries: aSymbolList
       category: aCategoryString
       environmentId: 0 .
    nil
  ] onException: CompileError do:[:ex | ex errorDetails ]
```

Two consequences for us:

1. **The environment is not ours to choose.** It is `0`, always, with no parameter to pass.
2. **We compile through a selector its own authors flag as possibly going away.**

The environment-aware form is the six-keyword one, present in the base image of every
supported release (verified in `upgrade/projects/gemstoneBaseImage/rowan/src/Filein1A/Behavior.extension.st`
on both 3.6.2 and 3.7.5):

```smalltalk
compileMethod: source
dictionaries: aSymbolList
category: categorySymbol
intoMethodDict: nil
intoCategories: nil
environmentId: envId
```

It answers the compiled `GsNMethod` on success and **signals** `CompileError` on failure,
rather than answering an Array of error descriptors.

## Why the signalling form is the interesting part

That difference is worth more than the environment parameter.

The engines shipped a whole class of defect — 17 sites across 12 engines — that existed only
because a compile failure arrived as a *return value* nobody was obliged to look at. A method
that would not compile was silently dropped, or worse, deleted from its source class while the
refactoring reported success. That work is tracked in
https://github.com/GemTalk/Jasper/issues/412 and is fixed by funnelling every compile through
one checked helper.

But that fix rests on discipline: one method is now careful, and a future engine that reaches
for `compileMethod:` directly re-opens the hole. Compiling through a form that **signals**
removes the discipline requirement — an ignored failure becomes impossible rather than merely
discouraged, since the apply loops already guard every change with `on: Error do:`.

## Current state

After the compile-checking fix, `gs-src/refactoring/engine/` contains exactly ONE send of
`compileMethod:`, in `GsRefactoringEnvironment>>compileFailureFor:into:category:`. Everything
[...]

PR #493, opened 2026-08-25
Check every compile result in the refactoring engines

### LINKED ISSUE #412: RB: A failed compile can lose a method — 17 unchecked compileMethod: sites across 12 refactoring engines
## Summary

`Behavior>>compileMethod:dictionaries:category:` reports a compile failure in its **return
value** — `nil` (or an empty `Array`) on success, an `Array` of error tuples otherwise — and
does **not** signal. Any send that ignores the result therefore reads as success while the
method is never installed.

PR #411 fixed this for three engines by introducing
`GsRefactoringEnvironment>>compile:into:category:`, which raises on a failed compile so the
failure lands on the path every apply loop already guards with `on: Error do:` and is recorded
in the apply result's `failed` list.

**That helper has only three senders.** Every other engine still sends `compileMethod:`
directly and discards what comes back — 17 sites across 12 engines. The user is told the
refactoring succeeded in all of them.

Senders of the safe helper (correct today):

- `GsRenameClassRefactoring.class.st:665`
- `GsRenameMethodRefactoring.class.st:342`
- `GsRenameClassVariableRefactoring.class.st:505`

Also correct today, checking the return value directly:
`GsInstVarRefactoring.class.st:685` and `:713`,
`GsRenameInstanceVariableRefactoring.class.st:416`.

### Tier 1 — the method is lost from both classes

`GsPushUpMethodRefactoring>>applyMethodAdd:` (`GsPushUpMethodRefactoring.class.st:516`)
discards the compile result, then runs `appliedAddSelectors add:` and `^true`
unconditionally. `applyMethodRemove:` guards only on that set, so a failed compile still
authorizes `removeSelector:` on the source class. The method exists on neither class
afterwards and the refactoring reports success.

This is the most severe case in the group — worse than the class-rename one below, where the
superseded class version still holds the method in class history.

### Tier 2 — silently dropped from the new class version

A new class version starts with an empty method dictionary, so the copy-forward loop is what
carries behaviour across. A method that will not recompile is dropped and nothing warns:

| Site | Method |
| --- | --- |
| `GsRenameClassRefactoring.class.st:716` | `copyMethod:from:to:` |
| `GsExtractSuperclassRefactoring.class.st:773` | `copyMethod:from:to:` |
| `GsSplitClassRefactoring.class.st:706` | `copyMethod:from:to:` |
| `GsInstVarStructureRefactoring.class.st:1000` | `copyMethod:from:to:` |

These need a different remedy from the raising helper: the copy-forward runs inside a
per-class change, so raising would abort an entire rename or split over one bad method.

### Tier 3 — false success; the refactoring silently did not happen

The method keeps its old source, or never lands on its target, while the apply counts it as
applied:

| Site | Method |
| --- | --- |
| `GsChangeSignatureRefactoring.class.st:657` | `applyChange:` |
| `GsExtractMethodRefactoring.class.st:892` | `applyMethodCompile:` |
| `GsExtractTemporaryRefactoring.class.st:629` | `applyMethodRecompile:` |
| `GsInlineMethodRefactoring.class.st:650` | `applyMethodRecompile:` |
| `GsInlineTemporaryRefactoring.class.st:565` | `applyMethodRecompile:` |
| `GsRenameTemporaryRefactoring.class.st:496` | `applyMethodRecompile:` |
| `GsInstVarStructureRefactoring.class.st:927` | `applyMethodAdd:` |
| `GsInstVarStructureRefactoring.class.st:1014` | `applyMethodRecompile:` |
| `GsExtractSuperclassRefactoring.class.st:787` | `applyMethodAdd:` |
| `GsSplitClassRefactoring.class.st:721` | `applyMethodAdd:` |
| `GsMoveMethodRefactoring.class.st:419` | `applyMethodAdd:` |
| `GsPushDownMethodRefactoring.class.st:474` | `applyMethodAdd:` |

Move-method and push-down are protected from actual method loss by their remove guards —
move-method checks `targetBehavior includesSelector:`, push-down additionally requires every
immediate subclass to understand the selector, so a partial push degrades to a copy-down.
Both still report clean success on a compile that never landed.

## Steps to reproduce

1. Compile a method that references a global, then remove that global, so the method still
   runs but its source no longer compiles (error 1031, undefined symbol).
2. **Tier 1:** push that method up to a superclass. The push reports success; the method is
   on neither the superclass nor the subclass.
3. **Tier 2:** rename the class that owns the method. The rename reports success; the method
   is absent from the new class version and nothing warns.
4. **Tier 3:** run any of the listed refactorings over it. The apply reports it as applied
   while the method still holds its old source.

## Expected behavior

1. No refactoring removes a method from its source until the compile onto the target is
   confirmed to have succeeded.
2. A method that cannot be recompiled is reported in the apply result's `failed` list; the
   rest of the refactoring still completes.
3. Every failure entry names the method, in the `Class>>selector` / `Class class>>selector`
   form `failureLabel` already produces.

## Actual behavior

1. Push-up deletes the source method after a compile that failed.
2. The class-rename, extract-superclass, split-class and instance-variable-structure
   copy-forward loops drop the method from the new class version and report clean success.
3. The remaining engines report a refactoring as applied when the compile never landed.

## Potential fix

For Tier 1 and Tier 3, route the send through `GsRefactoringEnvironment>>compile:into:category:`,
which already raises into the apply loop's existing `on: Error do:`. For push-up specifically,
the `appliedAddSelectors add:` must move behind the confirmed compile, since the paired remove
is gated on it.

For Tier 2, mirror what `GsRenameInstanceVariableRefactoring` does: `copyMethod:from:to:` there
takes an `into: failures` collection and records a failed method rather than raising, so the
copy continues. That means threading a failures collection through `copyMethodsFrom:to:` and
its callers, and folding the collected entries into the apply envelope.

Worth landing before the shared-engine-base extraction in
https://github.com/GemTalk/Jasper/issues/274 — that change moves large blocks across all of
these files and would bury a correctness diff.

## Already fixed — not part of this issue

The original second half of this issue (failure labels naming a class rather than a method)
was fixed in the review pass on PR #411. All three engines now build their failure entries
with `change failureLabel`, and `GsRefactoringChange>>failureLabel`
(`GsRefactoringChange.class.st:268`) renders `Class>>sel` / `Class class>>sel`, falling back to
the class name for structural changes. The remaining work is the compile-result checking above.

## Environment

- **Jasper:** 1.8.13
- **GemStone server:** all supported releases (engine-side; reproduced on 3.6.2 and 3.7.5)

ericwinger:
Fixes #412.

`Behavior>>compileMethod:dictionaries:category:` reports a compile failure in its **return
value** — `nil` (or an empty `Array`) on success, an `Array` of error tuples otherwise — and
does **not** signal. Any send that ignores the result reads as success while the method is
never installed.

PR #411 introduced `GsRefactoringEnvironment>>compile:into:category:` to raise on that, but
only three engines ever used it. Seventeen sites across twelve engines still discarded the
result.

## What was broken

**The method was destroyed.** `GsPushUpMethodRefactoring>>applyMethodAdd:` recorded the
selector in `appliedAddSelectors` unconditionally, and `applyMethodRemove:` is gated on that
set — so a method that would not compile on the superclass was stripped from the subclass
anyway. It ended up on neither class, and the apply reported success.

The same shape turned up in inline-method while auditing every removal for this PR: the apply
loop catches per change and keeps going, so a caller that would not recompile never stopped the
staged removal of the now-unused target. The target was destroyed while the caller still held
its old body calling it. (Opt-in in the UI, which limits how often it is reached, but not
whether it loses the method when it is.)

**The method was silently dropped.** Four copy-forward loops — class rename, extract
superclass, split class, instance-variable structure — carry a class's methods onto a new
version, which starts with an empty method dictionary. A method that would not recompile was
dropped from the new version and the refactoring reported clean success.

**The refactoring silently did not happen.** The remaining twelve sites counted a method as
applied when the compile never landed.

## The fix

Every compile now goes through `GsRefactoringEnvironment`, which holds the only send of
`compileMethod:` left in the engine tree:

- `compileFailureFor:into:category:` — answers the error text, or nil.
- `compile:into:category:` — raises on top of it, for the apply paths, where the failure lands
  on the `on: Error do:` every apply loop already has.
- `copyMethod:from:to:source:meta:into:` — records the failed method and keeps going, for the
  copy-forward loops. Those run inside a single per-class change, where raising would abort a
  whole rename or split over one bad method.
- `copyMethod:from:to:meta:into:` — the verbatim form, shared by the three engines that were
  left with byte-identical wrappers once the checking landed.

The two engines that already checked their results fold onto the same helpers.

Engines that reported failures by class name now use the change's `failureLabel`, as the rename
engines already did. These sites report compile failures for the first time, and a bare class
name does not say which method to go fix.

Inline-method's removal is now gated on the caller having recompiled, the way push-up and
move-method gate theirs, and it counts only changes that actually applied.

## One behaviour change worth reviewing

The instance-variable-structure copy-forward now skips selectors staged for a
`#methodRecompile`. Converting a method temporary into an instance variable leaves the old
source declaring a temporary the new version declares as an ivar, and the compiler refuses that
shadowing declaration. The verbatim copy was always failing there; it was simply invisible
until the result was checked. That rewrite always runs — the refactoring is all-or-nothing — so
its source is what belongs on the new version.

## Also here

`GsRefactoringEnvironment>>environmentId` answers 0 and has no sender yet: the selector every
engine compiles through takes no environment argument, so there is nowhere to pass it. It is a
documented seam so the change is one method rather than a sweep when #492 lands — GemStone
marks that selector "May be deprecated in the future", and the environment-aware form
**signals** rather than answering error descriptors, which would make this whole bug class
impossible rather than merely guarded.

## Testing

Four SUnit regression tests, each verified to fail against the pre-fix payload:

- push-up keeps the source method when the superclass compile fails;
- class rename reports the method and still carries the rest of the class forward;
- move-method reports a method that will not compile on the target;
- inline-method keeps the target when the caller will not recompile.

Plus `client/src/refactoring/__tests__/engineCompileFunnel.test.ts`, a source-level guard
asserting the funnel holds: exactly one `compileMethod:` send in the whole engine tree, inside
the checked helper. Nothing in the language stops a new engine from sending it directly and
reopening the hole, and a reopened hole is invisible — the refactoring still reports success.
It strips Smalltalk comments (tracking string literals in the same pass) rather than padding a
threshold to absorb the doc-comment mentions, which would make it blind to the regression it
exists to catch. Confirmed to fire by reintroducing a direct send.

Full suite green (6082 + 322 + 92). Payloads regenerated; the loader's completeness check
passes. All three user-visible tiers were also verified by hand in the IDE against a 3.6.2
stone.


client/src/refactoring/__tests__/engineCompileFunnel.test.ts
```diff
@@ -0,0 +1,139 @@
...
+ * Needed because the engines *discuss* `compileMethod:` at length in their doc-comments —
+ * a plain grep reports a dozen hits that are prose, not sends, which would make this test
+ * either permanently red or (if the threshold were padded to accommodate them) blind to a
+ * real regression.
+ *
+ * Smalltalk quoting: `"…"` is a comment and `'…'` a string literal; each doubles its own
+ * delimiter to escape it. A comment can contain apostrophes and a string can contain double
+ * quotes, so the two must be tracked together in one pass rather than stripped separately.
+ */
+const stripComments = (source: string): string => {
```
MatiasFernandez: 🚫 `stripComments` doesn't know about Smalltalk character literals, and there are `$"` literals in the engine tree today — so the funnel this test exists to guard has a blind spot.

`GsChangeSignatureRefactoring.class.st:118` is `(ch == $" or: [ch == $\]) ifTrue: [...]`. That `$"` opens a "comment" that runs to the next `"` in the file, and the polarity stays inverted for a long stretch afterwards. I injected `target compileMethod: src dictionaries: sl category: cat.` line by line into that file and ran it through this parser: the send is invisible for ~270 of its 685 lines — including real code from 118 through ~290 and again around 500. A new engine (or a regression in change-signature itself) that sends `compileMethod:` inside one of those regions passes this test green, which is exactly the invisible reopened hole the test is written to catch. `GsRefactoringChange` and `GsRefactoringJson` carry `$"` too.

Handling `$x` first fixes it and keeps the suite green:

```ts
if (ch === '$') { out += ch; i++; out += source[i] ?? ''; i++; continue; }
```

With that in, the injected send is visible at 118/200/500 and `senders` is still `['GsRefactoringEnvironment.class.st']`. Worth pinning in the parser case below as well — something like `expect(stripComments('a $" b "c" d')).toBe('a $" b  d')` — since that case is what makes the verdict trustworthy.

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
```diff
@@ -700,21 +697,16 @@ GsInstVarRefactoring >> compileAccessors: accessorPairs onFailures: failures [
...
 		sel := (pair at: 1) asString.
 		(target includesSelector: sel asSymbol) ifFalse: [
-			result := target
-				compileMethod: (pair at: 2) asString
-				dictionaries: System myUserProfile symbolList
-				category: 'accessing'.
-			(result isNil or: [(result isKindOf: Array) and: [result isEmpty]])
-				ifFalse: [failures add: (Array
+			(environment compileFailureFor: (pair at: 2) asString into: target category: 'accessing')
+				ifNotNil: [:text | failures add: (Array
```
MatiasFernandez: 💭 Now that `compileFailureFor:` hands the error text back, should these two sites use it? Both bind `:text` and then discard it — here the recorded message stays the generic `'accessor ', sel, ' could not be compiled onto the new class version'`, and the `dropped` entry above (line 689) carries no reason at all. The rest of the PR deliberately went the other way (`'did not recompile: ', text`) on the grounds that the failure list is what the user opens to go fix the thing, and "could not be compiled" doesn't say why. Appending `text` would also give the block argument something to do.

gs-src/refactoring/engine/GsExtractSuperclassRefactoring.class.st
```diff
@@ -675,11 +676,15 @@ GsExtractSuperclassRefactoring >> applyDeselected: deselectedIds [
 	oldToNew := IdentityDictionary new.
 	newClass := nil.
 	failures := OrderedCollection new.
+	"The copy-forward loop runs deep inside a per-class change and must not raise (one
+	 method that will not recompile would abort the whole change), so it reports through
+	 this alias instead -- entries land in the same `failed` list as everything else."
+	copyFailures := failures.
```
MatiasFernandez: 💭 Four engines now grow a `copyFailures` instance variable aliased to the local `failures`, each carrying the same four-line comment (split class, rename class and inst-var structure get it too). Since `failures` is already a method temp in all four `applyDeselected:` implementations, would promoting *it* to the instance variable and dropping the alias be simpler? Then there's one name for one collection and no comment needed to explain why there are two. `copyFailures` also reads like a separate copy-forward list, which is the one thing it isn't.

gs-src/refactoring/tests/GsInlineMethodRefactoringTest.class.st
```diff
@@ -537,6 +537,30 @@ GsInlineMethodRefactoringTest >> testApplyKeepsTargetWhenRemovalDeselected [
...
+	UserGlobals at: #GsIMGone put: 7.
+	self compile: 'gsimDoomed ^ count' in: self baseFixture.
+	self compile: 'useDoomed ^ self gsimDoomed + GsIMGone' in: self baseFixture.
+	UserGlobals removeKey: #GsIMGone.
+
+	json := (self inlineIn: self baseFixture selector: #useDoomed atSendOf: 'gsimDoomed')
+		applyDeselected: #().
+
+	self deny: json includesSubstring: '"failed":[]'.
+	self assert: json includesSubstring: 'did not recompile'.
```
MatiasFernandez: 💭 The other three new tests each assert the failure names the method (`'GsPUSuper>>usesGone'`, `'GsMMTarget>>usesGone'`, `'GsRCRenamed>>copyMe'`). This one doesn't, but inline-method also switched from `change className` to `change failureLabel` in this PR — should it pin that too, e.g. `self assert: json includesSubstring: 'GsIMBase>>useDoomed'`? Otherwise the label change is the one part of this engine's diff that no test would catch regressing.

gs-src/refactoring/engine/GsRefactoringEnvironment.class.st
```diff
@@ -332,24 +332,105 @@ GsRefactoringEnvironment >> instanceMethodsShadowing: anInstVarName inClass: aCl
...
+	 must NOT abort the surrounding work -- the copy-forward loops that carry a class's whole
+	 method dictionary onto a new version -- use copyMethod:from:to:source:meta:into: instead."
+	| text |
+	text := self compileFailureFor: source into: aBehavior category: aCategory.
+	text isNil ifTrue: [^self].
+	self error: 'did not recompile: ', text
+]
+
+{ #category : 'compiling' }
+GsRefactoringEnvironment >> environmentId [
```
MatiasFernandez: 💭 I read the PR body on why `environmentId` lands now, and the one-place-to-change argument makes sense. Still, it has no senders, so nothing keeps this doc-comment honest — and the funnel test that guards the other invariant in this class won't notice if #492 ends up carrying the environment differently. Would the same information read better as part of the `compileFailureFor:` comment (which already points here) plus the note on #492, so there's no unsent method to go stale?

MatiasFernandez (approved):

client/src/refactoring/__tests__/engineCompileFunnel.test.ts
ericwinger: Confirmed, and fixed in 1efa3e9a — thank you, this was a real hole.

I reproduced your finding before changing anything: with `target compileMethod: src dictionaries: sl category: cat.` injected at line 200 of `GsChangeSignatureRefactoring.class.st`, the pre-fix test passed 4/4. Genuinely blind, exactly as you described.

Took your fix, handling `$x` ahead of both the comment and string-literal branches. Post-fix I re-ran the injection at lines 130, 200, 280 and 500 — all four now fail the test, and `senders` is still `['GsRefactoringEnvironment.class.st']` on the clean tree.

Also pinned it in the parser case as you suggested, plus the mirror case: `$'` has the same defect in the other direction — it opens a fake *string literal*, so doc-comments leak into the scan instead of code being hidden from it. That one is in `GsExtractSuperclassRefactoring:553`, `GsSplitClassRefactoring:467` and `GsRenameClassRefactoring:374`. It happens not to be red today, but it would have made the test fail for reasons that had nothing to do with a regression. Your one-line fix covers both:

```ts
expect(stripComments('a $" b "c" d')).toBe('a $" b  d');
expect(stripComments("a $' b \"c\" d")).toBe("a $' b  d");
```

gs-src/refactoring/engine/GsInstVarRefactoring.class.st
ericwinger: Agreed on both — done in 1efa3e9a.

You were right that the two sites had drifted from the rest of the PR. The compiler's text now rides along at both:

- **The accessor failure (line 709)** now reads `'accessor ', sel, ' could not be compiled onto the new class version: ', text`. That one was pure upside — the message is already user-facing through the `failed` triple, so nothing else had to change.
- **The `dropped` entry (line 689)** now carries the reason as a third element, through `"reason":` in the JSON and `BrokenMethod.reason?` on the client.

That second one turned out to need a bit more than the tuple change, and it is worth flagging since it widens the diff past what you asked. The reason would have been dead data: `dropped` was only ever surfaced as a **count** in a toast — the class and selector were not displayed anywhere, let alone why each failed. So it now writes the full list to the durable "GemStone GCI" channel, which is what the rename family already does with its recompile failures via `formatRenameFailureLog`. The toast keeps the count and points at the channel.

Happy to split that reporting into its own PR if you would rather this one stayed narrow.

gs-src/refactoring/tests/GsInlineMethodRefactoringTest.class.st
ericwinger: Good catch — added in 1efa3e9a.

`testApplyKeepsTheTargetWhenTheCallerWillNotRecompile` now asserts `'GsIMBase>>useDoomed'` alongside the existing `'did not recompile'`, which matches what the other three do.

Verified it has teeth rather than just passing: reverting line 626 back to `change className`, rebuilding the payload and reinstalling makes the in-stone suite fail; restoring `change failureLabel` makes it green again. So the label change now has exactly one test standing behind it, which was your point.

gs-src/refactoring/engine/GsRefactoringEnvironment.class.st
ericwinger: Fair — an unsent method whose comment nothing checks is a comment waiting to go stale. Addressed in 1efa3e9a, though not quite either way you framed it.

Rather than folding the note into `compileFailureFor:` and dropping the method, I gave it senders where it can honestly have them: this class's own five `compiledMethodAt:`/`categoryOfSelector:` reads now go through `self environmentId`. The seam stays a single method to change, but it is now exercised on every preview and every copy-forward, so it cannot quietly stop meaning what it says.

The **engines'** `environmentId: 0` sites are deliberately still literal. Routing those here would advertise a configurability that does not exist yet, and they have to change in the same sweep that finally gives this method a value other than 0.

The doc-comment is updated to say that, and I fixed the line in `compileFailureFor:` that had asserted "`#environmentId` has no sender yet" — true when you read it, not any more. That method still cannot honour it, since the selector it sends takes no environment argument at all, and it now says so.

PR #496, opened 2026-08-25
Open the debugger the same way every time

### LINKED ISSUE #494: Enhanced debugger layout is nondeterministic: source pane lands below or beside at random
## What's wrong

When the enhanced debugger opens, you can't predict what you'll get. Sometimes the source
pane sits below the call stack and variables, the way it's meant to. Other times it lands off
to the side, and the debugger ends up as panes scattered around the window. Same code, same
session — it just varies, and it's worse on a smaller window. Because you can't predict it,
every halt starts with a moment of re-orienting yourself instead of debugging.

What we want: **the debugger opens the same way every single time** — call stack and variables
in one pane, source directly below it, sized the same — no matter how big the window is, what
was focused when the error hit, or what else is already open.

## Why it happens

The layout isn't decided once. It's assembled from several steps that each depend on timing
or on a VS Code setting, so any one of them can land differently. Three independent sources of
nondeterminism, all in `client/src/debuggerPanel.ts` (symbols as of `1b66673b` on main):

**1. Where the panel goes is a user setting, not our decision.**
`DebuggerPanel.create` opens the webview with `vscode.ViewColumn.Beside`. `Beside` means "the
side group", and VS Code resolves *which side* from `workbench.editor.openSideBySideDirection`
(`right` | `down`). If that's ever `down` — set by the user or by a synced profile — the
debugger itself opens below, and then source-below-panel gives a three-row stack. `Beside`
also **reuses** an already-existing adjacent group rather than creating one, so the panel can
land in a group that already holds the user's editors.

**2. The split that creates the source pane races the focus change it depends on.**
Both `showSourceEditor` and `openTemplateEditor` do this:

```ts
this.panel.reveal(this.panel.viewColumn, false);            // focus the panel's group…
await vscode.commands.executeCommand('workbench.action.newGroupBelow'); // …split below it
```

`reveal()` returns `void` — there is nothing to await. It queues a focus change in the
workbench; the command on the next line splits **whatever group is active at that instant**.
Win the race and the source lands below the panel. Lose it and the new group is created below
the *previously* active group (typically the left-hand editor column), and `showTextDocument`
opens into that column — which is exactly the "source is beside the debugger" case.

The source open is triggered by the webview's `ready` message (`handleMessage` → `postInit` →
the frame-source path), so the timing depends on how long the webview took to boot: machine
load, window size, cold vs. warm start. Hence the randomness. `browseImplementors` uses the
same `reveal`-then-`Beside` idiom and is fragile in the same way.

**3. The sizes come from relative step commands that clamp against window size.**
`create` fires a **non-awaited** async loop of `workbench.action.increaseViewWidth`
(`PANEL_WIDEN_STEPS`), which interleaves unpredictably with the `newGroupBelow` above and with
`applySourcePaneRatio`'s `getEditorLayout` read. And the `decreaseViewHeight` steps
(`SOURCE_SHRINK_STEPS`) hit VS Code's minimum group size on a small window, so the same step
count produces a different fraction on a different window. This is the "it's worse when the
screen is squished" half of the symptom.

## Proposed fix: declare the layout instead of composing it

Build the target grid in **one** `vscode.setEditorLayout` call *before* either the panel or the
source editor exists, then open both into explicit, computed `ViewColumn` numbers. No `Beside`,
no `newGroupBelow`, no `reveal` race, no relative step commands — nothing that depends on what
was focused or how wide the window is.

Most of the machinery is already there: `flattenLayoutLeaves` maps leaf index → `ViewColumn`,
and leaves flatten depth-first in the same order VS Code numbers columns.

```ts
/** Carve the debugger's column out of the current grid: panel on top, source below it.
 *  One atomic setEditorLayout; returns the two ViewColumns to open into. */
async function makeDebuggerColumns(
  ratio: number,
): Promise<{ panel: vscode.ViewColumn; source: vscode.ViewColumn }> {
  const cur = await vscode.commands.executeCommand<EditorGroupLayout>('vscode.getEditorLayout');
  const existing = cur?.groups ?? [];
  // Nested branches alternate orientation, so a child of a HORIZONTAL root splits
  // top/bottom on its own: panel above, source below.
  const branch: EditorGroupNode = { size: 0.6, groups: [{ size: 1 - ratio }, { size: ratio }] };
  const target: EditorGroupLayout =
    cur?.orientation === HORIZONTAL || existing.length <= 1
      ? { orientation: HORIZONTAL, groups: [...existing, branch] }            // append a column
      : { orientation: HORIZONTAL, groups: [{ groups: existing }, branch] };  // wrap a vertical root
  await vscode.commands.executeCommand('vscode.setEditorLayout', target);
  const n = flattenLayoutLeaves(target).length; // the new leaves are last, in order
  return { panel: (n - 1) as vscode.ViewColumn, source: n as vscode.ViewColumn };
}
```

Then `create` uses `createWebviewPanel(id, title, { viewColumn: cols.panel, preserveFocus: false }, …)`
and `showSourceEditor` uses `{ viewColumn: this.sourceColumn, preserveFocus: true }` with
`sourceColumn` seeded from `cols.source`. The `firstOpen` branch no longer has to synthesize a
group at all, and `PANEL_WIDEN_STEPS` / `SOURCE_SHRINK_STEPS` and their fallbacks go away.

One detail to sort out in the sketch above: `getEditorLayout` returns existing sizes in pixels
while the new branch is written as a fraction. VS Code treats sizes as relative weights, so the
existing groups' sizes need normalizing to weights summing to the share they should keep
(~0.4) rather than being passed through raw.

## Things to get right

- **`savedSourceRatio` must keep working.** The 2s sampler and `sourceRatioFromLayout` still
  read the source leaf's fraction of its branch, so a drag of the panel↔source divider still
  sticks across opens. Only the point where it's *applied* moves earlier.
- **`setEditorLayout` is global.** The System Browser tests already warn that it reorganizes
  *all* editor groups and would clobber other panels. The target has to be composed from the
  *current* tree, not a wholesale replacement, so a System Browser / GT Inspector / enhanced
  inspector that's open survives the debugger opening. This is the part to test hardest.
- **Two live debuggers.** `DebuggerPanel.panels` is a `Set` per session, so two halts can be
  open at once. The second panel should reuse the existing column pair rather than carving
  another one — otherwise the grid grows with every error.
- **Closing should restore.** Disposing the panel should leave the grid as the user had it
  before the debugger opened, not with empty groups behind.
- **Prose sweep.** Doc comments that describe the current step-nudge approach need updating in
  the same change: the `PANEL_WIDEN_STEPS` / `SOURCE_SHRINK_STEPS` block, `showSourceEditor`'s
  "we focus the panel and split a new group beneath it", `openTemplateEditor`'s mirror of that,
  and `applySourcePaneRatio`'s fallback description.

## Acceptance

- Opening the debugger on a narrow window, a wide window, a maximized window, and with a full
  editor grid already open produces the same arrangement every time.
- With `workbench.editor.openSideBySideDirection` set to `down`, the arrangement is unchanged.
- Other Jasper panels that are open when the debugger opens are still there afterwards.
- Closing the debugger returns the editor grid to its prior shape.
- Two simultaneous halts share one debugger column pair.

ericwinger:
Closes #494.

## What was wrong

The enhanced debugger's layout was assembled from several steps that each depended on timing or on a user setting, so the same halt could land the source pane below the panel or beside it, at sizes that varied with the window. Three independent sources of nondeterminism: `ViewColumn.Beside` (whose direction is the user's `workbench.editor.openSideBySideDirection`, and which reuses an adjacent group), a `reveal()`-then-`newGroupBelow` pair that raced its own focus change, and relative step commands that clamp against window size.

## What it does now

**The layout is declared, not composed.** The panel opens into a column of its own by explicit number, and one `setEditorLayout` splits that column into panel-over-source. Nothing depends on what was focused, how fast the webview booted, or how big the window is.

**The grid math is pure and tested** — `client/src/debuggerLayout.ts`. Every shape gets a unit test with no editor running: a flat row of columns, a panel nested inside branches earlier sessions left behind, a vertically stacked editor area, an unmeasured grid, and the cases where it declines to touch the grid at all.

**Sizes are whole pixels** taken from the grid VS Code just reported. VS Code decides for itself whether a `size` field is pixels or a fraction, and a fraction it reads as pixels collapses that group to the 70px minimum — the panel becoming a sliver the moment it opened.

**Column numbers are never stored.** VS Code renumbers ViewColumns whenever groups come and go, so a second halt reads the live pair off the open panel instead of a number that has since moved. The debugger also retires its own groups once they are empty, which is what left the grid nesting for the next one to trip over.

**The panel sizes its own column.** It measures what it needs after rendering and the host pushes the source pane down to fit, so the message saying why execution stopped is readable the moment the debugger opens. The host cannot compute this itself — it does not know how many rows the toolbar wrapped to, whether a "Create #selector" bar is up, or how many lines the error ran to.

**It degrades instead of clipping.** VS Code enforces a minimum editor-group size and then cuts off whatever a webview cannot fit. The panel now gives up padding, then labels, in that order, and scrolls as a last resort; a maximize button is the way out when the column is genuinely too small.

## Visual

The panel is the **GemStone Debugger** now, framed in purple with a tab icon. That is not decoration: it and its companion source editor sit in the same column, one above the other, and a webview tab is otherwise indistinguishable from a source-editor tab. VS Code has no API for tab colour, so the icon is the only per-tab visual an extension controls. The source editor carries the same wash.

The eval bar is one row — expression on the left, answer on the right, with a clear button. The answer used to sit under the input in a resizable block, which made it the first thing squeezed out of sight on a short column.

## Known limits

- The source editor's wash covers the document's **lines** only; the blank area below the last line stays plain. There is no per-editor background in the API and `workbench.colorCustomizations` would repaint every editor in the window.
- The "Browse it" path (`debuggerPanel.ts`, `browseFrame`) still uses the old `reveal`-then-`Beside` idiom to place a System Browser. Same fragility, different feature — left for a separate change rather than widening this diff.

## Testing

`npm run lint && npm run format:check && npm run compile && npm test` all green (6559 client, 322 server, 92 mcp). Exercised by hand against a live stone across narrow, wide and maximized windows, with a full editor grid already open, with two simultaneous halts, and with `workbench.editor.openSideBySideDirection` set to `down`.


npapagna (changes requested):
Thanks @ericwinger.
Left some comments, let me know your thoughts.

npapagna (commented):
Follow-up on the fixes in d0db132d — two edge cases in the new code that don't look blocking but are worth a look before merging.

npapagna (approved):
Thanks @ericwinger!

PR #497, opened 2026-08-26
Stop grouping unrelated install-script bumps, pin @types/node and @types/vscode

MatiasFernandez:
## Summary
- Remove the `install-scripts` dependabot group. It mixed low-risk patch bumps (esbuild, koffi, lefthook, @vscode/vsce, ovsx) with unrelated major-version jumps in one PR; those packages now fall through to the standard `dev-dependencies`/`prod-dependencies` grouping like everything else — minor/patch batched, majors reviewed individually.
- Add an `ignore` rule for `@types/node` and `@types/vscode`. Both are part of the coordinated VS Code/Node/TypeScript version floor documented in `docs/how-to/raising-the-version-floor.md` and should only move by hand, together with the rest of that file list — not via an automated bump.

## Related (not in this diff)
- Created `major`/`minor`/`patch` labels on the repo so Dependabot's built-in semver-label feature auto-tags future PRs by update type. No `dependabot.yml` config needed for that — GitHub applies it automatically once matching labels exist.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,18 +7,13 @@ updates:
     open-pull-requests-limit: 5
     cooldown:
       default-days: 7
+    ignore:
+      # Part of the coordinated VS Code/Node/TypeScript version floor — see
+      # .claude/rules/version-pins.md and docs/how-to/raising-the-version-floor.md.
+      # Only bumped by hand, together with the rest of that file list.
+      - dependency-name: '@types/node'
+      - dependency-name: '@types/vscode'
     groups:
-      # Listed first so it wins the match: keeps every ESTRICTALLOWSCRIPTS-
-      # triggering bump in one recognizable PR per cycle. @vscode/vsce-sign
-      # and keytar are deliberately excluded — they're transitive, and
-      # Dependabot never names them directly.
-      install-scripts:
-        patterns:
-          - 'esbuild'
-          - 'koffi'
-          - 'lefthook'
-          - '@vscode/vsce'
-          - 'ovsx'
       dev-dependencies:
         dependency-type: 'development'
         update-types:
--- a/docs/how-to/add-a-dependency-with-install-scripts.md
+++ b/docs/how-to/add-a-dependency-with-install-scripts.md
@@ -34,4 +34,4 @@ Approvals are version-pinned (`allow-scripts-pin=true`), so a Dependabot PR that
 2. `npm rebuild`.
 3. Commit the updated `package.json` on the branch and push.
 
-Packages that receive this treatment regularly are grouped into Dependabot's `install-scripts` group (see [docs/reference/supply-chain-controls.md](../reference/supply-chain-controls.md)) so they arrive together, in one recognizable PR per cycle.
+A bump of an approved package arrives in the ordinary `dev-dependencies` or `prod-dependencies` group PR, and fails that PR's install until it's re-approved.


ericwinger (approved):
Approving — both changes are right in intent, and the `@types/vscode` ignore closes a real silent-breakage path that I don't think the description gives it enough credit for. Notes below on the three things you asked about: what the ignore actually buys/costs, how we hear about majors now that Dependabot won't tell us, and which packages in this tree lie about semver.

**One 🚫**, on the `groups:` line: two docs still describe the `install-scripts` group in the present tense, and they're the ones someone reads while staring at a red build. Everything else is 💭 / nit — merge whenever.

## What the `ignore` rule actually does

**It's doing more than "don't auto-bump a pin".** `@types/vscode` is `^1.101.0` and VS Code never leaves `1.x`, so *every* future release (1.102, 1.103, …) satisfies that caret. With npm's default versioning strategy Dependabot would have been widening that range weekly, and merging any one of those PRs raises the type-level VS Code API surface above `engines.vscode: ^1.101.0` — the compiler starts accepting APIs that aren't on the declared floor, which is exactly the failure `raising-the-version-floor.md` warns about ("a partial bump lets the type checker or bundler assume APIs that don't exist on the shipped runtime floor"). Ignoring it is correct.

`@types/node` at `^22.15` is a milder case: the only bump that changes the manifest is a `23`/`24` major, which is by definition a floor move. So the ignore there suppresses one PR per major, not weekly noise.

**Cost, so it's a conscious choice:** an `ignore` entry with a bare `dependency-name` and no `versions`/`update-types` is total silence — and ignore conditions in `dependabot.yml` apply to *security* updates too, not just version updates. For two types-only devDependencies with no runtime code that's a non-risk, but it's worth knowing the rule is broader than "no version bumps".

**Blanket is the right shape here**, incidentally, because DefinitelyTyped isn't semver: the `major.minor` tracks upstream but the patch component is DT's own counter, and DT ships breaking type changes in patches. Scoping the ignore to `semver-major` would not have been safe.

## Being alerted to majors we now have to do by hand

Dependabot has no notify-only mode, so this rule is genuinely silent — no PR, no alert, ever. Three thoughts:

1. **The lost signal was the wrong signal anyway.** Per `raising-the-version-floor.md`, the trigger to move the floor isn't "a new `@types/node` shipped" — it's whichever of *the floor release's bundled Node went EOL* or *the floor release turned ~1 year old* lands first. Both are calendar/EOL facts, not registry facts. A `@types/node` bump PR was only ever an accidental proxy, and often a misleading one (`@types/node@24` exists long before we'd want the floor there).
2. **So the alert should be time-based.** A monthly scheduled workflow that reads `engines.vscode`, looks up the release date and bundled Node, and opens an issue when either bound trips would be a real trigger. A much cheaper 80% version: a monthly job running `npm outdated @types/node @types/vscode` that opens an issue on a major delta.
3. 💭 **At minimum, write the gap down.** `docs/reference/supply-chain-controls.md` already has a **Known gaps** table for exactly this class of thing (the actionlint digest, the `npm@11.17.0` literals) — "these don't bump automatically and nothing checks them". A row for `@types/node` / `@types/vscode` pointing at `raising-the-version-floor.md` puts the silence in the register where someone will find it.

## Packages here that don't honour semver

Removing the pattern group means these now fall into the `dev-dependencies` / `prod-dependencies` minor+patch groups, i.e. they arrive **batched**:

| Package | Why semver doesn't hold | Effect under the new config |
|---|---|---|
| `esbuild` `^0.28.1` | `0.x` by design — esbuild documents each minor as its breaking release | `0.28 → 0.29` is `semver-minor` to Dependabot, so a breaking bundler change lands inside a batch of unrelated bumps |
| `ovsx` `^0.10.5` | `0.x`, same story | Batched; and a break in the marketplace publisher surfaces at release time, the worst moment to find it |
| `typescript` `^5.7.2` | TS explicitly does not follow semver; minors routinely change type checking | Batched — the group PR fails `npm run compile` and you get to work out which member did it |
| `@types/*` | DT patch component is DT's own counter | Not an issue for the two ignored ones; is one for `@types/jsdom`, `@types/express` |
| `koffi` `^2.15.1` | semver-ish, but it's a native FFI binary with a Node ABI | Now batched into `prod-dependencies`, *and* it's `allowScripts`-pinned, so that PR also fails install |

💭 Worth considering `exclude-patterns: ['esbuild', 'ovsx', 'typescript']` on the two groups so those three always arrive as individual PRs — a batch is fine for things where minor means minor.

Not blocking either way; the `major`/`minor`/`patch` labels you added help a lot here, since a `0.x` minor at least arrives visibly tagged.

PR #507, opened 2026-08-26
Undo for GemStone code changes — manual/UX review first, please

### LINKED ISSUE #174: All user actions in Jasper should be undoable
"Undo the World" was my talk at ESUG this year, encouraging everyone to make everything undoable. So I figured I'd better file this issue for our own IDE, and discussed with James.

Basically, every action that a Jasper user can take should be undoable. 
Clearly, there are a very few exceptions, like

- Debugger stepping
- Killing a process

But certainly anything that changes code or files should be undoable. Examples of undoable things:

- Save a method
- Revert a method
- Delete a class
- Create a class
- Add a method
- Delete a method
- Rename a class, method, or variable
- Any other refactorings offered

### LINKED ISSUE #297: Roadmap theme: Undoable IDE actions
**Component:** IDE · **Tier:** Differentiating · **Roadmap:** Later

**Before:** destructive IDE actions — deleting a method, a class, or worse — take effect immediately and irreversibly, so a mis-click can cost real work. **After:** user actions in Jasper are undoable ("Undo the World"), making the IDE safe to explore.

### Sub-items

- [ ] [#174](https://github.com/GemTalk/Jasper/issues/174) — make all user actions in Jasper undoable

### Done when

Destructive IDE actions can be undone after the fact.

---
_Generated by [Claude Code](https://claude.ai/code)_

IDE actions become undoable ([#174](https://github.com/GemTalk/Jasper/issues/174)).

### LINKED ISSUE #434: RB: Undo a refactoring — at minimum the last one, ideally a buffer of recent ones
## Summary

There is no way to undo a refactoring. Once Apply runs, putting things back is manual — and for
anything that re-versions a class or moves methods across classes, that is a lot of careful manual
work. Every refactoring in the family previews what it will do, so the user can decline beforehand,
but nothing helps them after they said yes.

At minimum, **the last refactoring applied should be undoable**. Better still would be a buffer of
recent refactorings, so the user can look back over what they did in this session and undo one from
further up the list.

This is about refactorings specifically. The broader "everything in the IDE should be undoable"
themes are [#297](https://github.com/GemTalk/Jasper/issues/297) and
[#174](https://github.com/GemTalk/Jasper/issues/174); this issue is the refactoring-shaped slice of
that, and is worth doing on its own because the engine already has most of the machinery.

## Acceptance criteria

- [x] After applying any refactoring, an **Undo** action is available and restores what the refactoring
  changed: methods it added are removed, methods it removed come back with their original source and
  category, a class it re-versioned goes back to its previous shape, names it rebound are rebound to
  what they were.
- [x] Undo is offered for at least the **most recent** applied refactoring, and it is clear what it will
  undo before it runs — ideally the same preview treatment the forward direction gets.
- [x] Where an undo **cannot** be complete, it says so up front rather than half-undoing. Instance
  migration and history deletion are the known cases; see below.
- [x] Undo of a refactoring that did not commit does not commit either; undo of one that did commit
  needs an explicit commit from the user, the same as everywhere else.
- [x] The Explorer refreshes and selection lands sensibly afterwards, as it does after an apply.
- [x] [GS SUnit] coverage on **both 3.6.2 and 3.7.5**: for each refactoring covered, apply then undo, and
  assert the class and its methods are back to their original state — including methods the
  refactoring had no reason to touch.

**Nice to have** (not met — see the status comment): a list of the last N applied refactorings with what each one did, and the ability to
undo one from the list (with whatever ordering rules turn out to be safe — undoing out of order may
have to be refused).

## Implementation notes

Possible directions, nothing settled:

- The engine already builds a **change set** of typed changes (`#methodAdd`, `#methodRemove`,
  `#methodRecompile`, `#classDefinitionEdit`, `#classReparent`, `#classAdd`) and applies them as a
  unit. An undo record could be the inverse change set, captured at apply time while the old state is
  still readable — the old source for every method it is about to replace or remove, the previous
  class definition, the previous parent.
- Class shape may not need inverting at all: GemStone's `classHistory` keeps previous versions, and
  the rename-class work already implements `revertClassNamed:toIndex:` for the Class Definition
  History viewer's Restore ([#191](https://github.com/GemTalk/Jasper/pull/191)). That is effectively
  undo for the class-shape half, and could be the foundation rather than a parallel mechanism.
- Where the record should live is an open question — a session-lifetime structure is simplest, but a
  persistent one (in the `GsRefactoring` dictionary) would survive a reconnect and let the buffer
  outlive a session. Persistence also raises the question of whose undo it is on a shared stone.
- **Known not-undoable cases, which must be detected rather than attempted:** an apply that migrated
  instances has moved user data, and a by-name reverse migration would not restore it faithfully; an
  apply that removed old versions from the class history has destroyed the thing an undo would need.
  Both are the committing options, so a first cut could simply refuse to record an undo for those and
  say why.
- Worth deciding whether this rides VS Code's own undo affordance in any way, or is purely a
  GemStone-side action in the Explorer and palette. The refactorings change the stone, not the
  editor buffer, so the latter is likely.

Part of #295.

### LINKED ISSUE #452: Undo for every code-changing action, not just refactorings
## Summary

**Nothing in Jasper is undoable today.** Any action that changes code — saving a method, deleting a
class, renaming a dictionary, applying a refactoring — takes effect immediately, and the only way
back is aborting the transaction, which discards every other uncommitted change you have made too.

[#434](https://github.com/GemTalk/Jasper/issues/434) is work in progress to make the **refactorings**
reversible. This issue is everything else.

Together they are the engineering slice of [#174](https://github.com/GemTalk/Jasper/issues/174): #174
states the goal ("every user action should be undoable"), these are about what it costs to get there.

## Status of #434, which this builds on

**In progress. Not merged, not reviewed, not manually verified.** Nothing here should be started on
the assumption that it works. This issue is filed now to capture the design questions while they are
fresh, not to queue work behind something finished.

## Why it may be worth building on rather than beside

The recording being built in #434 is not refactoring-shaped: it snapshots the state of the method
slots a change touches, before and after, and diffs them. If that generalises, any action that
compiles or removes methods could record an undo the same way, and reuse the preview UI rather than
growing a second mechanism that would later have to be reconciled with the first.

That is an expectation, not a result. If it does not hold, better to find out before adding callers.

## The actions this issue covers

- **Methods** — save, revert, add, delete, delete a category's methods, rename a method category
- **Classes** — create, delete, save a definition, save a comment
- **Variables** — add a class variable, add accessors
- **Dictionaries** — create, rename, remove, move a class between dictionaries
- **Class history** — restore a version

**Save a method** is the one users would feel first, and is among the cheapest.

## The parts that are not cheap

- **Deleting a class** needs the whole definition plus every method captured up front — once it is
  gone there is no earlier version to fall back to.
- **Anything reversed by restoring a class version** discards work written since, because a restore
  brings that version's methods with it. Whatever #434 settles on for naming what would be lost, the
  same obligation applies here.
- **Removing a version from the class history is irreversible** — it destroys what any undo would
  need, so it belongs on a refuse-to-record list rather than being attempted.

## Open questions

- One undo, or a buffer? Ordinary editing would generate records constantly, which probably forces a
  buffer and a depth policy.
- What clears it? An abort invalidates every uncommitted record, and nothing would notice.
- Does `ctrl+z` bind to this? It cannot ride VS Code's own undo stack — that is per-document, and
  these changes are in the stone — so it would need its own keybinding, scoped so it never shadows
  buffer undo in an editor.

Part of [#297](https://github.com/GemTalk/Jasper/issues/297).

ericwinger:
> [!IMPORTANT]
> **Please review this by USING it, not by reading it.** Press F5, work through the script below,
> and tell me whether the feature feels right — where the buttons belong, whether you can find them,
> whether the messages say the right thing.
>
> **Not yet:** file layout, code structure, naming, test coverage, bugs in the diff. There is a lot
> of it (160 files) and none of it is worth your time until the shape of the feature is settled. I
> will fix whatever comes out of this pass, and *then* ask for a proper code review.

Editing GemStone code has been one-way. This makes it reversible: one **Undo** covering everything
you have done since you connected, most recent first.

## Try it (about ten minutes)

Press F5, connect to a stone, and open the GemStone Explorer.

1. **Save a method.** Change a method, save it. A notice appears with an **Undo** button — press it.
   The old source comes back, and the editor showing it reloads.
2. **Dismiss the notice, then undo anyway.** Save a method again, dismiss the toast, and look at the
   **left end of the status bar**: a purple **↩ Undo**. Hover it — it names the exact change. Click it.
3. **Delete a method**, then undo. It comes back with its category, and the Explorer selects it.
4. **Add a class variable** and say yes to accessors. Undo once — the variable *and* both accessors go.
5. **Add a class variable, then add a method that uses it**, then undo the variable. It warns you
   first, naming the method that would be left reading nil.
6. **Rename a method category**, then undo. Same methods, renamed back, nothing recompiled.
7. **Remove a dictionary**, then undo. It comes back at the position it held, with its classes.
8. **Move a class to another dictionary** — right-click a class in the **Classes** or
   **Hierarchy** pane → *Move to Dictionary…* — then undo. It goes back where it came from.
   (Both this and *Move to Category…* were System-Browser-only before; they are on the Explorer's
   class menu now.)
9. **Rename a class category onto one that already exists** — the classes merge — then undo. Only
   the classes that actually moved go back; the ones already there stay put.
10. **Change a class definition** (add an instance variable), then undo. Note the button now says
   **Revert**, and it tells you what that costs before doing it.
11. **Apply a refactoring** (rename a method, say), then undo. This one opens a preview, because it
   can have rewritten methods all over the hierarchy.
12. **Press Ctrl+K U** a few times and walk backwards through what you did.

Nothing commits, in either direction.

## What I would most like your opinion on

**1. Where should the Undo button live?** There are five ways to reach it right now, which may be
four too many:

| | |
|---|---|
| the **notice** that follows an action | the one that actually gets used — it is where you are already looking |
| the **status bar**, far left | purple, stays put while connected, dims when there is nothing to undo, tooltip names the change |
| the **Explorer Methods pane** title bar | ↩ icon |
| the **editor** title bar | ↩ icon |
| the **Command Palette** | *Undo Last Change…* |

Which of these earn their place? Is there somewhere better I have not thought of? The status-bar one
deliberately **stays visible but dimmed** when the stack is empty, on the theory that a control which
vanishes cannot be learned — you find it once by accident and then never again. Tell me if that is
wrong, or if it is just clutter.

**2. Is `Ctrl+K U` the right shortcut?** It follows the `Ctrl+K` prefix the other Jasper bindings use,
and it deliberately does **not** take `Ctrl+Z` — that stays with the editor, so undoing your typing
and undoing a GemStone change never get confused. The cost is that the obvious key does not do the
obvious thing when your cursor is in a GemStone pane rather than an editor. Would you rather `Ctrl+Z`
did GemStone undo when focus is in the Explorer, and accept the risk of getting the wrong one?

## Where this is honestly not "undo"

Worth knowing before you report these as bugs — each is deliberate, and each says so in the UI:

- **Changing a class definition is a REVERT, not a rollback.** GemStone has no transaction savepoints
  and re-versions a class on every shape change, so what happens is that the *earlier version is bound
  again*. The class history grows; it never shrinks. Anything written on the newer version since is
  left behind on it, and a modal names those methods before it happens. The button says **Revert**
  rather than **Undo** wherever this is what you will get.
- **Renames, and instance-variable add/remove, apply the opposite operation.** Reversing a rename is a
  fresh rename in the other direction, not a rollback — which is actually *safer* here, since work you
  did after the rename is carried forward rather than discarded.
- **A still-empty method category exists only in the pane.** Creating one with **+** does not touch the
  stone until something is filed in it, so undoing that is a pane-only change. If you file a method
  into it first, the undo notices and acts on the stone instead.
- **Some things refuse, and say why.** A refactoring that migrated instances or deleted class-history
  versions records nothing — both commit and both are irreversible. Undoing the creation of a method
  category that has since been filled refuses rather than acting, because GemStone's `removeCategory:`
  deletes the methods in a category along with it.
- **Undo is per session and per connection.** A reconnect starts empty, and an abort clears it — an
  abort has already rewound the stone underneath every entry. The stack holds the last 25 things.

## What is covered

Saving a method, adding one, deleting one, moving one to another category; creating a class, changing
its definition, removing it (with its subclasses, as one entry), saving its comment; adding a class
variable and its accessors; creating and renaming method categories; restoring a class version;
creating, renaming and removing a dictionary; moving a class between dictionaries; renaming a class
category and moving a class into another; and the refactorings — change signature, extract/inline
method, extract/inline temporary, move method, push up/down, rename method, rename temporary, plus
class renames, instance/class-variable renames, instance-variable add/remove, extract superclass,
insert superclass and split class.

## Reading the commits, if you want to

Grouped by area rather than chronologically — 42 commits, and the first of each group is the one worth
reading:

- **the stack itself** — a generic per-session stack that knows nothing about refactorings; the
  refactoring engine is one kind of entry on it
- **the reversers** — one per kind: method slots, class bindings, class comment, class variable,
  method category, symbol-list dictionary
- **the recording sites** — the Explorer, the System Browser, the file-system provider, the file-in
  mirror
- **the engine** — `GsRefactoringUndo` and its 71 SUnit tests, filed in with the rest of the payload

## Testing

Full suite green on **both 3.6.2 and 3.7.5** with the engine installed: 7 214 passing across 463
files. The engine's own SUnit suites run in-stone on each version. The generic undo has its own
integration suite that is deliberately ungated — nothing in it needs the refactoring engine, which is
the claim that ordinary undo works on a bare stone.

Closes #434. Closes #452. Closes #174. Closes #297.

Every command in Jasper that changes code is now undoable. Two items on #452's list have no command
to record — *delete a category's methods* and *revert a method* (the latter looks like editor revert,
which is VS Code's own undo). If something has been missed, it is worth a fresh issue rather than
holding these open.


ericwinger:
Copying from slack so we don't lose good feedback: 

I found 3 of the buttons I think. I certainly prefer a single button in the new navigation & actions section. And I think we can't color it if we want to follow VSCode UI guidelines. But all our buttons there are already uncolored, so no issue with that. (going to put this in the new navigation/actions row) 

Reverting a "create method category" does not refresh the Methods section, so the category is there until you select a different class.

separate but related thing, clicking again a selected class does not trigger a methods refresh. not sure if that's a problem, I guess we only see that because there was a bug in the methods section refresh

I think undo may probably need a confirmation button for every action. When you perform an action that is not reversible, the legend when hovering shows that the action you are reverting is "the one before the previous one", but a quick click would mean the user reverts the wrong action without knowing

client/src/undo/queries/classSlotQueries.ts
```diff
@@ -0,0 +1,170 @@
...
+ *  - re-sending `Object subclass: …` does NOT carry methods forward. It answers a new,
+ *    EMPTY version. Reverting by recompiling a definition would put the shape back and
+ *    throw away every method on it.
+ *  - a removed class is only unbound, not destroyed, so the very same version can go back
+ *    exactly as it was — history, methods, instances and all. Holding it in SessionTemps is
+ *    also what keeps it reachable, since an unbound, unreferenced class version can go.
+ *
+ * The stash lives as long as the session. Entries are capped by the stack (25), so at worst
+ * that pins 25 class versions — and for everything except a removal the version is still in
+ * the class's own history anyway, so nothing extra is held.
```
MatiasFernandez: 💭 I do not think this bound holds. Nothing ever removes a `JasperUndoStash_*` key — I grepped the whole repo and there is no `removeKey:` against one, in the client or in the `.gs` payload. The stack cap bounds *entries*, but an entry leaving the stack (evicted by `stack.shift()` in `pushUndoEntry`, popped after a successful undo, dropped by `dropUndoEntry`, or cleared wholesale by `clearUndoStack`) does not release the object its key pins. So the pin count tracks the number of class edits and dictionary removals in the session, not 25.

For an ordinary redefinition that costs nothing real, exactly as you say — the version is in the class history anyway. But it bites precisely in the two cases where the doc says the stash is load-bearing:

- remove a class subtree, then do 25 more method saves. The `classEdit` entry is evicted, the stash keys stay, and those class versions are pinned in SessionTemps for the rest of the session with no undo left that could ever use them.
- same for `beginDictionaryRemoval` — a removed dictionary plus every class it held, held live by a key nothing will ever read again.

An abort makes it worse rather than better: `clearUndoStack` drops every entry, but SessionTemps is temp object memory and survives the abort, so the whole set is now unreachable *and* still pinned. A long session that removes a lot of classes or dictionaries could plausibly walk into temporary-object-memory pressure.

Would it be worth giving the stack a release hook — something the stack calls with the departing entry so `classEdit`/`dictionaryEdit` can `SessionTemps current removeKey:` their keys on eviction, pop and clear? That would also make the 25 bound in this comment true as written.

client/src/extension.ts
```diff
@@ -1545,6 +1576,39 @@ export function activate(context: vscode.ExtensionContext) {
       await extractMethodCommand(sessionManager);
     }),
 
+    // Undo the last thing done in this session -- a method edit or an applied refactoring
+    // (#434). Reached five ways: the purple status-bar button, the Undo button on the
```
MatiasFernandez: nit: this comment is stale in two ways — it says "five ways" and then lists four, and the first of them ("the purple status-bar button") no longer exists. `undo/undoUi.ts` records that the status-bar item, the Methods-pane icon and the editor-title-bar icon all went on the review of #507. `refactoring/refactoringAppliedToast.ts` has the corrected list (pane button, palette, keybinding, toast) — worth matching it here.

client/src/undo/reverseClassEdit.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  if (failures.length > 0) {
+    const first = failures[0];
+    void vscode.window.showErrorMessage(
+      failures.length === results.length
+        ? `Revert of ${entry.label} failed: ${first.error}`
+        : `Revert of ${entry.label} was partial — ${first.op.slot.className}: ${first.error}`,
+    );
+    return true;
+  }
```
MatiasFernandez: 💭 The contract at the top of the file says the answer is "false when the user backed out or it could not run at all", but a reversal where *every* op came back with an error returns true here, so the entry is popped and cannot be retried. `reverseClassEdit.test.ts` shows that is deliberate — "reports a version the session no longer holds" asserts `spent` is `true`, and the sibling test draws the real line: a *throw* means nothing was attempted, a *reported* error means the stone refused and the entry is spent.

That distinction makes sense, it is just not visible from the code or the doc. `reverseMethodEdit` and `reverseClassVarEdit` both carry an inline comment at their equivalent `return true` explaining it ("Partial or total, the recorded 'before' is no longer a description of anything the stone holds"); this one has none, and its header doc reads as if it says the opposite. Could the same note go here, and "it could not run at all" be tightened to something like "the reversal threw before anything was attempted"?

Worth a second look at whether total-failure-spends-the-entry is right for *this* kind specifically, too. For a method edit the source is still on the stack, so losing the entry costs a retry. For a class edit the stashed earlier version is the only route back to that class version, so an entry spent on a failure that happened to be transient is the user's last path to it.

client/src/gemstoneExplorer.ts
```diff
@@ -4636,6 +4843,26 @@ export class ExplorerController {
...
+    const session = this.session();
+    if (!session) return;
+    let entries: queries.ClassNameEntry[];
+    try {
+      entries = queries.getAllClassNames(session);
+    } catch {
+      return;
+    }
+    const chosen = entries.find((e) => e.className === className);
+    if (!chosen) return;
```
MatiasFernandez: 💭 This takes the first class of that name anywhere on the symbol list, ignoring the dictionary — but `MethodSlot` carries `dict`, and `afterUndo.revealMethod` drops it on the way here.

What makes me raise it is that `closeEditorsForRemovedMethods` in the same feature goes to real lengths over exactly this hazard — `sameDictionary` plus a paragraph on why a class name is not unique in a session and why closing the tab for the other `Account` would be worse than the stale tab it is there to remove. Same setup here: undo a save on `Account>>#balance` in dictionary 4, and if `Account` is also bound in dictionary 1 the Explorer cascades to the one that was never touched, so the user is shown a class where nothing happened while the method that actually came back is elsewhere.

It is reveal-only and best-effort, so nothing breaks — but should the slot's `dict` be threaded through so it prefers the matching dictionary and only falls back to first-match when it cannot resolve one?

client/src/undo/undoLastCommand.ts
```diff
@@ -0,0 +1,175 @@
...
+  // stone's record is per session and a reconnect clears it), and dropping it should fall
+  // through to whatever is under it rather than answer "nothing to undo" over a stack
+  // that still has entries.
+  for (;;) {
+    const entry = peekUndoEntry(session.id);
+    if (!entry) {
+      // Nothing changed here, so the stack has nothing to announce — but reaching this
+      // point at all means a button or menu item was showing over an empty stack. Correct
+      // it, so the refusal is the last time it happens.
+      refreshUndoUi(session);
```
MatiasFernandez: 💭 "reaching this point at all means a button or menu item was showing over an empty stack" is not quite true any more once the keybinding is in the picture. `Ctrl+K U` is contributed as `gemstone.undoLast` `when: gemstone.hasActiveSession && !gemstone.revertAvailable`, and with an empty stack `refreshUndoUi` sets *both* keys false — so `!revertAvailable` is satisfied and the chord fires straight into this branch. That is the normal, correct path for someone pressing the chord speculatively, not evidence of a stale affordance, and the `refreshUndoUi(session)` "correction" is a no-op there.

The palette entries are gated on the keys and the pane button is `gated`, so those really cannot reach it. Is it worth either saying so in the comment, or gating the keybinding on `(gemstone.undoAvailable || gemstone.revertAvailable)` so the chord is inert with nothing to undo? I lean towards keeping the chord live and firing the message — a chord that silently does nothing is worse than one that tells you the stack is empty — in which case it is just the comment that wants adjusting.

client/src/undo/__tests__/undo.integration.test.ts
```diff
@@ -0,0 +1,901 @@
...
+
+      expect(driftedSlots(slots, after, now)).toHaveLength(1);
+      // Drift does not change what the reversal is FOR: it still puts `before` back.
+      expect(planReversal(slots, before, now)[0].source).toBe('watched\n  ^ 1');
+    });
+  });
+
+  // ── Layer 3: the GemStone facts the class design rests on ──────────────
+
+  describe('what GemStone does to a class on redefinition', () => {
```
MatiasFernandez: This block is the part of the suite I would most want kept if anything ever gets trimmed — pinning "a shape-changing redefinition answers a new EMPTY version", "an identical one is a no-op" and "an unbound class survives to be bound again" as tests means the three GemStone facts the whole revert design rests on cannot quietly change underneath it. Nice.

ericwinger:
More comments - need to fix

I created a class, then a method category, then a method. I reverted the method, then the method category. I saw a few things:

• the method list was not refreshed after undoing the method category creation. The category is still listed even when there was a successful notification.
• On a previous attempt, the method category undoing attempt failed because it said the method already existed, but the method list didn't show the method. But after failing that step, the undo button changed to reflect the previous action. It was really weird.

I'm going to send claude to review the PR now because the overall UX now matches what we discussed, but I would suggest playing a lot more with stuff like that to find if everything works good because it seems there's some flakiness/imprecision around

ericwinger:
> [More comments - need to fix](https://github.com/GemTalk/Jasper/pull/507#issuecomment-5609475249)

Both fixed, plus the "rename class couldn't be undone" report from the same session, in
[`23f09755`](https://github.com/GemTalk/Jasper/commit/23f0975594b4ab3c5a4f277ad5e1cea3ef1c6f6c).

**The category row that stayed after undoing its creation** (also reported on 2 Sep) — a category
created with **+** lives in the Explorer's overlay; filing a method into it makes it real on the
stone, and the name is then in *both* places. The pane hides the overlay copy only while a real
category has that name, so removing the real one left the row on show over a category the stone no
longer had, until you browsed a different class. The stone-side reversals now correct the overlay
too. Verified by hand.

**The undo button moving on after a failure** — four reversers counted a *total* apply failure as
spending the entry, the same as a partial one. Nothing is written in the total case, so the entry
now stays on offer. Related: a "now holds N methods" refusal refreshes the Methods pane first, so it
shows the methods it is counting rather than contradicting them — that is the half of the report
where the pane showed no method.

**Rename class offered no Undo** — not a bug in the branch. The stone had `GsRenameClassRefactoring`
but no `GsRefactoringUndo`: its engine was installed before the undo work. Every undo query reaches
that class through `objectNamed:` and takes a nil-tolerant branch, so the rename applied, the
recording answered `'ok'` having recorded nothing, and the status probe answered
`{"available":false}` — indistinguishable from a session that has applied nothing. Re-installing the
engine fixes it, and rename-class undo is verified working after that. The silence is fixed too: the
probe now distinguishes "nothing recorded yet" from "this engine cannot record", and Jasper says so
once per session with an **Install GemStone Support…** button.

**Worth knowing for the review:** any long-lived stone whose engine predates this branch behaves the
same way. The availability check only looks for `GsRenameInstanceVariableRefactoring`, so a stale
engine still counts as installed.

client/src/undo/queries/classSlotQueries.ts
ericwinger: You're right, and it was worse than the pin count: `clearUndoStack` on an **abort** dropped every entry while SessionTemps survived, so the whole set ended up unreachable *and* still held. Fixed in c6397e2f with the release hook you suggested.

The stack now announces entries **leaving** it — `onUndoEntriesReleased(sessionId, entries, reason)`, with `reason` one of `evicted` / `spent` / `dropped` / `cleared` — and a new `undo/releaseStash.ts` hangs off that to run `removeKey:ifAbsent:` for the departing keys. It's a listener rather than the stack doing it itself because releasing needs a live session and a query executor, and the stack is deliberately plain data (the same reason the reversers, not the stack, know what a `classEdit` means).

Two things beyond the eviction/pop/clear you named:

**Keys that never reach an entry.** A capture straddles the edit, so an edit that throws — or one `commit` declines as "changed nothing" — leaves a key that has already pinned something and that no entry can name. `newStashKey` now takes a session id and records into a per-session issued-key registry, and a `cleared` releases the whole registry rather than just the departing entries' keys. `beginClassEdit`'s decline paths release theirs directly. `cleared` is also the one reason a zero-entry release is announced at all.

**Logout.** `clearUndoStack` also fires when the session goes away, where there is nothing left to ask — the registry is emptied regardless, so it can't grow across a session's lifetime, but no doit is attempted.

`removeKey:ifAbsent:` throughout, because a key issued for a slot that turned out unbound was never written, and a release must never be the thing that raises. Verified on a live 3.6.2 stone in 68147849 — the key is gone from SessionTemps afterwards and a never-written key is a no-op — plus unit coverage of the four release reasons and the abandoned-recording path.

The comment you flagged now says the bound follows the stack's depth rather than the session's history, and points at `releaseStash.ts`; the same correction went into the `ClassEditUndoEntry` and `DictionaryUndoEntry` doc-comments and `dictionaryQueries.ts`, which all made the same claim.

client/src/extension.ts
ericwinger: Fixed in c6397e2f — it now reads "four ways: the Undo button on the notice that follows the action, the Actions & Navigation pane's button, the palette entry and the Ctrl+K U chord", listing four under the word four, with the status-bar button gone.

I matched the count to `refactoringAppliedToast.ts` and the toast test rather than to `undoUi.ts`, whose "five ways" is the historical note about what was removed and is still accurate as written.

client/src/undo/reverseClassEdit.ts
ericwinger: The code half of this had already changed by the time you read it — 23f09755 made a total failure return `succeeded.length > 0`, so nothing-landed now keeps the entry rather than spending it, with the inline note you were asking for. Your second point is what drove it: for a class edit the stashed version is the only route back, so an entry spent on a transient failure is the user's last path to that version.

The header doc was still the old contract, and that's fixed in c6397e2f. It now says explicitly what is true and what isn't: true when the revert landed in whole or in part, and when the class was found already back the way it was; false when the user backed out at a prompt, when the current state could not be read, and when the reversal was refused outright. It also spells out why that line sits differently here than for a method edit, and points at the note by the failure return for where "partial" and "nothing landed" divide.

The test you cited (`reports a version the session no longer holds`) now asserts `spent` is `false`, which is what made the doc the only thing left saying the wrong thing.

client/src/gemstoneExplorer.ts
ericwinger: Threaded through in c6397e2f — the slot's `dict` now reaches the Explorer and it prefers the matching dictionary, falling back to first match only when there is no dictionary to go on or it resolves to nothing on the live symbol list.

`afterUndo.revealMethod` takes an optional `dict` and passes it on the command; `revealMethodByName` matches it against `ClassNameEntry` by index or by name, whichever kind the recording site captured — the same either-is-conclusive split `sameDictionary` uses, since a slot carries `dictIndex ?? dictName`. Callers: `reverseMethodEdit` passes `landOn.slot.dict`, and `undoRefactoringCommand` passes the change row's `dictName`, which was already on `UndoChange` and also being dropped.

Your read on why it mattered was the useful part — `closeEditorsForRemovedMethods` going to those lengths over the same hazard while the reveal three functions away ignored it is exactly the kind of inconsistency that says one of them is wrong. The doc-comments on both `revealMethod` and `revealMethodByName` now cross-reference `sameDictionary` so the next reader finds the pair together.

Covered by a new `explorerRevealMethodByName.test.ts` — prefers by index, prefers by name, falls back with no dictionary, falls back when the recorded dictionary is gone, and leaves the panes alone for an unbound class.

client/src/undo/undoLastCommand.ts
ericwinger: Agreed on both counts, and I took the option you leaned towards: the chord stays live and fires the message, because a chord that silently does nothing is worse than one that tells you the stack is empty. So it's the comment that changed, in c6397e2f.

It now says the empty-stack branch is the normal path for a speculative `Ctrl+K U` — gated on `hasActiveSession && !revertAvailable`, and an empty stack leaves both keys false, so the chord lands here by design and not because of a stale affordance. It also records what you established about the other entry points: the palette entries are gated on the keys and the pane's button on the same state, so neither can reach it. And it says what `refreshUndoUi` is still there for, since you're right that it's a no-op on the chord path — a button drawn over a stack that has since emptied is the case it can genuinely fix.

MatiasFernandez (approved):

PR #509, opened 2026-08-26
Method history: view, diff, and restore prior versions of a method

ericwinger:
## What

Adds **per-method source history**: every version of a method as it is compiled in this stone, with the ability to view, diff, and restore prior versions. This is the method-level companion to the existing class-definition history viewer, and closes the gap developers arriving from ENVY or Jadeite notice first — being able to see *how* a method changed on its way to a problematic version, not just two disconnected snapshots.

## How it works

Method history stands on its own — it does **not** depend on the optional refactoring engine and works on a bare stone:

- **Storage** is a plain `Dictionary` in the user's `UserGlobals` (`#JasperMethodHistoryStore`), so history is this-stone-only, per-user, and persistent — it survives across logins once the user commits (it rides the user's own transaction; nothing here auto-commits).
- A small server helper class (`JasperMethodHistory`) is compiled into the session at login and held via `SessionTemps` — the same pattern as the Jade-style Transcript sink. No plugin, no SystemUser, nothing committed for the class itself.
- **Capture** rides the ordinary compile path. Around each (re)compile Jasper seeds the about-to-be-replaced source the first time a method is edited (so the original survives) and records the newly-compiled source as a version stamped with the time and the userId. Identical-to-latest source is not re-recorded. The selector is parsed with the base-kernel compiler (compiling into throwaway dictionaries in the class's own context), so it needs no AST add-on and correctly handles methods that reference instance variables.

## The viewer

Right-click a method → **Method History…**, or use the button / context-menu / palette entry that appear while editing a method's source.

- Versions listed newest-first; the **current** version is clearly badged.
- Each version shows its timestamp (in your locale), author, and category, with an **inline diff against the current version** and a **side-by-side editor diff**.
- **Restore this version** recompiles a chosen version's source as the new current version (not committed) — and because the restore is itself recorded, it is undoable by restoring the version that was current before.
- **One tab per method** (a repeat request reveals the open tab), and the open panel **live-refreshes** when the method is recompiled elsewhere (editor save, debugger edit), preserving the diff you're viewing.

## Testing

- Unit + jsdom coverage for the model, the panel (render / restore / diff / synthetic-current / preserve-diff-on-refresh), the queries, the compile-capture hook, the one-tab / live-refresh controller logic, and the in-editor entry gating.
- GCI integration tests drive the real capture + read path on the stone with **no plugin present** (proving the bare-stone story), plus the seed / dedup / forget paths.
- CI-mirror matrix (both plugin worlds, on-demand GCI) on **3.6.2 and 3.7.5**:

method history and the regular `npm test` suite (client + server + mcp) pass in every cell — 3.6.2 and 3.7.5, each with the server plugin absent and present. (The separate on-demand GCI e2e project keeps its documented pre-existing 3.6.x gaps and is unaffected by this change.)

## Scope / relationship to the roadmap

Closes #253 (method history) and the roadmap theme #290. The theme's other sub-item, #255 (class version "browser" / load class versions on demand), is already provided by the existing class-definition history viewer, so it is not part of this change.

## Known limitations (by design, for now)

- Only compiles made through Jasper are captured — not the refactoring engine's own recompiles (rename/move/extract/inline/push), nor topaz file-in or MCP compiles in other sessions.
- History accrues from the point the helper is installed (login) onward; the seeded "original" is stamped at first-edit time, not its true original compile time.


client/src/methodHistory/methodHistoryServer.ts
```diff
@@ -0,0 +1,276 @@
...
+  // --- store & keys -----------------------------------------------------------
+  `store
+  "The per-user history Dictionary, kept in UserGlobals so it is persistent and
+   owned by this user. Created on first use. Keyed by keyFor:selector:, valued by
+   an OrderedCollection of version arrays, oldest first."
+  | ug |
+  ug := System myUserProfile symbolList objectNamed: #UserGlobals.
+  ^ug at: #JasperMethodHistoryStore ifAbsentPut: [Dictionary new]`,
+
+  `keyFor: aBehavior selector: aSelector
```
npapagna: The persistent method-history store key (`keyFor:selector:` -> `aBehavior name asString, '>>', aSelector asString`) is purely name-based, with no dictionary/environment component.

Two classes both named `Foo` in different SymbolDictionaries (a case this codebase already treats as real, see `compileMethod`'s `dict`-disambiguation) would have their `Foo>>bar` history interleaved into the *same* stored entry: silent, persistent cross-class data corruption, either user could restore/forget the other's recorded versions.

client/src/methodHistory/queries/methodHistory.ts
```diff
@@ -0,0 +1,58 @@
...
+  ifTrue: ['{"error":"${HELPER_MISSING}"}']
+  ifFalse: [${body}]`;
+}
+
+// The per-method source history for one method, as the raw JSON the
+// JasperMethodHistory helper returns (parsed by ../methodHistoryModel.ts). One object per recorded
+// version, newest first, each carrying the version index, timeStamp, userId,
+// category, an isCurrent flag, and its source. Kept in the user's UserGlobals in
+// this stone, so it is this-stone-only and per-user. Read-only.
+export function getMethodHistory(
```
npapagna: `getMethodHistory`/`removeMethodHistory` take no `dict` parameter and resolve the class unscoped, unlike `compileMethod` which supports `dict` specifically to disambiguate shadowed class names.

With two same-named classes in different dictionaries, viewing or forgetting history resolves the class via an unscoped lookup that can pick the wrong dictionary's class, while Restore (which threads `dict` through) correctly targets the intended one: a read/write split-brain.

client/src/gemstoneExplorer.ts
```diff
@@ -3660,6 +3678,177 @@ export class ExplorerController {
...
+  async openMethodHistory(
+    session: ActiveSession,
+    className: string,
+    selector: string,
+    isMeta: boolean,
+    dict: number | string | undefined,
+  ): Promise<void> {
+    // One tab per method: if a viewer for this exact method is already open, reveal
+    // it instead of opening a duplicate.
+    const already = this.methodHistoryPanels.find(
```
npapagna: The 'one tab per method' open-panel check omits `dict`, matching only sessionId+className+selector+isMeta.

Opening history for `Foo>>bar` in dict A, then a different `Foo>>bar` in dict B, incorrectly `reveal()`s the dict-A panel instead of opening a second tab: the user views/restores against the wrong class with no indication anything is off.

client/src/methodHistory/methodHistoryPanelHtml.ts
```diff
@@ -0,0 +1,222 @@
...
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;')
+    .replace(/"/g, '&quot;');
+}
+
+// Render the engine's locale-neutral ISO timestamp (yyyy-mm-ddTHH:MM:SS) in the
+// user's own locale (this runs in the extension host, so toLocaleString uses the
+// user's machine locale). Falls back to the raw string if it isn't parseable, and
+// to '' when there is no stamp (the synthetic current version).
+export function formatLocalTimestamp(raw: string): string {
```
npapagna: `formatLocalTimestamp` parses the engine's timezone-naive ISO string as if it were already in the client machine's local timezone.

The stone's `DateTime now` (e.g. a server set to UTC) is emitted with no zone info; `new Date(raw)` on a developer's machine in a different timezone parses those digits as local wall-clock time, shifting the displayed instant by the zone difference. Verified empirically: `TZ=America/Argentina/Buenos_Aires` renders the same digits unchanged, just relabeled GMT-0300 (no actual conversion happens). Every recorded version's timestamp is silently wrong for any developer not in the stone's timezone.

client/src/gemstoneExplorer.ts
```diff
@@ -5868,6 +6057,10 @@ export class ExplorerController {
   // / class appears in the panels without a manual refresh.
 
   onExternalMethodCompiled(sessionId: number, className: string): void {
```
npapagna: `onExternalMethodCompiled` refreshes every open method-history panel for the compiled class, discarding a selector that's already available one call frame up.

`extension.ts:939-949` already parses `parts` (including `selector`) from the compile-completion URI but forwards only `className` to `explorer.onMethodCompiled`. Compiling one method with N history panels open for unrelated methods of the same class fires N unnecessary blocking `getMethodHistory` GCI round trips instead of 1.

client/src/gemstoneExplorer.ts
```diff
@@ -3660,6 +3678,177 @@ export class ExplorerController {
...
+    );
+    if (already) {
+      already.panel.reveal();
+      return;
+    }
+
+    // The method-history helper is installed at login (SessionTemps, no plugin).
+    // Ensure it once more here in case that login install was skipped or failed;
+    // it is idempotent and cheap.
+    installMethodHistory(session);
```
npapagna: `openMethodHistory` unconditionally re-installs the method-history helper on every open, even though `sessionManager.ts` already installs it once at login.

Every 'Method History' invocation costs a full GCI round trip compiling ~14 Smalltalk methods just to hit the server-side 'already installed' short-circuit, wasted latency on the overwhelmingly common already-installed path. Cheaper: only attempt (re-)install after a `getMethodHistory` call first fails with the 'not available' envelope.

client/src/methodHistory/queries/methodHistory.ts
```diff
@@ -0,0 +1,58 @@
...
+      `h forClassNamed: '${escapeString(className)}' selector: '${escapeString(
+        selector,
+      )}' meta: ${isMeta ? 'true' : 'false'}`,
+    ),
+  );
+}
+
+// Forget all recorded versions of one method. Does NOT commit (the user commits).
+// Returns the raw JSON result ({"removed":bool,...} or {"error":..}).
+export function removeMethodHistory(
```
npapagna: `removeMethodHistory` (plus its model parser and `browserQueries` export) is fully implemented end-to-end but never invoked by any command, menu, or webview message.

No 'Forget history' action wires it up in `gemstoneExplorer.ts`, `extension.ts`, `package.json`, or the panel/webview files: ~40 lines of Smalltalk/TS/model surface reachable only from tests. Is a UI action planned but not yet wired up, or should this be trimmed?

client/src/methodHistory/methodHistoryServer.ts
```diff
@@ -0,0 +1,276 @@
...
+     (self pad2: aDateTime second truncated) ]
+    on: Error
+    do: [:e | [aDateTime printString] on: Error do: [:e2 | '']]`,
+
+  `pad2: anInteger
+  "anInteger as a two-digit, zero-padded decimal string (e.g. 7 -> '07')."
+  ^(anInteger < 10 ifTrue: ['0'] ifFalse: ['']), anInteger printString`,
+
+  // --- JSON string escaping (inlined; base kernel only) --------------------
+  `hex2: anInteger
```
npapagna: `hex2:`/`jsonEscape:`/`jsonQuote:` reimplement `GsRefactoringJson`'s JSON-escaping helpers, invisibly to the repo's existing dedup guard test.

The repo already paid down this exact duplication once (a structural guard test fails if a second real copy of these helpers appears among `.class.st` files); this feature reintroduces an equivalent copy compiled from a TS string, so the guard can't see it. May be a deliberate tradeoff (this helper must work with no plugin on a bare stone, unlike `GsRefactoringJson`), but worth confirming with the author rather than leaving as an invisible regression of the dedup effort.

client/src/methodHistory/methodHistoryDiff.ts
```diff
@@ -0,0 +1,62 @@
+/**
```
npapagna: Introduces a second, independent 'virtual read-only text provider' pattern duplicating `debuggerPanel.ts`'s existing `gemstone-debug` scheme for the same purpose.

Two divergent copies of 'register a scheme, stash text by synthetic URI, serve it read-only, bound the map' now exist in the codebase; the next feature needing this will likely add a third instead of reusing a shared helper, compounding maintenance cost.

npapagna (changes requested):
Thanks @ericwinger!
Nice one!
Left a couple of comments. They look "small" to me, they might be worth fixing before merging. Let me know your thoughts!

client/src/methodHistory/methodHistoryServer.ts
ericwinger: Fixed. `keyFor:` now prefixes the defining dictionary, resolved **by identity** so a shadowed name finds the class actually in that slot rather than the first one named. A metaclass is matched through the bound class, since `Foo class` is never itself bound.

Worth recording: the obvious `(aBehavior isKindOf: Metaclass)` test does not work here. These methods compile into a *throwaway* SymbolDictionary, so a bare global is an undefined symbol at compile time — `a CompileError occurred (error 1001), undefined symbol Metaclass` — which aborts the whole install doit and surfaces only as "Method history support is not available in this session", never as a compile error. Caught on a real 3.6.2 stone. It now references no global but `System`, and `methodHistoryServer.test.ts` guards that.

client/src/methodHistory/queries/methodHistory.ts
ericwinger: Fixed. Both take a `dict` and resolve through `classLookupExpr`, exactly as `compileMethod` does, so read and write agree.

The lookup is deliberately not repeated server-side — `queries/util.ts` says that helper is the one place allowed to know the dict-aware idiom, and there is a duplication guard test. So `forClassNamed:` / `removeHistoryForClassNamed:` became `forClass:named:...` / `removeHistoryForClass:named:...`, taking a class the client already resolved.

Covered by a new `dictionary scoping` suite (index, name, unscoped fallback, quote escaping, remove parity) plus two integration cases that build a real shadow fixture — a second SymbolDictionary inserted *ahead* of UserGlobals. Reverting just the key makes the first of those fail on a live stone, so it reproduces your scenario rather than merely asserting the new shape.

client/src/gemstoneExplorer.ts
ericwinger: Fixed. `dict` is part of the panel identity and of every history read, so dict-A `Foo>>bar` and dict-B `Foo>>bar` get their own tabs. Two tests in `explorerMethodHistory.test.ts` cover it; removing just the `e.dict === dict` term makes the first go red.

client/src/methodHistory/methodHistoryPanelHtml.ts
ericwinger: Good catch, and it went further than a labelling problem. Fixed properly rather than papered over.

`DateTime now asStringISO8601` emits the stone offset (`2026-09-08T13:23:35-0700`) and is present on **both 3.6.2 and 3.7.5** — verified on live stones — so the engine now records that and the client genuinely converts into the reader timezone. Records written before this carry a bare zone-less stamp; there is nothing to convert from, so those keep their wall-clock components and are labelled as stone time instead of passing as the reader own.

Unit tests cover the compact `±HHMM`, colon and `Z` forms plus the legacy shape, and an integration test pins that the engine really emits an offset.

client/src/gemstoneExplorer.ts
ericwinger: Fixed. `extension.ts` now forwards `parts[5]`, and `refreshOpenMethodHistoryPanels` filters on it, so one compile refreshes one panel. A caller with no selector still refreshes the class panels — correct, just chattier. Two tests cover both paths.

client/src/gemstoneExplorer.ts
ericwinger: Fixed, and adopted your suggestion exactly: read first, install only on the helper-missing envelope, then retry. Three tests — success installs nothing, the missing-helper envelope installs once and still opens the panel, and an unrelated failure installs nothing.

client/src/methodHistory/queries/methodHistory.ts
ericwinger: Confirmed unreferenced — no command, menu, or webview path reaches it; only its own export chain and tests. Left in place and made dictionary-scoped along with `getMethodHistory`, so read and forget target the same class. Wiring a "Forget history" action, or trimming it, is a call for @ericwinger rather than something to settle inside this review.

client/src/methodHistory/methodHistoryServer.ts
ericwinger: Checked this on a stone rather than reasoning about it: on a bare 3.7.5 stone with no server plugin, `symbolList objectNamed: #GsRefactoringJson` answers **nil**. So the duplication is necessary, not incidental — method history is designed to run on a bare stone, and referencing that class here would not merely be slower, it would fail to resolve. In this compile context a bare global is a compile-time error that takes the whole install down, as on the `keyFor:` thread.

Since the structural dedup guard scans `.class.st` files and cannot see Smalltalk built from a TS string, I documented the reason at the helpers instead, so the next reader finds the answer rather than re-deriving it.

client/src/methodHistory/methodHistoryDiff.ts
ericwinger: Agreed that this is a second copy of the pattern, and it is worth consolidating. I have left it alone here: `debuggerPanel.ts` is already-merged code, and extracting a shared virtual-document helper is a refactor touching both features rather than a fix to the change under review. Flagged to @ericwinger to schedule separately so it is not lost.

npapagna (approved):
Thanks @ericwinger !

PR #512, opened 2026-08-27
Replace drifted GCI optionality sources with a typed header parser and registry

MatiasFernandez:
Replaces the four drifted, hand-maintained sources of GCI optionality truth in `gciLibrary.ts` and its tests with two typed modules, each checked against the vendored headers by a different mechanism, plus the tests and docs that depend on them.

- `headerDeclarations.ts` parses `vendor/gci-headers/<revision>/gcits.hf` for `EXTERN_GCI_DEC` declarations, tracking `#if` nesting so it can report `FLG_UNIX`-gating too. Self-checks its own declaration count against raw regex occurrences so a silently-broken match can't pass unnoticed.
- `optionalFunctions.ts` is the hand-written, machine-verified registry of which `GciTs*` bindings may be absent, and why (`addedIn` / `absentOn: 'win32'` / `removedIn`). `optionalFunctions.headers.test.ts` asserts every field against the parsed headers in both directions.
- `gciLibrary.ts`'s optional bindings are now keyed off the registry's type, so an unregistered optional binding (or a registry entry with no binding) is a compile error instead of a runtime surprise.
- Existing GCI tests (`gciOptionalFunctions`, `gciSpecials` I32ToOop, transcript sink) are migrated onto the integration harness and out of the `client/src/__tests__/gci/` project, per the ongoing migration.
- `docs/explanation/gci-version-compatibility.md` documents the whole chain — vendored headers → parser → registry → `gciLibrary.ts` — and which mechanism catches which mistake.

## One production behavior change

`GciI32ToOop` / `GciTsI32ToOop` move from `optionalFunc` to required `this.lib.func`, so a
library missing either now fails `GciLibrary` construction (a failed login) instead of
degrading to a stub that throws only if called.

The optional binding existed to keep a pre-3.6.2 library loadable; the deleted comment named
3.4.5. But of the 84 symbols `gciLibrary.ts` binds as required, seven aren't declared in
3.4.5's headers at all — and four are on the non-blocking execution path with production
callers: `GciTsNbExecute` (`codeExecutor`, notebook, `browserQueries`), `GciTsNbResult`
(`nbRunner`, `transcriptSink`, `debugQueries`), `GciTsNbPerform` (`debugQueries`) and
`GciTsSocket` (`nbRunner`). The other three — `GciTsI32ToOop`, `GciTsNbLogout`,
`GciTsDirtyObjsInit` — have no production callers.

So a 3.4.5 library would have constructed fine and then failed on the first evaluation,
debugger step or transcript read. The tolerance bought nothing but the suggestion otherwise.
3.6.2 is the declared minimum (`versionManager.ts`) and is CI-covered on Linux and Windows,
so this is a deliberate narrowing: a hard floor for construction, not just for features.

## Known gaps

- `gciVersionGated.test.ts` only gates the 3.6.2 floor (`addedIn`), not every optional function, and doesn't verify `absentOn: 'win32'` gating at all. Broadening it is separate follow-up work.
- The same gate is currently a test rather than a lint rule, because `eslint.config.mjs` can't import the TypeScript registry. Worth revisiting if the config ever moves to TS.


docs/explanation/gci-version-compatibility.md
```diff
@@ -0,0 +1,94 @@
+# GCI cross-version compatibility
+
+How Jasper tracks which GCI symbols are missing on which release or platform, and enforces that the tracking stays accurate: vendored headers, a parser, a hand-written registry, and compiler checks tying them together. Worth reading before adding a GCI binding, calling a GciTs* function from production code, or supporting a new GemStone release.
+
+## The problem
+
+Jasper binds one native library — `libgcits` — that ships inside each GemStone distribution, and the extension must work against every release it claims to support. That set spans 3.6.2 through 3.7.5, on Linux, macOS and Windows. **The exported symbol set is not the same across them.**.
```
npapagna: hardcoded

docs/explanation/gci-version-compatibility.md
```diff
@@ -0,0 +1,94 @@
...
+
+`gciLibrary.ts` binds symbols by name through koffi, at construction time. That gives two distinct failure modes:
+
+- A **required** binding (`lib.func`) for a symbol the loaded library doesn't export throws while constructing `GciLibrary` — Jasper doesn't load at all, for that user, on that release.
+- An **optional** binding (`optionalFunc`) survives construction and throws `"<name> is not available in this GCI library"` at the call site — a broken feature rather than a broken extension, but only if production code has a guarded fallback.
+
+Neither is detectable on a developer machine running a single recent GemStone. Both are quiet until they reach a user on an older release or on Windows.
+
+## Three independent reasons a symbol can be missing
```
npapagna: function?

client/src/gciLibrary/headerDeclarations.ts
```diff
@@ -0,0 +1,496 @@
...
+     * either, so it cannot surface later as a truncation error.
+     */
+    if (!currentlyLive()) continue;
+
+    occurrences += (line.match(DECLARATION_OCCURRENCE_RE) ?? []).length;
+
+    if (!capture) {
+      if (!line.includes(DECLARATION_MACRO)) continue;
+      const unixOnly = currentlyUnixOnly();
+      const unclassifiableFrame = unixOnly ? undefined : currentlyUnclassifiable();
```
npapagna: **[BLOCKER] Unix-only check ignores proven UNIX-excluded frames**

A declaration is exempted from the "unclassifiable platform condition" check only when an enclosing frame is proven UNIX-only (`currentlyUnixOnly()`, backed by `frame.unixOnly`/`effect.requiresUnix`), not when one is proven UNIX-*excluded* (`effect.excludesUnix`, e.g. `#if !defined(FLG_UNIX)`), even though the latter equally settles `unixOnly=false`.

Concretely, a declaration nested like:
```c
#if !defined(FLG_UNIX)
  #if defined(FLG_MSWIN32)
    ... declaration ...
  #endif
#endif
```
would throw "sits inside an unsupported platform-flag condition" at generation time, even though the outer frame already guarantees this isn't UNIX-only. No test covers this nested shape.

A fix would add an `excludesUnix` field to `ConditionalFrame` (set from `effect.excludesUnix` in `openedBy`), and OR it into the exemption check at line 421-422 alongside `unixOnly`.

scripts/generate-gci-optional-functions.mjs
```diff
@@ -0,0 +1,63 @@
...
+  console.error(`error: ${error.message}`);
+  console.error('optionalFunctions.generated.ts was NOT regenerated.');
+  process.exit(1);
+}
+
+const output = renderOptionalFunctionsModule(entries, revisions);
+let previous;
+try {
+  previous = readFileSync(OUTPUT_PATH, 'utf8');
+} catch {
```
npapagna: **[BLOCKER] Catch swallows all read errors, not just ENOENT**

The `try { ... } catch { ... }` around reading the previous generated file (lines 51-55) swallows every error, not just ENOENT (file-not-found). If the output file exists but is unreadable for another reason (permissions, transient I/O error in a locked-down CI checkout), `previous` silently stays undefined, the script assumes first-ever generation, and unconditionally overwrites the file, masking the real error.

Note: `scripts/generate-roadmap.mjs:252-254` has the identical pattern, so this was copied from there rather than introduced fresh. Worth fixing in both.

Related: both scripts also duplicate the whole surrounding read/compare/write-if-changed/log-summary block (`generate-roadmap.mjs:251-259` vs this file's lines 50-63), a conscious "follow the existing convention" copy per this file's own header comment. Extracting a shared helper would fix the error-handling bug in one place and resolve the duplication at the same time.

client/src/gciLibrary/optionalFunctionsFromHeaders.ts
```diff
@@ -0,0 +1,213 @@
...
+      fail(
+        `${name} is declared in ${declaringRevisions.join(', ')}, which is not a contiguous run ` +
+          `ending at ${newest} — no single addedIn floor describes it.`,
+      );
+    }
+
+    const unixOnlyIn = declaringRevisions.filter(
+      (revision) => declarations.get(revision)!.get(name)!.unixOnly,
+    );
+    // `absentOn` is unconditional, so it cannot say "Unix-only from 3.7.2 on".
```
npapagna: **[BLOCKER] Consistency guards can't catch a self-consistent misclassification**

`deriveOptionalFunctions`'s consistency guards (lines 117-127) only fail when a symbol's UNIX-only classification *disagrees* across the vendored revisions that declare it. They cannot catch a classification that is wrong but identical across every declaring revision.

Concretely: a parser regression in `headerDeclarations.ts`'s `FLG_UNIX` classification that is self-consistent (always misreads one symbol the same way) passes both the contiguity guard and this uniformity guard silently, committing an incorrect value into `optionalFunctions.generated.ts`.

This is inherent to any generate-from-parsed-source design (there's nothing to disagree with if every input agrees on the wrong answer), and not fixable within this PR's scope. Flagging as an FYI for future maintainers of this generator, not asking for a code change here.

client/src/__tests__/gciVersionGated.test.ts
```diff
@@ -1,54 +1,49 @@
...
-const gciLibrarySource = fs.readFileSync(
-  path.join(repoRoot, 'client', 'src', 'gciLibrary.ts'),
-  'utf-8',
+
+/**
+ * Gated symbol -> the release that first exports it. `absentOn`/`removedIn`-only
+ * entries (`GciTsNbLogin`, `GciTsEncrypt`) do ship in 3.6.2 — they're gated on
+ * platform or removal, which this test says nothing about.
+ */
+const GATED_FLOORS = new Map(
```
npapagna: **No forward-compatibility gate for `removedIn` symbols**

`GATED_FLOORS` only includes registry entries with `addedIn` (the 3.6.2-floor gate), deliberately excluding `removedIn`-only entries like `GciTsEncrypt` (per this file's own comment at lines 19-20). That exclusion is intentional for this test's purpose (backward-compat to 3.6.2), but there's no *other* mechanism playing the `ALLOWED_POST_362` role in the opposite direction: forcing a conscious opt-in before production code calls a `removedIn` symbol that will break on the next major GemStone release.

Currently dormant: only `gciLibrary.ts:934` (the binding itself) and test files call `GciTsEncrypt` today, no other production code depends on it directly. But if that changes, nothing would flag "this breaks on the next major version." Worth considering a `removedIn`-analog to `ALLOWED_POST_362` at some point, not urgent.

npapagna (approved):
Left a couple of findings, I think it's good to go one blockers have been resolved. Thanks!

client/src/__tests__/gciVersionGated.test.ts
MatiasFernandez: I have another PR for this

client/src/gciLibrary/headerDeclarations.ts
MatiasFernandez: Fixed. Mirrored `excludesUnix`/`fallThroughExcludesUnix` alongside `unixOnly`/`fallThroughUnixOnly` through all three frame transitions and ORed it into the exemption, since the `#else` of `#if defined(FLG_UNIX)` has the same gap as a negated condition — `openedBy` alone wouldn't have covered it. Tests added for both shapes.

scripts/generate-gci-optional-functions.mjs
MatiasFernandez: Fixed — the catch now rethrows anything that isn't `ENOENT`, so an unreadable output file surfaces instead of being mistaken for first generation.

Left `generate-roadmap.mjs` and the duplicated write-if-changed block alone for now — extracting the shared helper touches a script outside this PR, so I'd rather do it as its own change than widen this one.

client/src/gciLibrary/optionalFunctionsFromHeaders.ts
MatiasFernandez: Noted — thanks. Agreed there's nothing to change here; the guards catch disagreement, not a uniformly wrong answer.

Issue #511, opened 2026-08-27
Session configuration lives inside the GemStone Manager webview; give it its own panel opened from the session row


ericwinger:
## Summary
The stone/gem configuration viewer/editor (#232) currently exists only inside the GemStone
Manager webview (draft PR #454), which bundles connection, versions, databases, processes,
logs and configuration into one surface. Extract it into a standalone **Session
Configuration** panel, opened from a gear action on the session row in **Logins & Sessions**,
so it survives the Manager and can be reviewed and landed on its own.

## Why
From the Aug 26, 2026 feedback discussion with Matias Fernandez: the agreed architecture keeps
navigation in the sidebar and moves advanced configuration and management into separate web
views, refactored piecemeal with one PR per isolated component — the configurator was named as
the first such component. The countervailing constraint from the same discussion is
discoverability: anything reachable only through the Command Palette is effectively hidden, so
the panel needs a visible affordance on the row it belongs to. Configuration is session-scoped,
so the session row is that row.

## Acceptance criteria
- A session row in **Logins & Sessions** carries an inline gear that opens **Session
  Configuration** for that session; the command is also in the palette.
- The panel reads and edits configuration over the clicked session's GCI — not "the active
  session".
- One panel per session, so two sessions can be compared side by side; logging a session out
  closes its own panel.
- Editable vs read-only reflects what this session can actually change, with a tooltip saying
  why a value is read-only.
- Ping moves off the session row into the panel (the row is losing buttons, not gaining them).
- The panel does not depend on the GemStone Manager and keeps working if the Manager is
  removed.

## Out of scope
Versions, OS configuration, logs, and process management moving to their own web views — same
architectural decision, separate issues. Undo/redo for a configuration change is #510.

Related: #232 (closed — the viewer this extracts), #510, #482 (Manager reorg), #298, PR #454 (draft).

ericwinger:
Implemented in #513 — https://github.com/GemTalk/Jasper/pull/513

All six acceptance criteria are on that branch:

- Inline gear on the session row in **Logins & Sessions** opens **Session Configuration**; the command is also in the palette (and accepts a connected login's row).
- The panel reads and edits over the clicked session's own GCI, resolved by session id — not "the active session".
- One panel per session, so two can be compared side by side; a session logging out disposes its own panel.
- Editable vs read-only reflects what this session can actually change, with the tooltip giving the reason — config-file parameter, needs SystemUser, or a value kind that cannot be edited in place.
[...]

Issue #510, opened 2026-08-27
Session Configuration: undo/redo for a configuration setting change


ericwinger:
## Summary
In the Session Configuration panel, after a user changes a stone or gem runtime setting, offer
an Undo to restore the value it had before the change (and a Redo to re-apply). This makes a
mistaken edit a one-click recovery instead of a manual re-type, and pairs naturally with the
panel's existing "editable vs read-only" model. This is about undoing a *live configuration
set*, not general editor undo.

## Acceptance criteria
- After an editable value is successfully changed, the panel offers Undo to set it back to
  the value it held immediately before the change.
- Undo goes through the same set path as a normal change, so it honors the same rules: a
  stone setting still requires SystemUser, and a value the stone accepts but does not apply
  still surfaces the "accepted without error, but the session still reports …" warning
  (undo is not assumed to succeed just because it was requested).
- Redo re-applies the change Undo reverted, under the same rules.
- Undo/redo history is scoped to the session's panel and the current session; it does not
  outlive the session (a logout closes the panel).
- Read-only parameters offer no undo (there was nothing to change).

## Implementation notes
The Session Configuration panel already has the pieces this needs. After each set,
`client/src/configurationPanel.ts` re-reads the report and reports the settled value, so the
prior value is known at set time and could be pushed onto a per-panel history stack; Undo
would re-dispatch through the existing `setConfiguration` message + `setSessionConfiguration`
query (`client/src/queries/configurationReport.ts`), reusing the same authority checks and
the settled-value comparison that already produces the success/warning banner in
`client/src/configurationView.js`. One open question is how much history to keep (a single
level vs a stack) and whether Redo is worth it for a first cut.

## Depends on

**#513** — the Session Configuration panel itself. The three files this work edits
(`configurationPanel.ts`, `configurationView.js`, `queries/configurationReport.ts`) do not
exist on `main`; they arrive with that PR. There is nowhere to do this work until it lands,
or until the work is based on its branch.

**#507**, if the configuration undo rides the generic per-session undo stack it introduces
(`client/src/undo/`, the status-bar Undo, `Ctrl+K U`) rather than keeping a private history
inside the panel. That is the open question above; a panel-local one-level history needs only
#513.

ericwinger:
[...]

Issue #482, opened 2026-08-24
Extract the GemStone Manager's startup handlers into tested functions, and remove the dead sidebar-views switch


ericwinger:
## Summary

The extension's `activate()` in `client/src/extension.ts` is a single ~3,600-line function (roughly lines 564–4166). Several GemStone Manager command handlers are defined inline inside it, wired to a dozen collaborators from the surrounding closure, which makes them effectively impossible to unit-test without standing up an entire fake VS Code host. Lifting each handler into its own small, dependency-injected function would let the feature logic be tested directly and shrink `activate()` toward a thin wiring layer.

Separately, `package.json` gates five sidebar views on the context key `gemstone.adminViewsInSidebar`, which is **never set anywhere in the code** — so those views never appear (their functionality lives in the GemStone Manager panel now). The key and its now-unreachable views/wiring are dead code.

## Motivation

- A 3,600-line startup function is hard to read, hard to change safely, and untestable as a unit.
- The Manager's headline flows (create-with-defaults, delete-with-orphan-cleanup, auto-open-on-last-logout) currently have **no coverage of their orchestration/ordering** — only the pure helpers underneath them are tested. Their "do step A, then B, then C, and stop if one fails" wiring is exactly where a regression would slip through green.
- The dead `adminViewsInSidebar` switch reads like a bug someone will later "fix" by turning it on, resurrecting views that were intentionally retired.

## Acceptance criteria

- The following Manager command handlers are each extracted from `activate()` into their own function (in `extension.ts` or a small dedicated module), taking their collaborators as explicit parameters, and covered by unit tests for their branch/ordering behavior:
  - `gemstone.createDatabaseDefaults` — the zero-question create → default-name → create-DB → add-login → start-stone → start-NetLDI chain, including the "stop the chain and surface the stone's message when a start fails" paths.
  - `gemstone.createDefaultLogin` — create-or-find the DataCurator login, save only when new, then log in.
  - `gemstone.deleteDatabase` — snapshot orphaned logins **before** deletion, delete them after, and the singular/plural confirmation toast.
  - `openManagerForEmptyEnvironment` — open the Manager (without focus) only when the last session is gone, not on a non-last logout.
- The `gemstone.adminViewsInSidebar` context key and the five permanently-hidden views it gates (Configure OS, Versions, Databases, Processes, Logins & Sessions) are removed, along with their startup wiring, any dead view-welcome content, and the assertions in the affected tests.
[...]

PR #513, opened 2026-08-27
Session Configuration: a session's stone and gem settings in their own panel

ericwinger:
## Summary

Pulls the stone/gem configuration viewer/editor out of the all-in-one GemStone Manager webview
(draft #454) and gives it a standalone **Session Configuration** panel, opened from a gear on
the session row in **Logins & Sessions**.

This is the **first step in integrating the Manager into the sidebar** — the architecture agreed
in the Aug 26 feedback discussion: navigation stays in the sidebar, advanced configuration and
management move into separate web views, refactored one isolated component per PR. The
configurator was named as the first such component. It depends on nothing in the Manager and
keeps working if the Manager is removed.

Addresses #511 (the issue of record for this step). Implements the viewer asked for in #232.

## What it does

- **Gear on the session row** opens **Session Configuration** for *that* session, as an editor
  tab. Also **GemStone: Session Configuration** in the Command Palette, which falls back to the
  session the rest of Jasper is working with.
- **Reads and writes over the clicked session's own GCI**, not "the active session" — so one
  panel per session, two sessions can be compared side by side, and a panel disposes itself when
  its session logs out. A row is resolved by session id, which is unique and never reused, so a
  row left over from a logged-out session cannot open a panel onto a dead connection.
- **Editable vs read-only reflects this session**, and the tooltip says *why*: `ALL_CAPS`
  config-file parameters were fixed before the stone started; a stone parameter needs SystemUser;
  a value of a kind that cannot be edited in place says so. An ⓘ per row carries the type, what
  you can do with it, and the purpose text from `system.conf` (pins on click).
- **The stone stays the authority on a set.** Its verdict comes back in its own words, rendered
  beside the row that changed — success self-clears, a failure persists with Copy and Dismiss —
  so a refused or silently-ignored set is visible rather than assumed.
- **Filter box** narrows a long report, matching anywhere in a parameter name, so the word in the
  middle of a name ("cache", "timeout") finds it. **Refresh** re-reads from the session.
- **Ping moved off the session row into the panel** — the row is losing buttons, not gaining
  them.

## Shape

Everything lives under `client/src/configuration/`, the feature-folder convention `omniSearch/`
and `refactoring/` already follow.

| File | |
|---|---|
| `configuration/queries/configurationReport.ts` | Query layer for `System stoneConfigurationReport` / `gemConfigurationReport` and the two setters. All emitted Smalltalk is ASCII-only (3.6.x `ComStrmSetCursor`), and every part that could carry user text is gated to hold that — including a string value, which is refused with a clear message rather than failing obscurely on the gem. |
| `configuration/configurationPanel.ts` | The webview host. Takes a dependency bag rather than importing `extension.ts`, to avoid a circular import. |
| `configuration/showConfigurationCommand.ts` | Resolves which session a row means, then opens or reveals that session's panel. |
| `configuration/configurationView.js` | All webview behavior, injected under a nonce with a strict CSP — the `debuggerPanel.ts` convention. |
| `configuration/gemstoneConfigDescriptions.ts` | Parses `system.conf` for the per-parameter purpose text. |
| `client/src/extension.ts`, `package.json`, `.vscodeignore` | Command registration, the row gear, and shipping the view script. |

## Tests

`configurationPanel.test.ts` (34) covers the host: which session a panel is bound to, two panels
open at once each reading its own connection, self-disposal on logout, the settle check that
catches a value the stone accepted and then ignored, the rule that stops a non-SystemUser being
offered a doomed stone editor, refusals, Ping, Copy, and the `system.conf` cache.

`configurationView.test.ts` (35), `configurationReport.test.ts` (44),
`gemstoneConfigDescriptions.test.ts` (9), and `configurationReport.integration.test.ts` (5)
against a live stone — which now closes the setter path too: a gem parameter set to the value it
already holds answers OK, and a stone parameter set as a non-SystemUser comes back refused in the
stone's own words. Plus the session-row ordering test updated for the gear.

Gate green: lint, format:check, compile, and client (6960) + server (322) + mcp (92) tests.
Client-only diff, so no server matrix run.

## Out of scope

Versions, OS configuration, logs and process management getting their own web views — same
architectural decision, separate issues (#482). Undo/redo for a configuration change is #510.


client/src/configurationPanel.ts
```diff
@@ -0,0 +1,602 @@
...
+      appendSysadmin(`Configuration: set ${scope} configuration ${key} = ${value}`);
+
+      const config = this.readConfiguration(session);
+      void this.panel.webview.postMessage({ command: 'configuration', config });
+
+      const settled = (scope === 'stone' ? config.stoneParams : config.gemParams).find(
+        (p) => p.key === key,
+      );
+      const now = settled ? settled.value : '(unknown)';
+      const took = settled !== undefined && configValuesMatch(valueType, value, settled.value);
```
MatiasFernandez: 💭 I couldn't find a `configurationPanel.test.ts` — is one planned? The webview side is covered really well (705 lines of `configurationView.test.ts` 👏), but the host is 600 lines with the two decisions I'd most want pinned:

- this settle check — the "accepted without error but silently ignored" case is the headline safety claim of the PR, and nothing exercises `configValuesMatch` against a `readConfiguration` result;
- `toConfigParam`'s `isEditable(entry) && (scope === 'gem' || isSystemUser)` rule, which is what stops a non-SystemUser being offered a doomed stone editor.

`debuggerPanel.test.ts` and `loginEditorPanel.test.ts` already set up the pattern (`vi.mock('vscode', () => import('../__mocks__/vscode.js'))` plus a stubbed session manager), so a host test with a fake `QueryExecutor` should be cheap here. Worth adding in this PR, or is it better as a follow-up?

client/src/extension.ts
```diff
@@ -1947,6 +1948,32 @@ export function activate(context: vscode.ExtensionContext) {
...
+    // command palette works too.
+    vscode.commands.registerCommand(
+      'gemstone.showConfiguration',
+      (item?: GemStoneSessionItem | GemStoneLoginItem) => {
+        let session: ActiveSession | undefined;
+        if (item instanceof GemStoneSessionItem) {
+          session = item.activeSession;
+        } else if (item instanceof GemStoneLoginItem) {
+          const label = loginLabel(item.login);
+          session = sessionManager.getSessions().find((s) => loginLabel(s.login) === label);
```
MatiasFernandez: 💭 This resolves a login row to a session by comparing `loginLabel` strings, but `loginLabel` is `gs_user on stone (gem_host)` — it drops `netldi`. `sameLoginTarget` in `loginTypes.ts` is the identity check the rest of the file uses for exactly this (`extension.ts:731`, `extension.ts:3798`), and it includes NetLDI.

With two logins that differ only in NetLDI and both connected, the gear on one login's row would open the panel for the other one's session — and since the panel then reads and *writes* configuration over that session, the write lands on the wrong stone. Should this be `sessionManager.getSessions().find((s) => sameLoginTarget(s.login, item.login))`?

client/src/queries/configurationReport.ts
```diff
@@ -0,0 +1,276 @@
...
+      throw new ConfigValueError(`Expected true or false, got "${value}"`);
+    }
+    case 'integer': {
+      const v = value.trim();
+      if (!/^-?\d+$/.test(v)) throw new ConfigValueError(`Expected an integer, got "${value}"`);
+      return v;
+    }
+    case 'string':
+      // Double every quote — the one escape a Smalltalk string literal needs.
+      return `'${value.replace(/'/g, "''")}'`;
```
MatiasFernandez: 💭 The module header says "All emitted Smalltalk is ASCII-only (the 3.6.x `ComStrmSetCursor` bug)", and everything else in here holds to it — the key is gated by `KEY_SHAPE`, booleans and integers are gated by their own regexes. But a `string` value is whatever the user typed (or pasted) into the editor, spliced in with only the quote doubled, so a `°` or a smart quote in a path or an argument list goes straight to the gem and puts us in exactly the case the header says we avoid.

Should `configValueLiteral` reject non-ASCII here — a `/^[\x20-\x7e]*$/` test throwing `ConfigValueError`, the same way a bad integer does — so the failure is a clear message in the panel rather than whatever 3.6.x does with it? If it's actually safe for string literals, it'd be worth a line in the header saying so, since the current text reads as an invariant this branch doesn't hold.

client/src/queries/__tests__/configurationReport.integration.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    expect(entries.some((e) => e.key === 'StnMaxSessions')).toBe(true);
+  });
+
+  it('reads and parses the live gem report', () => {
+    const entries = parseConfigReport(execute(buildGemReportCode()));
+    assertWellFormed(entries);
+  });
+
+  it('answers whether the session is SystemUser as a real boolean', () => {
+    expect(typeof sessionIsSystemUser(execute)).toBe('boolean');
```
MatiasFernandez: 💭 Nice to see the report path closed against a real stone. The setter half isn't though — `buildSetConfigCode` / `setConfiguration` are only ever run against hand-written answer strings, so nothing proves the generated `[ System gemConfigurationAt: #K put: v. 'OK' ] on: Error do: [...]` actually compiles and evaluates on a live stone, or that a refusal really does come back through `messageText` in the shape the panel unwraps.

Is it worth one more case here — set a harmless gem parameter to the value it already has (so nothing changes) and assert `{ ok: true }`, plus a stone key as a non-SystemUser to assert the sentinel path returns `ok: false` with a non-empty message?

client/src/configurationView.js
```diff
@@ -0,0 +1,614 @@
...
+    setNotice = null;
+  }
+
+  // A case-insensitive prefix match on the parameter name: typing "Gem" shows the
+  // Gem* parameters, not every key that merely contains "gem" (e.g. StnGemTimeout).
+  // Done in place so typing never triggers a full re-render (which would drop focus).
+  function applyConfigFilter() {
+    const needle = configFilter.trim().toLowerCase();
+    els.root.querySelectorAll('tr.config-item').forEach((row) => {
+      const match = !needle || row.dataset.configKey.startsWith(needle);
```
MatiasFernandez: 💭 The prefix match is deliberate and tested, so this is just a question about which one serves the report better: with ~160 stone rows, most of what I'd type to narrow it is a *middle* word — "cache", "sessions", "timeout" — and a prefix filter finds nothing for any of those (`StnGemTimeout` doesn't match "timeout", `SHR_PAGE_CACHE_SIZE_KB` doesn't match "cache"). The stated win, that "Gem" shouldn't drag in `StnGemTimeout`, feels like the narrower case.

What do you think about `includes` instead, or prefix-first ordering with substring matches after?

client/src/configurationPanel.ts
```diff
@@ -0,0 +1,602 @@
...
+        retainContextWhenHidden: true,
+        // Glyphs are inline SVG, so the webview loads no external resource.
+        localResourceRoots: [],
+      },
+    );
+    ConfigurationPanel.panels.set(sessionId, new ConfigurationPanel(panel, deps, sessionId));
+  }
+
+  /** Close every open Session Configuration panel; a no-op when none are open. */
+  static close(): void {
```
MatiasFernandez: nit: `close()` doesn't look like it has any callers — I grepped and `ConfigurationPanel` is only referenced from `extension.ts:1973` (`show`) and inside this file. Is something meant to call it (a deactivate hook, a "log out all sessions" path), or can it go?

client/src/configurationView.js
```diff
@@ -0,0 +1,614 @@
+// Webview-side script for the standalone Session Configuration panel (configurationPanel.ts).
```
MatiasFernandez: 💭 Could you group the files related to the new configuration panel in their own folder?

MatiasFernandez (approved):

client/src/extension.ts
ericwinger: This is a real bug.  Yeek.

client/src/configuration/configurationPanel.ts
```diff
@@ -0,0 +1,602 @@
...
+      appendSysadmin(`Configuration: set ${scope} configuration ${key} = ${value}`);
+
+      const config = this.readConfiguration(session);
+      void this.panel.webview.postMessage({ command: 'configuration', config });
+
+      const settled = (scope === 'stone' ? config.stoneParams : config.gemParams).find(
+        (p) => p.key === key,
+      );
+      const now = settled ? settled.value : '(unknown)';
+      const took = settled !== undefined && configValuesMatch(valueType, value, settled.value);
```
ericwinger: Added — `client/src/configuration/__tests__/configurationPanel.test.ts`, 34 tests, following the `debuggerPanel.test.ts` pattern you pointed at (mocked `vscode`, a fake `QueryExecutor` per session).

Both decisions you named are pinned:

- the settle check — a fake GCI whose report reflects sets it applied, and one whose report does not, so `configValuesMatch` against a real `readConfiguration` result is exercised in both directions. There is also a case for a value that settles under a different spelling (`True` vs `true`), which would otherwise warn on every boolean.
- `toConfigParam` — a stone runtime key comes back `settable: true, editable: false` for a non-SystemUser and `editable: true` for SystemUser; an `ALL_CAPS` key and an `other`-typed value are never editable in either scope.

I broke each rule in turn to confirm the tests actually fail without it, rather than trusting coverage.

client/src/extension.ts
ericwinger: Good catch, and chasing it turned up something worse: the branch was unreachable. `gemstone.showSessionConfiguration` is registered in `package.json` only under `view == gemstoneLogins && viewItem == gemstoneSession` — there is no login-row menu entry anywhere — so a `GemStoneLoginItem` could never arrive, and the label comparison you found could never actually fire.

Rather than fix the comparison, the whole login-row path is gone. A login row is the *parent* of its session rows, so it names no one session; when a login is connected twice there is no right answer to pick. The command now takes a session row and resolves it by `activeSession.id`, which `SessionManager` hands out as `nextId++` and never reuses — so no label, no NetLDI, no identity comparison at all.

One thing kept from your reasoning: it resolves *through* `sessionManager.getSession(id)` rather than reading `item.activeSession` directly, because a tree item outlives its session and the panel writes over whatever it is handed. A row left over from a logged-out session now says "log in" instead of opening onto a dead connection. That case has a test.

The command moved out of `extension.ts` into `configuration/showConfigurationCommand.ts` so it could be tested at all.

client/src/configuration/queries/configurationReport.ts
```diff
@@ -0,0 +1,276 @@
...
+      throw new ConfigValueError(`Expected true or false, got "${value}"`);
+    }
+    case 'integer': {
+      const v = value.trim();
+      if (!/^-?\d+$/.test(v)) throw new ConfigValueError(`Expected an integer, got "${value}"`);
+      return v;
+    }
+    case 'string':
+      // Double every quote — the one escape a Smalltalk string literal needs.
+      return `'${value.replace(/'/g, "''")}'`;
```
ericwinger: You read it right — the header stated an invariant this branch did not hold. Fixed rather than documented away.

`configValueLiteral` now tests a string value against `/^[\x20-\x7e]*$/` and throws `ConfigValueError`, the same way a bad integer does, so it surfaces as a message in the panel. Control characters are excluded along with everything above `0x7e`: a configuration value has no use for a tab or a newline, and either would split the doit.

The module header now says which parts are gated and how, instead of asserting the property globally.

Worth noting the path is reachable: on a live 3.6.2 stone there are no editable string parameters on the gem side, but four on the stone side — `StnLogFileName`, `StnGemInitialTransactionMode`, `StnCurrentTranLogNames`, `StnStandbyRole`. `StnLogFileName` is a path, which is your smart-quote-pasted-into-a-path case exactly. SystemUser only, but reachable.

client/src/queries/__tests__/configurationReport.integration.test.ts
ericwinger: Added both cases you asked for, in `configurationReport.integration.test.ts`:

- a gem parameter set to the value it already holds, asserting `{ ok: true }` and that the report still reads the same afterwards — so a green run changes nothing about the gem it borrowed, and what is under test is that the emitted doit compiles and evaluates at all.
- `StnGemTimeout` set as a non-SystemUser, asserting `ok: false`, a non-empty message, that it carries the stone words (`SystemUser`), and that the `GS-ERROR:` sentinel has been stripped before it would reach the panel.

The harness logs in as DataCurator, so the refusal path is the real one rather than a simulated sentinel. Confirmed against a live 3.6.2 stone: `a SecurityError occurred (error 2213), An operation was attempted that may only be performed by SystemUser.`

client/src/configurationView.js
ericwinger: Agreed, and changed to `includes`. Your framing was right: the stated win was the narrower case. With ~160 stone rows the word someone reaches for is usually in the middle of the name, and prefix matching found nothing for any of them — `StnGemTimeout` for "timeout", `SHR_PAGE_CACHE_SIZE_KB` for "cache".

I went with plain substring rather than prefix-first ordering: the rows are hidden in place (`style.display`) so typing never forces a re-render and never drops focus, and reordering would mean giving that up. The cost is the one you named in reverse — "Gem" now keeps `StnGemTimeout` on screen — which seems the better trade.

Tests updated: "cache" and "timeout" each find their row, and "Gem" keeps both.

client/src/configurationPanel.ts
ericwinger: Nothing calls it — grepped the same way and got the same answer. Removed.

The case it looks like it was written for is already covered: each panel subscribes to `onDidRemoveSession` and disposes itself when its own session logs out, so there is no state that a "close them all" path would need to clean up. If a deactivate hook ever wants one, it is three lines to add back with a caller attached.

client/src/configuration/configurationView.js
```diff
@@ -0,0 +1,614 @@
+// Webview-side script for the standalone Session Configuration panel (configurationPanel.ts).
```
ericwinger: Done — everything now lives under `client/src/configuration/`, following `omniSearch/` and `refactoring/`:

```
configuration/
  configurationPanel.ts
  configurationView.js
  showConfigurationCommand.ts
  gemstoneConfigDescriptions.ts
  queries/configurationReport.ts
  queries/__tests__/
  __tests__/
```

`readWebviewScript` already took a `subdir` for exactly this, so the view script is loaded as `readWebviewScript('configurationView.js', 'configuration')` and `.vscodeignore` un-ignores the new path. Worth noting this makes the other threads on this PR point at the old paths.

PR #519, opened 2026-08-27
Explorer: the Methods pane filters with VS Code's own find box, and a filtered click lands on the right row

### LINKED ISSUE #518: Clicking a row in a filtered Explorer pane selects the wrong one — the click dismisses the filter box, the full list returns, and the row under the pointer changes
## Summary

With a filter active in an Explorer pane, clicking the row you want selects a different one. The
click dismisses the still-open filter input box; dismissal is treated as a cancel, so the filter is
undone and the full list snaps back before the click resolves — and the click lands on whatever row
has moved under the pointer. Filtering a long list is exactly when you most need to click a result,
so the filter is unusable for its main purpose. All four filterable panes are affected (Dictionaries,
Class Categories, Classes, Methods).

## Environment

- **Jasper:** 1.8.14
- **GemStone server:** N/A (pure UI — no stone interaction involved)

## Steps to reproduce

1. Open the GemStone Explorer and select a class with a long method list (`Array`, for example).
2. Click the funnel in the **Methods** pane title bar and type `a`. The pane filters live to the
   matching methods; the filter input box stays open with "Filters as you type — Escape to cancel".
3. Without pressing Enter, click `add:` in the filtered list.

## Expected behavior

`add:` is selected and opened — clicking a filtered result is a *use* of the filter, not a
cancellation of it.

## Actual behavior

The filter is cleared, the unfiltered list returns, and `_at:put:` is selected and opened — the
method that happens to occupy that row position in the full list. The filter chip is gone and the
pane is back to showing everything.

## Potential fix

Nothing settled; the mechanism is in `client/src/gemstoneExplorer.ts`, in `beginFilter` (the input
box) and its `onDidHide` handler. Filtering is live, so a cancel has to be undone explicitly, and
`onDidHide` fires for a click-away just as it does for Escape — there is no `accepted` flag on that
path, so a click-away is undone as if it were a cancel. Some directions:

- Treat only Escape as a cancel and any other dismissal as an accept, so a click-away leaves the
  filter in place and the list does not move under the pointer. Escape-to-restore would keep
  working; the filter chip's ✕ already exists to clear it afterwards.
- Or keep the filter box out of the way entirely — a filter affordance that does not need to be
  dismissed cannot reorder the list on the way out.
- Whatever the route, the invariant worth testing is that the row under the pointer at mouse-down is
  the row that ends up selected.

Related: the current cancel-restore semantics came from #388 / PR #390, and the guards in
`onDidHide` were tightened for the case where selecting a class clears the Methods filter — worth
re-reading before changing them.

ericwinger:
Fixes #518.

## What changed since this PR was opened

The first version of this PR fixed the click without changing the widget: the filter's
quick-input box gained `ignoreFocusOut` so a click elsewhere could not dismiss it, and a row click
committed the box instead of cancelling it. That work is unchanged and still here — it is what the
Dictionaries, Class Categories and Classes panes use.

What is new is the **Methods pane**, which now narrows with VS Code's own find box instead. The
earlier fix compensated for a floating input having to guess what a click somewhere else meant; the
find box removes the guess, because it lives inside the pane. Seeing it working elsewhere is what
prompted the change, and it is the better answer for the panes we can move to it. #523 tracks the
remaining three.

## The bug

With a filter active in an Explorer pane and its input box still open, clicking the row you wanted
selected a different one. The click dismissed the box, dismissal was the cancel path, so the filter
was undone and the full list snapped back — under the pointer, before the click resolved — and the
click landed on whichever row had moved into that spot. Filtering a long list is exactly when you
most need to click a result, so the filter was unusable for its main purpose.

## The Methods pane: VS Code's find box

The find box opens **inside** a tree view, at its top right, with a toggle that hides non-matching
rows instead of merely highlighting them. Nothing moves under the pointer, and VS Code, not us, maps
a clicked row back to its element. The toggle stays on for that pane once set, and closing the box
keeps the text for next time.

The pane's Filter button focuses the pane and runs `list.find` (`openPaneFindWidget`). There is no
extension API for the widget: `list.find` acts on whichever list was focused last, which is why the
focus call has to come first, and nothing about the widget — its query, its mode, whether it is even
open — is readable afterwards.

## The other three panes: the original fix, unchanged

`beginFilter` sets `ignoreFocusOut`, so a click elsewhere no longer hides the box and nothing moves.
That is also what makes "only Escape is a cancel" implementable: VS Code fires the same `onDidHide`
for Escape and for a click-away, and with focus-out ignored only Escape (cancel), Enter (accept) and
our own commit can reach it. A row click in any pane then **commits** the box —
`commitFilterInput()`, wired to the five panes by `commitFilterOnRowSelection` — keeping the filter
and putting the box away.

Two selections deliberately don't count as that click: an empty one (filtering can narrow the
selected row out of the pane, and closing the box mid-word would make the filter unusable a second
way) and the filter row itself, whose click is the one that opens the box.

#523 is not a repeat of this change. Each remaining pane needs a decision about prefix-`*` versus
fuzzy matching, about the chip and title label having nothing to read the widget's state from, and —
for Class Categories — about the built-in box filtering only the rows the tree has already loaded,
where our filter flattens the whole pane.

## What stays on our own filter state

The `reads:`/`writes:`/`accesses:` filters, which run a GemStone query no text search could and are
seeded from an instance variable's context menu. They still set the Methods pane's filter, still show
the r/w glyphs and highlight the variable in the opened source. The two compose: ours picks the rows,
VS Code's box searches within them. The one thing lost is editing an active `reads:foo` token as
text — ✕ clears it and the menu item re-runs it — so the Methods filter chip's tooltip now says
"click to search within it" rather than "click to edit".

The Filter button keeps its funnel icon and title. The button still means *narrow this pane*, and
`explorerFilterUx387.manifest.test.ts` pins the convention from #387 that the funnel means filter and
the magnifier means find.

## Tests

- `explorerMethodsFindBox.test.ts` — new: focus before find (getting the order wrong opens the box
  over some other tree), no input box for that pane, a box left open elsewhere gets committed first,
  an instance-variable filter survives, both chip tooltips, and that a failing focus/find command
  comes back out of `beginFilter` instead of being swallowed.
- `explorerFilterCancel` / `explorerFilterClickAway` / `explorerFilterUx387` — the box suites moved
  off the Methods pane and onto Dictionaries, so they exercise a path a user can still reach. Their
  `beginFilter` calls are awaited, as `no-floating-promises` requires once it returns a promise.
- Each of the five changed behaviours was mutation-tested: dropping the Methods branch, swapping the
  focus/find order, dropping the commit-open-box call, collapsing the tooltip, and restoring the
  `void` on the find-box call each fail a test.

## Review

`beginFilter` reached the find box through a voided call, so a rejection from `${viewId}.focus` or
`list.find` went nowhere and the Filter button just looked inert. It is `async` now and returns that
promise; the command handler already passed its return value to VS Code, so the failure surfaces
there. The other three panes' path stays synchronous.

`main` is merged in, and `npm run lint && npm run format:check && npm run compile && npm test` all
pass on the merged tree against a freshly provisioned 3.6.2 stone — 7011 client, 322 server, 92
mcp-server, 114 skipped. Diff is client-only. Manually F5-verified by Eric: the find box opens in the Methods pane,
its toggle hides non-matching rows and stays on, and clicking a result keeps the narrowing.


client/src/gemstoneExplorer.ts
```diff
@@ -1095,12 +1124,26 @@ export class ExplorerController {
         this.setFilterState(viewId, filterBeforeEdit);
       }
       this.filteringView = undefined;
+      // Deregister this box only. beginFilter commits an open box before creating the next, so
+      // there is normally just one — but the check costs nothing and means a hide delivered
+      // late (after another box has opened) can't leave that newer box with no way to commit.
+      if (this.openFilterBox === entry) this.openFilterBox = undefined;
```
MatiasFernandez: 💭 This guard says a hide can be delivered late, after another box has opened — but `filteringView` three lines up isn't guarded the same way. If VS Code delivers `onDidHide` asynchronously, two consecutive `beginFilter` calls run as: old `hide()` → new box sets `filteringView = viewId` → old `onDidHide` fires and sets `filteringView = undefined` + `syncTitles()`. The new box's pane then loses its live `Filter: …` header while the user is still typing in it (`syncTitles` only shows it when `filteringView === viewId` and the pane has a selection). The mock fires hide synchronously, so no test can see this.

Worth giving it the same ownership check (`if (this.filteringView === viewId) this.filteringView = undefined;`)? It would also make the "keeps the order deterministic: the old box's onDidHide has finished before this one records itself" claim on the `commitFilterInput()` call in `beginFilter` true regardless of when VS Code delivers the event — as written, that comment and this one assume opposite things about the same event.

client/src/gemstoneExplorer.ts
```diff
@@ -5757,6 +5800,38 @@ class ClassDropController implements vscode.TreeDragAndDropController<ClassNode>
...
+  for (const view of views) {
+    view.onDidChangeSelection((e) => {
+      const node = e.selection[0];
+      // Two selections are not the click this is about. An EMPTY one isn't a click at all —
+      // typing in the box can narrow the selected row out of the pane, and that must not close
+      // the box mid-word. And the filter row's click is the one that OPENS the box (its command
+      // is the pane's filter command), so committing on it would race the very edit it asked
+      // for — whether VS Code fires the selection before or after the row's command is its
+      // business, not ours; skipping the row settles it either way.
+      if (node === undefined || node instanceof FilterChipItem) return;
```
MatiasFernandez: 💭 The empty-selection half of this guard tells me VS Code does fire `onDidChangeSelection` when a keystroke narrows the selected row out of the pane. Does it also fire when the selected row *survives* the filter? `setFilterState` → `providerFor(viewId).refresh()` and `getChildren` builds fresh node objects each time, so if VS Code re-reports the selection against a new instance that's a non-empty selection which isn't a `FilterChipItem` — it would commit and close the box on the first keystroke, in what's probably the most common flow (select a method, then open the funnel to find another one).

Could you confirm that one in the dev host? The mock can't reach it, so it wouldn't show up in the new tests either way.

client/src/gemstoneExplorer.ts
```diff
@@ -1050,40 +1053,66 @@ export class ExplorerController {
...
     let accepted = false;
     box.title = 'Filter';
+    // Keep the box open when focus leaves it. VS Code otherwise hides it on any click
+    // elsewhere, and hiding is the cancel path: the pre-edit filter goes back, the pane
+    // re-expands to the full list — under the pointer, before the click that dismissed the box
+    // resolves — and the click lands on whichever row has moved into that spot. Clicking a
+    // result is exactly what a filter is for, so the box must survive the click. With focus-out
+    // ignored, the box closes only on Escape (cancel), Enter (accept), or commitFilterInput
+    // (accept), so a dismissal is no longer ambiguous between the three.
+    box.ignoreFocusOut = true;
```
MatiasFernandez: 💭 With focus-out ignored, what puts the box away when the user leaves the Explorer instead of clicking a result? Concretely: open the Methods filter, type `si`, then click a different session in the Sessions pane — that isn't one of the five panes, so nothing commits, `ctl.reset()` clears every filter, and the box is left hanging over a reset Explorer describing a filter that no longer exists. Escape at that point is also not obviously available, since focus is in the tree. Clicking into the editor and carrying on there is the same shape.

Should `reset()` (and maybe `clearFilter`) commit the box too, so the reachable states stay just "box open over the pane it filters" and "no box"?

client/src/__tests__/explorerFilterClickAway.test.ts
```diff
@@ -0,0 +1,240 @@
...
+        listener = h;
+        return { dispose: vi.fn() };
+      },
+      select: (node: unknown = { label: 'a row' }) => listener?.({ selection: [node] }),
+      selectNothing: () => listener?.({ selection: [] }),
+    };
+  }
+
+  it('commits the open filter box from a selection in any pane', () => {
+    const ctl = { commitFilterInput: vi.fn() };
```
MatiasFernandez: 💭 Both halves of the fix are covered, but nothing joins them: these wiring tests stub `commitFilterInput`, and the controller tests above call it directly. Is it worth one test that wires the real thing — `commitFilterOnRowSelection(ctl, pane)` after `ctl.beginFilter(METHODS)` + `__type('si')`, then `pane.select()` asserting the filter is still `'si'` and the box hidden? That's literally the #518 path, and it's the only place a mismatch between the two sides (argument order, a renamed method on the `Pick<…>`) would surface.

client/src/__mocks__/vscode.ts
```diff
@@ -443,6 +446,9 @@ export const window = {
         await onAccept?.();
       },
       __hide: () => fireHide(),
+      __clickAway: () => {
```
MatiasFernandez: Having `__clickAway()` honour `ignoreFocusOut` rather than just firing a hide is the right call — it means the tests fail if the flag is ever dropped, instead of encoding the fix twice. Thanks for taking that extra step ❤️

client/src/gemstoneExplorer.ts
```diff
@@ -1041,49 +1051,103 @@ export class ExplorerController {
...
   // Enter and Escape and onDidAccept only for Enter, so the accepted flag is what tells them
   // apart. On cancel we restore the filter captured when the box opened, which is the
   // previously accepted filter when the user was editing an existing one (the box is seeded
   // from it) rather than simply clearing.
+  //
+  // A click away from the box is neither: `ignoreFocusOut` keeps the box open through it, and
+  // the click's own handler commits the box instead (commitFilterInput) — see the note there.
   beginFilter(viewId: string): void {
+    if (viewId === VIEW_METHODS) {
+      void this.openPaneFindWidget(viewId);
```
MatiasFernandez: 💭 The `void` here makes a rejection invisible: if `${viewId}.focus` or `list.find` throws (a command id changed on VS Code's side, the pane not registered yet), the Filter button silently does nothing and nothing is logged. Since `beginFilter` is what the command handler calls, could it be `async` so VS Code reports the failure, or at least `.catch()` with a log?

There is a knock-on in the tests too: `explorerMethodsFindBox.test.ts` needs `vi.waitFor` to see the commands run, where an awaited call could just be asserted directly.

MatiasFernandez (approved):

client/src/gemstoneExplorer.ts
ericwinger: Good catch — it is `async` now, your first option.

`beginFilter` returns the promise instead of voiding it, and the command handler
already passed its return value back to VS Code, so a failing `${viewId}.focus`
or `list.find` surfaces as a failed command rather than a button that quietly did
nothing. The other three panes' path is entirely synchronous, so the promise they
return is already resolved and nothing about them changes.

The knock-on you predicted went the way you expected: `vi.waitFor` is gone in
favour of asserting straight after the `await`, and the other `beginFilter` call
sites are awaited too (`no-floating-promises` covers test files, so they had to
be). One case is new — a rejecting focus/find has to come back out of
`beginFilter` rather than being swallowed; restoring the `void` fails it.

PR #520, opened 2026-08-27
GemStone Search: rank the selector scan, reset on a session switch, add a refresh

### LINKED ISSUE #517: GemStone Search: reset on session switch, add a refresh, and fix missing Array>>at:
GemStone Search (`client/src/omniSearch/`) keeps cached corpora and on-screen state that can outlive
the thing they describe. Three related gaps:

## 1. Reset the search when the selected session changes

Switching the selected session (`SessionManager.selectSession` → `onDidChangeSelection`) is not wired
to GemStone Search at all. `OmniSearchViewProvider.ensureEngine()` does compare
`builtForSession` against the current session id and rebuild the engine — but only when something
else happens to call it (a reveal, a config change, the next search). Until then the panel keeps
showing the previous session's query, results, scope, and references pivot, and the preview pane
shows source fetched from the old stone. The docked panel is a long-lived view, so this is the normal
case, not an edge case.

Wanted: on a selection change, the search resets — engine rebuilt (or marked stale for the hidden
panel, matching the existing `syncPending` gate), query and results cleared, any references pivot
dropped, scope back to its default — so nothing on screen belongs to a session you are no longer in.
The editor-tab Spotter (`OmniSearchPanel`) has the same exposure: its `deps.sessionId` is fixed at
construction and it only compares ids to *ignore* notifications, so a pinned Spotter survives a
session switch bound to the old session.

## 2. A refresh button

Cached corpora (Classes, Dictionaries, Globals — primed once per engine) are only kept honest by the
hooks listed in `omniSearch/DESIGN.md`: local compiles, Explorer class removals, symbol-list changes,
and a full resync on commit/abort/file-in. Anything done by *executing code* — a class created or
removed in a workspace, a method compiled by evaluating a `compileMethod:`, a new global — is
invisible until the next commit or abort. That staleness window is documented as by-design, but there
is currently no way for the user to close it on demand.

Wanted: a refresh control on the panel/Spotter title bar (and a `gemstone.search.refresh` command, so
it can be bound and invoked from the palette) that re-primes every provider — the existing
`engine.resync()` path — and re-runs the current query, with the usual busy indication while the
image-wide GCI executes run.

## 3. Investigate: `Array>>at:` is missing when searching `at:`

Typing `at:` in GemStone Search does not return `Array>>at:` anywhere in the Methods results, where it
should be an obvious top hit.

First place to look is the bounded server scan in the Methods provider
(`omniSearch/providers/methodsProvider.ts` → `queries/searchSelectors.ts`): the scan is a substring
match walked in **symbol-list order, not relevance order**, and it short-circuits the moment it has
`maxServerScan` matches. For a selector as common as `at:` the ceiling is certainly hit long before
`Array` is reached, so the class the user actually wants is cut off by ordering, not by ranking —
`SERVER_OVERFETCH` widens the pool but cannot fix an order that never gets there. Worth confirming
before designing a fix, and worth checking whether the truncation was surfaced to the user at all in
this case.

If that is the cause, ranking needs an input the current scan cannot give it. Options to weigh: rank
server-side, seed the scan from the selector's implementors (`implementorsOfSelector`) when the query
is an exact selector, or raise/segment the scan so well-known dictionaries are covered first.

ericwinger:
Closes #517. Three defects in GemStone Search, all of them the panel answering out of state that is no longer true.

## 1. Typing `at:` never returned `Array>>at:`

The Methods scope asks the stone for a bounded number of matching selectors so searching-as-you-type stays fast. The stone collected the **first** `limit` matches it walked into — dictionary-hash order, nothing to do with relevance. Measured on a 3.6.2 base image: 1142 selectors *contain* `at:` and only 31 **are** `at:`, so with a slice of 80 the exact implementors sat far past the cutoff and were never sent. The slice filled with `instVarAt:put:` and `floatAt:put:` from a couple of incidental classes.

`buildSelectorSearchCode` now sorts matches into tiers — the selector **is** the term, **starts with** it, merely **contains** it — and returns them best-first, each tier capped. What a cut-off drops is now the least relevant tail. The walk no longer short-circuits on a match count (a better-tier hit can be anywhere in the image); the one early exit left is a full exact tier, where nothing later can displace a row. Cost for `at:`: ~27 ms against ~2 ms — and a full walk was already the price of every precise term, which never reached the old cutoff at all.

**One trap worth knowing about if you touch that query.** `GciLibrary.execute` compiles our source with `Utf8` as its string class, so every literal in it is a `Utf8` while `sel asString` answers a `String`. Comparing the two raises `ArgumentError 2718, Unicode argument disallowed in String comparison` and takes the whole search down. The tier test therefore uses integer sizes plus `includesString:` and **no string `=` anywhere**; a unit test guards against a "clearer" rewrite. The live-stone integration test caught this — the unit tests could not.

## 2. Making another session active did nothing to the search

The docked panel rebuilt its engine only when something else happened to ask it to (a reveal, a settings change, the next keystroke); an open Spotter never rebuilt at all. Until then both kept answering — and *opening rows* — out of the session the user had left.

Both hosts now listen to `SessionManager.onDidChangeSelection`. Two decisions worth recording:

- **The webview is wiped, not just the engine.** Dropping the engine is enough for a settings change, because the rows on screen are still true. Here they are not: they came from the old session, they still look live, and activating one would open a document against the session that is now current.
- **The wipe is not deferred while hidden, though the re-prime still is.** The `visible` gate exists to keep image-wide GCI executes off a background path, and that still holds for re-priming. It does not hold for clearing the screen: a reveal cannot un-show stale rows.

The Spotter is **re-pointed in place** rather than closed and reopened — it is an editor tab the user put there, possibly pinned, and the tab, its pin and its HTML are all session-independent. `show()` for a different session takes the same path, which replaced its old dispose-and-recreate branch. Logging out of the last session is the same event with nothing to bind to: both hosts reset and say *Log in to a GemStone session to search*.

## 3. Nothing could pick up code created by executing it

A `subclass:` evaluated in a workspace, a method compiled by a `compileMethod:` doit, a global assigned in a doit — none of them announce anything the panel can listen for, so short of a commit or abort the cached class/dictionary/global/category corpora stayed stale with no upper bound.

A **⟳** in the search bar (both surfaces), in the panel title bar, and `gemstone.search.refresh` in the palette reloads every corpus and re-runs the current term. It ignores the visibility gate the sync hooks use — the user asked for it now — and clears any deferred sync so one click is one walk.

It is a separate engine call (`refresh`) rather than the button wired to `resync`, and the difference is the pivot: `resync` deliberately leaves a references list alone, which made the ⟳ a dead button for anyone reading one. `refresh` re-asks the stone who references the row the pivot was taken from, keeps whatever filter is typed into it, and leaves the pivot if that row has been deleted.

Also fixed, found while testing: collapsing the panel disposes the view, and the fresh webview a reopen creates was never sent its `config` message (empty tab row, zero debounce until the first search refilled it).

## Testing

`npm test` green: 6641 client + 322 server + 92 mcp. New coverage:

- `methodsRelevance.integration.test.ts` — live stone, both releases: the bounded scan returns the exact implementors of `at:` including `Array>>at:`, they lead the slice, and a substring-only term still comes through (the tiers rank, they must not filter).
- `searchSelectors.test.ts` — tier caps, best-first order, the exact-tier-only early exit, and the no-string-`=` guard.
- `omniSearchViewProvider.test.ts` — session switch wipes + rebinds (visible and hidden), no-op on re-selecting the same session, logout resets, refresh reloads once when a hidden sync was also owed, refresh builds nothing for a view never instantiated, config re-sent on reopen.
- `omniSearchPanel.test.ts` — the Spotter rebinds in place, ignores a re-selection of its own session, resets on logout, and builds nothing when closed.
- `omniEngine.test.ts` — `refresh` vs `resync` over a pivot, pivot filter preserved across a re-fetch, pivot left when its target is gone.
- `omniRefreshAndReset.test.ts` — the ⟳ posts `refresh` and keeps the typed term; `reset` clears query, rows, preview, breadcrumb, error, references chip, and returns the scope to All.

Manually verified on a live 3.7.5 stone against a fixture (a class with exact/prefix targets plus 250 substring decoys, above the server ceiling): ranking A1–A5, refresh B1–B6, session switching C1–C5, and no-regression checks D1–D5.


client/src/omniSearch/omniSearchPanel.ts
```diff
@@ -117,6 +117,39 @@ export class OmniSearchPanel {
...
+   *  `resolveDeps` is a thunk so nothing is built when no Spotter is open. It answering null means
+   *  there is nothing left to search (the last session logged out): the Spotter stays open — it is an
+   *  editor tab the user put there — but resets and says so, rather than showing a departed session's
+   *  results. */
+  static onSessionSelectionChanged(resolveDeps: () => OmniPanelDeps | null): void {
+    const panel = OmniSearchPanel.current;
+    if (!panel) return; // nothing open — don't build providers for a Spotter that isn't there
+    const deps = resolveDeps();
+    if (!deps) {
+      panel.reset();
```
MatiasFernandez: 🚫 On the logout path we wipe the webview but leave `this.engine` and `this.deps` bound to the session that just went away — and unlike the docked panel, nothing here can catch that later.

The docked provider handles it: `onSessionSelectionChanged` sets `this.engine = undefined`, so the next webview message hits `if (!this.engine && !(await this.ensureEngine())) return;`, `ensureEngine` finds no context and bails with the "log in" notice. The Spotter has no such gate — `onMessage` goes straight into `dispatchEngineMessage(this.engine, m)`.

So: last session logs out, the Spotter tab is still open (we deliberately keep it), the user types. The engine is still holding session 1's primed class/dictionary/global corpora, so it happily answers with rows out of the session that's gone — and `activate` calls `this.deps.activate(result, …)`, which closes over the departed session's GCI handle. That's the same "rows still look live" failure this PR is fixing, just reached through logout instead of a switch.

Should this branch also drop the engine (and have `onMessage` re-check), the way `ensureEngine` does on the docked side? The `omniSearchPanel.test.ts` logout case asserts the `reset` + `error` messages go out, so it would pass either way — worth an assertion that a `query` after logout doesn't reach the engine.

client/src/omniSearch/omniSearchViewProvider.ts
```diff
@@ -92,6 +101,61 @@ export class OmniSearchViewProvider implements vscode.WebviewViewProvider {
...
+   *
+   *  Unlike `onSessionSynced` this is NOT gated on visibility or deferred: the user asked for it now.
+   *  It clears any pending hidden sync, since a full reload is strictly more than that sync owed. And it
+   *  calls the engine's `refresh` rather than its `resync`, which is what makes it re-fetch an open
+   *  references list instead of leaving it stale (see the engine). */
+  async refresh(): Promise<void> {
+    // Never instantiated: there is nothing on screen to refresh, and building an engine here would pay
+    // three image-wide GCI executes for a panel the user has not opened (the `gemstone.search.refresh`
+    // command reaches both hosts, so this fires even when the Spotter is the chosen UI).
+    if (!this.view) return;
```
MatiasFernandez: 💭 `this.view` is set once in `resolveWebviewView` and never cleared — the class comment above `viewResolved` says so explicitly ("it only reports whether the view has EVER been built"). So this guard catches "the Spotter is the chosen UI and this view was never opened" (which is what the test covers), but not "the user collapsed the panel", where the view is disposed and `this.view` is still truthy.

In that case `gemstone.search.refresh` from the palette pays the three image-wide GCI executes the comment right above is arguing we shouldn't pay, and posts the results to a webview nobody is looking at. Every other catch-up path in this file gates on `this.view?.visible` instead. Is `visible` the right guard here too — or, if the intent is "refresh now so it's fresh when I reopen", is that worth saying in the comment?

client/src/omniSearch/omniSearchPanel.ts
```diff
@@ -171,6 +204,33 @@ export class OmniSearchPanel {
...
+  private reset(): void {
+    this.panel.webview.postMessage({ command: 'reset' });
+  }
+
+  /** Reload every cached corpus and re-run the current term. See the static `refresh` for why. */
+  private async refresh(): Promise<void> {
+    this.panel.webview.postMessage({ command: 'busy', on: true });
+    // `refresh`, not `resync`: an open references list is as stale as the corpora, so it is re-fetched
+    // rather than left alone the way a commit leaves it.
+    const view = await this.engine.refresh(this.deps.onError);
```
MatiasFernandez: 💭 If `engine.refresh()` rejects, the busy spinner stays on for good — we post `busy: true` up front and only clear it on the two resolve paths.

`reprimeAll` swallows per-provider failures, but the pivot branch does a bare `await deps.resolveReferences(pivotSource)`, and resolving senders of a common selector against a busy session can throw. Coming in via the webview button that's survivable — `onMessage`'s catch posts `error`, and the webview's `error` handler calls `setBusy(false)`. But `OmniSearchPanel.refresh()` (the palette command / title-bar ⟳) is `void current?.refresh()` with no catch, so there the rejection is unhandled and the panel is left permanently faded. `OmniSearchViewProvider.refresh()` has the same shape.

Would a `try { … } finally { post busy:false }` around the engine call — or a catch in the two static entry points — be worth it here?

client/src/omniSearch/__tests__/omniSearchPanel.test.ts
```diff
@@ -35,8 +47,8 @@ function deps(sessionId: number): OmniPanelDeps {
   } as unknown as OmniPanelDeps;
 }
 
-describe('OmniSearchPanel.show', () => {
-  it('refocuses on the same session, but replaces + rebinds the engine on a different one', () => {
+describe('OmniSearchPanel session binding', () => {
+  it('refocuses on the same session, rebinds on a different one, and never opens a second panel', () => {
```
MatiasFernandez: 💭 This `it` now covers six distinct behaviours (same-session refocus, explicit switch, selection-change rebind, no-op re-selection, logout reset, closed-Spotter no-op), and the header explains it's one block because `OmniSearchPanel.current` leaks between tests.

The last few lines of the test itself suggest the way out though — `panelA.dispose()` clears the singleton via the panel's own `onDidDispose`. Could an `afterEach` doing that let these become six named `it`s? The names are the documentation here, and right now a failure reports "refocuses on the same session, rebinds on a different one, and never opens a second panel" no matter which of the six broke. It also means the first failing assertion hides the other five.

client/src/omniSearch/__tests__/methodsRelevance.integration.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    const rows = searchSelectors(exec, 'at:', { limit, ignoreCase: true });
+
+    // The scan really is up against its bound — this is the condition under which the bug appeared.
+    expect(rows).toHaveLength(limit);
+    // Yet the exact implementors are what came back, `Array>>at:` among them.
+    expect(rows.some((r) => r.className === 'Array' && r.selector === 'at:')).toBe(true);
+    // And they lead: everything before the first non-exact row is an `at:` implementor.
+    const firstInexact = rows.findIndex((r) => r.selector !== 'at:');
+    expect(firstInexact).toBeGreaterThan(1);
```
MatiasFernandez: 💭 `toBeGreaterThan(1)` is doing two jobs at once and it took me a minute: it's guarding against `findIndex` answering `-1` (which would make the `slice(0, -1)` on the next line pass vacuously) *and* asserting "at least two exact hits lead". Where does the `1` come from — is it just "more than one", or a property of the image?

Same question about `toBeLessThan(10)` below: with `maxResultsPerCategory` at 20, is 10 "comfortably within the first page" or is there something specific about it? Both would read better as named constants, or as an explicit `expect(firstInexact).not.toBe(-1)` plus a comment on the threshold — this test is the regression guard for #517 and the next reader shouldn't have to reverse-engineer the numbers.

client/src/omniSearch/__tests__/omniSearchViewProvider.test.ts
```diff
@@ -271,3 +272,172 @@ describe('GemStone Search docked panel — reporting whether a reveal landed', (
...
+  async function open(visible = true) {
+    const provider = new OmniSearchViewProvider(vi.fn(async () => fakeContext()));
+    const { view, on } = fakeView(visible);
+    provider.resolveWebviewView(view as never);
+    void on.message({ command: 'ready' });
+    // The handler is fire-and-forget (`void this.onMessage(m)`), so wait for the engine it builds AND
+    // let the rest of the handler drain — otherwise its own `flushPendingSync` lands mid-test and looks
+    // like the code under test resyncing.
+    await vi.waitFor(() => expect(createOmniEngine).toHaveBeenCalled());
+    await new Promise((resolve) => setTimeout(resolve, 0));
```
MatiasFernandez: nit: this is the same macrotask yield as the `settle()` helper at the top of the "session sync while hidden" block (and it's inlined again in the "reopening the view" test below), each with its own paragraph explaining why it's needed. Worth hoisting `settle()` to module scope and calling it in all three, so the explanation lives in one place?

MatiasFernandez (approved):

client/src/omniSearch/omniSearchPanel.ts
ericwinger: Yes — fixed, and you found a second bug behind it.

The logout path now calls a new `clearSession()` that drops the engine as well as wiping the webview, and `onMessage` gates on it: with no engine every message but `close` gets the notice instead of reaching `dispatchEngineMessage`. That is deliberately the same shape as the docked side's `ensureEngine` check, just without a resolver to rebuild from — a later login rebinds us through `onSessionSelectionChanged`.

Which is where the second bug was. The rebind guard compared session ids only:

```ts
if (panel.deps.sessionId === deps.sessionId) return;
```

After a logout the engine is gone but `deps` still holds the old id, so logging back in and getting the SAME id handed back read as a no-op — and the Spotter stayed engine-less for good. It now compares the engine too.

Test added as you suggested: a `query` after logout must not reach `engine.search`, plus a case for the same-id relogin.

Worth recording what it takes to reach this at all, because it is narrower than it looks and I had it wrong at first: the docked view has a `"when": "gemstone.hasActiveSession"` clause, so logging out your last session removes it from the panel entirely — there is nothing left to show a notice in. An UNPINNED Spotter disposes on focus-out, so clicking away to log out closes it first. Only a PINNED Spotter tab survives to take this path. That is now said in the doc-comment rather than left for the next reader to derive.

client/src/omniSearch/omniSearchViewProvider.ts
ericwinger: Good catch — `visible` is the right guard, and it is now the one used.

I did not want to just drop the request though: skipping it outright means the panel you reopen is the stale one you asked to refresh, with nothing to catch up later (`ensureEngine` short-circuits on a reopen because the session has not changed, so `flushPendingSync` has nothing to flush). So there is now a `refreshPending` flag alongside `syncPending`, flushed on reveal and on `ready` — and when both are outstanding the refresh wins, since it rebuilds every corpus and re-fetches the references list on top.

That meant splitting the entry points: public `refresh()` keeps the visibility gate (palette command, title-bar button), and a private `reload()` does the work with no gate. The webview's own ⟳ calls `reload()` directly — a message from the webview is proof enough that someone is looking.

New test: `defers the command to the next reveal when the panel is collapsed`.

client/src/omniSearch/omniSearchPanel.ts
ericwinger: Right, and it stranded the spinner on both hosts. Both now wrap the engine call:

```ts
let view: OmniViewData | null = null;
try {
  view = await engine.refresh(this.deps.onError);
} catch (e: unknown) {
  const message = e instanceof Error ? e.message : String(e);
  this.deps.onError?.(message);
  this.panel.webview.postMessage({ command: 'error', message });
}
if (view) this.postView(view);
else this.panel.webview.postMessage({ command: 'busy', on: false });
```

A catch rather than a `finally`, because the two no-view cases want the same ending anyway: superseded and threw both need a bare `busy: false`, and only a real view clears the spinner by replacing the results. The webview's `error` handler also calls `setBusy(false)`, so the explicit post is belt-and-braces — but it keeps the invariant readable at the call site instead of depending on the other end.

Test: `takes the spinner off when the reload throws`.

client/src/omniSearch/__tests__/omniSearchPanel.test.ts
ericwinger: Done — it is seven named `it`s now, and you were right that `panel.dispose()` was the way out. An `afterEach` disposes whatever the test opened, and a small `open()` helper returns the created panel, so each test starts from no open Spotter regardless of shuffle order.

The extra two beyond your six are the ones this round of review turned up: the engine being dropped on logout, and the same-id relogin.

The header comment lost its "it is all one `it` because the singleton leaks" paragraph and now explains the `afterEach` instead.

client/src/omniSearch/__tests__/methodsRelevance.integration.test.ts
ericwinger: Fair — both numbers were carrying more meaning than they showed.

`toBeGreaterThan(1)` really was doing two jobs, so they are now separate:

```ts
const firstInexact = rows.findIndex((r) => r.selector !== 'at:');
expect(firstInexact).not.toBe(-1);
expect(firstInexact).toBeGreaterThan(1);
```

with a comment saying why each is there — the first because `slice(0, -1)` would pass vacuously, the second because a single leading hit would also be true of a scan that merely stumbled on `at:` early.

`toBeLessThan(10)` is now `NEAR_TOP_OF_PAGE`, defined as `OMNI_DEFAULTS.maxResultsPerCategory / 2`. To answer directly: nothing in the image fixes it. The ranking only promises the exact implementors come first; how many there are (31 on a 3.6.2 base image) and how their class names sort are both image facts I did not want to pin a test to. So the bound is a readability claim — "near the top of the page you are looking at" — and the constant's comment says exactly that, rather than implying a measurement.

client/src/omniSearch/__tests__/omniSearchViewProvider.test.ts
ericwinger: Hoisted — `settle()` is at module scope now with the explanation in one place, and all three sites call it.

PR #521, opened 2026-08-28
Redesign the SUnit run-limit guard test for CI (Phase 12)

MatiasFernandez:
## Why

The blocking-call guard's only assertion branched on how many TestCase subclasses the live stone happened to hold, so CI's bare vendor extent always ran the within-cap path — every kernel SUnit suite in one synchronous, un-interruptible GCI call (~2s on 3.6.2, ~46s on 3.7.5, against vitest's 5s default).

## What changed

Added a bounded, deterministic fixture (`runLimitProbeFixture.ts`) that creates `MAX_RUN_CLASSES + 1` throwaway TestCase subclasses in one doit, so both directions of the guard are asserted identically on every matrix cell in ~15ms. Deleted `querySunitRunLimit.smoke.test.ts` now that its assertion lives in the CI-run integration suite.

## Testing

`npm run lint`, `npm run format:check`, `npm run compile` and `npm test` all pass.


client/src/queries/__tests__/querySunit.integration.test.ts
```diff
@@ -162,13 +163,34 @@ describe('SUnit queries (integration)', () => {
...
-    // it end-to-end means branching on the live image's size, and the
-    // within-cap branch runs every discovered suite in one un-interruptible GCI
-    // call — too slow for CI on some stones, and unkillable by a vitest
-    // timeout. It's parked in the on-demand gci project
-    // (src/__tests__/gci/querySunitRunLimit.smoke.test.ts) pending a redesign
-    // that trips the guard deterministically; see that file's header.
+    // The blocking-call guard (MAX_RUN_CLASSES). Both directions are asserted with
+    // a fixture sized on purpose, so the same assertion runs on every matrix cell.
+    it('refuses a discover-all selection over the cap, without running any suite', () => {
+      installRunLimitProbeClasses(exec, MAX_RUN_CLASSES + 1);
```
npapagna: Are we missing the `MAX_RUN_CLASSES - 1` case (below the threshold)

client/src/queries/__tests__/querySunit.integration.test.ts
```diff
@@ -162,13 +163,34 @@ describe('SUnit queries (integration)', () => {
...
+
+      // Whatever the stone already held, the selection is now over the cap.
+      expect(discoverAllTestClasses(exec).length).toBeGreaterThan(MAX_RUN_CLASSES);
+      expect(() => runFailingTests(exec)).toThrow(/too many to run[\s\S]*Narrow the run/);
+    });
+
+    it('runs an at-cap selection and reports its failures', () => {
```
npapagna: **[altitude] At-cap test cost scales unbounded with `MAX_RUN_CLASSES`**

This test installs 100 real TestCase subclasses then calls `runFailingTests(exec, names)`, which per `runFailingTests.ts` actually runs `cls suite run` for all of them (200 method sends) and re-runs each failing test a second time to capture messageText (100 more sends) — ~300 real GCI-side executions in one blocking call, directly proportional to `MAX_RUN_CLASSES` with no independent test-side cap.

If `MAX_RUN_CLASSES` is ever raised (e.g. to fit a legitimately larger customer suite), this test's runtime grows right along with it, silently reintroducing the '46s on GemStone 3.7.5'-style cliff this PR's own description cites as the original bug — except now deterministically, on every CI run across the whole release matrix.

client/src/queries/__tests__/querySunit.integration.test.ts
```diff
@@ -162,13 +163,34 @@ describe('SUnit queries (integration)', () => {
...
-    // (src/__tests__/gci/querySunitRunLimit.smoke.test.ts) pending a redesign
-    // that trips the guard deterministically; see that file's header.
+    // The blocking-call guard (MAX_RUN_CLASSES). Both directions are asserted with
+    // a fixture sized on purpose, so the same assertion runs on every matrix cell.
+    it('refuses a discover-all selection over the cap, without running any suite', () => {
+      installRunLimitProbeClasses(exec, MAX_RUN_CLASSES + 1);
+
+      // Whatever the stone already held, the selection is now over the cap.
+      expect(discoverAllTestClasses(exec).length).toBeGreaterThan(MAX_RUN_CLASSES);
+      expect(() => runFailingTests(exec)).toThrow(/too many to run[\s\S]*Narrow the run/);
```
npapagna: **[test-coverage] Discover-all happy path is now completely untested**

The only remaining no-args call to `runFailingTests(exec)` is in this over-cap test, which only exercises the throw branch. Nothing in the suite calls it with no args and lets it run to completion (under/at cap) against a live stone — the deleted smoke test's weak 'within cap: returns an array' assertion for this exact path has no replacement. A regression specific to the `DISCOVER_ALL_TEST_CLASSES` fragment actually being run (as opposed to the explicit-classNames path the other two new tests exercise) would pass this whole suite yet break for any real caller invoking 'run everything under the cap'.

client/src/queries/__tests__/querySunit.integration.test.ts
```diff
@@ -162,13 +163,34 @@ describe('SUnit queries (integration)', () => {
...
-    // The blocking-call guard (MAX_RUN_CLASSES) is NOT covered here. Asserting
-    // it end-to-end means branching on the live image's size, and the
-    // within-cap branch runs every discovered suite in one un-interruptible GCI
-    // call — too slow for CI on some stones, and unkillable by a vitest
-    // timeout. It's parked in the on-demand gci project
-    // (src/__tests__/gci/querySunitRunLimit.smoke.test.ts) pending a redesign
-    // that trips the guard deterministically; see that file's header.
+    // The blocking-call guard (MAX_RUN_CLASSES). Both directions are asserted with
+    // a fixture sized on purpose, so the same assertion runs on every matrix cell.
+    it('refuses a discover-all selection over the cap, without running any suite', () => {
```
npapagna: **[test-coverage] Test title's 'without running any suite' claim isn't independently verified**

The body only asserts `toThrow(...)`; it never checks a side effect that would only exist if a suite had actually run. The guarantee depends entirely on the guard preceding the run loop in `runFailingTests.ts`'s Smalltalk source — a future reordering of those two statements would silently break it without this test catching it.

client/src/queries/__tests__/runLimitProbeFixture.ts
```diff
@@ -0,0 +1,39 @@
...
+// One doit for the whole batch, so cost is flat in `count` — ~4ms for 100 on
+// 3.6.2. Installed inside the caller's transaction, so the harness's per-test
+// abort reclaims them; no teardown.
+
+import { QueryExecutor } from '../types';
+
+const PROBE_CLASS_PREFIX = 'JasperRunLimitProbe';
+
+/** Names of the classes {@link installRunLimitProbeClasses} creates, in order. */
+export function runLimitProbeClassNames(count: number): string[] {
```
npapagna: **[correctness] Two independent name-generation mechanisms must stay in sync**

Class names are built twice: once inside the Smalltalk doit (`i printString` in a `1 to: count do:` loop) and once separately in JS (`Array.from({length: count}, (_, i) => ...i+1)`). They agree today, but nothing enforces it — an independent edit to either (zero-based indexing, zero-padding) would silently desync the returned names from the classes actually created in GemStone.

Relatedly: this export has exactly one caller in the repo (itself, from `installRunLimitProbeClasses`) — worth considering whether it needs to be public at all.

client/src/queries/__tests__/runLimitProbeFixture.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/**
+ * Creates `count` trivial TestCase subclasses in UserGlobals, each with one
+ * passing and one failing test, and returns their names.
+ */
+export function installRunLimitProbeClasses(exec: QueryExecutor, count: number): string[] {
+  exec(`[| c |
+1 to: ${count} do: [:i |
+  c := TestCase subclass: '${PROBE_CLASS_PREFIX}', i printString
```
npapagna: **[correctness, plausible] Missing `ifAbsent:` guard vs. sibling fixture**

`sunitProbeFixture.ts` guards the identical `TestCase subclass: ...` operation with `UserGlobals at: #'JasperProbeTest' ifAbsent: [...]`; this fixture redefines unconditionally. Normally harmless under the harness's per-test transaction abort, but a crashed prior run or a long-lived stone bypassing the abort (e.g. `dev:fresh:keep-installs`) could hit a redefinition conflict instead of failing cleanly.

client/src/queries/__tests__/querySunit.integration.test.ts
```diff
@@ -162,13 +163,34 @@ describe('SUnit queries (integration)', () => {
...
+      installRunLimitProbeClasses(exec, MAX_RUN_CLASSES + 1);
+
+      // Whatever the stone already held, the selection is now over the cap.
+      expect(discoverAllTestClasses(exec).length).toBeGreaterThan(MAX_RUN_CLASSES);
```
npapagna: **[efficiency] Over-cap test walks the full symbolList twice**

`discoverAllTestClasses(exec)` here and the identical internal walk inside `runFailingTests(exec)`'s no-args path both execute `DISCOVER_ALL_TEST_CLASSES`. Unlike the other two new tests (explicit class lists, O(1) in image size), this test's runtime still scales with however many TestCase subclasses the live image actually holds.

client/src/queries/__tests__/runLimitProbeFixture.ts
```diff
@@ -0,0 +1,39 @@
...
+    instVarNames: #() classVars: #() classInstVars: #() poolDictionaries: #()
+    inDictionary: UserGlobals.
+  c compileMethod: 'testPasses  self assert: true'
```
npapagna: **[efficiency] Over-cap test compiles 202 methods it never executes**

The over-cap test only needs 101 classes to exist so the pure count guard trips before any suite runs — neither `discoverAllTestClasses` nor the guard's count check inspects methods, so a bare class shell would trip it identically. `installRunLimitProbeClasses` always compiles both `testPasses`/`testFails` regardless of caller, making these 202 compiles pure waste for this call site.

Also: `System myUserProfile symbolList` is re-evaluated on every `compileMethod:` call inside the loop (200 times for count=100) instead of hoisted once into a temp, unlike the sibling pattern already used in `runFailingTests.ts`'s `DISCOVER_ALL_TEST_CLASSES`/`buildPatternFilter`. Minor, but easy to fix.

client/src/__tests__/gci/querySunitRunLimit.smoke.test.ts
```diff
@@ -1,59 +0,0 @@
-// On-demand smoke test for runFailingTests' blocking-call guard (MAX_RUN_CLASSES).
-//
-// PARKED HERE DELIBERATELY. The rest of the SUnit-query smoke tests live in the
-// automatic suite (`src/queries/__tests__/querySunit.integration.test.ts`);
-// this one does not, because as written it cannot run in CI:
-//
-// Correct behavior depends on image size, so the test branches on how many
```
npapagna: **[altitude] Historical diagnostic data lost with nothing preserved**

This header recorded specific figures (CI's bare vendor extent has 7 kernel SUnit classes; ~2s vs ~46s on GemStone 3.7.5; vitest's 5s timeout) that justified `MAX_RUN_CLASSES` and this guard's existence. The replacement comment in `querySunit.integration.test.ts` is two generic sentences with none of that data — it's now only recoverable via git blame on this deleted file.

npapagna (changes requested):
Thanks @MatiasFernandez!
Left a couple of comments.

client/src/queries/__tests__/runLimitProbeFixture.ts
MatiasFernandez: Fixed — the names are generated once in JS and interpolated into the doit as a literal array, so the doit can't create a name the caller didn't get back. Also made `runLimitProbeClassNames` module-private.

client/src/queries/__tests__/runLimitProbeFixture.ts
MatiasFernandez: Added the `ifAbsent:` guard for consistency with `sunitProbeFixture.ts`. FWIW the conflict itself didn't reproduce — re-running the unguarded fixture twice on a live 3.7.5 stone just makes a new class version, no error and no duplicate in the discovered set — but the guard is free and makes the fixture idempotent by construction.

client/src/queries/__tests__/querySunit.integration.test.ts
MatiasFernandez: The over-cap test does evaluate the `DISCOVER_ALL_TEST_CLASSES` fragment inside the production doit — it has to, to compute `classes size` before the guard raises — so a compile or result-shape regression in the fragment fails there (or in the `discoverAllTestClasses` tests). The run-loop-and-parse half is covered end-to-end by the explicit-classNames test and the `classNamePattern` test, which walks the symbolList with the same shape and runs to completion; the two halves don't interact, since the loop just consumes an OrderedCollection of classes whichever expression produced it. What's left uncovered is that seam, and the only test that closes it is the unbounded live no-args run this phase deleted on purpose (every kernel SUnit suite in one un-interruptible GCI call: ~46s on 3.7.5 against vitest's 5s timeout, with no way to bound the discovered set).

client/src/queries/__tests__/querySunit.integration.test.ts
MatiasFernandez: Good catch — the title was asserting something the body didn't check, and the only thing holding it up was the order of two adjacent statements in the generated doit.

The probe classes' `testPasses` now writes a marker global (`#JasperRunLimitProbeSuiteRan`), which the fixture removes at install time, and the over-cap test asserts the marker is still absent after the throw:

```ts
expect(() => runFailingTests(exec)).toThrow(/too many to run[\s\S]*Narrow the run/);
expect(anyRunLimitProbeSuiteRan(gci, handle)).toBe(false);
```

`anyRunLimitProbeSuiteRan` lives in `runLimitProbeFixture.ts` and is a thin wrapper over the existing `GciLibrary#isIncludedInUserGlobals`. Confirmed it has teeth: asserting `true` in the at-cap test (where suites genuinely run) passes, so reordering the guard after the run loop would now fail this test. Cost is one extra tiny GCI call in the over-cap test only — no extra `compileMethod` and nothing added to the at-cap test's runtime or the symbolList walks.

client/src/queries/__tests__/querySunit.integration.test.ts
MatiasFernandez: The guard is `classes size > MAX_RUN_CLASSES`, so the boundary pair is N and N+1 — `MAX_RUN_CLASSES - 1` sits on the same side as the at-cap test and can only fail if that one already does. The at-cap test ("runs an at-cap selection and reports its failures") is the below-threshold case; a `MAX - 1` variant would add ~99 more live suite runs for no extra coverage. Worth noting too that a below-cap selection isn't constructible on the discover-all path at all, since the fixture can only add to whatever the stone already holds — which is why that test goes through explicit `classNames`.

client/src/queries/__tests__/querySunit.integration.test.ts
MatiasFernandez: Measured it. The mechanics are right — 200 test-method executions plus 100 re-runs, linear in `MAX_RUN_CLASSES` — but the cost isn't there.

Local, install vs. `runFailingTests`, same instrumentation on both stones:

```
            3.7.5                       3.6.2
N=10   install=1.4ms run=1.3ms     install=1.5ms run=1.2ms
N=25   install=1.2ms run=1.5ms     install=1.3ms run=1.7ms
N=50   install=1.8ms run=2.3ms     install=1.8ms run=2.6ms
N=100  install=3.3ms run=4.1ms     install=3.4ms run=4.5ms
N=100  install=3.1ms run=4.1ms     install=3.5ms run=4.4ms
N=100  install=3.6ms run=4.1ms     install=3.6ms run=4.6ms
```

3.7.5 is marginally *faster*, not slower: ~31 µs/class on the run half vs ~37 µs on 3.6.2. The at-cap test itself reports 13 ms on 3.7.5, 10 ms on 3.6.2.

CI agrees, on hardware neither of us controls. Per-file durations from run 33537673189 (all 14 tests plus every `beforeEach` install, so an upper bound on the at-cap test):

```
ubuntu  3.6.2    202 / 216 ms      windows  3.6.2    405 ms
ubuntu  3.6.8    260 ms            windows  3.6.8    479 ms
ubuntu  3.7.2    163 ms            windows  3.7.2    424 ms
ubuntu  3.7.4.3  244 ms            windows  3.7.4.3  458 ms
ubuntu  3.7.5    147 / 168 ms      windows  3.7.5    335 / 427 ms
```

Worst cell in the matrix is 479 ms for the whole file, and 3.7.5 is among the fastest.

The 46s cliff was a real image's TestCases doing real setUp/tearDown work on the discover-all path — it scaled with the *weight* of the tests, not the class count. The probes are two one-line methods each. At the measured slope the cap would have to reach roughly 60,000 before this test alone approached vitest's 5s timeout; at a raised cap of 1,000 it costs ~50 ms.

Leaving it as is. An independent test-side cap would also stop this from being the at-cap boundary case, which is the property I leaned on in the `MAX - 1` thread.

client/src/queries/__tests__/runLimitProbeFixture.ts
MatiasFernandez: Hoisted `System myUserProfile symbolList` into an `sl` temp, so it's evaluated once per install instead of twice per class (202 times for the over-cap call) — matches the pattern already in `DISCOVER_ALL_TEST_CLASSES` and `buildPatternFilter`. Measured before/after on a 3.6.2 stone, 4 runs each: the difference is within run-to-run noise, so this is a consistency fix rather than a real saving.

I've left the method compilation alone. Since you wrote this, the over-cap test gained an `expect(anyRunLimitProbeSuiteRan(gci, handle)).toBe(false)` assertion (from another thread on this round) — it checks that no probe's `testPasses` wrote its marker global, which is what actually pins the refusal to *before* the run loop. With bare class shells there'd be no `testPasses` to write the marker, so "marker absent" would be vacuously true and the assertion would prove nothing. Those 202 compiles are what give it teeth, so they're no longer waste for this call site.

client/src/queries/__tests__/querySunit.integration.test.ts
MatiasFernandez: Confirmed, and measured: one full walk is ~0.6 ms warm on both 3.6.2 and 3.7.5 (108 classes discovered after the fixture installs), so the redundant one is ~0.2% of the file's ~400 ms. Leaving it — the second walk is `runFailingTests`' own and irreducible, and the first is what shows the DISCOVER_ALL fragment is exercised at all in this test (see the discover-all thread).

The interesting bit from measuring: on a freshly started 3.7.5 stone, one discover-all test really does cost ~30 ms — but it's one-shot warm-up, and vitest's shuffle pins it on whichever discover-all test runs first. This time that was the dedup test, which does a single walk and no install, so the double walk isn't what's paying for it. Steady state the over-cap test is 8–11 ms, tied with the at-cap one.

client/src/__tests__/gci/querySunitRunLimit.smoke.test.ts
MatiasFernandez: Good catch — those figures were worth keeping. Split them by what each one justifies: the ~2s-vs-~46s-on-3.7.5 cliff for a bare extent's 7 kernel SUnit classes now sits directly above `MAX_RUN_CLASSES` in `runFailingTests.ts`, since it's the only evidence the cap needs to exist at all, and it's framed so the 100-trivial-probes-in-~4ms number can't be misread as the cap being over-tight. The CI-parking history and vitest's 5s timeout went into the guard tests' comment in `querySunit.integration.test.ts` (including the ~60,000 headroom figure from the at-cap-cost thread), and while I was in `runLimitProbeFixture.ts` I refreshed its stale "~4ms for 100 on 3.6.2" with the numbers measured during this round: ~3.3-3.4ms to install 100 on both versions, ~4.1ms/~4.5ms to run them on 3.7.5/3.6.2. The 7-class figure was already in that fixture header and stays there.

npapagna (approved):
Thanks @MatiasFernandez!
🚢

PR #522, opened 2026-08-28
Migrate the transcript sink tests to CI (phase 9a)

MatiasFernandez:
The 9 live-stone tests for the Transcript sink lived in `client/src/__tests__/gci/gciTranscriptSink.test.ts` — the `gci` vitest project, which `npm test` excludes. They only ever ran when someone remembered `npm run test:gci`, so in practice they never ran in CI at all. This moves them onto the `useIntegrationTest` harness in the default project, and takes the opportunity to stop the live-mode tests hand-rolling machinery that production already shares.

## Why the old file needed a rewrite, not just a move

Everything the harness does, that file did itself: its own `GciTsLogin`/`GciTsLogout` in `beforeAll`/`afterAll`, its own session object, its own transaction discipline, and a version/platform gate. None of that survives the move — the harness owns login, the per-test transaction and its abort, and the commit guard.

The gate went too. `GciTsNbPoll` appeared only in that file's *own* polling helper, never in production code, so there was nothing version-specific left to gate once the polling moved to the shared runner.

## The one production change

`runNbCall`'s `onReady` widens from `() => T` to `() => T | Promise<T>`, matching `pollNbToCompletion`, which has always awaited it and documents the async case explicitly (the transcript-forwarding settle loop is the reason it does). Without the widening, an async `onReady` infers `T` as a promise and the call answers a nested `Promise<Promise<…>>`. Every existing caller passes a sync function, so nothing else moves.

## Verification

`npm run lint`, `npm run format:check`, `npm run compile` and `npm test` all pass against a live stone — 6718 client / 322 server / 92 mcp-server tests green, with the 8 migrated tests among them (they were invisible to `npm test` before this PR).


client/src/__tests__/transcriptSink.integration.test.ts
```diff
@@ -0,0 +1,167 @@
...
+  // nbRunner keys its per-session state off `session.id`, so that stays fixed
+  // for the file, while `handle` has to track the live one -- a wrapper
+  // memoized on first use would keep handing tests a handle that a test's
+  // logout had invalidated.
+  let theSession: ActiveSession;
+
+  useIntegrationTest((testContext) => {
+    gci = testContext.gciLibrary;
+    handle = testContext.session;
+    theSession = { id: 1, gci, handle } as unknown as ActiveSession;
```
ericwinger: 🧪 `client/src/__tests__/testActiveSession.ts` already builds this, and ~17 integration files use it — including `browserQueries.integration.test.ts:34` and `explorerRemoveMethod.integration.test.ts` in this same directory. It returns a real `ActiveSession` (with a genuine `login` from `resolveTestConnection`, and a `stoneVersion` cached per `GciLibrary`), so no `as unknown as` is needed.

The cast is what makes this worth changing rather than just tidier: it hides that the object has no `login` and no `stoneVersion`. If `installTranscriptSink` / `settleNbResult` / `runNbCall` ever start reading either — `pluginFeatures.isApplicable` version-gates off `stoneVersion` — this file gets `undefined` at runtime with no compile error to catch it.

```ts
import { testActiveSession } from './testActiveSession';

const session = (): ActiveSession => testActiveSession(gci, handle);
```

Rebuilding per call from the live `gci`/`handle` also makes the memoization comment above unnecessary — the staleness it guards against can't arise. `testActiveSession` uses `id: -1`, still fixed for the file, so `nbRunner`'s per-session keying is unaffected.

client/src/__tests__/transcriptSink.integration.test.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * The UTF-8 contents of `oop`, failing rather than silently dropping a fetch
+   * error -- a negative `bytesReturned` covers a string too long for the
+   * buffer as well as an outright GCI failure.
+   */
+  const fetchString = (oop: bigint): string => {
+    const { bytesReturned, data, err } = gci.GciTsFetchUtf8(handle, oop, 256);
+    expect(err.number).toBe(0);
+    expect(bytesReturned).toBeGreaterThanOrEqual(0n);
```
ericwinger: 🧪 This assertion doesn't cover the truncation case the comment above claims for it, and is redundant against the one on line 49.

Every vendored header, 3.6.2 through 3.7.5 (`vendor/gci-headers/*/gcits.hf`), documents `GciTsFetchUtf8` as:

> Returns -1 **if an error was returned in `*err`**, otherwise returns the number of bytes stored starting at `*dest` […] If `*requiredSize` is > destSize, then `*dest` contains an incomplete result not terminated with a zero byte.

So `bytesReturned < 0` happens **only** when `err.number != 0` — which line 49 already asserts — and a string too long for the 256-byte buffer comes back silently truncated with `err.number == 0` and `bytesReturned == 256`. `fetchString` would return the truncated prefix, and `toBe('no error')` would fail with a confusing diff (or a `toContain` would quietly pass).

The wrapper already hands back `requiredSize`, which includes the terminator, so completeness is `requiredSize <= destSize`:

```ts
const BUFFER_BYTES = 256;
const { requiredSize, data, err } = gci.GciTsFetchUtf8(handle, oop, BUFFER_BYTES);
expect(err.number).toBe(0);
expect(requiredSize).toBeLessThanOrEqual(BigInt(BUFFER_BYTES));
return data;
```

The doc comment needs to change with it — as written it states the opposite of what the header says.

client/src/__tests__/transcriptSink.integration.test.ts
```diff
@@ -0,0 +1,167 @@
...
+    return data;
+  };
+
+  // The harness clears SessionTemps after every test, and the sink is
+  // registered only there -- so it is installed per test, exactly as
+  // sessionManager does per login, rather than once for the file.
+  beforeEach(() => {
+    expect(installTranscriptSink(session())).toBe(true);
```
ericwinger: 🧪 When this fails, the only thing reported is `expected false to be true`. `installTranscriptSink` catches the real cause and routes it to `logError`, and in this file `gciLog`'s output channel is the `vscode` mock's `appendLine: vi.fn()` — the message is discarded, not printed.

The old `gci/` file mocked `gciLog` to the console for exactly this reason, and said so:

```ts
// transcriptSink logs through gciLog, which needs the vscode module — absent
// in the gci project. Console logging keeps failures diagnosable here.
vi.mock('../../gciLog', () => ({ … }));
```

The first half of that reason is gone now that `vscode` is mocked, but the second half still holds, and it applies to every test in the file since the install runs in `beforeEach`. Keeping the console shim (with the comment trimmed to the surviving reason) costs three lines and keeps a bad install self-explaining.

client/src/__tests__/transcriptSink.integration.test.ts
```diff
@@ -0,0 +1,167 @@
...
+    // non-ASCII character is built at runtime via codePoint:, never literal.
+    exec("Transcript nextPutAll: 'caf', (Character codePoint: 233) asString. 'ok'");
+
+    expect(drainTranscript(session())).toContain('café');
+  });
+
+  it('switching to live mode returns any buffered residue', () => {
+    exec("Transcript nextPutAll: 'residue'. 'ok'");
+
+    const residue = setTranscriptLive(session(), true);
```
ericwinger: 💭 This test — and the three below it — leave the sink in live mode. The old file always put it back in a `finally`; here the reset is delegated to the harness clearing `SessionTemps` in `afterEach`.

That works, but it lands the harness's own teardown doits in a combination `transcriptSink.ts`'s module doc calls out as unsafe: `assertCommitGuardIsStillArmed` and the `System abortTransaction` unwind both run on the `GciTsExecuteFetchBytes` path, *before* the `SessionTemps` reset, with the sink still live — and

> This is the only safe mode for GciTsExecuteFetchBytes-based calls […] a forwarder send on that path degenerates to rtErrExpectedClass with no continuable context, killing the call.

It's safe today only because neither teardown doit happens to write to `Transcript`. If one ever does, it fails as `sessionCleanupFailed` poisoning every remaining test in the file, with nothing in the message pointing at the transcript. An

```ts
afterEach(() => setTranscriptLive(session(), false));
```

next to the existing `beforeEach` makes the file independent of that, and reads as the counterpart to installing per test.

client/src/nbRunner.ts
```diff
@@ -337,7 +337,7 @@ export function pollNbToCompletion<T>(
 export function runNbCall<T>(
   session: ActiveSession,
   start: () => { success: boolean; err: GciError },
-  onReady: () => T,
+  onReady: () => T | Promise<T>,
```
ericwinger: 🧪 The widening is right, and the `Promise.resolve(ready).then(…)` branch it feeds is well documented — but nothing at the mocked boundary exercises it. `nbRunner.test.ts`'s five `runNbCall` cases all pass a sync `onReady`, and a grep for an async one across `client/src/**/*.test.ts` finds nothing, so the contract the widening exists to serve is now proved only by a live-stone test.

The part worth pinning is the behaviour, not the type (a narrowing back to `() => T` would fail `tsc` anyway): that a promise-returning `onReady` keeps the run **unsettled** until that promise settles — which is what keeps the progress notification's soft/hard-break Cancel live for the whole settle loop, per the doc on `pollNbToCompletion` — and that a rejecting one rejects the run. Two cases next to the existing ones cover both.

Worth noting this is a pre-existing gap being surfaced, not one this PR introduced: `pollNbToCompletion` has accepted `T | Promise<T>` all along; this PR is just the first caller to use it.

ericwinger (approved):
Nice migration — this genuinely closes a coverage hole (9 tests that never ran
in CI), and dropping the hand-rolled login/transaction/polling machinery in
favour of the shared harness and nbRunner is the right shape. The runNbCall
widening is correct and matches what pollNbToCompletion has always documented.
Full matrix green on 3.6.2 and 3.6.8 confirms the shared runner's socket-poll
fallback carries these tests on pre-3.7 stones.

Five notes inline, none blocking. The one I'd act on first is the fetchString
truncation guard: every vendored gcits.hf (3.6.2 → 3.7.5) says GciTsFetchUtf8
returns -1 only when *err is set, so a result too long for the 256-byte buffer
comes back silently truncated — the assertion as written is equivalent to the
err.number check above it, and the doc comment states the opposite of the
header.

One thing not in the diff: the description says the old file had "a version/
platform gate" that was removed. main's copy has no gate — no skipIf, no version
or platform check, and none in gciTestConfig.ts either. What it had was an
undeclared dependency on GciTsNbPoll in its own polling helper, which would just
have failed on a pre-3.7 stone. And GciTsNbPoll *is* in production code —
nbRunner.pollNbResultReady calls it when available. The accurate point is the
better one: production wraps it behind a GciTsSocket fallback, so moving onto
the shared runner removes the version dependency outright.

client/src/__tests__/transcriptSink.integration.test.ts
MatiasFernandez: Good catch 🙏 — fixed. `fetchString` now asserts `requiredSize <= 256` (which is what actually catches truncation) instead of the redundant `bytesReturned >= 0`, and the doc comment says what the header says.

client/src/__tests__/transcriptSink.integration.test.ts
MatiasFernandez: Good catch — the cast was doing real damage by hiding the missing `login`/`stoneVersion`. Switched to `testActiveSession(gci, handle)`, rebuilt per call, and dropped the memoization comment since the staleness it guarded against can't arise anymore. Confirmed your point about `id`: `nbRunner` only uses `session.id` as a `draining` map key and `transcriptSink` only logs it, so the fixed `-1` is fine. Lint, format, compile clean; the 8 tests still pass against a live stone.

client/src/__tests__/transcriptSink.integration.test.ts
MatiasFernandez: Good catch — the safety here was incidental, resting on neither teardown doit ever writing to `Transcript`. Added the counterpart hook right below the per-test install:

```ts
// Counterpart to installing per test: the harness's teardown doits run on the
// blocking execute path before it clears SessionTemps, and a live forwarder
// send there has no continuable context (see transcriptSink.ts's module doc).
afterEach(() => {
  setTranscriptLive(session(), false);
});
```

Ordering works out: no `sequence.hooks` in `client/vitest.config.ts`, so the default `'stack'` applies and this `afterEach` runs before the harness's — the reset lands while the sink is still installed, ahead of `assertCommitGuardIsStillArmed` and the abort unwind. `setTranscriptLive` also goes through `runFetchString`, which swallows errors and returns `''`, so it can't itself fail teardown after a test that logged out. File passes 8/8 against the stone.

client/src/__tests__/transcriptSink.integration.test.ts
MatiasFernandez: Good catch — restored it. Verified the chain first: `installTranscriptSink` swallows the exception into `logError`, and `getGciLog()` here resolves to the vscode mock's `appendLine: vi.fn()`, so the cause was genuinely being dropped for all 8 tests.

Two deliberate deviations from the old shim:

- **`logError` only, not `logInfo`.** The install logs a success line through `logInfo` once per test — noise on green runs, and the diagnosability argument is only about the failure path.
- **Spread the real module** rather than a bare factory, matching `gemstoneFileSystemProvider.test.ts`. Now that `vscode` is mocked the real `gciLog` loads fine, so this can't leave `logWarning`/`getGciLog` undefined if the import graph ever grows to reach them.

Lint, format:check, and compile are clean; the file's 8 tests still pass against a live stone.

client/src/nbRunner.ts
MatiasFernandez: Nico is working on a revamp of non-blocking calls, and the nbRunner will be affected. So this is going to change soon. I'll leave it as it is for now and can revisit it in the future if Nico's changes still leave this gap

PR #524, opened 2026-08-31
Stop leaked layout samplers between debugger tests

ericwinger:
## The failure

`main`'s post-merge Health Check failed on `windows-latest 3.7.2` after #481 merged:

```
FAIL src/__tests__/debuggerPanel.test.ts > DebuggerPanel > reuses the first
debugger's column pair when a second halt opens
AssertionError: expected [ [ 'vscode.getEditorLayout' ], …(1) ] to have a length of 1 but got 2
```

[Failing run](https://github.com/GemTalk/Jasper/actions/runs/33417679086/job/99572198702)

It is not the merge, and not a race in production code. It is a leaked timer in the test fixture.

## Cause

Every `DebuggerPanel` that opens a companion source editor starts a sampler (`startLayoutSampler`):

```ts
this.layoutSampler = setInterval(() => void this.captureSourceRatio(), 2000);
```

Every 2s it calls `vscode.commands.executeCommand('vscode.getEditorLayout')` so a drag of the
panel/source divider is remembered. That interval is cleared in exactly one place — `dispose()`.

The tests never dispose their panels. The `beforeEach` clears the static `panels` registry, but
that only drops the map entries; the panel objects stay alive with their timers still running. Those
timers write to the same `executeCommand` mock the tests count, so a tick landing mid-test inflates
the next `getEditorLayout` assertion.

**Why Windows only:** the interval is 2s. On Linux this file spends ~0.7s in tests, so it never fires
once. The Windows job spent 68.88s in tests, giving it dozens of chances.

**Why it looked like a flake:** `sequence.shuffle` is seeded from `Date.now()`, so every run gets a
different order. The PR run and the merge run were never running the same order, and the same commit
can pass then fail 8 minutes later.

## The change

Turn the samplers off before dropping the registry. Test-only; **production is unaffected**, because
VS Code fires `onDidDispose`, which clears the interval.

## Verification

Leaving production code alone and only shortening the sampler interval — which just gives the leaked
timers more turns on a fast host:

| sampler interval | before | after |
|---|---|---|
| 2000ms (real) | passes on Linux | passes |
| 100ms | neighbouring layout-persistence test fails | passes |
| 50ms | **reproduces the CI failure verbatim** | passes |
| 10ms x 10 seeds | 0/10 pass | **10/10 pass** |

Full suite against a live 3.6.2 stone: **436/436 files, 6859 passed, 0 failed** (plus server 322 and
mcp-server 92).

## One thing worth a reviewer's opinion

This lifts a single line out of `dispose()` rather than calling `dispose()`. I built and tested the
`dispose()` version too — it also passes, but it needs to be sequenced above `vi.clearAllMocks()`
(its `debug.clearStack` call otherwise trips a test that asserts clearStack is never called), and it
fires `void closeEmptyGroups(...)`, a floating promise with its own 2s timer whose tail calls
`tabGroups.close` in a *later* test. 20 assertions in this file watch `tabGroups.close`, three of
them counting exactly — so that version re-introduces the same class of bug, currently harmless only
because `tabGroups.all` is emptied a few lines later.

The tradeoff: what is here starts no new async work and does not depend on line order, but it only
knows about `layoutSampler`. A future timer added to `DebuggerPanel` would start this flaking again.
Happy to add a guard that fails loudly on an unexpected live timer field if that is preferred.


client/src/__tests__/debuggerPanel.test.ts
```diff
@@ -623,9 +623,27 @@ describe('DebuggerPanel', () => {
...
+    // never close theirs. Every one that opened a companion source editor also
+    // left a layout sampler running — a 2s setInterval calling
+    // `vscode.getEditorLayout`, cleared only in dispose() — so stop those first.
+    // On a host slow enough for the interval to fire mid-test (Windows CI, where
+    // this file takes ~70s rather than ~1s) a leaked sampler lands an extra
+    // getEditorLayout on the shared executeCommand mock and inflates whichever
+    // test is counting those calls. Then clear the registry, so a leftover panel
+    // can't hand its carved column pair to the next test's debugger (see
+    // liveDebuggerColumns).
+    const panelRegistry = (DebuggerPanel as unknown as { panels: Map<number, Set<unknown>> })
```
npapagna: **Reuse:** `DebuggerPanel.disposeForSession` (client/src/debuggerPanel.ts:1364-1370) already tears down every panel for a session, clearing `layoutSampler` as a side effect of `dispose()`, and that path is already exercised by an existing test ("disposeForSession disposes every panel created for the session"). This block reimplements the sampler-stopping step inline via a private-field cast instead of calling it. If `dispose()`'s teardown logic changes later, this test-only copy can silently go stale.

That said, I know the PR description already tried the full `dispose()` route and rejected it, since it also fires `closeEmptyGroups`'s floating promise, which pollutes a later test's `tabGroups.close` counts, so a straight swap to `disposeForSession` isn't risk-free either. Flagging for awareness rather than as a drop-in fix.

client/src/__tests__/debuggerPanel.test.ts
```diff
@@ -623,9 +623,27 @@ describe('DebuggerPanel', () => {
...
+      .panels;
+    for (const sessionPanels of panelRegistry.values()) {
+      for (const panel of sessionPanels) {
+        const sampled = panel as { layoutSampler?: ReturnType<typeof setInterval> };
```
npapagna: **Simplification:** the private-internals cast is split across two statements, a registry-level cast to `Set<unknown>` on the outer `panelRegistry`, then a per-iteration `panel as {...}` cast here into `sampled`. Since the real shape is already known, both could collapse into one cast at the point the registry is obtained, e.g. `Map<number, Set<{ layoutSampler?: ReturnType<typeof setInterval> }>>`, removing the intermediate `unknown` typing and this extra binding.

client/src/__tests__/debuggerPanel.test.ts
```diff
@@ -623,9 +623,27 @@ describe('DebuggerPanel', () => {
...
+        if (sampled.layoutSampler) {
```
npapagna: **Simplification:** this `if (sampled.layoutSampler)` guard is unnecessary, both Node's and the DOM's `clearInterval` accept `undefined` as a no-op, and assigning `undefined` to an already-`undefined` field is also a no-op. Unconditional `clearInterval(sampled.layoutSampler); sampled.layoutSampler = undefined;` would behave identically for every panel, including ones that never started a sampler. Not a bug, just avoids implying that clearing an unset timer is unsafe.

npapagna (approved):
Thanks @ericwinger! left a couple of comments.

client/src/__tests__/debuggerPanel.test.ts
ericwinger: Agreed on the staleness risk, and I looked at swapping it in — two things keep it out. `disposeForSession` goes through `panel.dispose()`, which runs the whole teardown including the fire-and-forget `closeEmptyGroups` at debuggerPanel.ts:4675 — the exact pollution the PR description rejected. And it is keyed by one `sessionId`, while this `beforeEach` has to sweep every session left in the registry, so it would need a loop over the keys anyway rather than a one-line swap. Leaving the targeted sampler stop here, with the comment block explaining why it exists.

client/src/__tests__/debuggerPanel.test.ts
ericwinger: Done in 4000ac6f — one cast at the registry (`Map<number, Set<{ layoutSampler?: ReturnType<typeof setInterval> }>>`), and the `sampled` binding is gone.

client/src/__tests__/debuggerPanel.test.ts
ericwinger: Right — dropped in 4000ac6f. It now calls `clearInterval(panel.layoutSampler)` unconditionally, with a short comment noting that clearing an unset timer is a no-op so nobody re-adds the guard. Type-checks clean against client/tsconfig.json.

PR #527, opened 2026-08-31
Databases & Versions: one panel for the releases and the databases made from them

### LINKED ISSUE #257: create new database but click away from vs code loses changes
If you try to create a database, you'll get 4 popups - stone version, name, etc.  But if you click away from VS Code to say, lookup an available netldi name, you'll lose all your previous changes. 

Should either let the user see running stones/netldis in the "add database" tool or get the info from the user in a way that they can click away from VS code and not lose their work.

---
_Migrated from GitLab: [GemStone/grail#68](https://code.gemtalksystems.com/GemStone/grail/-/work_items/68)_

### LINKED ISSUE #526: Integrate the Manager's Databases and Versions into the sidebar
## Summary

Versions and databases live in two places that disagree. The sidebar has a **Versions**
section, a **Databases** section and a **Processes** section; the GemStone Manager webview
(draft #454) shows much of the same information again, in more detail, on a surface of its
own. Bring them together: the Manager's databases and versions become one panel reached
from the sidebar's Databases section, and the sidebar sections they replace go away.

The same change replaces the four-Quick-Pick database creation flow with a form, which is
what #257 asks for — a chain of popups loses every answer already given the moment focus
leaves VS Code.

This is one piece of a larger effort to move the GemStone Manager into the sidebar a
component at a time (see Context below); it is not that whole effort.

## Acceptance criteria

- The **Versions** sidebar section is gone; installed releases are a list in the new panel,
  with install, register-a-local-build, remove, open-folder and open-terminal on each row.
- Getting a release is one action: downloading and unpacking are not two steps, and removing
  a release takes away everything installing it put there.
- The panel is reached from a visible affordance on the sidebar's **Databases** section, not
  only from the Command Palette.
- Creating a database is a single form: every answer visible at once, names already in use
  shown beside the fields that must avoid them, and nothing lost when focus leaves VS Code.
  (#257)
- A database shows its logins, the sessions opened from them, its stone and NetLDI, and its
  files — so the **Processes** sidebar section is no longer needed and is removed.
- The sidebar's Databases section keeps working as it does today.

## Context

The agreed architecture, from a design discussion in August 2026: navigation stays in the
sidebar where VS Code users expect it, advanced management moves into separate web views,
and the Manager is refactored piecemeal — one isolated component per pull request, each
reviewable on its own and each leaving Jasper working if the rest never arrives. The
countervailing constraint from the same discussion is discoverability: anything reachable
only through the Command Palette is effectively hidden, so each panel gets a visible
affordance on the row or section it belongs to.

The first component to move was **Session Configuration** (#511, PR #513). This is the
second. Still to come: a **Quick Start** panel gathering the first-run actions, and the
Manager's remaining sections — operating-system prerequisite checks, and connections. The
all-in-one Manager webview is retired once the last component has moved.

## Implementation notes

The panel would follow the conventions established by `client/src/configuration/`: a host
module owning the webview and its message protocol, a companion plain-JS view file injected
as a nonce'd script under a strict CSP, and a dependency bag rather than an import of
`extension.ts`.

Lifting the Manager's two sections needs more than the two render functions — the draft
Manager branch also added supporting API those sections depend on, including a split of
`VersionManager` into a network half and a disk half.

One inconsistency to settle while doing this: the Session Configuration panel draws its
glyphs as inline SVG codicon paths, while the Manager uses the `@vscode/codicons` webfont.
One of the two should probably win.

ericwinger:
## Summary

The **Versions** section leaves the sidebar. Its twelve row actions only ever worked by clicking a row — typed into the Command Palette they threw, because each handler expects a tree row and got nothing, so "it's still in the palette" was never a real fallback. They now live in a **Databases & Versions** panel, opened from the Databases section title bar.

This is the second slice of moving the GemStone Manager into the sidebar piecemeal, after Session Configuration (#511 → #513). The panel's two sections are lifted from the draft Manager (#454) and trimmed; its OS checks, Connect section, guided tour and Configuration section are deliberately not here.

## Versions

A release is a row in a table — version, status, size, released — rather than a line of hover icons:

- **Install** downloads *and* unpacks, and deletes the archive once the unpack lands, so **Remove** is its exact inverse. Previously Remove left the archive behind and the release reappeared offering to install itself again, which reads as the removal having failed.
- **Remove** says Remove whichever kind of row it is — an archive, an installed product tree, or a link to a build you compiled. The tooltip says which is about to happen.
- Register Local Version, Show in Finder, Open Terminal, Install Windows Client.

## Creating a database is a form

No longer four Quick Picks in a row, so going to look up a free NetLDI name no longer throws away the three answers already given (**closes #257**). Names in use are listed beside the fields you have to avoid them in; a clash or a bad character is refused where it was typed; the NFS warning that was a modal is now a line in the form with the override — and a "choose another folder" button — beside it.

## Each database

Its logins, the sessions opened from them, its stone and NetLDI, and its files (newest first, with timestamps). A session row offers what the Logins & Sessions row shows, plus **Ping**, which moves here out of Session Configuration — a liveness check isn't a setting, and here the answer can name the session that asked.

**New: back up a stopped database's extents.** Jasper's two existing backups both run through a live session, so neither could be taken of a database sitting there stopped — exactly when copying extents is safe and cheap. Kept in `backups/extents/`, apart from the logical backups, because both are `.dbf` files restored in completely different ways; one list offering one restore over both is how a database gets destroyed. Restore is the existing **Replace Extent → Browse**, which already refuses a live stone.

## Also here

- **Logging in no longer drags the sidebar to the Explorer.** Two causes: an explicit view switch, and the Explorer's auto-select ending in `TreeView.reveal`, which makes VS Code *show* that view and brings its container forward.
- **Open Workspace is no longer per-session** — it worked on the *active* session, and the per-session variant quietly re-selected your session as a side effect. Moved to the Logins & Sessions title bar.
- **Processes leaves the sidebar.** A database's own processes are on its panel row. Note this was the only place a stone Jasper *didn't* create was visible.
- **Show in Finder opens the folder** for a version and a database alike; they differed, and arrived differently relative to the editor window.

## Worth your eye

1. **This panel uses the `@vscode/codicons` webfont; the Session Configuration panel from #513 uses inline SVG codicon paths and says so in a comment.** The dependency was added because the lifted code was built around the font, but the two now disagree. Happy to convert either way.
2. **`showConfigurationCommand` now also accepts `{ sessionId }`**, not just a tree item — the panel has no tree row to hand over.
3. **Session commands are allow-listed** in the host. The message comes from a webview; passing its string to `executeCommand` would let anything reachable by name run with a live session.
4. **Supporting code came from #454 too**, not just the two panes: `VersionManager` splits into a network half and a disk half (which also fixes downloaded releases vanishing from the list when the download site is unreachable), plus `wslStatFilesSync` for file times and a `SessionManager.onDidAddSession` event.

## Testing

Client 6959, server 322, mcp 92, all green; lint, format and compile clean. Client-only diff, so no CI matrix. New behaviour is mutation-tested — the form surviving a mid-typing refresh, Remove taking the archive with it, the stopped-stone backup gate, the session-command allow-list, and the no-jump-on-connect fix each fail when the fix is reverted.

Closes #257
Closes #526


ericwinger:
Pushed a review pass (`e7fcb290`) before you start, so you are not reading code I already knew was wrong.

I reviewed the branch against `main` and audited the webview message protocol in both directions — every action the view can send against every case the host handles. Six real defects, all fixed and covered:

- **The Databases section's two New Database buttons created a broken database.** Both posted a bare `createDatabase`, which the host reads as a filled-in form; every field arrived undefined and a `db-N` directory was written whose `database.yaml` recorded a version of `"undefined"`. They open the form now, and the host refuses an incomplete create message outright rather than half-honouring it.
- **Opening the panel straight into the form did nothing on a cold open** — `beginCreate` was posted before the webview had a listener. It waits for the `ready` handshake now.
- **The Versions list was permanently empty on Windows without WSL**, which also made the Windows-client actions unreachable since they only exist on a rendered row.
- **A downloaded archive the catalogue does not list carried no file name** — that is every downloaded archive when the download site is unreachable — so Install resolved to the root directory and Remove silently did nothing.
- **The Extent backups folder button opened the log directory.**
- **A comment claimed the extent dropdown carried the user's choice into the replace flow.** It never did; the unused argument is gone and the comment now says what actually happens.

Also removed nine dead message cases and a dead field left by the trim, and swept the README and WSL guide, which still described the Versions and Processes sidebar views section by section.

Nineteen tests added, each mutation-checked by reverting its fix. Coverage also extended to the create path's success case — it writes a database directory, a `database.yaml` and a login, and only its refusals had been tested.

Still green: client 6984, server 322, mcp 92. The codicon webfont-vs-inline-SVG question in the description is unchanged and still yours to call.

client/src/extension.ts
```diff
@@ -317,13 +318,6 @@ export async function handleClassDefinitionCompiled(event: ClassDefinitionCompil
  * otherwise, with several sessions open, Execute It would target whichever
  * session happened to already be active.
  */
```
MatiasFernandez: 🚫 This JSDoc belonged to `openWorkspaceForSession`, which the PR deletes — it's now a floating block describing a function that no longer exists, right above the Getting Started constants.

It's not the only one; the lift out of #454 seems to have left a few docs detached from what they document:

- `client/src/databaseManager.ts` — the `stillRunningMessage` doc block ("Why a destructive operation was refused…") now sits above the new `extentBackupDir`, with a second doc block under it.
- `client/src/manager/databasesPanel.ts` — `DatabasesPanel.show` carries two stacked doc blocks (the `preserveFocus` one and the `startCreating` one); `listFiles` has one saying "sorted by name" directly above one saying "newest first", which is what the code actually does; `bringUp` is documented as "Run one of the allowed session commands against a live session"; `buildLoginTargets` is documented as "The selected session, without touching the GCI…"; and there's a dangling `// Parsed system.conf descriptions, one map per version…` comment with no field under it.

Could you sweep these? A wrong doc costs more than a missing one — a reader (or an agent) trusts `listFiles`'s first line and gets the opposite of the truth.

client/src/extension.ts
```diff
@@ -3665,11 +3679,10 @@ export function activate(context: vscode.ExtensionContext) {
         const wslInfo = await getWslInfoAsync();
         vscode.commands.executeCommand('setContext', 'gemstone.wslAvailable', wslInfo.available);
       }
-      // eslint-disable-next-line @typescript-eslint/no-floating-promises -- FIXME: unhandled floating promise; needs investigation to decide await vs. void vs. .catch before this rule is enabled repo-wide
-      versionProvider.loadVersions();
+      refreshVersions();
     }),
 
-    vscode.commands.registerCommand('gemstone.downloadVersion', async (item: VersionItem) => {
+    vscode.commands.registerCommand('gemstone.downloadVersion', async (item: VersionTarget) => {
```
MatiasFernandez: 💭 The PR description's motivation for the move is that the Versions row actions "typed into the Command Palette threw, because each handler expects a tree row and got nothing". That's still true after this change: `gemstone.downloadVersion`, `extractVersion`, `deleteDownload`, `deleteExtracted`, `unregisterLocalVersion`, `openVersionFolder`, `openVersionTerminal`, `downloadWindowsClient`, `openWindowsClientFolder` and `deleteWindowsClientExtracted` are all still in `contributes.commands` with no `commandPalette` gate, so they're still listed and still hit `item.version` on `undefined`.

The row that used to invoke them is gone, so the palette is now the *only* way to reach them — which makes the throw easier to hit than before, not harder. Should they get `"when": "false"` entries in `menus.commandPalette` now that the panel is the way in?

Same question, smaller stakes, for `gemstone.createDatabase`: the palette now offers both "GemStone Admin: Create Database" (the four Quick Picks) and "GemStone: New Database…" (the form). I understand why the command stays registered — quickSetup and the walkthrough call it — but does it need to stay *visible*?

client/src/manager/databasesPanel.ts
```diff
@@ -0,0 +1,1907 @@
...
+    if (!v) return;
+    await vscode.commands.executeCommand(command, { version: v });
+    await this.postState();
+  }
+
+  /**
+   * Run a version command on the release with this number. They take a tree item
+   * and read only its `version`, so that is what the panel hands them.
+   */
+  private async versionCommand(command: string, version: string): Promise<void> {
```
MatiasFernandez: 💭 `versionCommand` and `runVersionCommand` (just above) look identical to me — find the release in `lastVersions` by number, `executeCommand(command, { version: target })`, `postState()`. The message switch picks between them seemingly at random: `extractVersion`/`deleteDownload`/`unregisterLocalVersion`/`openVersionFolder` go through one, `openVersionTerminal`/`installWindowsClient`/`openWindowsClientFolder`/`deleteWindowsClient` through the other.

Is there a difference I'm missing? If not, could we keep one? Two names for the same thing invites a future change to land in only one of them, and the next reader will spend a while looking for the distinction that justifies the split.

client/src/manager/databasesPanel.ts
```diff
@@ -0,0 +1,1907 @@
...
+      rootPath: risk ? risk.rootPath : this.deps.storage.getRootPath(),
+    };
+  }
+
+  private async openDbSubfolder(dirName: string, folder: string): Promise<void> {
+    const db = this.lastDatabases.find((d) => d.dirName === dirName);
+    if (!db) return;
+    // Whitelisted rather than joined blindly: `folder` arrives from the webview.
+    const ALLOWED = ['conf', 'backups', 'backups/extents', 'log'];
+    const sub = ALLOWED.includes(folder) ? folder : 'log';
```
MatiasFernandez: 💭 When `folder` isn't on the allow-list this opens the `log` directory instead — and the test pins that (`['../../etc', 'log']`). Since the only sender is our own view, an off-list value means the view and the host have drifted, and silently opening a *different* folder makes that look like the click worked.

Would `if (!ALLOWED.includes(folder)) return;` be better here? It fails closed, and a button that does nothing is a much louder bug report than one that opens the wrong thing. The comment above says "whitelisted rather than joined blindly", which reads like refusal is what was intended.

client/src/manager/databasesPanel.ts
```diff
@@ -0,0 +1,1907 @@
...
+  // ── State ─────────────────────────────────────────────────────────────────
+
+  /**
+   * Render twice: first from what is already on disk, then again once the
+   * download catalog answers. Only the version list differs between the two, so
+   * the second post reuses the first state rather than re-running the OS,
+   * process and database scans — and an unreachable catalog costs the panel
+   * nothing but its available-versions rows.
+   */
+  private async postState(): Promise<void> {
```
MatiasFernandez: 💭 `rebuild()` goes to some trouble to serialise itself (`rebuilding`/`rebuildAgain`) so "two rebuilds [don't] post their states out of order" — but `postState()`, which nearly every message handler calls, has no such guard and posts twice with an `await` in between.

Two actions in quick succession (say Start Stone then Log in) give two overlapping `postState()`s; if the first one's remote pass resolves last, it posts a state built before the second action's effects. The panel would then sit showing stale rows until something else triggers a redraw.

The window is small once the catalog is cached, but is it worth routing `postState` through the same guard, or giving it a monotonic sequence number so a late state is dropped?

client/src/wslFs.ts
```diff
@@ -256,3 +257,88 @@ export function wslChmodSync(p: string, mode: number | string): void {
...
+/**
+ * The plain files directly inside `p`, sorted, with directories left out.
+ *
+ * Callers that want this list have to have it in one round trip: a readdir
+ * followed by a wslIsFile per entry costs one `wsl.exe` spawn per file, which
+ * a directory of stone logs turns into seconds of a frozen extension host.
+ * `ls -A1p` marks directories with a trailing `/`, and readdir's Dirent answers
+ * the same question natively, so neither side needs a stat per name.
+ */
+export function wslListFilesSync(p: string): string[] {
```
MatiasFernandez: 💭 Is `wslListFilesSync` used anywhere? I only see it referenced from `wslStatFilesSync`'s doc comment — `databasesPanel.listFiles` goes through `wslStatFilesSync`. If it was a stepping stone to the stat version, should it come out? Otherwise it's a WSL-only code path with no caller and no test to keep it honest.

client/src/manager/databasesView.js
```diff
@@ -0,0 +1,1378 @@
...
+          .map((l) => {
+            const open = l.sessions || [];
+            const head = `<div class="db-line db-login">
+                <span class="db-line-name"><span class="session-mark"></span><span class="db-login-user">${esc(l.user)}</span></span>
+                <span class="db-line-actions">${btn('editLogin', 'Edit login', 'edit', null, {
+                  login: l.label,
+                  iconOnly: true,
+                })}${btn('connectLogin', 'Log in', null, 'btn-secondary', {
+                  login: l.label,
+                  title: `Log in to ${esc(db.stoneName)} as ${esc(l.user)}`,
```
MatiasFernandez: nit: `btn()` already escapes `a.title` (`title="${esc(a.title || label)}"`), so pre-escaping inside the template double-escapes it — a stone named `A&B` shows as `A&amp;B` in the tooltip. Same at line 508 (`Log out session ${esc(...)}`) and 615 (`Remove the stale lock file for ${esc(proc.name)}`). Dropping the inner `esc()` at those three sites is enough; the surrounding calls that pass raw strings are already right.

client/src/manager/__tests__/localVersion.test.ts
```diff
@@ -161,108 +160,87 @@ describe('VersionManager.deleteExtracted', () => {
 
 // ── VersionTreeProvider (local version display) ─────────────
 
-describe('VersionItem (local version)', () => {
```
MatiasFernandez: 💭 I see there is code like this that now may be just tied to the new view, which is organized under the "manager" folder. Could we do a pass to see if more files/code can be moved within that folder now that we are replacing the sidebar versions section with the new webview?

MatiasFernandez (approved):
Approved with some comments worth reviewing

ericwinger:
## UX review, and what came of it

@MatiasFernandez sent this UX pass from using the panel, separately from the code review. Recording it here with the outcome, since it arrived outside the PR and the fixes are now on the branch (`afc4eefe`).

> - When clicking on the + button on the sidebar, the "Databases & Versions" view opens, but there's a noticeable delay in changing the view to the "New Database" one. I would guess there's a way to make that faster, and I'm not sure if it could be even slower on other users' machines (I have a very recent and powerful MacBook).
> - play & stop buttons within a database within the web view are not colored.
> - Hovering over the icons next to the "Extent" dropdown takes ages to show the tooltip. I guess that's because we are using the native HTML tooltips. This could be a follow-up, but probably tweaking that UX would be helpful for new users..
> - Clicking the + icon on a fresh extension install or clicking the New Database... button in the "databases" sidebar before installing any gemstone version opens the "new database" webview but in a broken state that is not easy to solve from there. You will need to cancel the database creation to see the main "Databases & Versions" view, where you can download a version and try again.

All four are fixed. Your machine was not the slow one — the delay was real, and worse on a slower network than on a slower CPU.

**The delay before the form.** `beginCreate` was posted *after* the first state, and building that state ends by asking the download site for its catalogue. So a whole network round trip sat between the click and the form, with the lists on screen for the duration — which is why it scaled with the network rather than the machine, and why an offline or throttled connection made it much worse than what you saw. It is posted before the state now, and the view draws the form with the first state that arrives. The lists never render at all.

**Uncoloured play and stop.** The database's own power control was already red-stops/green-starts; the Stone and NetLDI rows directly beneath it were plain glyphs. They take the same colouring now.

**Slow tooltips.** Native `title`, exactly as you guessed — the browser holds it back for about a second and a half. Not deferred: every explanation in the panel is now `data-tip` and the panel draws its own bubble after 150 ms, immediately for a reader who arrived by keyboard. It is a fixed element on the document body, because the `overflow` rule that gives each section its rounded corners would otherwise clip it.

**The broken form on a fresh install.** The form is built from installed releases, so with none there was an empty release list, an empty extent list, a Create that could never be pressed — and, because the panel had been opened only for the form, a Cancel that closed the whole window rather than revealing the panel behind it. The machine that most needed the Versions list was the one that could not reach it. Asked for the form with nothing to copy, the panel now shows the lists, with Versions leading them as it already does in that state, above a line saying why the form did not open.

**One more, found by hand.** That line first carried an **Install Version…** button of its own. Manually validating these fixes in a clean-slate window is what caught it: three **Install Version…** on a single screen, because the Versions section header and its empty state each already had a pair. The line is words only now, and the header gives its pair up to the empty state below it, where the sentence explaining them is. One of each. Worth noting that the doubled header/empty-state pair predates this branch's UX work — it only became obvious once a third one was sitting above it.

Twenty-three tests added, each checked by restoring the old behaviour and watching it fail. The tooltip change also removed the last code referencing the pinned-info popover, which turned out to have been unreachable for some time — nothing renders the attribute that opened it — so its stylesheet block went too: twenty-three rules for a Configuration section this panel does not have, left behind when that section was trimmed out of the lift. Say the word if you would rather review that separately and have it put back.

Still green: client 6999, server 322, mcp 92.

Your last point — missing the panel's whole-database play/stop in the sidebar — is a gap in the sidebar rather than a fix to this panel, so it is filed as [#535](https://github.com/GemTalk/Jasper/issues/535) rather than growing this branch. Nothing was tracking it before that.

client/src/extension.ts
ericwinger: Swept, and you were right that the first line is the dangerous part.

- The Workspace-per-session block is gone along with the function it documented.
- `stillRunningMessage`'s doc is back on `stillRunningMessage`; `extentBackupDir` keeps its own.
- `DatabasesPanel.show` has one block now, with both paragraphs in it.
- `listFiles` no longer opens by claiming the opposite of what it does.
- `bringUp` says it starts whichever servers are not up. The session-command sentence it was carrying belonged to `sessionCommand`, which had no doc at all, so it moved there.
- `buildLoginTargets` describes the login rows it builds.
- The dangling `system.conf` comment is gone — and so is the line in the `refresh` handler saying it drops that cache, since the cache went in the trim too.

Two more the sweep turned up: `listBackups` and the `logFiles`/`confFiles` field both cited "the sidebar tree", whose ordering no longer matches the panel's.

client/src/extension.ts
ericwinger: Good catch, and worse than a gate would fix — so they are out of the palette entirely.

Removing the `contributes.commands` entry is what actually takes a command out of the palette; the `registerCommand` call keeps it callable by name, which is how the panel drives it. So this removes ten declaration blocks instead of adding ten `"when": "false"` stanzas, and the diff shrinks. I checked each one first: no menu entries, no keybindings, no walkthrough or welcome-view links pointed at them. I left a comment above the handlers saying why they are deliberately undeclared, so nobody helpfully adds them back.

On `createDatabase` — it did not need to stay at all. `quickSetup` calls `databaseManager.createDatabaseDirect` directly, and the walkthrough and the empty-view welcome now link `gemstone.newDatabase` and `gemstone.quickSetup`. The comment above the handler claiming otherwise was stale. Nothing reached the command, so nothing reached the 172-line Quick Pick method behind it either, and both are deleted.

One knock-on your comment led me to: the walkthrough still told users to run **Create Database** from the **Versions** view. Both are gone, so it now describes the panel.

client/src/manager/databasesPanel.ts
ericwinger: Nothing you are missing — they were byte-identical. `runVersionCommand` is gone and its four call sites now use `versionCommand`, which is the one that carried the doc comment.

client/src/manager/databasesPanel.ts
ericwinger: Agreed, and the comment was indeed describing the intent rather than the code. It is `if (!ALLOWED.includes(folder)) return;` now.

The test that pinned the old behaviour is split in two: the four allowed folders still assert the path they open, and a new case asserts that an off-list folder opens nothing at all. Mutation-checked by restoring the `log` fallback — the new case fails.

client/src/manager/databasesPanel.ts
ericwinger: Worth it, and done — a sequence number rather than routing through `rebuilding`, since the two want different things: `rebuild` needs to coalesce and run one more pass, while a superseded `postState` just needs dropping.

Every pass stamps `stateSeq` on the way in and checks it is still the current one before each post, so an overtaken pass is discarded instead of repainting. `rebuild` shares the same counter — it also builds asynchronously and posts at the end, so a late pass from either side can be the stale one, and guarding only `postState` would leave half the race open.

Neither Eric nor I could reproduce the stale rows by hand, even with the catalogue unreachable to widen the window, so the fix rests on the test rather than on a sighting. It holds two passes in flight on held catalog answers and finishes them in the losing order; reverting the guard fails it.

client/src/wslFs.ts
ericwinger: It was a stepping stone, and it is out. `wslStatFilesSync`'s doc comment referred to it by name, so that sentence is rewritten rather than left pointing at nothing.

client/src/manager/databasesView.js
ericwinger: Right on all three — `btn()` escapes the explanation it is handed, so pre-escaping showed `A&amp;B`. The inner `esc()` is gone at all three sites.

Two tests cover it now, one for a stone and login name and one for a process name, both using `A&B`. Restoring the double escape fails both.

client/src/manager/__tests__/localVersion.test.ts
ericwinger: Done in this branch. Seven files moved to `client/src/manager/`, with their tests — including this one:

- `versionManager.ts`, `versionMatch.ts`
- `databaseManager.ts`, `extentBackupManager.ts`
- `processManager.ts`
- `databaseTreeProvider.ts`, `processTreeProvider.ts`

Import paths were rewritten by resolving each specifier against the file's old location and recomputing it from the new one, so only paths that had to change did. Sixteen other files updated their imports; git tracks all fourteen as renames.

The pass turned up one thing worth more than a move. `ProcessTreeProvider` was no longer bound to any view — `gemstoneProcesses` left `contributes.views` in this branch and nothing registers it — but it was not dead either: behind its event it re-ran `gslist` and re-probed the WSL address, which does not survive a restart, firing a second time when that probe lands so the panel picks up the new address without a manual Refresh. So it is now a `ProcessWatcher` with those two jobs and the event, and no tree. `ProcessItem` went with it; it had shrunk to a type annotation on two handlers, which take a `{ process }` target now — what the panel was already sending them.

That drops the client total by 10: fifteen tests asserting the tooltip, description and `contextValue` of a row nothing renders, replaced by five on what the watcher does.

Eric is deciding whether he wants you to look again, since this landed after your approval.

ericwinger:
## Review fixes

All eight review comments are addressed on `ccb8332f`; each is answered in its own thread with what changed and why.

**In short**

- **Docs.** Seven detached or contradicting blocks put back on what they describe, plus two the sweep turned up (`listBackups` and the `logFiles` field both cited a sidebar ordering that is no longer the panel's).
- **Palette.** The ten version commands lose their `contributes.commands` entries, which takes them out of the palette while leaving them callable by name — so this removes ten blocks rather than adding ten `when: false` stanzas. The four-Quick-Pick **Create Database** goes entirely: nothing called it, and nothing called the 172-line method behind it either.
- **Duplicates and dead weight.** `runVersionCommand` and `versionCommand` were the same function twice; `wslListFilesSync` had no caller.
- **Failing closed.** An unrecognised folder now opens nothing instead of quietly opening the log directory.
- **Ordering.** State passes carry a sequence number, so a pass overtaken by a newer one is dropped rather than repainting with stale rows. Shared with `rebuild`, since a late pass from either side can be the stale one.
- **Escaping.** Three tooltips escaped text the button helper escapes again — a stone named `A&B` introduced itself as `A&amp;B`.

**Also in this round**

The manager code moves in beside the panel: the version, database and process managers and the two tree providers now live under `client/src/manager/` with their tests. `ProcessTreeProvider` became a `ProcessWatcher` on the way — its view left the sidebar in this branch, but the `gslist` re-read and the WSL address re-probe behind its event are still needed, so those stay and the tree goes.

**Testing.** Three new tests, each mutation-checked by reverting its fix. The client total moves 7002 → 6992: fifteen tests asserting the appearance of a tree row nothing renders any more, replaced by five on what the watcher does. `main` is merged in. CI green on all legs.

One test needed a second pass — it drove two refreshes to get two state passes in flight, but Refresh re-probes WSL first and only on Windows, where that shells out to `wsl.exe` and takes seconds, so the test was measuring the probe. The panel tests hold `needsWsl` to false now, so every platform walks the same path.

PR #533, opened 2026-09-01
Migrate gciLogin's session-only tests to the integration harness

MatiasFernandez:
Phase 9c of the GCI test migration: moves the six `gciLogin.test.ts` tests whose subject isn't the login call itself onto `useIntegrationTest`, so they run in CI instead of only on demand.

- New `gciLibrary/__tests__/gciSession.integration.test.ts` covers `GciTsSessionIsRemote`, `GciTsNbLogout`, `GciTsAbort`, `GciTsBegin`, `GciTsContinueWith`. Abort/Begin assert the rollback of an uncommitted change rather than `transactionLevel`, since GemStone's default `#autoBegin` mode keeps the level pinned at 1 across both calls.
- `gciLogin.test.ts` trimmed from 17 to 11 tests (only the raw-login-subject tests remain).

### Known trade-off: the `GciTsNbLogout` test prints a warning

`withTransientSession` always logs its transient session out in the `finally`. The `GciTsNbLogout` test logs that session out itself, so teardown's logout lands on an already-dead session and prints one line to stderr on every run:

```
GciTsLogout failed [4100]: argument is not a valid GciSession pointer
```

The test passes and nothing is broken — the line just reads like a failure when scanning CI output.

An earlier revision of this branch silenced it by skipping teardown's logout when `GciTsSessionIsRemote` answered `-1`. That guard was dropped, because it keys on *"the session is invalid"* rather than *"the callback logged out"*, and so it also swallowed the warning for a session that died for a real reason — a stone restart or a dropped connection mid-test, which is precisely the case the warning exists to surface. It also left a stale `_utfCache` entry behind: that cache is purged inside `GciTsLogout`, which the guard skipped.

So the redundant logout is the cheaper end of the trade: one noisy-but-green stderr line, in exchange for keeping both the diagnostic and the cache cleanup. If the noise becomes a problem, the fix is to have the callback *declare* that it logged out, rather than to infer it from session validity.


client/src/gciLibrary/__tests__/gciSession.integration.test.ts
```diff
@@ -0,0 +1,91 @@
...
+      expect(gci.isIncludedInUserGlobals(session, key)).toBe(false);
+    });
+  });
+
+  describe('GciTsContinueWith', () => {
+    it('reports an error when there is no suspended process to resume', () => {
+      const { result, err } = gci.GciTsContinueWith(session, OOP_NIL, OOP_ILLEGAL, null, 0);
+
+      expect(result).toBe(OOP_ILLEGAL);
+      expect(err.number).not.toBe(0);
```
ericwinger: 🧪 This assertion passes for *any* error — a broken koffi signature, a wrong arity, an invalid session — so the test would stay green straight through a regression in the binding it exists to cover. It's also the only loose assertion in the file; every other test here pins an exact value.

Separately, the name may not describe the case being tested. Per `gcits.hf` the signature is `GciTsContinueWith(sess, gsProcess, replaceTopOfStack, ...)`, so the `OOP_NIL` on line 85 is an invalid **gsProcess** rather than "no suspended process to resume" — the error GemStone returns is "nil isn't a GsProcess", not "nothing is suspended". (`OOP_ILLEGAL` as `replaceTopOfStack` is correct and benign — the header documents it as "TopOfStack will not be changed".)

Suggest asserting the specific GemStone error number, and renaming to the invalid-`gsProcess` case.

client/src/gciLibrary/__tests__/gciSession.integration.test.ts
```diff
@@ -0,0 +1,91 @@
...
+    });
+  });
+
+  // GemStone's default transactionMode is #autoBegin, under which an abort
+  // re-enters a transaction immediately -- System transactionLevel stays
+  // pinned at 1 across both calls (verified live against a 3.6.2 stone:
+  // 1 -> abort -> 1 -> begin -> 1). The rollback each performs is the only
+  // observable effect that is stable across GemStone releases.
+  describe('GciTsAbort / GciTsBegin', () => {
+    it('abort discards uncommitted changes and opens a fresh transaction', () => {
```
ericwinger: 🧪 This name and line 73's both promise "and opens a fresh transaction", but each test asserts only the rollback — nothing checks that a transaction is open afterward. If `GciTsBegin` ever degraded to abort-without-begin, both tests would stay green while their names claim otherwise.

Given the comment just above explains why `transactionLevel` deliberately isn't asserted, trimming the phrase from both names is probably the honest fix. Asserting that a write still succeeds after the call would be the other way.

client/src/gciLibrary/__tests__/gciSession.integration.test.ts
```diff
@@ -0,0 +1,91 @@
...
+      // so 1 (RPC) is the only value an active session here can answer.
+      expect(gci.GciTsSessionIsRemote(session)).toBe(1);
+    });
+
+    it('reports a logged-out session as invalid', () => {
+      const loggedOutSession = logout();
+
+      expect(gci.GciTsSessionIsRemote(loggedOutSession)).toBe(-1);
+
+      login();
```
ericwinger: 🧪 If the assertion on line 41 fails, this never runs — and the harness's recovery then logs a `console.warn` scolding the test for not logging back in, which lands on top of the real assertion failure and reads like a second, unrelated problem.

Free to avoid by capturing, re-logging in, and then asserting:

```ts
const loggedOutSession = logout();
login();

expect(gci.GciTsSessionIsRemote(loggedOutSession)).toBe(-1);
```

That also matches the pattern in the harness's own self-test (`useIntegrationTest.integration.test.ts:73-74`), which does `logout(); login();` back-to-back with the assertions after.

ericwinger (approved):
Couple comments added. Seemed reasonable to mention. Approved.

client/src/gciLibrary/__tests__/gciSession.integration.test.ts
MatiasFernandez: Good catch on both counts — fixed.

You're right about the signature: `OOP_NIL` was going in as `gsProcess`, not as "nothing suspended". I probed a live stone to get the actual failure: `result` comes back `OOP_ILLEGAL` and `err.number` is **2092**, message `The Process nil to continue from is invalid,'argument is not a GsProcess'`. That's `RT_ERR_NO_PROCESS_TO_CONTINUE` in `gcierr.ht`, and it's the same value in every vendored header from 3.6.2 through 3.7.5, so pinning it is safe across the CI matrix.

The test now pins that number via a named local constant (matching `GS_ERR_UNCONTINUABLE` in `debuggerPanel.ts` and the exact-`2249` assertion in the harness test) and is renamed to the invalid-`gsProcess` case:

```ts
const RT_ERR_NO_PROCESS_TO_CONTINUE = 2092;

it('rejects a gsProcess argument that is not a GsProcess', () => {
  // OOP_ILLEGAL as replaceTopOfStack is the documented "leave TopOfStack
  // unchanged" value, so the only thing wrong with this call is gsProcess.
  const { result, err } = gci.GciTsContinueWith(session, OOP_NIL, OOP_ILLEGAL, null, 0);

  expect(result).toBe(OOP_ILLEGAL);
  expect(err.number).toBe(RT_ERR_NO_PROCESS_TO_CONTINUE);
});
```

One aside for the record: GemStone's own symbol for 2092 is `RT_ERR_NO_PROCESS_TO_CONTINUE`, so the old name wasn't pulled out of thin air — but the message it actually returns is about the argument not being a `GsProcess`, which is what the test exercises, so the new name is the honest one.

client/src/gciLibrary/__tests__/gciSession.integration.test.ts
MatiasFernandez: Good catch — trimmed the phrase from both names; they now read `abort discards uncommitted changes` and `begin discards uncommitted changes`.

Worth noting the write-succeeds alternative wouldn't have caught the degradation either: under the default `#autoBegin` mode an abort re-enters a transaction automatically, so a follow-up write succeeds whether or not `GciTsBegin` actually begins anything. Between that and `transactionLevel` staying pinned at 1 (the reason for the comment above), "opens a fresh transaction" isn't observable from this harness at all — so the rollback is the only guarantee the names can honestly claim.

client/src/gciLibrary/__tests__/gciSession.integration.test.ts
MatiasFernandez: Tried this one against a live stone and it doesn't hold up: `GciTsSessionIsRemote` takes a `GciSessionPtr` (`gciLibrary.ts:422`) — a raw pointer, not a session id. `logout()` frees the struct and the immediately following `login()` reuses that block, so the stale handle resolves to the fresh session and answers `1`. The reordered version fails with `Expected -1 / Received 1`. The self-test escapes this because its assertions are about the new session, not the dead handle.

Capturing the *value* before the re-login (`const isRemote = ...; login(); expect(isRemote).toBe(-1);`) does work, but I'd rather leave it as-is: the extra warning only shows up when the test is already failing, and it isn't a false alarm — a logout really did happen without a matching login, which is exactly what the harness is reporting. Happy to revisit if it turns out to be a real annoyance in practice.

PR #537, opened 2026-09-02
Fix five small Explorer, search, and SUnit bugs (#532)

### LINKED ISSUE #532: Miscellaneous small bugs (catch-all)
Umbrella issue for small, unrelated bugs found in passing that don't each warrant their own ticket. Each item gets a checkbox above and a short section below. Anything that grows teeth gets spun out into its own issue.

- [x] **GemStone Search: Source-scope results come back in no recognizable order**
- [x] **Step point numbers: hover help doesn't say how to turn the view off**
- [x] **Workbench tabs: badges, and "GemStone"-prefixed names**
- [x] **Explorer: drag a class onto a class category to move it**
- [x] **A new method should land in the selected method category**
- [x] **Run Test on a newly added test method says "No tests found"**

---

## GemStone Search: Source-scope results come back in no recognizable order

**Environment** — Jasper 1.8.14; GemStone server: N/A (client-side ordering).

**Steps to reproduce**
1. Open the GemStone Search panel, scope to **Source**.
2. Search for a common identifier, e.g. `anIndex`.

**Expected** — a recognizable order (class A→Z, or shortest-label-first — anything stable).

**Actual** — arbitrary order. Observed: `Array>>at:`, `Bag>>at:`, `Array>>at:put:`, `Array>>_basicAt:put:`, `AbstractDictionary>>_at:`, `Array>>_at:`, … Effectively the stone's own `ClassOrganizer>>substringSearch:ignoreCase:` traversal order, passed through untouched.

**Cause** — two things compound:

1. Source hits are stamped `score: 0` (`client/src/omniSearch/references.ts:57`), because the query matched the method *body* on the server and the matcher never runs against the label. So in `omniRank` (`client/src/omniSearch/omniEngine.ts:218`) every ranking rule ties — no selector prefixes the query, all rows are method-like, all scores are equal — and ordering falls through to `compareMatches`' tiebreaks (shorter label, then A→Z).
2. Those tiebreaks contradict the same-selector-by-class rule (`omniEngine.ts:229-239`), making the comparator non-transitive. A cycle from the list above: `AppendableString>>at:put:` before `Array>>at:put:` (same selector, class A→Z), `Array>>at:put:` before `Array>>_basicAt:put:` (shorter), `Array>>_basicAt:put:` before `AppendableString>>at:put:` (shorter). With a comparator that cycles, `Array.prototype.sort` may return anything — here it hands back the input order nearly unchanged. Drop the duplicate-selector rows from the same set and the remainder sorts cleanly shortest-first, confirming it.

**Potential fix** — likely needs both halves; either alone leaves it broken. One direction: give Source (and Literals — same `methodRowsToResults` path) rows a real order of their own, e.g. class name then selector, since there is no label match to score. And potentially make `omniRank` transitive by folding the class-name comparison into one consistent sort key rather than a conditional override that only fires for same-selector pairs.

---

## Step point numbers: hover help doesn't say how to turn the view off

**Environment** — Jasper 1.8.14; GemStone server: N/A (client-side UI).

**Steps to reproduce**
1. Show step point numbers in a method (e.g. start debugging with `gemstone.stepPoints.display` at its `debugging` default, or turn them on explicitly).
2. Hover a step point number and look for a way to hide the numbers.

**Expected** — the hover tells you how to turn the view off, ideally as a clickable action right there.

**Actual** — the tooltip says only "Step point **N** of M — click to toggle a breakpoint here" (`client/src/stepPointHints.ts:116-118`). The way off does exist — **Toggle Step Point Numbers** in the editor right-click menu — but it sits far down that menu (group `3_gemstoneBreakpoints@4`), which is easy to miss entirely. The other routes are the same command from the palette and the `gemstone.stepPoints.display` setting; there is no keybinding.

**Potential fix** — one direction: extend the hint tooltip with a command link, e.g. a "Hide step point numbers" line invoking `gemstone.breakpoints.toggleStepPoints`. That needs `isTrusted` set on the tooltip's `MarkdownString` for a `command:` link to fire. A plain click on the number is already the breakpoint toggle, so the hide action probably has to live in the hover (or behind a modifier-click) rather than take over that gesture.

---

## Workbench tabs: badges, and "GemStone"-prefixed names

**Environment** — Jasper 1.8.14; GemStone server: N/A (client-side UI).

**Wanted** — badges on the GemStone tabs (a count where one is meaningful, e.g. search results), and names that say whose they are: "Test Results" → "GemStone Test Results", and so on.

**Notes** — `WebviewView.badge` / `TreeView.badge` already exist, so badges are wiring, not new API. Renaming reaches our own contributed containers and views (`contributes.viewsContainers` / `contributes.views` in `package.json`) and the SUnit controller label (`client/src/sunitTestController.ts:225`); VS Code's built-in tabs (Testing, Test Results, Problems, Output) can't be renamed by an extension, so those need a different answer.

---

## Explorer: drag a class onto a class category to move it

**Environment** — Jasper 1.8.14; GemStone server: N/A (client-side UI).

**Wanted** — drag a row from the Classes pane onto a row in the Class Categories pane and have the class land in that category.

**Notes** — classes are deliberately non-draggable today (`client/src/gemstoneExplorer.ts:5734-5739` — `dragMimeTypes` empty, `handleDrag` a no-op), and the Categories tree has no drop controller at all. The move itself already exists as `client/src/queries/recategorizeClass.ts`, so this is mostly wiring: a class MIME type on the Classes pane plus a `TreeDragAndDropController` on Class Categories, mirroring how methods already drag onto classes.

---

## A new method should land in the selected method category

**Environment** — Jasper 1.8.14; GemStone server: N/A (client-side UI).

**Wanted** — with a method category selected, a newly saved method files into that category, not into `as yet unclassified`.

**Notes** — partly true already: **New Method** from a category row uses that row's category, and from the palette it uses the Methods-pane selection when the side matches (`client/src/gemstoneExplorer.ts:4994-5023`). The `+` on the instance/class side node always passes `'as yet unclassified'` (`gemstoneExplorer.ts:5027-5033`), and the palette path falls back to the default whenever the selected category belongs to the other side. Worth a pass over every new-method entry point so they all read the same selection.

---

## Run Test on a newly added test method says "No tests found"

**Environment** — Jasper 1.8.14; GemStone server 3.6.2.

**Steps to reproduce**
1. In a `TestCase` subclass whose tests have already been listed once, add another test method.
2. Press ▷ **Run Test** on the new method's row in the Explorer's Methods pane.

**Expected** — the test runs.

**Actual** — a warning toast: "No tests found".

**Cause** — confirmed: **Refresh SUnit Tests** makes the new method runnable (manually verified by Eric), which places it in the method list, not in discovery of the class or in `testSelectors`.

`ensureClassItem` lists a class's test methods only when it has none yet: `if (classItem.children.size === 0) await this.resolveTestMethods(classItem)` (`client/src/sunitTestController.ts:570-572`). A class listed before the new method was written keeps its stale children, `itemForMethodNamed` finds no match, and `runTestItems` reports "No tests found" over an empty list (`sunitTestController.ts:1229-1231`).

Refreshing works around it, but writing a test and running it is the ordinary loop — it shouldn't need a manual refresh.

**Potential fix** — one direction: re-list a class's methods when running by name rather than only when the list is empty, and have the empty-list message say which class and selector it looked for.

ericwinger:
Closes #532 — five of the six items in that catch-all issue. Each is independent of the others.

### GemStone Search: Source-scope results came back in no recognizable order

Those hits match a method *body*, so they carry no label match to rank on and every ranking rule tied. The tiebreak meant to catch them — same-selector rows sort by class A→Z — contradicted the shorter-label rule beside it, which made the comparator **non-transitive**, and a cyclic comparator lets `Array.prototype.sort` return anything (here, the stone's traversal order nearly untouched).

Method-like rows now order by one key rather than a rule that fires for some pairs and not others: score, then class A→Z, then instance side before class side, then selector. That is the same key the Methods provider already caps its own page with, so a row does not shift when the engine re-ranks it. Method-like and name-like categories never share a rank bucket, which is what keeps the whole comparator total.

Review caught that the provider only *nearly* shared that key — it sorted score → class → **label**, and `'Array class>>at:'.localeCompare('Array>>at:')` is -1, so a label tiebreak puts the class side first. At the cap boundary that dropped an instance-side row while keeping its class-side sibling, and the engine then showed the instance side first. `compareMethodRows` now lives in `rank.ts` and **both** callers use it, so the rows the provider keeps and the order they are displayed in are one key.

### Step point numbers: hover didn't say how to turn the view off

The tooltip said only which step point it was and that clicking toggles a breakpoint. The way out exists — **Toggle Step Point Numbers** — but sits far down the editor's right-click menu with no keybinding. The hover now carries a **Hide step point numbers** link. `isTrusted` is scoped to that single command rather than granted wholesale.

### Workbench tabs: names (no badges)

Views dragged out of the GemStone containers showed as a bare "Classes" or "Methods" beside everyone else's tabs. Each contributed view now declares a `contextualTitle` VS Code uses there, and the Rowan pane contributed into the file Explorer — which has no container heading of its own to name it — is **GemStone Rowan** outright. A manifest test pins this so the next view copy-pasted from a neighbour cannot ship without one.

**The badges half of that item is deliberately not done**, so the checkbox stays open. A badge is set from the data the view last rendered, and VS Code only asks a tree for children while it is showing — so the count is right exactly when the rows are visible anyway, and stale the rest of the time. VS Code's own Testing, Test Results, Problems and Output tabs belong to the workbench and cannot be renamed by an extension.

### Explorer: drag classes onto a class category to move them

A class row — or a whole selection of them, since the Classes pane now allows a multi-row selection — drags from the Classes pane onto any Class Categories row, including a sub-category (it files into the full dashed path). Uncommitted, like every other Explorer write.

Multi-select was review's suggestion, and it made one existing label wrong. The Remove Class confirmation button said **Remove All**, meaning "this class *and its subclass subtree*, all-or-none" — but beside a dialog titled after one class, with several rows selected, it reads as "remove every row I selected". It now says **Remove With Subclass(es)**. The inline trash is a row control (VS Code hands it its own row and nothing else), so a selection-wide remove was never what it could have meant.

An editor open on that class's definition **re-reads**: the category is a line of definition source, so a stale buffer would have shown the old one and, on save, filed the class straight back — the definition save path applies its category line via `recategorizeClass`. The file system provider gained `notifyChanged(uri)` for stone-side changes nothing else announces.

### A new method lands in the selected method category

**New Instance Method** and **New Class Method** always passed `as yet unclassified` whatever the Methods pane had selected. All three entry points now read the same selection through one helper, still falling back when the selected category is on the other side — a category is per-side.

### Run Test on a newly added test method said "No tests found"

A class's test methods were listed once and never re-read, so a method written afterwards was missing: running it by name reported "No tests found", and running its whole class **silently skipped it**, until someone happened to hit Refresh SUnit Tests. Running by name now re-reads the class's methods first — one `discoverTestMethods` query per run, next to running the tests themselves, nothing. The empty-list message names the class and selector it looked for.

## Testing

`npm run lint`, `format:check` and `compile` clean. Client 7302 passed / 12 skipped, server 322, mcp-server 92, against a live 3.6.2 stone with the server plugin installed.

Each fix was mutation-checked — the fix reverted, the tests confirmed failing, the fix restored. The review fixes too: the two new provider-ordering tests fail against the old score → class → label sort, the drag test's category seeding fails both current-category tests when seeded empty, and the confirm-button label is pinned by a test that also asserts it says no "All". Two tests that first passed against the broken code were rewritten: the "leaves open definitions alone on failure" case (the whole block is skipped when nothing moved, so the per-class filter needed a *partial* failure to exercise) and a badge test that used a second provider, which starts with nothing to clear.

One pre-existing flake is worth knowing about: the in-stone refactoring SUnit suites (`refactoringClass`, `refactoringMethod`, `refactoringInlineMethod`) time out at vitest's 5s default under the full suite's parallel load. A different subset fails each run, each passes in isolation, and `origin/main` fails them the same way — I verified that against the same stone in a pristine worktree. Not from this branch.


client/src/omniSearch/omniEngine.ts
```diff
@@ -226,17 +251,18 @@ function omniRank(a: OmniResult, b: OmniResult, q: string, upperFirst: boolean):
...
-  if (
-    a.action.kind === 'openMethod' &&
-    b.action.kind === 'openMethod' &&
-    a.action.selector === b.action.selector
-  ) {
-    const byClass = a.action.className.localeCompare(b.action.className);
-    if (byClass !== 0) return byClass;
-  }
+  // Within the method-like bucket, order by the METHOD, not by the label text: equal score, then
+  // class A→Z, then instance side before class side, then selector. This is the same key the
```
MatiasFernandez: 🚫 "This is the same key the Methods provider caps its own page with, so a row's place doesn't shift when the engine re-ranks it" — I don't think that holds any more. `methodsProvider`'s own sort is score → class → **label**; it has no instance-vs-class-side step. And because the label is `Class class>>sel` for the meta side, the label tiebreak puts the class side *first*:

```
> 'Array class>>at:'.localeCompare('Array>>at:')
-1
```

So for `Array>>at:` vs `Array class>>at:` the provider orders class-side-then-instance-side and `compareMethodRows` orders instance-side-then-class-side — which is exactly what the new "puts a class-side method after the instance-side one" test pins. That's not just cosmetic: the provider does `out.slice(0, cfg.maxResultsPerCategory)`, so *which* rows survive the cap is decided by a different order than the one they're displayed in. At the cap boundary an instance-side row can be dropped while its class-side sibling is kept, and then the engine claims to have shown the instance side first.

Should `methodsProvider` grow the same side step (and reuse `methodSortKey`, if it can be exported), so the two really are one key? Otherwise the comment's claim needs softening to "nearly the same key".

client/src/sunitTestController.ts
```diff
@@ -508,7 +508,7 @@ export class SunitTestController implements vscode.Disposable {
     await this.discoverTests();
     const classItems: TestItem[] = this.itemsForClasses(dictName, classNames);
```
MatiasFernandez: 💭 `runClassesByName` is the one named entry point that doesn't go through `ensureClassItem` — it calls `itemsForClasses` and hands the class items straight to `runTestItems`. So `runClassTests` re-resolves only via its own `if (classItem.children.size === 0)` guard, which is the same stale-list check the PR just removed from `ensureClassItem`.

As far as I can tell that leaves the reported symptom alive on this path: write `testBrandNew`, select two test classes in the Explorer and Run Tests. The stone builds the suite itself so the method *does* execute, but `runClassTests` maps results back with `classItem.children.forEach(...)` and a result whose selector has no child item is simply never read — no row, no verdict, silently absent from the run. Same story for ▶ on a class row in the Testing view that was expanded before the method was written.

Is the multi-class path deliberately left out (one `discoverTestMethods` per class is a real cost where the single-class path pays one), or worth an `ensureClassItem` loop / dropping the `children.size === 0` guard in `runClassTests` too? If it's deliberate, the doc comment above `ensureClassItem` might be the place to say which runs re-read and which don't — it currently reads as though only the Testing view's lazy path is exempt.

client/src/omniSearch/DESIGN.md
```diff
@@ -287,6 +287,14 @@ Behaviour decisions (Eric's review of the first webview cut):
 - **Flat, globally relevance-ranked results — no category grouping.** Typing "foo" should surface the
   closest "foo" first regardless of kind, so `buildView` ranks every result together by match score.
   Each row wears a small **category tag** (Class / Method / Global / …) so you still see what it is.
+- **Ties break by kind.** Below the prefix and first-letter-case rules, method rows (Methods / Source /
```
MatiasFernandez: 💭 "Below the prefix and first-letter-case rules, method rows … order by class A→Z, then instance side before class side, then selector" — `compareMethodRows` actually compares the matcher score *first*, and only then class/side/selector. For Source and Literals that's invisible (everything is 0, which is the whole point), but for Methods rows the score is what decides, and the class A→Z order only ever breaks a tie below it. Worth naming score in the key here? The CHANGELOG entry omits it the same way, and it's the one part a reader would be surprised by — the new "still leads with the better match when scores differ" test exists precisely because score wins.

client/src/gemstoneExplorer.ts
```diff
@@ -6167,10 +6339,14 @@ export function registerGemStoneExplorer(
   });
   const categoryView = vscode.window.createTreeView('gemstoneExplorerCategories', {
     treeDataProvider: ctl.categoryProvider,
+    dragAndDropController: new CategoryDropController(ctl),
   });
   const classView = vscode.window.createTreeView('gemstoneExplorerClasses', {
     treeDataProvider: ctl.classProvider,
-    dragAndDropController: new ClassDropController(ctl),
+    // No canSelectMany here: every class action (remove, rename, comment…) is written
```
MatiasFernandez: 💭 With no `canSelectMany` on the Classes view, a drag can only ever carry one row — which makes the plural branch of the confirmation (`Moved N classes to class category …`), the per-payload failure accumulation, and the "reports every class it moved when several travel together" / "moves the classes it can when one of them fails" tests all cover a shape a user can't currently produce. I like the code being written for the general case, but is it worth flipping `canSelectMany: true` here so the multi-class behaviour is actually reachable (refiling a handful of classes into a new category seems like the common case for this feature)? The comment explains why it's off for the *other* class actions, but drag is the one action that reads naturally over a selection.

client/src/gemstoneExplorer.ts
```diff
@@ -5425,6 +5454,91 @@ export class ExplorerController {
...
+      }
+      this.categoryProvider.refresh();
+      this.classProvider.refresh();
+      this.syncTitles();
+      // A class definition carries its category as a line of source, so an editor open
+      // on one of these classes is now showing a category the class no longer has —
+      // and saving that buffer would file the class straight back (the definition save
+      // path applies its category line via recategorizeClass). Tell the file system the
+      // documents changed so any open tab re-reads.
+      for (const p of toMove.filter((m) => moved.includes(m.className))) {
```
MatiasFernandez: nit: `toMove.filter((m) => moved.includes(m.className))` re-derives payloads from names that came out of those same payloads a few lines up. Would pushing the payload rather than the name — `const moved: ClassDragPayload[] = []` … `moved.push(p)` — let this be a plain `for (const p of moved)`? The success message then reads `moved[0].className` / `moved.length`, and the `includes` scan goes away.

client/src/__tests__/explorer/explorerClassCategoryDrag.test.ts
```diff
@@ -0,0 +1,448 @@
...
+    undefined,
+    notifyDocumentChanged,
+  );
+  ctl.setViews(makeViews() as unknown as Parameters<ExplorerController['setViews']>[0]);
+  ctl.state.dictName = 'UserGlobals';
+  ctl.state.dictIndex = 3;
+  // The classes pane's own view of which class sits in which category.
+  (
+    ctl as unknown as { classCategoryEntriesStore: ClassCategoryEntry[] }
+  ).classCategoryEntriesStore = entries;
```
MatiasFernandez: 🧪  This reaches past the accessor into the private backing field through an `as unknown as` cast, so TypeScript can't tell you if `classCategoryEntriesStore` is ever renamed — and the failure mode is quiet: every `categoryOfClass` lookup returns `undefined`, so "files the dragged class into the dropped-on category", "files into the full dashed path" and the failure/notify tests all still pass, and only "does nothing when the class is already in that category" goes red. That's one test standing between a rename and a suite that no longer exercises the current-category read.

Could the setup go through whatever normally populates it instead — the mocked `getClassesWithCategory` already returns `[]` here, so having `makeController` seed `vi.mocked(queries.getClassesWithCategory).mockReturnValue(entries)` and then triggering the same refresh the pane does would keep the cast out of it. If there's no seam for that today, would a narrow test-only setter be worth it?

client/src/sunitTestController.ts
```diff
@@ -597,8 +609,11 @@ export class SunitTestController implements vscode.Disposable {
     return `${className} is not a test class.`;
   }
 
-  public noTestsFoundErrorMessage() {
-    return `No tests found`;
+  /** "No tests found" on its own leaves the user guessing which name was looked up — the very thing
+   *  they need to see when a just-written test won't run. `looked` names what was searched for;
+   *  without it (a run started from the Testing view's own selection) the bare message stands. */
+  public noTestsFoundErrorMessage(looked?: string) {
```
MatiasFernandez: nit: `looked` reads as a verb where it's being used as a noun — `noTestsFoundErrorMessage(looked)` / `looked ? ...`. `lookedFor` or `searchedFor` would say what it holds, and matches how the message reads (`No tests found for ${lookedFor}`).

MatiasFernandez (approved):

client/src/omniSearch/omniEngine.ts
ericwinger: You were right, and it was not just cosmetic. `methodSortKey` and `compareMethodRows` moved to `rank.ts` and the provider now sorts with the same comparator rather than a hand-rolled mirror of it, so the cap and the display are decided by one key.

Two tests pin it (`providers.test.ts`): the instance side leads for the same selector, and — the case you named — with `maxResultsPerCategory: 1` the surviving row is the instance side, not its class-side sibling. Both fail against the old score → class → label sort.

client/src/omniSearch/DESIGN.md
ericwinger: Named. `DESIGN.md` and the changelog entry now lead with match score and say why the class/side/selector steps are all you see in Source and Literals (every score there is 0). Both also note that the key is shared with the Methods provider.

client/src/gemstoneExplorer.ts
ericwinger: Flipped to `canSelectMany: true` — refiling a handful of classes is the ordinary shape of the gesture, and the single-row actions are unaffected since VS Code passes the clicked row as the command argument. A test reads the pane options out of the source so the multi-class paths cannot go unreachable again (activation is not run in unit tests).

Eric verified the multi-class drag by hand, and it turned up a label the flip made wrong: the Remove Class confirmation said **Remove All**, meaning "this class and its subclass subtree, all-or-none" — but beside a dialog titled after one class, with several rows selected, it reads as "remove every row I selected". It now says **Remove With Subclass(es)**. The six existing subtree tests answer with whatever confirm button the dialog offered rather than a hard-coded string, so the next rename cannot quietly turn them into passing "cancelled" runs.

client/src/__tests__/explorer/explorerClassCategoryDrag.test.ts
ericwinger: Cast gone. `makeController` seeds through `getClassesWithCategory` and goes in via `selectDict`, the same path a dictionary click takes — which also exercises the accessor rather than bypassing it. Two tests now guard the current-category read instead of one, and seeding an empty list makes both fail, so the read is genuinely covered.

Issue #498, opened 2026-08-26
Global Go Back / Go Forward across Jasper navigation landings


ericwinger:
## The problem

I lose methods. I am reading a method, something moves me somewhere else — I click a
different method in the Methods pane, I follow Implementors of / Senders of, Omni Search
lands me somewhere, a debugger frame or a test failure jumps the browser, a refactoring
repositions the selection — and now the method I was reading five landings ago is gone.
The only way back is a global search, and only if I still remember the selector.

Every other VS Code language experience answers this with **Go Back** / **Go Forward**.
Jasper should too, over its *own* navigation landings, not just over text editors.

## What I want

Two commands with the native VS Code names and behavior:

| Command | Title | VS Code equivalent |
|---|---|---|
| `gemstone.navigateBack` | **Go Back** | `workbench.action.navigateBack` (`Ctrl+Alt+-`) |
| `gemstone.navigateForward` | **Go Forward** | `workbench.action.navigateForward` (`Ctrl+Shift+-`) |

- A **chain**, not a single step: Back repeatedly walks the whole history, one landing per press.
- Forward re-walks what Back undid, and is discarded when a new landing happens after going
  back — standard browser/VS Code semantics, no branching.
- Reachable from the Command Palette, from a keybinding, and from `$(arrow-left)` /
  `$(arrow-right)` buttons in the System Browser title bar so they are visible, not just known.
- Greyed out / no-op at the ends of the chain.

## Preferred approach: push onto VS Code's own history, don't invent a parallel one

The reason Back/Forward feels broken today is almost certainly that a selection change
*inside* the browser webview never registers as a navigation location with VS Code — the
native Back/Forward buttons in the title bar only see the editor documents Jasper opens.
Someone hitting the native buttons gets inconsistent results depending on whether the
landing happened to open or pin an editor. Two sessions of first-use feedback already
circled this (see #469, "Navigation: confusing, then resolved").

So the first thing to try is: make every Jasper landing register as a real VS Code
navigation location, so the *native* Go Back/Go Forward — the ones already in the title
bar with the keybindings the user already knows — just work. A Jasper-private stack is the
fallback for landings that genuinely have no editor/document to anchor to, and if we build
one it should be keyed the same and feel identical.

## Landings that must be recorded

Anything that changes what the user is looking at, manual or automatic:

- Methods pane selection; class pane selection; category selection (`selectMethod`,
  `selectMethodCategory` in `client/src/systemBrowser.ts`)
[...]

PR #538, opened 2026-09-02
Go Back / Go Forward over the Explorer's navigation landings

ericwinger:
Addresses https://github.com/GemTalk/Jasper/issues/498.

**This is up for review as an approach, not just an implementation.** It deliberately
departs from what the issue proposed, and the departure is the thing worth arguing
about — see "The call I'd like reviewed" below.

## What it does

**Go Back** and **Go Forward** walk the places the GemStone Explorer has taken you —
dictionary, class category, class, and method — one press per landing, with the forward
trail dropped when you go somewhere new. Browser semantics, which are also VS Code's.

A landing is a **coordinate** (session, dictionary, class, side, selector), not a handle.
It is recomputed against the stone on the way back, so a method that has since been
recompiled still resolves; the dictionary is re-resolved by *name*, because a commit
elsewhere can shift every symbol-list index. A landing that no longer resolves is reported
and dropped from the trail rather than erroring, so a second press moves on instead of
sticking. The last 50 are kept.

The trail is listed in a new **Actions & Navigation** pane at the top of the Explorer,
newest first, with the landing you are on marked; click any row to jump to it. The same
list is available from the keyboard as **GemStone: Recent Locations…**.

## The call I'd like reviewed

The issue's preferred approach was to *push Jasper's landings onto VS Code's own
navigation history*, so the native Go Back/Forward in the title bar would just work, and
to treat a Jasper-private stack as the fallback. **This does the opposite: it builds a
private chain.**

The reason is that VS Code's history is a history of *editor documents*. Most Explorer
landings are pane moves that open no document at all — selecting a dictionary, a class
category, or a class — and the ones that do open a document open it in a single reusable
preview tab, which VS Code's history does not track as distinct locations. Registering
landings with the native history would have meant opening a real editor for every pane
move, which is a bigger behavioural change than the feature is worth.

The cost of the private chain is a second set of Back/Forward controls next to the
native ones, which is a genuine wart. If the reviewer prefers the native route, the
recording points are all funnelled through one `recordLanding` call, so the destination
is swappable without touching the call sites.

Two smaller calls, both mine, both worth challenging:

- **The buttons are in a new pane, not the System Browser title bar** the issue asked for.
  Refresh/Commit/Abort moved onto the same row, since they act on the Explorer and its
  session rather than the dictionary list they used to sit above. A pane's title-bar
  actions are drawn only while the pane is expanded *and* hovered or focused, so they
  vanish exactly when the pane is small enough to be worth keeping; these do not.
- **A drill-down is one landing, not three.** Clicking a dictionary, then a class, then a
  method records a single entry, because the coarser two are the cascade passing through
  rather than places the user asked to be. Without this, leaving a method costs three
  presses of Back.

## What is *not* recorded yet

The issue lists landings this does not cover, and it should not be read as complete:

- **Debugger frame selection**, and the jump the debugger makes when a session stops
- **Go to Definition from a peek pane** (a plain Go to Definition is covered)
- **Test-failure navigation** from the SUnit runner

Everything that opens a `gemstone://` editor *is* covered, because that activation drives
the Explorer's own sync — so Implementors of, Senders of, GemStone Search, Find Method in
Class, the method-results picker, plain Go to Definition, and post-refactoring
repositioning all record landings without needing their own call sites. The three above
do not go through that path. Happy to add them here or in a follow-up, whichever the
reviewer prefers.

The issue's "with several System Browsers open, Back should act on the one that produced
the landing" is moot: the System Browser is frozen, so several open browsers is not a case
to design for. One global chain, owned by the Explorer.

## Testing

Client-only diff, so the fast gate rather than the version matrix.

- `lint`, `format:check`, `compile` clean; client 448 files / 7197 tests, server 322,
  mcp-server 92 — all passing.
- The chain's rules are unit-tested against hand-built landings; a separate suite drives
  the **real pane cascade** and asserts the landings it actually produces, plus the way
  back (dictionary re-resolution, an unselected session declined rather than reconnected,
  a vanished dictionary, a removed method), the Recent Locations picker, and that the
  three commands have real handlers rather than only manifest entries.
- Those tests were mutation-checked rather than trusted for being green: ten mutations
  (removing each folding rule, the restore guard, the stale-entry prune, the dictionary
  re-resolution, the session guard, the method-gone report, the trail ordering, and the
  command registrations) were each verified to fail the suite.
- Manually F5-verified throughout by Eric; the design was driven by his live findings.


client/src/gemstoneExplorer.ts
```diff
@@ -4449,6 +4482,241 @@ export class ExplorerController {
...
+    if (!revealMethod) return true;
+
+    const info = this.selectorsFor(revealMethod.isMeta, ALL_METHODS_CATEGORY).find(
+      (i) => i.selector === revealMethod.selector,
+    );
+    if (!info) {
+      void vscode.window.showWarningMessage(
+        `${landing.className} no longer implements ${revealMethod.isMeta ? 'class method ' : ''}${revealMethod.selector}.`,
+      );
+      return false;
```
MatiasFernandez: 💭 When the class is still there but the method has gone, we've deliberately moved the panes to the class — the doc above says that beats refusing to move at all, and I agree. But returning false makes `step` prune this entry *and* rewind the cursor to where it was, so the panes end up on `Account` while the pinned line and the trail's current marker still name the method we came *from*, and Forward now points at somewhere we never left.

Should this return true instead, so the cursor follows where the panes actually landed? The class is a real place in the chain, and the warning already tells the user the method itself is gone. Worth noting the `still moves to the class when only the method is gone` test asserts the prune but not the pinned line, so it wouldn't have caught the disagreement either way.

client/src/explorerNavigationHistory.ts
```diff
@@ -0,0 +1,433 @@
...
+ *
+ * - **One chain per session.** A landing only means anything against the stone it
+ *   was read from, so each session gets its own chain and the pane shows the
+ *   selected session's. A session logging out takes its chain with it.
+ */
+export interface ExplorerLanding {
+  sessionId: number;
+  dictName: string;
+  /** 1-based symbolList position at record time; re-resolved by name on the way back. */
+  dictIndex: number;
```
MatiasFernandez: 💭 `dictIndex` is recorded on every landing, but I can't find anything that reads it back: `goToLanding` re-resolves the index itself from `getDictionaryNames(...).indexOf(landing.dictName)`, `landingKey` leaves it out on purpose, and none of the label helpers touch it. Given this field's own doc says it's re-resolved by name on the way back, is it worth dropping it from the coordinate (or at least marking it optional) so a future reader doesn't take the recorded index for something Back relies on?

client/src/__tests__/explorer/explorerNavigationView.test.ts
```diff
@@ -0,0 +1,390 @@
...
+    expect(html).toMatch(/\.location \{[^}]*display: none/);
+    expect(html).toContain('.location.shown { display: block; }');
+  });
+
+  it('gives the trail its own scrolling region under the fixed button row', () => {
+    const html = renderNavigationViewHtml('test-nonce');
+    expect(html).toContain('id="trail"');
+    expect(html).toMatch(/\.trail \{[^}]*overflow-y: auto/);
+    expect(html).toMatch(/\.toolbar \{[^}]*flex: 0 0 auto/);
+  });
```
MatiasFernandez: 💭 This handful asserts the literal source text of the template — `where.textContent = location`, `.location.shown { display: block; }`, `.toolbar { … flex: 0 0 auto` — so a Prettier reflow of the inline CSS or renaming the local `where` breaks them with no behaviour changing, and they'd still pass if `drawLocation` were never wired up at all.

The two just below earn their place as regression guards, because the contract *is* the text: the CSP nonce, and `not.toMatch(/\.innerHTML\s*=/)` standing between a selector out of the stone and an injection. For these, would rendering the html into jsdom and asserting the DOM — location line hidden until a state push names somewhere, one row per trail entry, the current row carrying the accent class — buy more than string matching? Or is the styling worth pinning this tightly deliberately, since a headless render is what caught the empty-toolbar bug?

package.json
```diff
@@ -2073,11 +2113,27 @@
...
+        {
+          "command": "gemstone.explorer.showHistory",
+          "when": "gemstone.hasActiveSession"
+        },
+        {
+          "command": "gemstone.explorer.clearHistory",
+          "when": "gemstone.hasActiveSession"
+        },
+        {
+          "command": "gemstone.explorer.showNavigationSelectorsOnly",
```
MatiasFernandez: 💭 Both halves of the toggle show in the Command Palette whenever a session is active, so whichever mode you're already in is offered as a visible no-op — the button row avoids exactly that with its `data-mode` gating. Since the mode is a real setting, the `when` can key off it directly:

```json
{
  "command": "gemstone.explorer.showNavigationSelectorsOnly",
  "when": "gemstone.hasActiveSession && !config.gemstone.explorer.navigationSelectorsOnly"
}
```

and the mirror of it for `showNavigationFullLocations`. Worth doing, or is having both always listed easier to find than a palette entry that comes and goes?

README.md
```diff
@@ -201,6 +201,8 @@ Beyond browsing, the Explorer is where the code-changing operations live:
 - Deleting first looks for what still references the target: nothing does, and it just goes (and says so); something does, and you are shown the methods before you decide
 - The refactorings — rename, extract/inline method and temporary, change signature, move/push up/push down method, instance-variable structure changes, extract superclass, split class — each previewed before it is applied
 - Browse senders, implementors, references, and the class hierarchy
+- Retrace your steps: **Go Back** / **Go Forward** (`Ctrl+Alt+-` / `Ctrl+Shift+-`) walk method by method through what you have been reading, reopening each tab; the dictionaries, classes and categories you passed through name the pinned line above the trail rather than taking steps of their own, and **Recent Locations…** lists everywhere you have been including them. A method is listed once, at its most recent visit, so returning to one moves its row instead of repeating it. The trail is per session — switching sessions switches it, logging out clears it, and **Clear Navigation History** empties it by hand. A press with nothing of Jasper's left that way falls through to VS Code's own Go Back / Go Forward
```
MatiasFernandez: nit: The mac bindings are `ctrl+-` / `ctrl+shift+-`, not `Ctrl+Alt+-` / `Ctrl+Shift+-` — and the same unqualified pair appears in the CHANGELOG entry. The README spells both out elsewhere, e.g. "**Cmd+S** (Ctrl+S)" a few paragraphs up. Worth matching that convention here so a mac reader doesn't try a chord that isn't bound?

client/src/__tests__/explorer/explorerNavigationHistory.test.ts
```diff
@@ -0,0 +1,628 @@
...
+
+    await history.back();
+    expect(history.entries().map((l) => landingLabel(l))).toEqual(['Array>>gone', 'Set>>add:']);
+    expect(history.currentIndex()).toBe(1);
+    expect(history.canGoBack()).toBe(true);
+  });
+
+  it('treats a throwing restore as unreachable rather than propagating', async () => {
+    history.record(method('Array', 'at:'));
+    history.record(method('Set', 'add:'));
```
MatiasFernandez: nit: These two records go into `history`, but every assertion below is about `boom` — looks like leftover setup. Dropping them would make it clearer that the throwing `go` is the whole point of the test.

client/src/__tests__/explorer/explorerNavigationLandings.test.ts
```diff
@@ -0,0 +1,698 @@
...
+      `${CLASS}>>balance`,
+      `${CLASS}>>deposit:`,
+    ]);
+    expect(pane.latest().trail.map((r) => r.context)).toEqual([DICT, DICT]);
+  });
+
+  it('drops the class to the dimmed column when the setting is on', async () => {
+    // Working inside one class, the repeated class name crowds out the selector,
+    // which is the only part that differs.
+    vscode.workspace.getConfiguration('gemstone').update('explorer.navigationSelectorsOnly', true);
```
MatiasFernandez: nit: This `update(...)` isn't awaited, unlike the `await ctl.setTrailLabelMode(...)` in the test just below. It works today only because the vscode mock's `update` mutates its store before hitting an await — worth an `await` here so this doesn't quietly start reading the old value if that mock ever grows a real async step.

MatiasFernandez (approved):

client/src/gemstoneExplorer.ts
ericwinger: Agreed, and taken a step further than `return true`.

`return true` would have kept the *method* entry: the pinned line and the accented row would still name `Account>>balance` while no such method exists and no editor opened — the panes and the chain would agree on a place that is a fiction.

So `go` can now answer with the coordinate it actually reached (`Promise<boolean | ExplorerLanding>`), and `step` rewrites the entry to it. The landing is demoted to its class, which is your "the class is a real place in the chain" made literal: the cursor follows the panes, the pinned line names the class, and — the class landing no longer being a method landing — it drops out of the trail rather than sitting there as a row pointing at nothing.

You were right that the test asserted the prune and nothing else. It now asserts the cursor, the rewritten entry, that the pinned line does not mention the selector, and that Forward is still live; a second test covers the trail having no row marked. Reverting the fix fails three tests. b2eb67b8

client/src/explorerNavigationHistory.ts
ericwinger: Dropped from the coordinate rather than made optional — nothing read it back, so optional would only have moved the question.

The one place it still matters is `recordLanding`, which keeps checking `state.dictIndex` because its absence means the panes are mid-cascade with no dictionary landed on yet. That reads oddly now that it is checked and not stored, so there is a comment saying so. The interface doc says the name is the whole coordinate and why. b2eb67b8

client/src/__tests__/explorer/explorerNavigationView.test.ts
ericwinger: Worth it — converted.

The pane's real chrome and its real inline script are mounted in a jsdom window, so these are behaviour now: the location line hidden until a state push names somewhere, one row per trail entry newest first, the accent on the current row, the empty-trail message, the chain index a click reports, and the arrows dimming at the ends. Five mutations (dropping the `reverse`, the accent class, the location hide, the empty message, the row index) each fail these and nothing else.

Two things I kept as text, deliberately:

- the CSP nonce and the `not.toMatch(/\.innerHTML\s*=/)` guard, for your reason — the contract *is* the text;
- the flex/overflow rules, because jsdom computes no styles, so it cannot see the empty-toolbar class of bug at all. Those are now whitespace-tolerant regexes so a Prettier reflow does not break them, with a comment saying why they stay pinned as CSS.

A jsdom *window* rather than the file-wide environment pragma the other webview tests use: this pane's script is inline in a module that imports `vscode`, and under that environment the bare specifier stops resolving and the whole file fails to load. b2eb67b8

package.json
ericwinger: Worth doing — applied as you wrote it, both halves.

Verified in a running window rather than just the manifest, since these are the first `config.` context keys in this extension and a wrong key fails silently rather than erroring: with the trail in full mode only **Show Only Selectors in the Navigation Trail** is offered, and after running it only **Show Full Locations in the Navigation Trail**. b2eb67b8

README.md
ericwinger: Right on both counts — fixed in the README and the CHANGELOG, following the `**Cmd+S** (Ctrl+S)` convention a few paragraphs up.

Back really is the asymmetric one: `ctrl+-` on mac against `ctrl+alt+-` elsewhere, while Forward is `ctrl+shift+-` on both. Both platforms are now spelled out rather than the unqualified pair. Not yet pressed on a mac — the manifest is all I have verified. b2eb67b8

client/src/__tests__/explorer/explorerNavigationHistory.test.ts
ericwinger: Leftover indeed — dropped. The throwing `go` is all the test is about now. b2eb67b8

client/src/__tests__/explorer/explorerNavigationLandings.test.ts
ericwinger: Awaited. b2eb67b8

ericwinger:
Review acted on, and the branch is up to date with `main`. Every inline thread has a reply and is resolved; this is the summary in one place.

## The call that was actually reviewed

The reviewer's substantive point was the one worth having: landing on a class whose **method has gone** returned `false`, which pruned the entry *and* rewound the cursor — so the panes sat on the class while the pinned line and the trail's marker still named the method we came from, and Forward pointed at a place we had never left.

Their suggestion was `return true`. That is better, but it keeps a *method* entry the stone no longer has: the pinned line and the accented row would name `Account>>balance` with no such method and no editor open. So `go` now answers with the coordinate it **actually reached** (`Promise<boolean | ExplorerLanding>`) and the chain rewrites the entry to it. The landing is demoted to its class — the cursor follows the panes, the pinned line names the class, and the entry stops being a method landing, so it leaves the trail rather than sitting there pointing at nothing.

## The rest

- **`dictIndex` is gone from the coordinate.** Nothing read it back; Back re-resolves the index from the name, because a commit elsewhere can shift every index. `recordLanding` still *checks* `state.dictIndex`, because its absence means the panes are mid-cascade — that now says so in a comment.
- **The palette offers only the half of the trail-label toggle that changes something**, keyed off the setting, matching what the button row already did with its own gating.
- **The pane's drawing is asserted against a real DOM.** Its chrome and inline script mount in a jsdom window, so the location line hiding until there is somewhere to name, one row per entry newest-first, the accent on the current row, the empty-trail message, the chain index a click reports, and the arrows dimming are behaviour now instead of matches against the template's source text. The CSP nonce and the `innerHTML` guard stay as text — there the contract *is* the text — and so do the flex/overflow rules, since jsdom computes no layout.
- **The mac keybindings are spelled out** in the README and CHANGELOG: Go Back is `ctrl+-` there, not `ctrl+alt+-`.
- Three test nits: leftover setup dropped, a settings write awaited, and the pane stand-in declared once instead of twice.

Two things the change made stale were swept with it: the module doc and the CHANGELOG both said a landing that has gone is "dropped from the trail", which is no longer true of the method-gone case, and two test comments referred to a recorded dictionary index that no longer exists.

## Verification

New tests were mutation-checked rather than trusted for being green: dropping the `reverse`, the accent class, the location hide, the empty-trail message, and the row index each fail these and nothing else; reverting the coarser-arrival fix fails three.

Manually verified in a running window — including that the palette really does offer one half at a time, since these are this extension's first `config.` context keys and a wrong key fails silently rather than erroring. Also re-checked the merge seams with `main`: Refresh/Commit/Abort on the pane, that logging in no longer drags the sidebar over, and that a method editor opened *by Back* still carries CodeLens, Refactor… and a breakpoint gutter after this cycle's narrowing of all three. The mac keybindings are the one item verified only against the manifest.

Full gate green against a live stone: client 472 files / 7810 tests, server 322, mcp-server 92.

PR #541, opened 2026-09-02
WIP: Extend optional functions usage verification

MatiasFernandez:


eslint.config.mjs
```diff
@@ -80,6 +86,45 @@ const TS_EXTENSION_IMPORT = {
...
+  // in `gci['GciTsNbPoll'](...)` the property is a `Literal`, which carries
+  // `.value` and no `.name`. Shaped as a call, not a bare member access -- a
+  // call is the thing that reaches the native library. The bare-identifier
+  // form is gated alongside them, as the test-session `login` selectors below
+  // are: `const { GciTsNbPoll } = gci` pulls the binding out of the library
+  // into a local, and calling *that* reaches the native library just the same.
+  return [
+    { selector: `CallExpression[callee.property.name='${name}']`, message },
+    { selector: `CallExpression[callee.property.value='${name}']`, message },
+    { selector: `CallExpression[callee.name='${name}']`, message },
```
ericwinger: 🚫 **The call-shape restriction has real holes, and unlike the deleted test this file no longer names them.**

`gciVersionGated.test.ts` carried this in its own doc-comment:

> Matched syntactically, on `.<name>(`, so an aliased or dynamically dispatched call slips through — the same blind spot a lint selector would have.

That sentence is gone, and the comment that replaces it claims the opposite direction only ("The AST selectors are stricter than the old `\.<name>\s*\(` scan"). Both are true, but a reader of this block now has no idea where the fence ends.

I checked against the real config, on a scratch file under `client/src/`. Caught: `gci.GciTsNbPoll(...)`, `gci['GciTsNbPoll'](...)`, `gci?.GciTsNbPoll?.(...)`, `gci!.GciTsNbPoll(...)`, `const { GciTsNbPoll } = gci; GciTsNbPoll(...)`. Not caught:

```ts
(gci.GciTsNbPoll as (s: unknown, t: number) => unknown)(session, 0);  // cast the function
gci.GciTsNbPoll.call(gci, session, 0);
const poll = gci.GciTsNbPoll.bind(gci); poll(session, 0);
const { GciTsNbPoll: poll } = gci; poll(session, 0);                   // renaming destructure
gci[`GciTsNbPoll`](session, 0);                                        // template-literal key
gci.GciTsNbPoll!(session, 0);   // only flagged today by no-unnecessary-type-assertion,
                                // which stops flagging it the moment the receiver is nullable
```

None of these is a regression — the old regex missed them too — so this isn't a blocker. But there is a cheap way to close all of them here, and I think it's worth taking.

The reason given for call-shape ("a call is the thing that reaches the native library") is inherited from the `RAW_LOGIN_NAMES` selectors below, where it's load-bearing because test files legitimately *name* a mocked binding (`expect(gci.GciTsLogin).not.toHaveBeenCalled()`). This block's `files`/`ignores` already exclude tests, mocks and `gciLibrary.ts` — so in the files it actually covers, that reason doesn't apply. A bare member access to one of these names in production code is a hazard no matter what shape the call around it takes.

Adding two selectors:

```js
{ selector: `MemberExpression[property.name='${name}']`, message },
{ selector: `MemberExpression[property.value='${name}']`, message },
```

catches the cast, `.call`, `.bind`, the renaming destructure and the template-literal key in one go. I ran it over `client/src server/src mcp-server/src` with the rest of the config intact: **zero** new findings — the only non-call mentions of registry names in production are the string literals in `enhancedInspector/enhancedInspectorPerfTracker.ts`, which are `Literal` nodes and don't match. So it's free today, and it's cheapest to take now rather than after someone writes the first `.bind`.

If you'd rather keep the selectors call-shaped, then please put the blind spot back into the comment — the deleted test documented it and the replacement shouldn't be quieter than what it replaced.

eslint.config.mjs
```diff
@@ -80,6 +86,45 @@ const TS_EXTENSION_IMPORT = {
...
+// internally would leave a cached file green after a new entry lands.
+//
+// Background: `docs/explanation/gci-version-compatibility.md`.
+const OPTIONAL_GCI_CALL = Object.entries(GCI_OPTIONAL_FUNCTIONS).flatMap(([name, reason]) => {
+  // One clause per axis the registry records, so a two-axis entry
+  // (`GciTsNbLogin_`) names both hazards rather than the first one found.
+  const hazard = [
+    reason.addedIn && `absent before ${reason.addedIn}`,
+    reason.absentOn === 'win32' &&
+      'absent from the Windows client library -- throws on every `windows-latest` cell and every Windows install',
```
ericwinger: 💭 **`absentOn` is compared against `'win32'` by hand, so widening the union degrades silently rather than failing.**

The type is `absentOn?: 'win32'` today, so this is safe right now. But the moment someone adds `'darwin'` — which is a one-word edit to `GciAbsenceReason` and compiles everywhere else — an entry carrying only `absentOn: 'darwin'` produces `hazard === ''`, and the message reads:

> `GciTsFoo may be absent from the loaded library (). Put the cross-version conditional inside client/src/gciLibrary/ ...`

Empty parens. The rule still fires, so nothing goes ungated — but the one thing this rewrite is *for* (the message naming the actual hazard, per axis) is what quietly stops working, and nothing in CI notices.

This registry is explicitly built so that the header generator refuses to write what it can't express, and so that a stale entry is a compile error. The hazard renderer should hold itself to the same standard. Something like:

```js
const clause = { win32: 'absent from the Windows client library -- throws on ...' };
...
reason.absentOn && (clause[reason.absentOn] ?? throwUnknownAxis(name, reason.absentOn)),
```

or, at minimum, assert `hazard !== ''` and throw with the entry name — a config that fails to load is a far better signal than an error message with a hole in it, and this is the only axis where the code has to restate a value the type already knows.

eslint.config.mjs
```diff
@@ -267,6 +312,18 @@ export default tseslint.config(
...
+    // The GCI optionality gate (see OPTIONAL_GCI_CALL above). Scoped to the
+    // three workspaces' production sources; tests and mocks are exempt because
+    // an absent-world test's whole job is to call the symbol and watch it
+    // throw. `gciLibrary.ts` is exempt because the bindings *are* its subject:
+    // it holds every `this._optional.GciTsX(...)` call there is. Once a
+    // cross-version helper lives under `client/src/gciLibrary/`, this exemption
+    // widens to that directory -- and belongs to the commit that puts one there.
+    files: ['client/src/**/*.ts', 'server/src/**/*.ts', 'mcp-server/src/**/*.ts'],
+    ignores: ['**/__tests__/**', '**/__mocks__/**', '**/gciLibrary.ts'],
+    rules: { 'no-restricted-syntax': ['error', ...OPTIONAL_GCI_CALL] },
```
ericwinger: 🚫 **A `*.test.ts` that isn't under `__tests__/` silently loses this gate — the deleted test covered that case.**

The old scan filtered by *directory* (`__tests__/`, `__mocks__/`, basename `gciLibrary.ts`), so a `client/src/foo.test.ts` was scanned. Here, the test-session block further down (`files: ['**/*.test.ts', '**/*.spec.ts', '**/*.test.tsx']`) also sets `no-restricted-syntax`, and in flat config a later entry **replaces** a rule's options rather than merging them. So for any such file the GCI selectors are dropped wholesale and only the login selectors apply.

Verified — a scratch `client/src/zzProbe.test.ts` containing `gci.GciTsNbPoll(s, 0)` lints completely clean under the PR's config.

There are zero such files in the repo today (`find client/src server/src mcp-server/src -name '*.test.ts' | grep -v __tests__` is empty), so this is latent, not live. But it's an invisible failure: the file looks covered by `files: ['client/src/**/*.ts']`, and nothing tells you the rule got replaced. Two options:

- add `'**/*.test.ts', '**/*.spec.ts', '**/*.test.tsx'` to this block's `ignores`, which makes the exemption explicit and matches what the comment already says the intent is ("tests and mocks are exempt"); or
- merge the two sets in the later block instead of restating one.

Either way the current pairing is load-bearing on file layout in a way neither comment mentions.

Minor, same block: `ignores: ['**/gciLibrary.ts']` is a repo-wide basename glob, so it would also exempt a future `server/src/gciLibrary.ts`. `'client/src/gciLibrary.ts'` says what's actually meant.

package.json
```diff
@@ -3498,9 +3498,9 @@
...
-    "lint:fix": "eslint . --fix --cache --cache-location node_modules/.cache/eslint/.eslintcache --cache-strategy content",
+    "//lint": "Results are cached under node_modules/.cache/eslint/, keyed on file content; run lint:cache:clear to force a full re-check. The NODE_OPTIONS prefix silences MODULE_TYPELESS_PACKAGE_JSON: eslint.config.mjs imports the GCI optionality registry as `.ts`, and client/package.json declares no module type. It is set through cross-env because a bare `VAR=value cmd` prefix is Unix shell syntax, and `npm run lint` has to run under cmd.exe on Windows too. The value re-expands $NODE_OPTIONS so a contributor's own flags (say --max-old-space-size) survive rather than being replaced; cross-env expands it itself on cmd.exe, where the shell will not. The double quotes are required -- cmd.exe does not treat single quotes as quoting, so the value would split in two.",
+    "lint": "cross-env \"NODE_OPTIONS=$NODE_OPTIONS --disable-warning=MODULE_TYPELESS_PACKAGE_JSON\" eslint . --cache --cache-location node_modules/.cache/eslint/.eslintcache --cache-strategy content",
```
ericwinger: 💭 **Is "the one cost" the right framing? What's being bought is the suppression of a cosmetic warning, and the bill is longer than one line.**

I checked what happens without any of this. On Node 24, `import ... from './client/src/gciLibrary/optionalFunctions.ts'` works, and the only consequence is four lines on stderr, once:

```
(node:442791) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of file:///.../optionalFunctions.ts is not specified and it doesn't parse as CommonJS.
Reparsing as ES module because module syntax was detected. This incurs a performance overhead.
```

`npm run lint` still exits 0. So the warning is noise, not a failure. Against that, the ledger is:

1. **A new dev dependency** (`cross-env`, pulling `@epic-web/invariant` and `cross-spawn`) in a repo that runs `lint:supply-chain`, `lint:lockfile`, `npm audit signatures` and a documented approval flow for new deps. Small, well-known package — but it's a supply-chain entry bought to silence stderr.
2. **Two shells' quoting semantics on the critical path**, and CI exercises neither. The `lint` job is `runs-on: ubuntu-latest` only, so the `cmd.exe` behaviour this indirection exists for is never run anywhere in CI. Previously the script was `eslint .` and had no shell surface at all.
3. **The `$NODE_OPTIONS` re-expansion doesn't hold on every Windows setup.** I read `cross-env@10.1.0`'s `varValueConvert` — it does resolve `$VAR` on all platforms, so under `cmd.exe` the comment is right. But npm's `script-shell` is configurable, and under PowerShell `$NODE_OPTIONS` is expanded *by the shell* to the empty string (PowerShell spells env vars `$env:NODE_OPTIONS`), so cross-env never sees it and the contributor's own flags are dropped — silently, which is exactly the outcome the comment says is avoided. Worth either narrowing the claim or using `%NODE_OPTIONS%`-safe handling.

None of this is wrong; I'd just like the PR to say what's actually being traded. The cheapest alternative is genuinely "print the warning" — and if that's unacceptable, it'd be good to see the one-line reason why in the script comment, since it's the option a future reader will reach for first when they're wondering why `lint` goes through a shim.

Also worth noting in the same breath: the long `//lint` doc-string now carries three paragraphs of shell trivia that have nothing to do with linting. If it stays, consider moving the cmd.exe/PowerShell explanation next to the `cross-env` devDependency or into the GCI doc, and leaving the script comment as a pointer.

lefthook.yml
```diff
@@ -6,7 +6,14 @@ pre-commit:
...
       glob: '*.{ts,mts,cts,js,mjs,cjs}'
-      run: npx eslint {staged_files}
+      # NODE_OPTIONS matches the `lint` script: the config imports the GCI
+      # optionality registry as `.ts`, and Node warns about the type-stripped
+      # import because client/package.json declares no module type. Set through
+      # `cross-env` because a bare `VAR=value cmd` prefix is Unix shell syntax,
+      # and this hook has to run under `cmd.exe` for a Windows contributor. The
+      # value re-expands $NODE_OPTIONS so a contributor's own flags survive; the
+      # double quotes are required, as cmd.exe does not quote with single ones.
+      run: npx cross-env "NODE_OPTIONS=$NODE_OPTIONS --disable-warning=MODULE_TYPELESS_PACKAGE_JSON" eslint {staged_files}
```
ericwinger: nit: `npx cross-env` in a pre-commit hook can reach the network.

Before: `npx eslint {staged_files}` — `eslint` is always in `node_modules/.bin` because it's a direct devDependency and nothing works without it.

After: `npx cross-env ...` — on a machine that has pulled this branch but not re-run `npm ci` (the common case: switch branches, edit a file, commit), `cross-env` isn't installed, and `npx` falls through to fetching it from the registry mid-commit. In a non-TTY hook that's either a surprise network round-trip or an outright failure, and the error it produces will point at `cross-env`, not at "run npm ci".

Not a big deal, but a pre-commit hook is the one place you least want an install step. Worth considering `node_modules/.bin/cross-env` (fails fast and legibly if deps are stale) or `npx --no cross-env` so it refuses to install rather than doing it silently.

docs/explanation/gci-version-compatibility.md
```diff
@@ -56,7 +56,7 @@ Both directions are compile errors: an entry with no binding, and an `optionalFu
...
-| Call a post-3.6.2 symbol from production code                              | [`gciVersionGated.test.ts`](../../client/src/__tests__/gciVersionGated.test.ts) names every call site, unless you add it to `ALLOWED_POST_362` — a reviewable diff | `npm test`, stone-free                            |
+| Call any registry symbol from production code                             | `eslint.config.mjs`'s `OPTIONAL_GCI_CALL` block fails and names the hazard the entry implies — absent before its `addedIn` floor, absent from the Windows client library, or removed in a server version that is supported by Jasper. Selectors are generated from the registry, so the gated set is never restated. The remedy is to put the conditional in `client/src/gciLibrary/` and call a helper from there; the fallback is an `eslint-disable-next-line` at the call site whose `-- reason` names the guard and the degraded path | `npm run lint`                                    |
```
ericwinger: 📝  Two things about this row, now that the gate lives in lint:

1. The command column says `npm run lint`, which is right — but it also means this row is now the **only** enforcement in the table that doesn't run under `npm test`, and §"The one gap that costs a developer time" further down (line 66) still describes a single gap, the stale generated file. There's now a second one with the same shape: `npm test` alone is green on an ungated call site too. That paragraph should name both, since it's the paragraph a developer reads to decide what to run.

2. Prettier doesn't format Markdown here (`format` covers `{ts,mts,cts,js,mjs,cjs}` only), so the table columns are now hand-aligned and this row breaks the alignment — the trailing `|` no longer lines up and the description cell is ~4x the width of any other. Renders fine on GitHub; just harder to edit. Consider trimming it to the outcome and the remedy and letting `.claude/rules/client/gci.md` carry the detail, which it already does almost verbatim.

eslint.config.mjs
ericwinger: The blocker really isn't a regression or blocker but probably should be fixed before merging.

ericwinger (approved):
I think these are legit if I read them right.  But some probably don't need to be done as they are nitty.

eslint.config.mjs
MatiasFernandez: Confirmed — reproduced your probe, and took option 1. Option 2 would over-apply: that block matches *every* test file including those under `__tests__/`, so merging would gate the absent-world tests whose job is to call an optional binding and watch it throw.

Two trims: no `'**/*.test.tsx'` (this block's `files` are all `*.ts`, so it could never match), and `'**/gciLibrary.ts'` narrowed to `'client/src/gciLibrary.ts'` — a future `server/src/gciLibrary.ts` should be gated like any other production source. The comment now says flat config replaces rather than merges, so it isn't left to block ordering.

eslint.config.mjs
MatiasFernandez: Good catch — `addedIn` and `removedIn` interpolate their values, so `absentOn` is the only axis restating something the type already knows. Took both guards: an `ABSENT_ON_CLAUSE` table whose lookup throws on an unknown value, plus a `hazard === ''` assert for a wholly new axis the renderer doesn't read at all. Both throw at config-load time, so it's a failed lint run rather than empty parens.

package.json
MatiasFernandez: Fair — "the one cost" was the wrong framing. Dropped the `$NODE_OPTIONS` re-expansion and the quotes, so items 2 and 3 go away: the value is a constant, nothing left for `cmd.exe` or PowerShell to get wrong, and no claim to narrow since no single spelling is right across all three shells. The cost is that a contributor's own `NODE_OPTIONS` is replaced rather than merged, for lint only.

The devDependency stays, and `//lint` now says the warning is cosmetic and lint exits 0 without it, rather than calling it one line. Shell trivia trimmed out of both it and the lefthook hook.

docs/explanation/gci-version-compatibility.md
MatiasFernandez: Both fixed. Row trimmed to outcome + remedy with a pointer to `.claude/rules/client/gci.md` (which already carries the detail verbatim), and padded back to the table's 76/164/51 so the pipes line up.

The section is now "The two gaps that cost a developer time" — registry staleness and the ungated call site, plus a note that the pre-commit hook is opt-in. I stopped short of calling the lint row the *only* thing `npm test` misses: the compile-error rows technically qualify, they just surface in the editor instead.

lefthook.yml
MatiasFernandez: Good catch — fixed, on both jobs: `prettier` is a direct devDependency too, so it had the same hole. Verified with deps uninstalled that `npx --no` exits 1 naming the missing package and installs nothing.

Went with `--no` over `node_modules/.bin/cross-env` for portability — on Windows the bare path is the sh shim, not the `.cmd` entry point, so it only works if lefthook hands the line to a POSIX shell.

PR #542, opened 2026-09-02
Start and stop a whole database from the Databases sidebar

### LINKED ISSUE #535: Start and stop a whole database from the Databases sidebar, as the Databases & Versions panel already does
## Summary

The **Databases & Versions** panel puts one Start/Stop button on a database's summary row: it
brings that database's Stone and NetLDI up or down together, in a single click. The
**Databases** view in the sidebar has no equivalent — the only start/stop there is per server,
on the Stone and NetLDI child rows, so bringing a database up means expanding it and clicking
two separate rows.

The sidebar is where most people watch their databases day to day, and it is now the one place
that has lost a control the panel offers.

## Acceptance criteria

- A database row in the **Databases** sidebar view offers a single Start (when the stone is
  down) or Stop (when it is up) action, and that action brings both the Stone and the NetLDI
  with it — the same behaviour as the panel's power button.
- The action follows the row's state the way the panel's does, so a running database offers
  Stop and a stopped one offers Start, rather than both being present at once.
- The existing per-server Start/Stop on the Stone and NetLDI child rows stay as they are; this
  adds the whole-database control, it does not replace the individual ones.
- A database Jasper cannot control — servers started outside its environment — is not offered
  the action, matching how the panel already withholds its power button in that case.

## Implementation notes

Nothing in the extension registers a whole-database start or stop today. It exists only as a
webview message inside the panel: `startDatabase` / `stopDatabase` in
`client/src/manager/databasesPanel.ts`, handled by a private `startStopDatabase(dirName, start)`
that calls the panel's own `bringUp` / `takeDown`. The individual `gemstone.startStone`,
`gemstone.stopStone`, `gemstone.startNetldi` and `gemstone.stopNetldi` commands are registered
and are what the sidebar's child rows use.

So one direction could be to lift that pair out of the panel into registered
`gemstone.startDatabase` / `gemstone.stopDatabase` commands and have both surfaces call them —
which would also make the whole-database action reachable from the Command Palette, where it
is absent today. The sidebar rows would then need `view/item/context` entries in
`package.json` gated on the database row's context value (currently a single `gemstoneDb`,
which would probably need to distinguish running from stopped the way the Stone row's
`gemstoneDbStoneRunning` / `gemstoneDbStoneStopped` already do).

Raised in review of [#527](https://github.com/GemTalk/Jasper/pull/527), which is what added the
panel's power button.

ericwinger:
Closes #535.

## The feature

The **Databases & Versions** panel puts one Start/Stop on a database's summary row: it brings that
database's Stone and NetLDI up or down in a single click. The **Databases** sidebar — where most
people watch their databases day to day — had no equivalent, so bringing a database up meant
expanding the row and clicking the Stone and the NetLDI separately.

A database row now offers that same single action:

```
GEMSTONE  ▸ Databases
  ▸ db-1   gs64stone_375 (3.7.5)   ▶  📁  >_  +      ← Start / Stop, leftmost on the row
      Stone: gs64stone_375   Stopped        🗘  ▶    ← unchanged
      NetLDI: gs64ldi        Stopped            ▶    ← unchanged
```

- Follows the row's state — a running database offers Stop, a stopped one Start, never both.
- The per-server buttons on the Stone and NetLDI rows are untouched; this adds to them.
- Withheld from a database whose servers were started outside Jasper's environment.
- Also reachable as **GemStone Admin: Start Database** / **Stop Database** in the Command Palette,
  where it asks which database. It was absent there entirely before.

The whole-database start/stop existed only as a webview message inside the panel. It is now a
registered `gemstone.startDatabase` / `gemstone.stopDatabase`, so the panel, the sidebar and the
palette share one implementation instead of the sidebar growing a second one.

**One correction to the issue.** It says the panel already withholds its power button for servers
started outside Jasper. It does not — only the *per-server* toggles were withheld; `powerControl`
rendered unconditionally. Rather than have the two surfaces disagree, the panel now withholds it
too. That is the one change here outside the issue's literal scope.

## What testing it turned up

The rest of this PR is a bug the feature made easy to reach but did not cause. It is included at
@ericwinger's direction rather than deferred, because the new button leads straight into it.

`stopstone` refuses while any session is logged in. Jasper offered that refusal to the user as
*"Enter the DataCurator password, or force-stop the stone"* — which reads as a credentials problem
when the password was never wrong, so Force Stop looked like the only way through. Force Stop
SIGKILLs the stone; its gems survive it, wedged in GemStone's coredump wait, holding `extent0.dbf`
open. The next start then fails with *"File is open by another process"*, naming nothing, and the
database cannot be brought back without finding and killing those gems by hand. Starting point:
be logged in while stopping.

- **Sessions are logged out, and their gems confirmed gone, before either server is stopped.** A
  `GciTsLogout` returning is not the gem being gone — the reconcile path already knew this and
  worked around it with a retry loop — so this waits on the extents, not on the call.
- **Anything still attached that Jasper did not open** (a topaz session, a gem of another
  checkout's stone) is listed with PID, user, start time and command line, and the stop asks before
  going ahead. Jasper never ends those processes: what is safe to kill depends on what is inside
  them. A **Show In Terminal** button opens a shell on the database with the inspecting command
  typed and waiting — typed, not run, and `ps` rather than `kill`.
- **A start blocked by a held extent now names the holders.** This lives in
  `ProcessManager.startStone`, so every route into starting a stone carries it — the sidebar, the
  panel, Quick Setup, and the offer that follows a failed login — not just the one command with a
  button.

## Notes for review

- **Probe cost is UI freeze.** `wslExecSync` is synchronous on the extension host, so the probe
  blocks everything including tree rendering. It is `fuser` first, `lsof` second: 0.2s against 1.5s
  measured, same PIDs. The wait ends when the holder list *stops changing* rather than when it
  empties, and is skipped entirely when nothing was logged out — a foreign session never leaves,
  and the first draft polled for one until a 10s deadline with a 1.5s blocking call each time.
- **`lsof -b` is deliberately not used.** It is fifteen times faster again and returns *nothing at
  all* on a real ext4 extent. A probe answering "nobody" for a database with live gems would have
  Jasper stop a stone on top of them, which is the failure this exists to prevent. There is a test
  asserting the flag is never passed.
- **The stone holds its own extents.** A running `stoned` and its housekeeping gems
  (`reclaimgcgem`, `symbolgem`) are always holders; they name their stone in argv and a session's
  gem does not. Without that filter every stop of a healthy database would report the stone itself
  as an unknown process.
- `stopstone -i` ("stop even if others are logged in") is knowingly not used. With sessions logged
  out first the graceful stop succeeds on its own, and `-i` would shove a foreign session off
  without asking — the opposite of the intent above. A future set of explicit stop options is the
  better home for it.

## Testing

- Unit: `extentHolders` 21, `databaseLifecycle` 15 (including the wait/converge algorithm, extracted
  from `extension.ts` so it could be tested at all), `findExtentHolders` 7, `startStone` extent-lock
  3, `openTerminal` 3, plus tree-provider context values, panel gating and menu wiring.
- Integration: `extentHolders.e2e.test.ts` drives the real `fuser`/`lsof`/`ps` pipeline against real
  processes holding a real file — the layer where a mock happily agreed with `lsof -b`. Every wait
  polls with a deadline rather than sleeping, each test owns a `mkdtemp` directory, and teardown
  kills the process group. Verified non-flaky over 12 consecutive runs and 3 full-suite runs on
  different shuffle seeds. Skips cleanly where neither tool is installed.
- Full CI gate green: `compile` (all four tsconfigs), repo-wide `lint`, supply-chain / gci-headers /
  lockfile, and `npm test` across client 7222 / server 322 / mcp 92.
- Start and Stop exercised in the running UI by @ericwinger.


client/src/manager/__tests__/processManager.test.ts
```diff
@@ -1672,6 +1672,231 @@ describe('ProcessManager', () => {
...
+    });
+
+    it('still names the PIDs when ps says nothing about them', () => {
+      vi.mocked(wslBridge.wslExecSync).mockImplementation((cmd: string) =>
+        cmd.startsWith('fuser') ? '444' : '',
+      );
+
+      const holders = new ProcessManager(makeStorage()).findExtentHolders(db());
+
+      expect(holders).toEqual([]);
```
MatiasFernandez: 🚫 The test name says the PIDs are still named, but the assertion says nothing is. Which of the two is right matters: `findExtentHolders` runs `ps ... 2>/dev/null || true`, so `ps` printing nothing is still exit 0, the `catch` never fires, and `parseHolderDetails('')` returns `[]` — the PIDs are dropped. That makes the fallback at processManager.ts:733 ("The PIDs are still worth reporting even when ps says nothing about them") and `describeHolder`'s no-user/no-startedAt branch reachable only when the shell itself blows up.

The visible consequence: a start blocked by a gem that exits between the `fuser` probe and the `ps` call gets "Jasper could not determine which process holds it — `lsof` and `fuser` were both unavailable or reported nothing", when it had the PID a moment earlier. Was the intent to keep the bare PIDs when `ps` comes back empty? If so the `pids.map(...)` fallback wants to move out of the `catch` and into a check on the parsed result; if not, the test name is the thing to fix.

client/src/extension.ts
```diff
@@ -3624,6 +3631,167 @@ export function activate(context: vscode.ExtensionContext) {
...
+        processManager.isNetldiRunning(name, version),
+      run: async (command: string, target: ServerTarget) => {
+        await vscode.commands.executeCommand(command, target);
+      },
+    };
+  }
+
+  /** The database a whole-database command acts on when it was invoked without
+   *  a row — from the Command Palette, where there is nothing selected. */
+  async function pickDatabase(placeHolder: string): Promise<GemStoneDatabase | undefined> {
```
MatiasFernandez: 💭 The sidebar withholds Start/Stop through its `when` clauses and `powerControl` withholds it in the panel, but the palette lists every database — so the one surface that cannot see a row's state is the one that will happily run Stop on a database whose stone is already down, or on one whose servers were started outside Jasper. The stop path pays a blocking extent probe and can put the holders prompt up before `takeDownDatabase` turns out to have nothing to do.

Would it be worth filtering (or at least labelling) the pick with the same reading the tree uses — `DatabaseTreeProvider.databaseContext` already computes Running/Stopped/External? Even putting the state in the QuickPick `description` would stop the palette contradicting the two surfaces the PR body says now agree.

client/src/manager/databaseLifecycle.ts
```diff
@@ -0,0 +1,124 @@
...
+): Promise<ExtentHolder[]> {
+  if (deps.reapSessions() === 0) return deps.attachedHolders();
+
+  const deadline = deps.now() + opts.timeoutMs;
+  let holders = deps.attachedHolders();
+  while (holders.length > 0 && deps.now() < deadline) {
+    await deps.wait(opts.pollMs);
+    const next = deps.attachedHolders();
+    const settled =
+      next.length === holders.length && next.every((h, i) => h.pid === holders[i].pid);
```
MatiasFernandez: 💭 This compares the two probes pid-by-index, so the same set of holders coming back in a different order reads as "still changing" and buys another poll — and a poll here is another synchronous shell-out. Neither `fuser` nor `lsof -t` promises an order. Worth comparing them as sets (e.g. same length plus every pid present) so only a genuine change keeps the wait alive?

client/src/extension.ts
```diff
@@ -3624,6 +3631,167 @@ export function activate(context: vscode.ExtensionContext) {
     processWatcher.refresh();
   }
 
+  /** Ceiling on waiting for a database's gems to exit after their sessions are
+   *  logged out. Only reached when a gem is genuinely slow to go: the wait ends
+   *  as soon as the set of holders stops changing, so the ordinary case costs
+   *  two probes rather than the whole span. */
+  const SESSION_EXIT_TIMEOUT_MS = 10_000;
```
MatiasFernandez: 💭 Since the deadline is checked against `now()`, the probe time counts toward it, so 10s really is the ceiling on wall clock — good. What I am less sure about is what those 10s cost: `attachedHolders` is `wslExecSync`, synchronous on the extension host, and on a machine without `fuser` every poll is the 1.5s `lsof` path. A holder list that keeps churning then holds the whole host frozen for the full 10s.

Is that acceptable because it only happens when gems are genuinely slow to exit, or is it worth a lower ceiling once the probe has had to fall back to `lsof`? The PR notes cover the "nothing was logged out" case, but not this one.

client/src/extension.ts
```diff
@@ -3894,77 +4088,117 @@ export function activate(context: vscode.ExtensionContext) {
...
+        } else {
+          vscode.window.showErrorMessage(msg);
+        }
       }
       refreshAdminViews();
     }),
 
+    vscode.commands.registerCommand(
+      'gemstone.stopStone',
+      async (node: DatabaseNode & { sessionsCleared?: boolean }) => {
```
MatiasFernandez: 💭 `sessionsCleared` is declared here as well as on `ServerTarget` (databaseLifecycle.ts:8), and only the `ServerTarget` one carries the doc comment explaining what the flag promises. Since everything that actually sets it builds a `ServerTarget`, would typing this as `DatabaseNode | ServerTarget` keep the flag's declaration and its explanation in one place?

client/src/extension.ts
```diff
@@ -3624,6 +3631,167 @@ export function activate(context: vscode.ExtensionContext) {
...
+    if (choice === SHOW_IN_TERMINAL) {
+      openTerminalOnHolders(db, holders);
+      // Not a stop. They have gone to look at the processes; the stop is one
+      // click away again once they have dealt with them.
+      return false;
+    }
+    return choice === 'Stop Anyway';
+  }
+
+  const SHOW_IN_TERMINAL = 'Show In Terminal';
```
MatiasFernandez: nit: `SHOW_IN_TERMINAL` is declared after both functions that use it — `confirmStopWithHolders` above and the `startStone` handler further down. Hoisting makes it safe, but it reads as if it were not defined yet at the point of use. Worth moving it up next to `SESSION_EXIT_TIMEOUT_MS`?

MatiasFernandez (approved):
Approved with some comments worth reviewing

client/src/manager/__tests__/processManager.test.ts
ericwinger: Fixed — the intent was to keep the bare PIDs, and you're right that the code did not.

The `pids.map(...)` fallback moved out of the `catch` and onto a check of the parsed result:

```ts
let detailed: ExtentHolder[] = [];
try {
  detailed = parseHolderDetails(await wslExec(`ps ... || true`, ...));
} catch {
  // The shell itself failed; fall through to the bare PIDs below.
}
return detailed.length > 0 ? detailed : pids.map((pid) => ({ pid, command: '' }));
```

So the case you described — a gem exiting between the `fuser` probe and the `ps` call — now reports "held by PID 444" instead of "Jasper could not determine which process holds it", and `describeHolder`'s no-user/no-startedAt branch is reachable through the ordinary path rather than only when the shell blows up.

Two tests now, since the two causes take different routes: `ps` printing nothing (exit 0, no throw), and the `ps` call itself failing.

client/src/extension.ts
ericwinger: Agreed, and taken the stronger of the two options you offered — filtering rather than labelling.

The tree's private `databaseContext` is extracted as `databaseAction(status)` in `databaseServerStatus.ts`, and the tree provider now calls it instead of computing its own. `pickDatabase` takes the action it is picking for and offers only the databases that action applies to, so Stop lists running databases, Start lists stopped ones, and External is withheld in the palette exactly as it is on the row and in the panel.

Labelling alone would have left the palette able to *run* the wrong thing, including the blocking extent probe and the holders prompt you point out it would pay for first. When nothing qualifies it says which of the two reasons it is rather than showing an empty picker.

The extracted function has its own tests pinning the whole table, so the three surfaces can't drift apart again silently.

client/src/manager/databaseLifecycle.ts
ericwinger: Fixed — compared as sets now.

```ts
function samePids(a: ExtentHolder[], b: ExtentHolder[]): boolean {
  if (a.length !== b.length) return false;
  const seen = new Set(a.map((h) => h.pid));
  return b.every((h) => seen.has(h.pid));
}
```

You're right that neither tool promises an order, so a shuffle was buying a poll — and after the change below, a poll is a shell-out with a 5s timeout on it. There's a test with the same two holders returned in reverse, asserting it settles in one poll.

client/src/extension.ts
ericwinger: You're right, and I've taken the fix rather than the lower ceiling — blocking the host was the wrong shape regardless of how long for.

`wslBridge` gains `wslExec`, a non-blocking counterpart to `wslExecSync` with the same WSL routing, and `findExtentHolders` is asynchronous throughout. `SessionClearDeps.attachedHolders` returns a promise and the wait awaits it, so the 10s is now a ceiling on how long the user watches a progress notification, not on how long the extension host is unavailable to everything else. A test asserts the sync bridge is never called on that path.

One thing that fell out of it and is worth flagging, since it's a hazard being asynchronous *created*: the probe had no timeout. Under `execSync` a wedged `lsof` froze the window, which nobody could miss. Awaited with no timeout it would leave the progress notification up for ever with the editor working normally around it — the stop would simply never happen and nothing would say so. Every probe now carries `PS_TIMEOUT_MS`, matching the other `ps` shell-outs in `ProcessManager`; a timeout falls through to the next probe, or to the bare PIDs, the same as a missing tool.

And one that I have *not* fixed, so it isn't discovered later: `clearSessionsForStop` calls `reapSessions()` before it probes, which is `GciTsLogout` — synchronous on the extension host, once per session on the database. Against a healthy stone that is milliseconds; against an unresponsive one it blocks, which is the same objection you're raising here. There is no non-blocking logout in the GciTs API, and moving it onto a koffi worker the way `GciTsContinueWithAsync` does is a session-lifecycle change well beyond this path. It's named in a comment where it happens. Happy to file it if you'd like it tracked.

client/src/extension.ts
ericwinger: Done — `gemstone.stopStone` now takes `DatabaseNode | ServerTarget`, so the flag keeps its single declaration and its doc comment on `ServerTarget`.

One consequence worth naming: the union means the property has to be reached through a guard, so the check reads `!('sessionsCleared' in node && node.sessionsCleared)`. That's arguably clearer about the actual rule anyway — a tree row carries no such promise, so absence means "not cleared".

ericwinger:
Thanks for the review — all five addressed in 3a364e54, with `main` merged in.

| Comment | Outcome |
| --- | --- |
| `ps` empty drops the PIDs | Fixed. The bare-PID fallback moved out of the `catch` onto a check of the parsed result, so it is reachable through the ordinary path rather than only when the shell blows up. |
| Palette contradicts the other two surfaces | Fixed by filtering. The tree's reading is extracted as `databaseAction` and shared; the picker offers only the databases the action applies to. |
| Probes compared pid-by-index | Fixed. Compared as sets, with a test for the same holders returned shuffled. |
| 10s of blocked extension host | Fixed properly rather than by lowering the ceiling — new non-blocking `wslExec`, `findExtentHolders` async throughout. |
| `sessionsCleared` declared twice | Fixed. `DatabaseNode \| ServerTarget`, one declaration on `ServerTarget`. |

Two things came out of the fourth one that are worth reading even though they aren't in the original comments:

**Going async needed a timeout, and I'd left it out.** A wedged `lsof` used to freeze the window — unmissable. Awaited with no timeout it would leave the progress notification up for ever with the editor working normally around it, so the stop silently never happens. Every probe now carries `PS_TIMEOUT_MS`, matching the other `ps` shell-outs in `ProcessManager`.

**One synchronous blocking call remains, and this PR introduced it.** `clearSessionsForStop` calls `reapSessions()` before probing, which is `GciTsLogout` — synchronous on the extension host, once per session. Milliseconds against a healthy stone, blocking against an unresponsive one. There's no non-blocking logout in the GciTs API and moving it to a worker thread is a session-lifecycle change well past this path, so it's named in a comment rather than quietly moved. Say the word and I'll file it.

Gate: `lint`, `format:check`, `compile` (all four tsconfigs), and `npm test` green on a fresh 3.6.2 stone — client 7294 / 452 files, server 322, mcp 92. Client-only diff, so the version matrix wasn't run.

Start and Stop still to be re-exercised in the running UI by @ericwinger after the async change; the unit and integration layers can't reach host responsiveness.

ericwinger:
Follow-up in 86202ac3: CodeQL failed the previous push with one new high-severity alert, and it was mine.

The first draft of `wslExec` copied `wslExecSync`'s command line verbatim — the script interpolated into a double-quoted stretch with `"` hand-escaped inside it. CodeQL already reports that construction as `js/incomplete-sanitization` on the synchronous original (it escapes quotes but not backslashes), so duplicating it added a second alert rather than inheriting one.

`wslExec` now hands the script to `wsl.exe` as a single argv element via `execFile`. There is no outer quoting left for the escape to defend, and it skips the local `cmd.exe` that would otherwise get its own turn at the string.

`wslExecSync` still carries the original construction and the original alert. Fixing it means editing the Windows path this branch can't exercise, for a warning that predates the branch, so I've left it alone rather than widen the diff — worth its own issue if you agree.

For the record on the previous push, the full matrix was green including all five `windows-latest` jobs; CodeQL was the only failure. Worth noting those jobs don't specifically exercise `wslExec`'s WSL branch — `findExtentHolders` is unit-tested against a mocked bridge and the integration suites never call it — so Windows remains the part of this that CI green does not actually cover.

PR #546, opened 2026-09-02
Add cross-platform native socket poller

npapagna:
## Summary
- Adds `NativeSocketLibrary`, a cross-platform way to check whether a socket has data ready to read right now: backed by libc's `poll()` on POSIX and `ws2_32`'s `WSAPoll()` on Windows through koffi FFI bindings, since Node's `net.Socket` never exposes a real, pollable handle on Windows.
- `createNativeSocketLibrary()` selects the right implementation for the current platform.
- Adds test infrastructure to exercise this against genuine loopback sockets (closed/reset connections) rather than mocking the native calls: a `process.platform` stub, Windows/POSIX `describe` gates, and a loopback-connection fixture.
- Nothing uses this yet. The next PRs will use it to add non-blocking support to `GciLibrary`, as a fallback for GCI versions where `GciTsNbPoll` isn't available.

## Test plan
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run compile`
- [ ] `npm test` (run on CI matrix — POSIX and Windows paths both need coverage)

### DIFF
--- a/client/src/sockets/__tests__/nativeSocketLibrary.test.ts
+++ b/client/src/sockets/__tests__/nativeSocketLibrary.test.ts
@@ -1,0 +1,126 @@
+import koffi from 'koffi';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { NativeSocketLibrary } from '../nativeSocketLibrary';
+import { NativePOSIXSocketLibrary } from '../nativePOSIXSocketLibrary';
+import { PosixSocketLibrary } from '../bindings/posixSocketLibrary';
+import { createNativeSocketLibrary } from '../factory';
+import { changeProcessPlatformDuring } from '../../__tests__/support/process';
+import {
+  onSupportedPosixDescribe,
+  onSupportedWindowsDescribe,
+} from '../../__tests__/platformGates';
+import { LoopbackConnection, openLoopbackConnection } from './support/loopbackConnection';
+
+describe('Native socket library', () => {
+  describe('Creating instances', () => {
+    it('throws an error when a platform is not supported', () => {
+      changeProcessPlatformDuring('aix', () => {
+        expect(() => createNativeSocketLibrary()).toThrow('Unsupported platform: aix');
+      });
+    });
+  });
+
+  describe('Checking whether a socket is readable', () => {
+    let fd: bigint;
+    let loopbackConnection: LoopbackConnection;
+    let library: NativeSocketLibrary;
+
+    beforeEach(async () => {
+      loopbackConnection = await openLoopbackConnection();
+      fd = loopbackConnection.fd;
+      library = createNativeSocketLibrary();
+    });
+
+    afterEach(async () => {
+      await loopbackConnection.close();
+    });
+
+    it('reports a socket readable when data is actually waiting on it', async () => {
+      await loopbackConnection.writeFromServer('hi');
+
+      expect(library.isReadable(fd)).toBe(true);
+    });
+
+    it('reports a socket not readable when nothing has arrived yet', async () => {
+      expect(library.isReadable(fd)).toBe(false);
+    });
+
+    it('throws an error when polling an invalid file descriptor', () => {
+      expect(() => library.isReadable(library.invalidSocketDescriptor())).toThrow(
+        library.invalidFileDescriptorErrorMessage(),
+      );
+    });
+
+    onSupportedWindowsDescribe('Error handling on Windows', () => {
+      // A stale-but-non-negative closed handle makes WSAPoll return
+      // SOCKET_ERROR outright (see the "cannot be polled" test below), a
+      // different code path than the one this test means to exercise.
+      // Resetting the peer's connection is this suite's one way to produce a
+      // dead-but-open handle through a real socket: it leaves the handle
+      // itself open but errored, which reliably clears POLLRDNORM from
+      // `revents` (the same as WSAPoll would for POLLHUP/POLLERR/POLLNVAL)
+      // without touching the handle's validity. On POSIX, resolving
+      // `resetFromServer()` only guarantees the peer has torn down its own
+      // socket, not that the RST has reached and been processed by the
+      // client fd's kernel state by the time `poll()` runs, so the same
+      // assertion is flaky there.
+      it('reports the socket as unusable when it has been reset by its peer', async () => {
+        await loopbackConnection.resetFromServer();
+
+        // POLLERR|POLLHUP (0x0001|0x0002 = 0x0003): the revents WSAPoll sets
+        // for a reset connection, verified live via WSAPoll on Windows.
+        expect(() => library.isReadable(fd)).toThrow(
+          library.socketUnusableErrorMessage(fd, 0x0003),
+        );
+      });
+
+      // Unlike the reset case above, this closes the handle itself rather
+      // than merely erroring the connection behind it. This is the one
+      // real-socket way to exercise a genuine WSAPoll() syscall failure (it
+      // returns SOCKET_ERROR outright for an already-closed handle) rather
+      // than a mere unready/errored result.
+      it('reports the WSA diagnostic when the socket cannot be polled', async () => {
+        await loopbackConnection.close();
+
+        // WSAENOTSOCK (10038): what WSAPoll fails with for an already-closed
+        // handle, verified live via WSAPoll on Windows.
+        expect(() => library.isReadable(fd)).toThrow(
+          library.pollFailedErrorMessage(fd, 'WSAGetLastError 10038'),
+        );
+      });
+    });
+
+    onSupportedPosixDescribe('Error handling on POSIX', () => {
+      // Closing the handle doesn't fail poll() itself: POLLNVAL comes back
+      // through revents on an otherwise-successful call, same as the
+      // Windows reset case reports POLLHUP/POLLERR.
+      it('reports the socket as unusable once its file descriptor has been closed', async () => {
+        await loopbackConnection.close();
+
+        // POLLNVAL (0x0020, standard across POSIX libc's <poll.h>), the one
+        // revents bit a closed fd sets; verified live via poll() on macOS.
+        expect(() => library.isReadable(fd)).toThrow(
+          library.socketUnusableErrorMessage(fd, 0x0020),
+        );
+      });
+
+      // No real socket condition reliably makes poll() itself fail (EFAULT/
+      // EINTR/ENOMEM aren't safely triggerable), so this fakes the binding
+      // directly rather than going through a real fd. koffi.errno() is the
+      // real function, primed with a real POSIX error code, to check that
+      // the thrown message reflects whatever poll() actually left there.
+      it('reports the diagnostic when the poll syscall itself fails', () => {
+        koffi.errno(koffi.os.errno.EBADF);
+        const fakeLibrary = new NativePOSIXSocketLibrary({
+          name: 'fake',
+          poll: () => -1,
+          POLLIN: 0x0001,
+        } as unknown as PosixSocketLibrary);
+
+        expect(() => fakeLibrary.isReadable(3n)).toThrow(
+          fakeLibrary.pollFailedErrorMessage(3n, `errno ${koffi.os.errno.EBADF}`),
+        );
+      });
+    });
+  });
+});
--- a/client/src/sockets/__tests__/support/loopbackConnection.ts
+++ b/client/src/sockets/__tests__/support/loopbackConnection.ts
@@ -1,0 +1,133 @@
+import { createServer, Socket } from 'net';
+import { createNativeSocketLibrary } from '../../factory';
+import { RawSocketConnection } from '../../nativeSocketLibrary';
+
+export type LoopbackConnection = {
+  fd: bigint;
+  writeFromServer: (data: string) => Promise<void>;
+  resetFromServer: () => Promise<void>;
+  close: () => Promise<void>;
+};
+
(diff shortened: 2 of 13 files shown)


client/src/sockets/__tests__/support/loopbackConnection.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    let peer: Socket | undefined;
+    let client: RawSocketConnection | undefined;
+    let closed = false;
+
+    const resolveOnceBothEndsAreUp = () => {
+      if (peer && client) {
+        resolve({
+          fd: client.fd,
+          writeFromServer: (data) => new Promise((writeResolve) => peer!.end(data, writeResolve)),
```
MatiasFernandez: 🚫 `peer.end(data)` writes **and** half-closes, so the client fd gets a FIN along with the bytes — and on POSIX a received FIN sets `POLLIN` on its own, whether or not any data arrived. I checked on macOS with the same `poll()` binding this PR adds:

- `peer.end('')` (pure FIN, zero bytes) → `revents=0x11`, POLLIN **set**
- `peer.write('hi')` → `revents=0x01`, POLLIN set
- nothing sent → `revents=0x00`

So `reports a socket readable when data is actually waiting on it` would pass even if zero bytes were ever delivered — what it's really asserting is "the peer closed", which is the same condition the POSIX `Error handling` block treats as *unusable*. Should `writeFromServer` use `peer.write(data, cb)` instead, so the test can only go green on real data?

One catch if you do: the doc above says "The client side is never read from, so any bytes sent stay at the OS level rather than being drained" — that doesn't hold for a `net.Socket` client. libuv starts reading as soon as the fd is readable, so the bytes move into `client.readableLength` and POLLIN clears on its own. In the same experiment, 100ms after the write callback the fd was back to `revents=0x00` with `client.readableLength === 2`, which is probably why `end()` works today (POLLHUP/EOF is sticky, a plain write isn't).

`clientSocket.pause()` in `connectRawSocket` stops libuv's reads and makes that doc claim actually true — with it, `write('hi')` still reported `revents=0x01` 200ms later.

client/src/sockets/nativeSocketLibrary.ts
```diff
@@ -0,0 +1,113 @@
+/** A raw, poll-able socket handle, and a way to close it. */
+export type RawSocketConnection = {
+  fd: bigint;
+  disconnect: () => void;
+};
+
+export abstract class NativeSocketLibrary {
```
MatiasFernandez: 💭 `NativeWindowsSocketLibrary` got a nice class-level doc, but the abstract base — the abstraction the follow-up PRs will actually program against — doesn't have one. Could you add a short one covering what a "native socket library" is for (a right-now readiness check, never a wait), why it exists at all (the `GciTsNbPoll` fallback the PR description mentions), and the `bigint` fd convention that lets POSIX fds and Windows `SOCKET` handles share one signature?

That last point especially: right now a reader has to reverse-engineer it from the `Number(fd)` narrowing comment in the POSIX subclass.

client/src/sockets/nativeSocketLibrary.ts
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * Builds the message reported when the underlying readiness check itself
+   * fails, as opposed to succeeding but reporting the socket unusable.
+   *
+   * @param fd - OS-level file descriptor the check was performed on.
+   * @param diagnostic - platform-specific detail about the failure, e.g. an
+   *   errno or a `WSAGetLastError` code.
+   * @returns The error message text.
+   */
+  public pollFailedErrorMessage(fd: bigint, diagnostic: string) {
```
MatiasFernandez: 💭 `pollFailedErrorMessage` and `socketUnusableErrorMessage` are `public`, but nothing outside this class calls them: the `throwX` helpers below are the only callers, and the tests match these two messages with literal regexes (`/socket is in an unusable state \(revents=0x/`) instead. `invalidFileDescriptorErrorMessage` *is* called from the test, so the suite ends up asserting two of the three messages one way and the third the other.

Should these two drop to `protected`, and the invalid-fd test switch to a literal too? Asserting a message through the very method that builds it makes that test pass for any text, so it pins down which error is thrown but not what it says — which is the part the regexes are protecting for the other two.

client/src/sockets/nativeSocketLibrary.ts
```diff
@@ -0,0 +1,113 @@
...
+   * Opens a real, poll-able client connection to 127.0.0.1:`port`. Production
+   * code never opens its own sockets through this method; it exists for test
+   * fixtures that need a genuine handle to check readiness against, rather
+   * than a fake result.
+   *
+   * @param port - TCP port on the loopback interface to connect to.
+   * @returns The connected handle, and a way to close it.
+   * @throws {Error} If the connection can't be established.
+   */
+  public abstract connectRawSocket(port: number): Promise<RawSocketConnection>;
```
MatiasFernandez: 💭 `connectRawSocket` is documented as existing only for test fixtures, but it's an abstract member of the production base class — so every implementation must provide it, and on Windows that means `windowsSocketLibrary.ts` binds `socket`, `connect` and `closesocket` (plus the `WS2_RawSockAddrIn` struct) into the shipped extension for a path production never takes.

Since nothing consumes this library yet, is this a good moment to split it — `NativeSocketLibrary` carrying just `isReadable`, and the raw-connection bits living in a test-support subclass under `__tests__/support/`? Fine by me if the extra indirection isn't worth it, but it would keep the production surface down to the one primitive the follow-up PRs actually need.

client/src/sockets/nativeSocketLibrary.ts
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * Throws the error for when the underlying readiness check itself fails,
+   * as opposed to succeeding but reporting the socket unusable.
+   *
+   * @param fd - OS-level file descriptor the check was performed on.
+   * @param diagnostic - platform-specific detail about the failure, e.g. an
+   *   errno or a `WSAGetLastError` code.
+   * @throws {Error} Always.
+   */
+  protected throwPollFailedError(fd: bigint, diagnostic: string) {
```
MatiasFernandez: 💭 Should these two be declared `: never`? They're documented `@throws {Error} Always`, but the inferred `void` means TS believes control continues past them. In both `isReadable` implementations that makes the `status < 0` branch look like it falls through into the `status === 0` check and then the `revents` test — the opposite of what the code means, and the reading an agent or a hurried reviewer would take from it. `: never` would line the compiler's view up with the doc comment, and would keep a future reordering of those branches honest.

client/src/sockets/nativeWindowsSocketLibrary.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    const address = {
+      family: this.ws2.AF_INET,
+      portBytes: [(port >> 8) & 0xff, port & 0xff],
+      addrBytes: [127, 0, 0, 1],
+      zero: [0, 0, 0, 0, 0, 0, 0, 0],
+    };
+
+    const status = this.ws2.connect(handle, address, 16) as number;
```
MatiasFernandez: 💭 `16` here is the `sockaddr_in` size that the `WS2_RawSockAddrIn` comment in `windowsSocketLibrary.ts` spells out field by field. If that struct ever gains or loses a field, the literal goes stale silently and `connect()` fails with a WSAEFAULT that points nowhere near here.

Worth deriving it instead? `koffi.sizeof('WS2_RawSockAddrIn')` returns 16. Or exporting it from the bindings module next to `AF_INET`/`SOCK_STREAM`, since that's where the rest of the Winsock constants already live.

client/src/sockets/__tests__/nativeSocketLibrary.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      // the thrown message reflects whatever poll() actually left there.
+      it('reports the diagnostic when the poll syscall itself fails', () => {
+        koffi.errno(koffi.os.errno.EBADF);
+        const fakeLibrary = new NativePOSIXSocketLibrary({
+          name: 'fake',
+          poll: () => -1,
+          POLLIN: 0x0001,
+        } as unknown as PosixSocketLibrary);
+
+        expect(() => fakeLibrary.isReadable(3n)).toThrow(/poll syscall itself failed \(errno 9\)/);
```
MatiasFernandez: nit: The errno is primed symbolically (`koffi.os.errno.EBADF`) but asserted as a hardcoded `9`. Could the expectation be built from the same source — e.g. `` new RegExp(`poll syscall itself failed \\(errno ${koffi.os.errno.EBADF}\\)`) `` — so the two can't drift, and a reader doesn't have to know that EBADF is 9?

client/src/types.ts
```diff
@@ -0,0 +1,5 @@
+/** Rejects a Promise-returning (async) type at the type level, resolving to `never` instead. */
+export type NotPromise<T> = T extends Promise<unknown> ? never : T;
+
+/** A callback that takes no arguments and cannot be a promise-returning (async) function. */
+export type SyncThunk<T> = () => NotPromise<T>;
```
MatiasFernandez: nit: `client/src/` currently scopes its type modules by domain (`loginTypes.ts`, `sysadminTypes.ts`) rather than having a shared catch-all, so a top-level `types.ts` reads like an invitation for anything type-shaped to land here later. Both of these are about the sync/async boundary — would something narrower like `syncTypes.ts` fit the existing convention better?

MatiasFernandez (approved):

client/src/sockets/__tests__/support/loopbackConnection.ts
npapagna: Good catch, thanks. Switched `writeFromServer` to `peer.write(data, cb)` so a bare FIN can no longer satisfy the "readable" assertion, and added `clientSocket.pause()` in `nativePOSIXSocketLibrary.connectRawSocket` so libuv doesn't silently drain the socket in the background (confirmed live that without `pause()`, POLLIN clears again ~100ms after the write). Also fixed the write callback to actually reject on error instead of discarding it.

client/src/sockets/nativeSocketLibrary.ts
npapagna: Added a class-level doc covering what this checks (readiness right now, never a wait), why it exists (fallback for `GciTsNbPoll`), and the `bigint` fd convention that lets POSIX fds and Windows `SOCKET` handles share one signature.

client/src/sockets/nativeSocketLibrary.ts
npapagna: Kept these `public` rather than dropping to `protected`, and went the other direction: switched every test in this file to assert through the public message-building methods instead of ad-hoc regexes, matching what `invalidFileDescriptorErrorMessage` already did. For the cases where the exact `revents`/diagnostic isn't knowable ahead of time (peer-reset, closed-fd, WSA failure), I verified the real values live on both macOS and a Windows VM (`POLLNVAL`/`0x0020` on POSIX; `POLLERR|POLLHUP`/`0x0003` and `WSAENOTSOCK`/`10038` on Windows) and asserted those exactly, so the tests now pin down the actual message text everywhere instead of just the invalid-fd case.

client/src/sockets/nativeSocketLibrary.ts
npapagna: Discussed this, leaving it as-is for now. The extra `socket`/`connect`/`closesocket` bindings only get declared once, lazily, the first time `NativeWindowsSocketLibrary` is constructed (nothing calls that yet), so the runtime cost is negligible even after a follow-up PR wires it up. The split would buy some production-surface cleanliness, but didn't seem worth the added indirection for one primitive right now, happy to revisit if a later PR makes the coupling more painful.

client/src/sockets/nativeSocketLibrary.ts
npapagna: Good point, added `: never` to both `throwPollFailedError` and `throwUnusableSocketError`. Confirmed it compiles cleanly and doesn't change any call-site behavior.

client/src/sockets/nativeWindowsSocketLibrary.ts
npapagna: Exported `SOCKADDR_IN_SIZE` from `windowsSocketLibrary.ts` via `koffi.sizeof('WS2_RawSockAddrIn')` and used it here instead of the literal `16`. Verified live on a Windows VM that it still evaluates to `16` and that `connect()` still succeeds with it.

client/src/sockets/__tests__/nativeSocketLibrary.test.ts
npapagna: Resolved as a side effect of the visibility discussion above: this test now asserts through `` fakeLibrary.pollFailedErrorMessage(3n, `errno ${koffi.os.errno.EBADF}`) `` instead of a hardcoded `9`, so it's built from the same symbolic source it primes `koffi.errno()` with.

client/src/syncTypes.ts
```diff
@@ -0,0 +1,5 @@
+/** Rejects a Promise-returning (async) type at the type level, resolving to `never` instead. */
+export type NotPromise<T> = T extends Promise<unknown> ? never : T;
+
+/** A callback that takes no arguments and cannot be a promise-returning (async) function. */
+export type SyncThunk<T> = () => NotPromise<T>;
```
npapagna: Renamed `client/src/types.ts` to `client/src/syncTypes.ts` and updated its two importers (`gciLibrary.ts`, the test's `process.ts` helper).

PR #549, opened 2026-09-02
Give method editors their own language id, so a breakpoint can only be set in method source

### LINKED ISSUE #515: Give method editors their own language id, so the breakpoint gutter is only offered where a breakpoint can be set
## Summary

VS Code decides where its breakpoint gutter appears by **language** (`contributes.breakpoints`), and offers no way to narrow that by URI scheme or document. Jasper gives the single `gemstone-smalltalk` language id to five different kinds of document, only one of which — a `gemstone://` method editor — can actually hold a breakpoint. The result is that VS Code invites a breakpoint in a workspace, a `.gst` file, and the debugger's read-only source panes, all of which must then refuse it.

Refusing after the click is the best that can be done today, and that is what Jasper now does. But an invitation that is always withdrawn is a worse experience than never extending it: the developer clicks, a dot appears, a warning fires, and the dot vanishes. Giving method editors a language of their own — say `gemstone-method` — and declaring `contributes.breakpoints` for that language alone would mean the gutter is simply never offered anywhere else, with no refusal needed.

## Acceptance criteria

- The breakpoint gutter is offered **only** in `gemstone://` method editors.
- Clicking the gutter in a workspace, a `.gst` file, or either of the debugger's read-only source views is not possible — nothing to refuse, no warning shown.
- Every editor that shows GemStone Smalltalk keeps its syntax highlighting, including the debugger's read-only panes. This rules out the cheaper workaround of withholding the language from those documents.
- Execute It / Display It / Inspect It and the other editor commands keep working everywhere they work today — a workspace, a `.gst` file, and a method editor alike.
- Completion, hover, diagnostics, formatting and semantic tokens continue to work in method editors.

### The five documents sharing the language today

| Document | Gets `gemstone-smalltalk` from | Should offer a breakpoint? |
| --- | --- | --- |
| `gemstone://` method editor | `client/src/extension.ts:824` (and `client/src/gemstoneExplorer.ts:3843`, set before the doc is shown so the CodeLens does not pop in) | **yes** |
| Workspace (`untitled:Workspace`) | `client/src/workspace.ts:39` | no |
| `.gst` file on disk | `extensions: [".gst"]` in the language contribution | no |
| Debugger panel's companion source tab | `client/src/debuggerPanel.ts:3564`, set explicitly so executed code is highlighted | no |
| Debug adapter's frame source view | previously the `mimeTypes` entry on the language contribution | no |

The last one is already closed — the debug adapter now withholds the mime type — but at the cost of syntax highlighting in that view. That trick is not available for the debugger panel's companion tab, where the highlighting is the point.

### A possible direction

One approach could be to declare a second language, `gemstone-method`, sharing the existing grammar and language configuration, and assign it wherever a `gemstone://` method document is opened; `contributes.breakpoints` would then name only `gemstone-method`. The two setters at `extension.ts:824` and `gemstoneExplorer.ts:3843` look like the only places that would need to change, since every other document is left on `gemstone-smalltalk`.

The bulk of the work is auditing what currently keys off the language id so nothing silently stops applying to method editors:

- **`package.json`** — around 21 references. The grammar and `[gemstone-smalltalk]` editor settings would need to cover both ids; roughly a dozen `when` clauses of the form `editorTextFocus && resourceLangId == gemstone-smalltalk` gate Execute It / Display It / Inspect It and friends and would need to match either id; the `gemstone` debugger contribution lists its languages; and `contributes.breakpoints` moves to the new id alone.
- **`client/src/extension.ts`** — the language client's `documentSelector` (~line 654) and several `{ scheme, language }` selector arrays (~lines 850, 876, 890) that register providers per scheme.

A cheaper alternative worth weighing first: if VS Code ever gains a way to scope `contributes.breakpoints` by scheme, none of this would be needed. Worth a check against current VS Code API before taking on the split.

### Background

The refuse-after-the-fact behaviour this would replace was added in https://github.com/GemTalk/Jasper/pull/481, which also has the fuller history of how each of the five documents ends up on the same language.

ericwinger:
**The idea in one line: a breakpoint should only be settable in a method's source — not in a class definition, a class comment, a workspace, a `.gst` file, a notebook cell, or the debugger's read-only source views.**

A GemStone breakpoint is a step point in a compiled method, so a method editor is the only document one can mean anything in. Everywhere else, the offer had to be withdrawn after the click: the dot appeared, a warning fired, the dot vanished. An invitation that is always withdrawn is worse than never extending one.

Fixes #515.

## Why a new language id

VS Code decides where the breakpoint gutter appears by **language** (`contributes.breakpoints`), and gives no way to narrow that by URI scheme or document. Jasper gave the single `gemstone-smalltalk` id to five kinds of document, only one of which can carry a breakpoint — so the gutter appeared in all five.

Method editors now carry `gemstone-method`, and `contributes.breakpoints` names that id alone. The gutter is simply never offered anywhere else, so there is nothing to refuse.

The split is about where a breakpoint may be set and **nothing else**. Both ids share the grammar, the language configuration, the editor defaults, the language client's document selector, and the editor-command `when` clauses — Execute It / Display It / Inspect It, Senders, Implementors, completion, hover, diagnostics, formatting and syntax highlighting all work exactly where they did.

`client/src/languageIds.ts` owns the rule. The language a document gets and the refusal that backstops it derive from **one** predicate — the same test `applyToUri` makes before touching the gem — so the gutter is offered exactly where a breakpoint can be armed, and the two cannot drift apart.

## Two things that got better along the way

- **The debugger's read-only frame source is syntax highlighted again.** Its mime type had to be withheld purely to keep the gutter away; with the gutter gone from that language, it can come back.
- **A stray breakpoint on a class definition or an override diff view is now taken back** with the same explanation as the rest, instead of being silently dropped on the floor.

## The subtle one, for review

`debuggerPanel.showSourceEditor` displays **either** a `gemstone://` method (when the frame is edit-and-continue-able) **or** the read-only stash — and it re-tagged whatever it was given. Left alone, it would have put method editors back on `gemstone-smalltalk` and silently undone the split in the one editor a developer is most likely to want a breakpoint in. It is now scoped to the read-only scheme, with a test.

## ⚠️ One out-of-scope change snuck in

**`Drop the connection-success banner from the Explorer`** is unrelated to #515 and is here because I asked for it mid-review, not because the issue called for it.

Logging in flashed a green `✅ Connected to <stone>` message atop the Dictionaries view for five seconds. It read as odd, and it announced something the Active Session status-bar item already says permanently. The connecting spinner still clears and Start Here still appears, so a connect is not silent. No changelog entry: the banner was added after `v1.8.14` and never shipped, so no user would recognise it being removed.

It is its own commit and reverts cleanly on its own if you would rather it went in separately.

A second commit — `Ask for the senders/implementors lens only in source editors` — is adjacent rather than out of scope: the CodeLens provider was registered against the whole `gemstone://` scheme, so every class comment and class definition asked it for lenses just to be told there were none. Those counts moved to the selector hover in #432, so that registration had been buying nothing since.

## Commits

| | |
| --- | --- |
| `Give method editors their own language id for the breakpoint gutter` | the change |
| `Ask for the senders/implementors lens only in source editors` | adjacent cleanup |
| `Drop the connection-success banner from the Explorer` | **out of scope** |
| `Offer the breakpoint commands in a method editor alone` | found in manual testing |
| `Tie the breakpoint keybinding gate to the menu's` | guards the above |

The fourth is worth a look: the gutter stopped being offered outside a method, but the five context-menu breakpoint actions did not — I had widened those `when` clauses to cover both ids instead of narrowing them to the one that means "a compiled method". `Shift+F9` had the same hole via a scheme gate that also matched comments and definitions. The follow-up test derives the command set from the menu, so the two doors to one command cannot drift apart again.

## Testing

Full suite green against a live 3.6.2 stone: **452/452 files, 7271 passed**.

`client/src/__tests__/methodGrammar.test.ts` tokenizes with the real TextMate engine and asserts a method editor produces theme scopes identical to a Smalltalk document — an `include` that silently resolves to nothing is not a TextMate error, so it is worth proving rather than assuming.

The `when`-clause and keybinding tests were **mutation-tested**: reverting either gate to its broken form fails a test in two files independently.

Manually verified in a dev window: gutter present in a method editor and in the debugger's editable companion pane; absent in a workspace, a `.gst` file, a class comment, a class definition and the Executed Code frame; the Executed Code frame source now highlighted; editor commands present in both a method and a workspace; breakpoint commands in a method alone; Refactor → Rename Temporary still offered; a saved new method gets a gutter; Topaz `.gs` still shows its senders/implementors lens; no banner on login.


client/src/extension.ts
```diff
@@ -857,30 +852,16 @@ export function activate(context: vscode.ExtensionContext) {
...
     vscode.languages.registerHoverProvider(providerSelectors, hoverProvider),
     vscode.languages.registerCompletionItemProvider(providerSelectors, completionProvider),
-    vscode.languages.registerCodeLensProvider(codeLensSelectors, codeLensProvider),
+    vscode.languages.registerCodeLensProvider(CODE_LENS_SELECTORS, codeLensProvider),
     codeLensProvider, // dispose() cancels pending count lookups + releases the emitter
     // Hosts "Rename Temporary/Argument…" under the native "Refactor…" menu in a
     // saved (scheme:gemstone) method editor.
     vscode.languages.registerCodeActionsProvider(
-      { scheme: 'gemstone', language: 'gemstone-smalltalk' },
+      { scheme: 'gemstone', language: METHOD_LANGUAGE },
```
MatiasFernandez: 💭 This one narrows more than the gutter, I think. The provider used to be registered for `{ scheme: 'gemstone', language: 'gemstone-smalltalk' }`, which before the split also matched a class definition, a `new-method` / `new-class` template and the override diff view — now only a compiled method's editor gets the native **Refactor…** menu at all.

The case I'd want to be sure about is a class definition editor: put the cursor on an instance variable name and `Refactor… → Rename…` was reachable there before this PR, and isn't now. Same for a `new-method` template you're still writing.

Was that deliberate? If those refactorings never resolved outside a compiled method anyway then this is a clean improvement and worth saying so in the doc comment above; if they did work, this is the one place the split changes something other than where a breakpoint may be set, and the PR description says it doesn't. Either way it seems worth a test in `languageContributions.test.ts`-style pinning which languages the code-action provider attaches to, since nothing currently covers it.

client/src/languageIds.ts
```diff
@@ -0,0 +1,90 @@
...
+ * Deliberately the same test `BreakpointManager.applyToUri` makes before it
+ * touches the gem: a saved, compiled method (`parseUri` kind `method`) that is
+ * not an override *diff view*. Everything else behind the scheme has no compiled
+ * method to hold a step point — a class definition, a `new-method` template that
+ * has never been compiled, a `new-class` template — and the read-only diff view
+ * shows two versions of a method at once, so a line in it does not name one.
+ *
+ * Both the language a document is given and the refusal that backstops it are
+ * derived from this, so the gutter is offered exactly where a breakpoint can be
+ * armed and the two can't drift apart.
```
MatiasFernandez: 💭 "Deliberately the same test `BreakpointManager.applyToUri` makes" and "the two can't drift apart" is the claim, but `applyToUri` still open-codes it:

```ts
const method = parseMethodUri(uri);
if (!method || method.diffView) { ... }
```

So there are two independent copies of the predicate, and nothing structurally stops one from changing — which is exactly the drift the comment says can't happen. `applyToUri` needs `method` afterwards so it can't collapse entirely, but could it be `if (!isMethodSourceUri(uri)) return …` followed by `const method = parseMethodUri(uri)!`, so the rule really does live in one place? Otherwise, should we soften the claim to "kept in step with" rather than "cannot drift apart"?

client/src/languageIds.ts
```diff
@@ -0,0 +1,90 @@
...
+ * read-only source — are given.
+ *
+ * Not every GemStone document: a Topaz `.gs` file and a Tonel `.st` file are
+ * GemStone Smalltalk source too, and are deliberately out. This set exists to
+ * answer "did a breakpoint here come from an offer of ours?", and the gutter has
+ * only ever been contributed for the ids listed here — so those are the only
+ * ones there is anything of ours to take back. A breakpoint on a Topaz or Tonel
+ * file came from somewhere else and is left where it is.
+ */
+export const SMALLTALK_LANGUAGES = [SMALLTALK_LANGUAGE, METHOD_LANGUAGE] as const;
```
MatiasFernandez: 💭 The name says "the Smalltalk languages", but the doc right above spends a paragraph explaining that it deliberately *isn't* that — Topaz and Tonel are GemStone Smalltalk source and are out, because what this really answers is "has Jasper ever contributed a breakpoint gutter for this language?". Should we name it for that instead — `GUTTER_LANGUAGES` / `BREAKPOINT_OFFER_LANGUAGES` / similar? Then the single caller in `inviteWeCannotHonour` reads as the question it's asking, and the next person adding a Smalltalk-ish language id doesn't reach for this list on the strength of the name.

client/src/breakpointManager.ts
```diff
@@ -1100,27 +1107,34 @@ const NOT_A_METHOD_REFUSAL =
...
-  if (uri.scheme === 'gemstone') return false;
+  if (isMethodSourceUri(uri)) return false;
 
   const uriStr = uri.toString();
   const open = vscode.workspace.textDocuments.find((d) => d.uri.toString() === uriStr);
-  if (open) return open.languageId === 'gemstone-smalltalk';
+  if (open) return (SMALLTALK_LANGUAGES as readonly string[]).includes(open.languageId);
 
   // Restored across a restart, before its editor is opened.
   return uri.path.endsWith('.gst');
```
MatiasFernandez: 💭 Is the restored-across-a-restart path now a bit narrower than the doc above claims? The comment says this "still catches a breakpoint VS Code restores from before the language split, when the gutter was offered in all of those" — but that only holds once the document is open and has a language to read. A breakpoint restored on a `gemstone://` class definition or an override diff view before its editor is opened falls through to this line, isn't a `.gst`, and gets left in the gutter arming nothing — which is the "dropped on the floor" case the PR advertises fixing for exactly those two documents.

Since a `gemstone://` URI can be classified without the document being open (that's what `gemstoneDocumentLanguage` / `isMethodSourceUri` already do off the URI alone), would something like `if (uri.scheme === 'gemstone') return true` here — reached only after the `isMethodSourceUri` early return, so a real method is already excluded — close that gap? Happy to leave it if restored `gemstone://` breakpoints aren't a thing in practice, but then the comment above probably shouldn't promise it.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ All notable changes to the **GemStone Smalltalk** extension will be documented i
...
...d it is the only one the breakpoint contribution names, so the gutter appears in a `gemstone://` method and nowhere else — as do the breakpoint commands themselves (**Toggle/Enable/Disable Breakpoint at Cursor**, **Clear Method Breakpoints**, **Toggle Step Point Numbers**, and `Shift+F9`), which were offered in the same editors the gutter was and could only ever be refused there. Everything else about those editors is unchanged: Execute It, Display It, Inspect It, Senders, Implementors, completion, hover, diagnostics, formatting and syntax highlighting all work exactly where they did. Two things got better along the way — the debugger's read-only frame source is syntax highlighted again (the mime type that gave it highlighting had to be withheld to keep the gutter away, and no longer does), and a stray breakpoint on a class definition or an override diff view, which used to be silently dropped on the floor, is now taken back with the same explanation as the rest. The refusal itself stays as a backstop for the routes that bypass the contribution: VS Code's `debug.allowBreakpointsEverywhere`, and DAP clients that are not VS Code. ([#515](https://github.com/GemTalk/Jasper/issues/515))
```
MatiasFernandez: 💭 Worth a sentence here about what a user with their own settings will notice? Anything they scoped to the old language id stops reaching method editors after the split — `"[gemstone-smalltalk]": { "editor.fontSize": … }` and friends in their `settings.json`, and `editor.tokenColorCustomizations` rules written against `source.gemstone-smalltalk` as a parent scope (the `methodGrammar.test.ts` comment already spots the latter and notes nothing *Jasper* ships relies on it — but a user's own customizations might).

It's a small thing and completely invisible from the code, which is what makes a changelog line valuable: "my method editors suddenly look different from my workspaces" is a confusing symptom to debug from the outside, and one line naming `[gemstone-method]` as the id to add turns it into a two-second fix.

client/src/__tests__/breakpointManager.test.ts
```diff
@@ -423,10 +424,91 @@ describe('BreakpointManager', () => {
...
 
+    it('takes back one on a gemstone:// class definition, which has no method to arm', () => {
+      // Smalltalk source behind our own scheme, but there is no compiled method
+      // under a class definition, so `applyToUri` can do nothing with it. It is
+      // offered no gutter now (its language is gemstone-smalltalk); reachable
+      // via allowBreakpointsEverywhere, and previously dropped on the floor.
+      const uri = 'gemstone://1/Globals/Array/definition';
+      const stray = bpOn(uri);
+      workspace.textDocuments = [
+        { uri: Uri.parse(uri), languageId: 'gemstone-smalltalk', isDirty: false },
```
MatiasFernandez: 💭 I saw in this PR that we have some constants to reference these strings. Does it make sense to use them all-around? I see that in some places we use `METHOD_LANGUAGE` and in other places we use `gemstone-method`. Can we unify all in something like a `GEMSTONE_METHOD_LANGUAGE` constant?

MatiasFernandez (approved):

client/src/extension.ts
ericwinger: Deliberate, and it turns out to be the clean-improvement branch of your question — but you're right that it was unstated, so it now says so and is pinned.

Every action in that menu runs a command that funnels through `resolveMethodEditor` (`renameAtCursorShared.ts`), which refuses anything but a `kind: 'method'` URI. All eight go through it: extract method/temporary, inline method/temporary, the unified rename, convert-temp-to-ivar and change-signature. So in a class definition editor, `Refactor… → Rename…` was reachable and answered *"This rename works in a method source editor."* — the full list was offered and every entry declined. Same for a `new-method` template: it parses as `kind: 'new-method'`, not `'method'`, so it refused too (`changeSignatureCommand` even has a special-case guard to make that refusal less generic).

The one case that did resolve is the read-only override diff view, `kind: 'method'` with a display label on the selector segment. It's out now as well, on purpose: the buffer is `FilePermission.Readonly`, so a rewrite has nowhere to land, and the two versions on screen don't name the one method to run it against — `parseUri` strips the `(base)` / `(session override)` label, so a rename fired from the base side would have gone at whichever method the plain URI resolves to.

So: nothing that ever completed is lost. The selector is now a named constant beside the provider carrying that reasoning (`REFACTOR_CODE_ACTION_SELECTOR`, the pattern `CODE_LENS_SELECTORS` already uses), pinned by three tests in `renameRefactorCodeActions.test.ts`, and it has its own changelog entry since the menu vanishing from those editors is user-visible. b2167642

client/src/languageIds.ts
ericwinger: Fair catch — the comment was describing an intention, not the code. Fixed rather than softened.

`applyToUri` couldn't call the boolean because it needs the parsed method immediately afterwards (className, isMeta, selector, environmentId), which is exactly why the copy existed. So the rule now answers *with* the method:

```ts
export function methodSourceRef(uri: vscode.Uri): MethodUriRef | null {
  const method = parseMethodUri(uri);
  return method !== null && !method.diffView ? method : null;
}

/** Whether `uri` is the source of a compiled method (see methodSourceRef). */
export function isMethodSourceUri(uri: vscode.Uri): boolean {
  return methodSourceRef(uri) !== null;
}
```

`applyToUri` is now `const method = methodSourceRef(uri); if (!method) return …` — no `!` needed, and one statement of the rule with the language decision and the refusal both reading from it. 266c5355

client/src/languageIds.ts
ericwinger: Agreed — renamed to `BREAKPOINT_GUTTER_LANGUAGES`, and the doc comment now leads with the question instead of denying its own name:

> The languages Jasper has ever contributed a breakpoint gutter for … Not "the Smalltalk languages", and deliberately not every GemStone document.

The call site in `inviteWeCannotHonour` reads as what it's asking now, and the `describe` block in `languageIds.test.ts` moved with it. 266c5355

CHANGELOG.md
ericwinger: Added, and you were right that it's the half you can't see from the code. The bullet now ends:

> One thing to know if you have tuned the editor yourself: a settings block scoped to the old id — `"[gemstone-smalltalk]": { "editor.fontSize": … }` in your `settings.json` — now covers workspaces, `.gst` files and the debugger's read-only panes but no longer reaches method source; add a matching `"[gemstone-method]"` block to cover method editors too. Themes and ordinary `editor.tokenColorCustomizations` rules are unaffected, since every token in a method editor still carries the same `.gemstone-smalltalk` scope names it always did — but a rule that uses `source.gemstone-smalltalk` as a *parent* scope no longer matches there, because the document is now rooted at `source.gemstone-method`.

The token-colour case is worth splitting out, I think: it's the one a reader would otherwise assume is broken. Only a rule that names the old root as a parent selector stops matching; a plain `keyword.control.gemstone-smalltalk` rule still hits, which is what the `methodGrammar.test.ts` assertions pin. 4cce7b3a

The Refactor… menu narrowing you spotted above got its own entry in the same pass.

client/src/__tests__/breakpointManager.test.ts
ericwinger: Yes — done, and it found a real gap rather than just tidying.

Every `languageId:` in the suite now comes from `languageIds.ts`, including the two manifest-reading tests (`languageContributions.test.ts`, `editorContextMenu.test.ts`). That last part is the gap: those files carried their own `const METHOD = 'gemstone-method'`, so they asserted package.json against a *second copy* of the string. Renaming the constant's value without touching the manifest would have left them green and the extension broken.

So the literal is now written in exactly one place — a test in `languageIds.test.ts` — because the ids are a published contract (the manifest spells them out, and so can a user's `settings.json`). A rename that forgets the manifest fails there.

I kept the existing names (`METHOD_LANGUAGE`, `SMALLTALK_LANGUAGE`, `CLASS_COMMENT_LANGUAGE`) rather than adding a `GEMSTONE_` prefix — the module already qualifies them and they're used unprefixed throughout `extension.ts`. Two things left as literals on purpose: the TextMate scope names (`source.gemstone-method`) are a different namespace, and `gemstone-topaz` / `gemstone-tonel` are resolved by file extension and never assigned by Jasper, so they don't belong in a module about the ids we hand out. 580fb6fc

PR #550, opened 2026-09-03
Stop the break tests asserting a premise they cannot construct

### LINKED ISSUE #540: Flaky integration test: "escalates to a hard break when the run ignores a soft one" races on soft-break delivery
## Summary

`client/src/queries/__tests__/sunitRunNb.integration.test.ts` → *"escalates to a hard break when the run ignores a soft one"* fails intermittently in CI. It failed on all five `windows-latest` jobs of one run and passed on all five `ubuntu-latest` jobs of the same commit; re-running the identical commit turned all five Windows jobs green with no code change.

The test asserts that a run which **ignores** a soft break is escalated to a hard break. That precondition is never established, and — as measured below — **cannot** be established: Jasper's own test-runner code traps the soft break and turns it into an ordinary result, and soft-break delivery latency is bimodal (~80 ms or ~1.9 s) and non-deterministic on both GemStone 3.6.x and 3.7.x. The test therefore passes or fails according to which delivery mode happens to occur.

## Environment

- **Jasper:** 1.8.14
- **GemStone server:** reproduced on 3.6.2 and 3.7.5 (local test stones). CI matrix covers 3.6.2, 3.6.8, 3.7.2, 3.7.4.3, 3.7.5 — the CI failure occurred on **all five** versions simultaneously on the Windows leg.

## Steps to reproduce

The failure is a race, so it does not reproduce on demand. To observe the underlying non-determinism directly:

1. Start a local test stone (`npm run test:server:start`, any matrix version).
2. Run `client/src/queries/__tests__/sunitRunNb.integration.test.ts`.
3. Instrument the run: start `runTestMethodNb` against a long-running test method, wait 250 ms, send a single soft break (the first `cancel()`), and measure how long until the returned promise settles.
4. Repeat. The latency is bimodal — roughly 80 ms or roughly 1.9 s. The test only passes when the slow mode occurs.

## Expected behavior

The test either deterministically exercises soft-break-then-hard-break escalation, or it does not assert on escalation at all.

## Actual behavior

When soft-break delivery lands in the fast (~80 ms) mode, the run settles before the test's second `cancel()` is issued. `requestCancel` is a documented no-op once the call has settled (`client/src/nbRunner.ts:222-225`), so no hard break is ever sent and the promise resolves instead of rejecting:

```
AssertionError: promise resolved "{ className: 'JasperProbeTest', …(4) }" instead of rejecting
- Error { "message": "rejected promise" }
+ { "className": "JasperProbeTest",
+   "durationMs": 269,
+   "message": "Break: a Break occurred (error 6003), A soft break was received.",
+   "selector": "testRunsLongEnoughToStop",
+   "status": "error" }
 ❯ src/queries/__tests__/sunitRunNb.integration.test.ts:190:21
```

`durationMs: 269` against the test's 250 ms pre-break wait means the gem stopped roughly 19 ms after the soft break was sent.

### 1. Jasper's own runner traps the soft break

`client/src/queries/runTestMethod.ts:36` wraps the test invocation:

```smalltalk
[testCase perform: #'<selector>'] on: AbstractException do: [:e | captured := e].
```

Queried against a live 3.6.2 stone, `Break allSuperclasses` returns `(Object, AbstractException, Exception, ControlInterrupt)` — so **`Break` is an `AbstractException`** and this handler catches it. The handler does not resume; it terminates the protected block and records the exception, which the query then reports as `status: 'error'`. So a soft break that lands produces a perfectly ordinary resolved result, settling the promise and disarming escalation. This is the mechanism behind the failure output above.

### 2. Soft-break delivery latency is bimodal and non-deterministic

Measured on the real `runTestMethodNb` path (see Experiment 2). Both modes were reproduced on both versions on the same machine, with near-identical timings — so this is not a version or platform boundary, it is a race.

### 3. The test's premise is unconstructible

To make "the run ignores a soft break" true, one of these would have to hold:

- **The gem ignores the soft break.** The only way to do that from Smalltalk is to trap and resume it — but that swallows the **hard** break too (Experiment 1), leaving the session wedged with an uncollected Nb result for the rest of the file.
- **Jasper's runner does not trap the soft break.** That is production behavior (`runTestMethod.ts:36`), not something a test can arrange.
- **Delivery is reliably slower than the test's window.** It is not (Experiment 2).

There is no fourth option, so no amount of sleep tuning or timeout inflation makes this test deterministic.

### Contributing factors in the test as written

- **The fixture contradicts the test.** `installSlowTest` (`sunitRunNb.integration.test.ts:79-86`) is shared by two tests with opposite needs — the test at line 130 needs the soft break to land, this one needs it not to. Its comment (lines 33-36) states the fixture "*Sends a message every iteration, so the gem reaches a safe point and a soft break lands*", which is the opposite of this test's premise.
- **The 500 ms sleep works around a hazard production already handles, and widens the race.** The comment at lines 184-186 says *"The runner refuses to send it back-to-back with the soft one … so wait past its safety gap before asking again."* The runner does not refuse — it **defers**, via `setTimeout(sendHardBreak, MIN_HARD_BREAK_GAP_MS - waited)` at `nbRunner.ts:239-247`. This was already true in the commit that introduced the test. Callers are meant to issue `cancel(); cancel();` back-to-back; sleeping instead hands the soft break an extra 500 ms to settle the promise first.
- **Both waits are sleep-as-synchronization.** The 250 ms stands in for "the gem has picked the call up"; the 500 ms stands in for a constraint that does not exist.

### Documentation checked

- `GS64-GemBuilderforC-3.6.pdf` — p.30 ("Interrupting GemStone Execution"), p.212 (`GciHardBreak`), pp.364-365 (`GciSoftBreak`)
- `GS64-GemBuilderforC-3.7.pdf` — p.30, p.110 (`GciCallInProgress`), p.215 (`GciHardBreak`), p.365 (`GciSoftBreak`)

Findings:

- The break sections are **identically worded between 3.6 and 3.7** — the manuals describe no behavioral change across the 3.6/3.7 boundary.
- **Neither manual documents `GciTsBreak`**, the thread-safe call Jasper actually uses (`client/src/gciLibrary.ts:1629`). Both document only the legacy `GciSoftBreak` / `GciHardBreak`. The documented semantics were therefore never authoritative for this code path.
- Documented claims: a soft break "interrupts only the Smalltalk virtual machine (if it is running)" and replies `RT_ERR_SOFT_BREAK`; a hard break "interrupts the Gem process itself, and **is not trappable through Smalltalk exception handlers**", replies `RT_ERR_HARD_BREAK`, is described as aborting the current transaction, and terminates the whole session if the gem is inside certain `Repository` methods. Both: "If GemStone is executing when it receives the break, it replies with an error message. Otherwise, it ignores the break."
- `GciCallInProgress` is documented as "intended for use within signal handlers" and "considers the highest (most recent) call context only" — i.e. it describes the blocking API, not an outstanding non-blocking call. This matters for the latent defect below.

### Live class model (queried on a 3.6.2 stone)

- `Break allSuperclasses` → `(Object, AbstractException, Exception, ControlInterrupt)`
- `Break new gsNumber` → `2706`; `Break new isResumable` → `true`
- `Break>>initialize` → `gsNumber := ERR_Break. gsResumable := true. gsTrappable := true.`
- `Break subclasses` → empty. `ControlInterrupt subclasses` → `Halt, Break, Breakpoint, ClientForwarderSend, TerminateProcess`.
- **There is exactly one `Break` class with one error number for both soft and hard breaks** — nothing Smalltalk-visible distinguishes them, so an exception handler cannot filter one from the other by class or number.
- No Smalltalk-side API exists to send a break (the only `System` selector matching "break" is `_breakSerialization`), so breaks can only be driven from a GCI client.

### Experiment 1 — raw-GCI break matrix, 3.6.2 vs 3.7.5

Method: a throwaway integration test using **raw `GciTs*` calls only** — `GciTsNbExecute`, `GciTsBreak`, and `pollNbResultReady` — deliberately bypassing `nbRunner` so its background drain could not collect a result out from under the measurement. The result was never collected until the end, so a "ready" poll stays ready. Three workload shapes × plain vs. `[...] on: Break do: [:ex | ex resume]`, on a local 3.6.2 stone and a local 3.7.5 stone.

| workload / variant | 3.6.2 | 3.7.5 |
|---|---|---|
| `tight` / plain — soft break | settled 1500 ms, `err=6003 "A soft break was received."` | settled 50 ms, same `err=6003` |
| `tight` / resuming — soft break | not honored within 4000 ms | not honored within 4000 ms |
| `tight` / resuming — **hard break** | **not honored within 15000 ms**; ran to natural completion (~78.9 s), `err=0` | **not honored within 15000 ms**; ran to natural completion (~73.8 s), `err=0` |
| `sleep` / plain — soft break | settled 1500 ms, `err=6003` | settled 300 ms, `err=6003` |
| `sleep` / resuming — **hard break** | **not honored within 15000 ms**; natural completion (~68.5 s) | **not honored within 15000 ms**; natural completion (~72.5 s) |

(`tight` = `1 to: 40 do: [:s | 1 to: 177000000 do: [:k | n := n max: k]]`, the shape the current fixture uses. `sleep` = `40 timesRepeat: [System sleep: 1]`, the shape GemTalk's own `GciSoftBreak` example uses. A third `sends` shape completed before the break was sent and yielded no information.)

Conclusions:

- **The mechanism is qualitatively identical on 3.6.2 and 3.7.5** in every cell — same error number, same trappability, same outcomes.
- **The documented "a hard break is not trappable through Smalltalk exception handlers" does not hold for `GciTsBreak` on either version.** `on: Break do: [:ex | ex resume]` swallowed the hard break, and the workload ran to completion. This is the finding that makes the test's premise unconstructible.
- The documented transaction-abort behavior was **not exercised** and remains unverified: in every cell where a hard break was sent it was swallowed, and in the cells where the soft break sufficed no hard break was sent. `System transactionLevel` read `1` throughout, but no effective hard break ever occurred, so this says nothing about the claim.

### Experiment 2 — soft-break settle latency on the real path

Method: the real `runTestMethodNb` path against the existing `testRunsLongEnoughToStop` fixture. Wait 250 ms (exactly what the test does), send **one** soft break, then poll until the returned promise settles. n=5 per version.

| | 3.6.2 | 3.7.5 |
|---|---|---|
| fast mode | **78 ms** | **75 ms, 77 ms, 78 ms** |
| slow mode | **1910, 1912, 1916, 1918 ms** | **1907 ms, 1909 ms** |
| under the test's 500 ms window | 1 of 5 | 3 of 5 |
| gem-reported `durationMs` | 250 (fast) / 1989-1999 (slow) | 250-252 (fast) / 1997 (slow) |

- **The two modes are numerically identical across versions** (~78 ms and ~1910 ms). Only the observed frequency differed, and at n=5 that difference is not meaningful.
- The fast mode matches the CI failure exactly (`durationMs: 269`), so the Windows leg simply landed in the fast mode.
- The ~2 s slow mode suggests a periodic break check in the gem, but the internals were not investigated further; the observable is bimodal and non-deterministic.

### Experiment 3 — end-to-end run outcomes

- **CI, commit `f4da4f94`:** all five `windows-latest` jobs (3.6.2, 3.6.8, 3.7.2, 3.7.4.3, 3.7.5) failed on this test; all five `ubuntu-latest` jobs passed. Re-running the **identical commit** turned all five Windows jobs green.
- **Local, 3.7.5:** the real test file passed 11/11 runs, even though Experiment 2 showed 3 of 5 soft breaks beating the window on the same stone.
- **Local, 3.6.2:** the real test file passed on every run.
- **CI topology note:** per `.github/actions/setup-gemstone-in-windows/action.yml`, the Windows leg runs a **native Windows GCI client against a stone inside a WSL guest**, whereas the Ubuntu leg has client and stone on one host. This differs from both local setups and plausibly biases delivery toward the fast mode, but Experiment 2 shows both modes occur locally on both versions, so topology is an aggravating factor rather than the cause.
- **Prior CI history:** this test has not failed in CI before, on any `main`-branch run since it was introduced. Two earlier Windows failures on `main` were unrelated (`lintSupplyChain.test.ts` on 2026-08-27, `debuggerPanel.test.ts` on 2026-08-31); this file passed in both.

### Provenance

The test and both of its sleeps were introduced in commit `4b14434b` (2026-08-20), merged via #474. `MIN_HARD_BREAK_GAP_MS` and its deferral were introduced in `aa1a9eda` the same day, i.e. the deferral already existed when the 500 ms sleep and its "the runner refuses" comment were written.

### `waitUntilIdle()` is a no-op

`sunitRunNb.integration.test.ts:71-77` loops on `gci.GciTsCallInProgress(handle).result === 0`. Measured live mid-run: `GciTsCallInProgress` returns `0` ("idle") while `pollNbResultReady` returns `0` ("started, not ready") — i.e. while an Nb call is demonstrably outstanding. The loop therefore returns on its first iteration, every time, and the cascade it documents preventing ("*a test that ends while the drain is still going hands back a session that refuses the abort, and every later test in the file is skipped*") is unprotected. That exact cascade occurred during this investigation: `session has a GciTsNb operation in progress` from the teardown commit-guard check, followed by `skipping this test — a previous test's cleanup failed` for the rest of the file.

`nbRunner.ts:349-352` already documents this precisely: *"GciTsCallInProgress is no help here: it can answer 'idle' while the abandoned Nb result is still uncollected."*

The same call backs the "Session is busy with another operation" guard in `browserQueries.ts:349-354` (and `:314-317`), which by the same measurement never fires for an outstanding Nb call. The real refusal comes from `GciTsNbExecute` itself. Worth a separate look.

### The `NbCancelledError` assertion is vacuous at the integration level

`sendHardBreak` (`nbRunner.ts:206-219`) rejects with `NbCancelledError` unconditionally after calling `GciTsBreak`, regardless of whether the gem obeyed. During Experiment 1 the gem demonstrably kept running for over 70 s while the promise had already rejected with `NbCancelledError`. So `await expect(run).rejects.toBeInstanceOf(NbCancelledError)` proves that TypeScript set a variable, not that GemStone stopped — it cannot fail for the reason it appears to test.

Relatedly, the drain budget is `DRAIN_ATTEMPTS × DRAIN_INTERVAL_MS` = 40 × 50 ms = 2 s (`nbRunner.ts:94-95`), which was not enough to collect the result in any of the swallowed-hard-break cells.

## Potential fix

A possible direction, given that the premise cannot be constructed:

1. **Retire the escalation assertion from the integration suite.** `client/src/__tests__/nbRunner.test.ts:116` and `:167` already cover soft-break → hard-break → `NbCancelledError` deterministically, with fake timers and a stub GCI, including the `MIN_HARD_BREAK_GAP_MS` deferral. Escalation is pure TypeScript branching, so that is arguably the level it belongs at.

2. **Potentially merge the two live-gem tests into one that asserts the guarantee rather than the mechanism** — what only a real gem can prove, and which cannot flake by construction:

```ts
it('stops a running test however the gem takes the break, and leaves the session usable', async () => {
  installSlowTest();
  const breaks = vi.spyOn(gci, 'GciTsBreak');

  let cancel: (() => void) | undefined;
  const run = runTestMethodNb(session(), SUNIT_PROBE_TEST_CLASS, SLOW_SELECTOR, undefined,
    (c) => { cancel = c; });

  cancel!(); // soft
  cancel!(); // hard — deferred internally past MIN_HARD_BREAK_GAP_MS; never sleep for it

  // A break surfaces either as a rejection or as a non-passing verdict, depending
  // on which of the gem's two break-delivery modes serviced it (~80ms or ~1.9s,
  // measured on 3.6.2 and 3.7.5 alike). What must never happen is an interrupted
  // test reported as having passed.
  const outcome = await run.catch((e: unknown) => e);
  expect(outcome).not.toMatchObject({ status: 'passed' });
  expect(breaks.mock.calls.map((c) => c[1])).toEqual([false, true]);

  // The real assertion: the next run works, without the caller knowing a break
  // just happened. This also *is* the synchronisation — runNbCall waits on the
  // drain promise before starting.
  await expect(
    runTestMethodNb(session(), SUNIT_PROBE_TEST_CLASS, SUNIT_PROBE_PASSING_SELECTOR),
  ).resolves.toMatchObject({ status: 'passed' });
});
```

3. **Potentially drop `waitUntilIdle()`** rather than repair it — `runNbCall` already gates the next call on the `draining` promise (`nbRunner.ts:355-357`), so the follow-up run is itself the synchronization.

4. **Possibly export `MIN_HARD_BREAK_GAP_MS`** and derive the unit tests' `advanceTimersByTimeAsync` from it, instead of the magic `400` currently repeated in three places.

## Open questions for the maintainers

1. **Should a stopped test be reported as a test error?** Because `runTestMethod.ts:36` traps `Break` via `on: AbstractException`, pressing stop once shows the test **red** with `Break: a Break occurred (error 6003), A soft break was received.` — indistinguishable from the test genuinely erroring. Should `Break` be special-cased there so a stopped test reads as stopped/cancelled rather than failed?

2. **Should pressing stop twice always report "cancelled"?** Today the outcome depends on the bimodal race: if the soft break is serviced in the fast mode the run resolves as a red test error; in the slow mode it rejects with `NbCancelledError`. Same user gesture, two different outcomes. Recording the cancellation intent and honoring it on every settle path would make this deterministic for users, not just for CI — something like:

```ts
// in requestCancel, when escalating:
cancellationRequested = true;

// in doPoll, on pollResult === 1:
if (cancellationRequested) {
  void drainAbandonedCall(session);
  settle(() => reject(new NbCancelledError()));
  return;
}
```
   Is that the desired semantics, or is surfacing the gem's actual `Break` error preferable?

3. **Is the 2 s drain budget right?** 40 × 50 ms did not suffice in any cell where the gem kept running after a break. Should it be longer, or should the drain distinguish "gave up" from "collected" so callers can tell?

4. **Should `GciTsBreak`'s divergence from the documented `GciHardBreak` semantics be raised with GemStone engineering?** A hard break that *is* trappable by a Smalltalk exception handler contradicts both the 3.6 and 3.7 manuals, and Jasper's stop button relies on the hard break being the escalation of last resort. If a user's test method has a broad `on: AbstractException` / `on: Break` handler, stop may be unable to interrupt it at all — which would be a genuine product limitation worth documenting even if the GCI behavior is intended.

5. **Is the `GciTsCallInProgress`-based "Session is busy" guard in `browserQueries.ts` worth keeping?** By the measurement above it never fires for an outstanding Nb call. Should it be removed, or replaced with something that actually observes Nb state?

ericwinger:
Closes #540.

Thanks for the write-up on this one — the diagnosis in the issue is the fix. This just implements suggestions 2–4 from it.

## The problem

`escalates to a hard break when the run ignores a soft one` asserted a premise that cannot be constructed. `runTestMethod.ts:36` wraps the invocation in `on: AbstractException do:`, and `Break` **is** an `AbstractException` — so a soft break that lands is trapped, recorded, and reported as an ordinary `status: 'error'` result. The call settles, and `requestCancel` is a documented no-op once it has. The second press then had nothing to escalate.

Since soft-break delivery is bimodal (~80 ms or ~1.9 s, on 3.6.2 and 3.7.5 alike), the run settled before the second press roughly as often as not, and the test passed or failed on which mode happened to occur. Hence all five `windows-latest` jobs red on one run and all five green on a re-run of the same commit.

## What changed

The two live-gem tests now assert only what a real gem can prove and what cannot flake by construction:

- ~~a stopped test is **never** reported as `passed`;~~ **Corrected after review:** a stopped run settles as one of its two legal endings — cancelled outright (`NbCancelledError`), or a `Break` the gem trapped and answered as an `error` — asserted as that positive set. `not.toBe('passed')` let a poll failure, a vanished session, or a follow-up call refused as busy read as a clean stop, which is the opposite of what these tests are for;
- a second press never puts the two breaks back-to-back — the sequence that used to fault the client process outright, and the reason `MIN_HARD_BREAK_GAP_MS` exists. A live GCI is required to show the deferral holds, since a stubbed one cannot fault;
- the session comes back usable afterwards.

Deliberately **not** asserted: that a hard break is sent at all. Whether it is depends on whether the gem services the soft break inside the gap or after it, which the client does not decide. Both endings are named as legal (`['soft', 'soft,hard']`), so the delivery mode can no longer decide pass/fail.

`waitUntilIdle()` is dropped rather than repaired — `runNbCall` already gates a new call on the drain promise, so the follow-up run *is* the synchronisation.

`MIN_HARD_BREAK_GAP_MS` is exported, so the unit tests derive their fake-timer wait from it instead of repeating a magic `400` in three places. Its value is unchanged at 300, and no production behaviour changes in this PR.

## Verification

- Full suite green: 452 client + 12 server + 3 mcp-server files, 7294 passed / 114 skipped.
- The integration file run 13× against a live 3.6.2 stone, 7/7 each time. Vitest reshuffles test order per run, so the ordering varied.
- Both delivery modes exercised. This machine hits the slow mode essentially every time, so the fast ~80 ms branch — the one that broke CI — was forced separately by temporarily widening the gap so the run always settles before the hard break comes due: green there too, with a probe confirming it took that branch. That instrumentation is not in this diff.
- No runtime cost. Measured per test on a live stone, old vs. new: 2159 → 2156 ms and 1820 → 1971 ms; whole file 4057 → 4198 ms. The delta is noise — break latency alone swings ~1.8 s between modes. The fixture's `1 to: 40 do:` is a hang ceiling, not a cost; it is only reached if a break never lands, which is a failure.

**Caveat worth stating plainly:** all local runs were Linux, one stone, one version. The failure you saw was Windows across the whole matrix, so CI is the verification that counts here, not the local repetition count.

## Left alone, deliberately

Open questions 1–5 in the issue are all real and none are addressed here — this PR only stops the test lying. Flagging the two I think are worth their own issues:

- **Q1/Q2 — semantics.** Pressing stop once shows the test **red** with `Break: a Break occurred (error 6003)`, indistinguishable from the test genuinely erroring; pressing twice gives a red error or an `NbCancelledError` depending on the same bimodal race. Same gesture, two outcomes, neither reading as "cancelled". Making that deterministic is a product change and wants its own decision.
- **Q3 — the drain budget.** `40 × 50 ms` = 2 s, and the drain cannot distinguish "collected" from "gave up". If the gem keeps running past a break, the drain resolves anyway and the *next* call is refused as busy. That is the most plausible remaining way these tests could go red, and it is a production fix rather than a test one.

## Review follow-up

Three findings from @MatiasFernandez, all fixed in `5491a9e3`.

**Any rejection read as a clean stop.** `settledStatus` mapped every `Error` to `'stopped'`, so paired
with `.not.toBe('passed')` the tests really only asserted "did not resolve as passed" — the strikethrough
above. Only `NbCancelledError` counts as a stop now; every other rejection is rethrown so it surfaces as
itself, and both sites assert `['stopped', 'error']` rather than everything-but-one.

**Nothing asserted the hard break is _withheld_.** Correct, and worth more than the finding's tone
suggests: all three unit sites advance *past* the safety gap before checking, and the integration test
deliberately does not assert a hard break at all — so dropping the deferral and sending the hard break
immediately passed the entire suite while faulting the client process on a user's machine. Both
cancellation tests now stop the fake clock just short of the gap and assert no hard break has gone out.
Added to **both** paths rather than one, since the notification's canceller and the external one handed
out via `onStart` are wired separately and a change could touch only one.

Confirmed by making exactly that mutation — `setTimeout(sendHardBreak, …)` → `sendHardBreak()` — which
now fails on the new line (`expected "vi.fn()" to not be called with arguments: [ { h: 1 }, true ]`) and
passed everything before it.

**`Promise<unknown>` forcing the cast.** `startSlowRun` and `settledStatus` now carry
`ReturnType<typeof runTestMethodNb>`, so the `(outcome as { status: string })` cast is gone and a change
to the result shape is a compile error instead of a test that quietly stops checking.

Also swept: the CHANGELOG entry restated the assertion that was narrowed here, so it was stale the moment
this landed; rewritten to the two legal endings and the new withholding check.

**Re-verified after the follow-up:** full suite green (452 client + 12 server + 3 mcp-server files, 7294
passed / 114 skipped on the client), the integration file 7× against a live 3.6.2 stone at 7/7 each, and
lint / format:check / compile clean. Same caveat as above stands — all local, Linux, one stone version.


client/src/queries/__tests__/sunitRunNb.integration.test.ts
```diff
@@ -85,6 +70,54 @@ describe('SUnit non-blocking runs (integration)', () => {
...
+   * non-passing verdict, depending on which of the gem's two break-delivery
+   * modes serviced it — both are correct, and which one happens is not the
+   * test's business.
+   */
+  async function settledStatus(run: Promise<unknown>): Promise<string> {
+    const outcome = await run.then(
+      (r) => r,
+      (e: unknown) => e,
+    );
+    return outcome instanceof Error ? 'stopped' : (outcome as { status: string }).status;
```
MatiasFernandez: 🧪  `outcome instanceof Error` maps *any* rejection to `'stopped'`, so with `.not.toBe('passed')` below, both tests now really only assert "the run didn't resolve as passed". A poll error, the `session ended while this call was still running` reject, or a next-call-refused-as-busy would all read here as a clean stop — and that last one is exactly the failure mode Q3 in the issue calls the most plausible remaining way these tests go red. Before this PR the second test at least pinned `NbCancelledError`.

Should we narrow it — keep `'stopped'` for `NbCancelledError` only and let anything else surface as itself — and then assert the positive set (`expect(['stopped', 'error']).toContain(...)`) instead of the negative one? Both legal endings are already named in the PR description, so we'd be asserting the same thing we documented rather than everything-except-one.

client/src/__tests__/nbRunner.test.ts
```diff
@@ -150,7 +158,7 @@ describe('runNbCall — cancellation', () => {
       );
 
       cancelHandler!(); // second cancel → hard break, once the safety gap has passed
-      await vi.advanceTimersByTimeAsync(400);
+      await vi.advanceTimersByTimeAsync(PAST_HARD_BREAK_GAP_MS);
```
MatiasFernandez: 💭 All three unit sites advance *past* the gap and then assert the hard break went out, and the integration test now deliberately doesn't assert a hard break at all — so as far as I can tell nothing in the suite asserts the break is **withheld**. If someone dropped the deferral and sent the hard break immediately in `requestCancel`, every one of these still passes, and the only thing that notices is a client process fault on a user's machine.

The fake clock makes it nearly free here — advance `MIN_HARD_BREAK_GAP_MS - 50`, `expect(session.gci.GciTsBreak).not.toHaveBeenCalledWith(session.handle, true)`, then advance the rest. Worth adding to at least one of the two cancellation tests?

client/src/queries/__tests__/sunitRunNb.integration.test.ts
```diff
@@ -85,6 +70,54 @@ describe('SUnit non-blocking runs (integration)', () => {
...
+   *
+   * - `vi.waitFor` on the canceller, because a run that follows a break waits for
+   *   the abandoned call to be drained before it starts, so `onStart` is not
+   *   always synchronous.
+   * - the short sleep, because a break that arrives before the gem has picked the
+   *   call up is documented as ignored. 250ms is far more than it needs: the gem
+   *   reports its own elapsed time as ~250ms when a break lands right after this
+   *   wait, i.e. it started executing essentially at once.
+   */
+  async function startSlowRun(): Promise<{ run: Promise<unknown>; cancel: () => void }> {
```
MatiasFernandez: nit: `run: Promise<unknown>` throws away the result type, which is what forces the `(outcome as { status: string })` cast in `settledStatus`. Could this be `ReturnType<typeof runTestMethodNb>` so the cast goes away and a future change to the result shape is caught by the compiler?

MatiasFernandez (approved):

client/src/queries/__tests__/sunitRunNb.integration.test.ts
ericwinger: Narrowed, and asserting the positive set — both in `5491a9e3`.

`settledStatus` now returns `'stopped'` only for `NbCancelledError`; every other rejection is rethrown so it surfaces as itself rather than being counted as a clean stop. Both sites then assert `expect(STOPPED_ENDINGS).toContain(...)` with `STOPPED_ENDINGS = ['stopped', 'error']`, so the same two endings the description names are what the tests actually check — and a surprise verdict now fails instead of quietly passing for not being `passed`.

You are right that the negative form gave up ground the old test held. Worth noting the busy case specifically: `session is busy` came back as a plain `Error`, so under the old mapping the exact Q3 failure mode would have read as a successful stop in a green run. That is a worse outcome than the flake this PR set out to fix.

client/src/__tests__/nbRunner.test.ts
ericwinger: Added, on both cancellation tests — `5491a9e3`.

Each now advances `MIN_HARD_BREAK_GAP_MS - 50`, asserts `not.toHaveBeenCalledWith(session.handle, true)`, and only then advances past the gap. Put on both paths rather than one, since the notification's canceller and the external one handed out via `onStart` are wired separately today — a change could touch just one of them.

I checked your claim by making it true: dropping the deferral in `requestCancel` (`setTimeout(sendHardBreak, MIN_HARD_BREAK_GAP_MS - waited)` → `sendHardBreak()`) passed the whole suite before this, and now fails on the new line with `expected "vi.fn()" to not be called with arguments: [ { h: 1 }, true ]`. So the gap the deferral exists to keep had no test of its own, in a PR whose subject is that these tests were asserting the wrong thing. Good catch.

client/src/queries/__tests__/sunitRunNb.integration.test.ts
ericwinger: Done in `5491a9e3` — `startSlowRun` returns `{ run: ReturnType<typeof runTestMethodNb>; cancel: () => void }` and `settledStatus` takes the same, so the `(outcome as { status: string })` cast is gone.

It also pulled its weight immediately: with the real type in place, narrowing the rejection handling for your first comment was a compile-checked change rather than a hopeful one.

PR #551, opened 2026-09-03
Move the release instructions to docs/, and correct them

jgfoster:
Publishing is a maintainer task — it needs a personal access token for the `gemtalksystems` publisher on each of the two registries — so the procedure is of no use to the public contributors `CONTRIBUTING.md` is written for. This moves it to `docs/how-to/publishing-a-release.md`, per that directory's [Diátaxis](https://diataxis.fr/) layout, and leaves a short pointer behind for the maintainer who looks in the old place.

It also carries the corrections and additions from cutting 1.8.11 through 1.8.14 in one session. The mechanics were right — steps 1, 3, 4, 5 and 7 behaved as written — but several things were wrong or missing.

## Corrections

- **The changelog sweep now uses `git log --first-parent`**, one line per landing on `main`. Filtering for merge commits sees only PRs that landed as merge commits: a squash-merged PR and anything committed straight to `main` are both invisible to it, and those are the changes least likely to have written their own entry. Measured over `v1.8.14..main`, `--first-parent` misses none of the 26 the merge filter found and adds the 2 direct-to-`main` commits it did not.
- **A changelog entry can land in an already-released section.** A branch cut before the previous `[Unreleased]` → `[X.Y.Z]` rename merges its bullet into that released section, claiming it shipped in a version that never contained it. #442 did exactly this into `[1.8.11]`; caught and moved during 1.8.13.
- **Step 2 was narrower than the gate every PR is held to** — `compile && test` rather than `lint && format:check && compile && test`.
- **The double-publish warning had the wrong reason.** It said re-running `npm run publish` makes `ovsx` "double-publish and fail with 'already exists.'" Re-running it twice (1.8.11 and 1.8.13) showed `ovsx` refuses a duplicate rather than creating one, reporting `already published, but currently isn't active and therefore not visible`. Re-running only the `vsce` half still stands as the advice; the stated reason did not.
- **Step 5's `.vsix` is not the artifact that gets uploaded.** Both publish commands repackage into their own temp file — `ovsx publish` self-describes as "packaging it first if necessary." Reading 5 → 6 in sequence implies otherwise.
- **`npm run package` and `npm run publish:vsce`** replace the spelled-out `npx` invocations, so the instructions cannot drift from the scripts.

## Additions

- **A new "A success message is not a live release" section.** Both CLIs print success when the *upload* is accepted; the version then takes ~2–22 minutes to become publicly queryable, and either registry can lead. Open VSX's `already published, but currently isn't active` means wait, not retry — it resolved on its own both times. This cost a misdiagnosis during 1.8.11, where the `ovsx` step was wrongly assumed not to have run. Includes commands to verify against each registry's API rather than trusting CLI output.
- **`ovsx login`**, previously undocumented — the Open VSX half can be stored the same way `vsce login` stores the Marketplace half, so neither publish step needs anything in the environment.
- **`verify-pat` as a pre-flight** for both registries, at the top of the document.
- **The Azure DevOps PAT requirements**: `All accessible organizations` (a single-org token fails with a 401 that reads like a bad token) and scope `Marketplace → Manage`. This is what stalled the first release of the session. Also notes that a lost PAT is replaced rather than recovered, and that an empty `vsce ls-publishers` means no credentials *on this machine* rather than an unregistered publisher.

## Note on history

An earlier revision of this went in as 7dc5233c straight to `main`, skipping review, and was reverted in cc666da8. It was then re-opened as #516, which is where the `--first-parent` sweep, the `npm run package` / `npm run publish:vsce` consistency fix, and one removal came from: #516 also claimed `npm version` reformats `package.json` and gave a `git checkout package.json` recipe to undo it. That claim does not hold — `package.json` is in npm's own style on `main`, so `npm version` rewrites only the version line (verified: a 2-line diff), and all four of the `v1.8.11`–`v1.8.14` release commits touch exactly 2 lines of it. Only the durable part is kept: confirm with `git diff package.json`, since `format:check` globs no JSON. #516 and its branch have been deleted in favour of this PR.

Docs-only — no build, extension, or CI surface. `lint`, `format:check`, `compile` and `npm test` all pass (client 452 files / 7294 tests, server 322, mcp-server 92).

### DIFF
--- a/docs/how-to/publishing-a-release.md
+++ b/docs/how-to/publishing-a-release.md
@@ -1,0 +1,64 @@
+# Publishing a release
+
+Releasing Jasper to the marketplaces is a maintainer task: it needs a personal access token for the `gemtalksystems` publisher on each of the two registries. Outside contributors need none of it — [CONTRIBUTING.md](../../CONTRIBUTING.md) covers the ordinary build-and-test loop.
+
+## Pre-flight
+
+Verify both tokens resolve before you build anything — a missing or expired one is far cheaper to find now than after the commit and tag exist:
+
+```sh
+npx @vscode/vsce verify-pat gemtalksystems
+npx ovsx verify-pat gemtalksystems
+```
+
+## Steps
+
+1. `npm version <X.Y.Z> --no-git-tag-version` — bumps `package.json`'s `version` and `package-lock.json`'s two root fields (`version` and `packages."".version`) atomically. Don't hand-edit these or find-and-replace the version string across the lockfile: the version can collide with an unrelated dependency's own version elsewhere in `package-lock.json` (e.g. `1.8.11` matches `typed-rest-client@1.8.11`), corrupting that entry. `--no-git-tag-version` skips npm's own commit/tag, since steps 3-4 below handle that.
+
+   npm owns `package.json`'s formatting, and the file is currently in npm's own style, so this rewrites only the version line. Confirm that with `git diff package.json` before committing — `npm run format` globs only `*.{ts,mts,cts,js,mjs,cjs}`, so `format:check` will not catch stray churn here.
+
+   Then promote the `[Unreleased]` section in `CHANGELOG.md` to a new dated `[X.Y.Z]` heading. Sweep `main` since the last release for changes that didn't add their own changelog entries — `git log --oneline --first-parent vX.Y.Z..HEAD` lists them, one line per landing on `main`. Use `--first-parent` rather than filtering for merge commits: it also catches a squash-merged PR and anything committed straight to `main`, which are exactly the changes least likely to have written their own entry. **Check the previous release's section too:** a branch cut before the last `[Unreleased]` → `[X.Y.Z]` rename merges its bullet into that now-released section, so the entry claims to have shipped in a version that never contained it. Move any such entry into the new section.
+
+2. `npm run lint && npm run format:check && npm run compile && npm test` — the full gate, matching the one every PR is held to.
+3. Commit the version + changelog changes (e.g. `Release X.Y.Z: <one-line summary>`).
+4. `git tag -a vX.Y.Z -m "Release X.Y.Z"` — annotated tag, on the release commit.
+5. `npm run package` — runs `vsce package`, producing `gemstone-ide-X.Y.Z.vsix` in the repo root. The previous version's `.vsix` is gitignored but stays on disk; delete it to keep the root tidy. Note this file is **not** what gets uploaded: both commands in step 6 repackage into their own temp `.vsix`. It is for local inspection and archival.
+6. `npm run publish` — runs `vsce publish` then `ovsx publish` for the VS Code Marketplace and Open VSX. If `vsce publish` times out on the Azure DevOps Gallery API (it happens), re-run just that half with `npm run publish:vsce` rather than `npm run publish`, so the ovsx half doesn't run twice. If it does run twice it is harmless — `ovsx` refuses a duplicate rather than publishing one — but re-running the whole script obscures which half actually succeeded.
+7. `git push origin main && git push origin vX.Y.Z` — push the commit and the tag (the tag does not piggyback on the branch push).
+
+## A success message is not a live release
+
+Both CLIs print success as soon as the **upload** is accepted. The version then takes anywhere from ~2 to ~22 minutes to become publicly queryable, and the two registries are independent — either can be first. During that window:
+
+- Open VSX answers `Extension not found` for the new version, and `ovsx publish` run again reports `already published, but currently isn't active and therefore not visible`. That message means *wait*, not *retry* — it has always resolved on its own. It is also proof the upload landed.
+- The Marketplace omits the version from gallery queries, so `vsce show` still reports the previous one.
+
+Verify against each registry rather than trusting the CLI output:
+
+```sh
+# Open VSX
+curl -s https://open-vsx.org/api/gemtalksystems/gemstone-ide | jq -r .version
+
+# VS Code Marketplace
+npx @vscode/vsce show gemtalksystems.gemstone-ide
+```
+
+## Credentials
+
+Each registry takes a personal access token, tied to your own account rather than to the publisher. Store both once and neither publish step needs anything in the environment:
+
+```sh
+npx @vscode/vsce login gemtalksystems   # VS Code Marketplace
+npx ovsx login gemtalksystems           # Open VSX
+```
+
+`vsce` reads `VSCE_PAT` from the environment or its stored login; `ovsx` reads `OVSX_PAT` or its stored token. `npx @vscode/vsce ls-publishers` shows what is stored locally — an empty list means no credentials **on this machine**, not that the publisher is unregistered.
+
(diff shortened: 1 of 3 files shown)


docs/how-to/publishing-a-release.md
```diff
@@ -0,0 +1,64 @@
...
+
+2. `npm run lint && npm run format:check && npm run compile && npm test` — the full gate, matching the one every PR is held to.
+3. Commit the version + changelog changes (e.g. `Release X.Y.Z: <one-line summary>`).
+4. `git tag -a vX.Y.Z -m "Release X.Y.Z"` — annotated tag, on the release commit.
+5. `npm run package` — runs `vsce package`, producing `gemstone-ide-X.Y.Z.vsix` in the repo root. The previous version's `.vsix` is gitignored but stays on disk; delete it to keep the root tidy. Note this file is **not** what gets uploaded: both commands in step 6 repackage into their own temp `.vsix`. It is for local inspection and archival.
+6. `npm run publish` — runs `vsce publish` then `ovsx publish` for the VS Code Marketplace and Open VSX. If `vsce publish` times out on the Azure DevOps Gallery API (it happens), re-run just that half with `npm run publish:vsce` rather than `npm run publish`, so the ovsx half doesn't run twice. If it does run twice it is harmless — `ovsx` refuses a duplicate rather than publishing one — but re-running the whole script obscures which half actually succeeded.
```
MatiasFernandez: 🚫 `publish` is `publish:vsce && publish:ovsx`, so if `vsce publish` times out and exits non-zero, the ovsx half never ran at all — the "runs twice" case only happens when `vsce` exits 0. Followed as written (re-run `npm run publish:vsce`, stop), a timed-out release ends up live on the Marketplace and absent from Open VSX. And the section right below makes that easy to misread as the propagation delay rather than a publish that never happened.

Should the step say to run `npm run publish:ovsx` afterwards as well, or at least to check the Open VSX API before considering the release done?

docs/how-to/publishing-a-release.md
```diff
@@ -0,0 +1,64 @@
...
+
+   npm owns `package.json`'s formatting, and the file is currently in npm's own style, so this rewrites only the version line. Confirm that with `git diff package.json` before committing — `npm run format` globs only `*.{ts,mts,cts,js,mjs,cjs}`, so `format:check` will not catch stray churn here.
+
+   Then promote the `[Unreleased]` section in `CHANGELOG.md` to a new dated `[X.Y.Z]` heading. Sweep `main` since the last release for changes that didn't add their own changelog entries — `git log --oneline --first-parent vX.Y.Z..HEAD` lists them, one line per landing on `main`. Use `--first-parent` rather than filtering for merge commits: it also catches a squash-merged PR and anything committed straight to `main`, which are exactly the changes least likely to have written their own entry. **Check the previous release's section too:** a branch cut before the last `[Unreleased]` → `[X.Y.Z]` rename merges its bullet into that now-released section, so the entry claims to have shipped in a version that never contained it. Move any such entry into the new section.
```
MatiasFernandez: 💭 `vX.Y.Z` here is the *previous* release tag, but `X.Y.Z` a few lines up (and in steps 3, 4 and 7) is the version being cut — same placeholder, two different values inside one step. Worth writing it as `v<previous>..HEAD` or `<last-release-tag>..HEAD` so it can't be pasted with the new version by mistake? Getting it wrong here fails silently: `vNEW..HEAD` just lists nothing, which reads exactly like "every change already has a changelog entry".

docs/how-to/publishing-a-release.md
```diff
@@ -0,0 +1,64 @@
...
+
+2. `npm run lint && npm run format:check && npm run compile && npm test` — the full gate, matching the one every PR is held to.
```
MatiasFernandez: 💭 `npm test` includes integration tests that log into a live stone, and when none is reachable it fails with a connection error rather than skipping. Since this doc is the cold-start path for a maintainer who may not have one up, is it worth a parenthetical pointing at `npm run test:server:list` / `npm run test:server:start`? Otherwise the first failure of the release looks like a broken build rather than a missing stone.

docs/how-to/publishing-a-release.md
```diff
@@ -0,0 +1,64 @@
+# Publishing a release
+
+Releasing Jasper to the marketplaces is a maintainer task: it needs a personal access token for the `gemtalksystems` publisher on each of the two registries. Outside contributors need none of it — [CONTRIBUTING.md](../../CONTRIBUTING.md) covers the ordinary build-and-test loop.
+
+## Pre-flight
+
+Verify both tokens resolve before you build anything — a missing or expired one is far cheaper to find now than after the commit and tag exist:
+
+```sh
+npx @vscode/vsce verify-pat gemtalksystems
```
MatiasFernandez: 💭 Steps 5 and 6 were moved onto `npm run package` / `npm run publish:vsce` so the instructions can't drift from the scripts — should the pre-flight get the same treatment? There's no script for it today, but something like `"verify-pat": "vsce verify-pat && ovsx verify-pat"` would work (both read the publisher from `package.json` when no namespace is given) and would keep the last spelled-out `npx` invocations in step with the rest. Non-blocking.

MatiasFernandez (approved):

docs/how-to/publishing-a-release.md
jgfoster: You're right on both counts, and the second half of your point is the worse one: the advice as written doesn't just leave ambiguity, it actively misdirects. Fixed in 5dcb24c9.

On `;` vs `&&` — I'd say no, because it trades a visible ambiguity for a silent failure. `npm run` executes via `sh -c`, so the script's exit status becomes that of the *last* command:

```
$ sh -c "false && true"; echo $?
1
$ sh -c "false ; true";  echo $?
0
```

With `;`, a failed `vsce publish` followed by a successful `ovsx publish` exits 0 — `npm run publish` reports success for a release that never reached the Marketplace. That's harder to catch than the current failure, which at least exits non-zero.

So I kept `&&` and rewrote the advice to not depend on knowing which halves ran:

- Step 6 now states up front that the script is `publish:vsce && publish:ovsx` and that a non-zero `vsce publish` means `ovsx publish` never runs at all, leaving the release live on the Marketplace and absent from Open VSX.
- It says to ask the registries and publish whatever is missing, rather than inferring it, and that a release isn't done until both report the version.
- The propagation section now leads with the distinction that decides the action, since as you say the two are indistinguishable from outside: a registry not showing the version is propagation **only if the publish command reported success**; if it never ran or exited non-zero, nothing is propagating and that half needs publishing.
- And where it's genuinely unclear which happened, re-running a publish settles it safely — `vsce` reports `already exists`, `ovsx` reports `already published, but currently isn't active`, and either message is itself proof the upload had landed.

One thing I did not do, and would value your read on: adding `--skip-duplicate` to `publish:vsce` and `publish:ovsx` would make re-running the whole script safe and let the advice collapse to "just re-run it." I left it out because it isn't a complete shield — during the propagation window the duplicate check can't see the version it's meant to detect, so it can still fail with `already exists` — and because it would silently accept a re-publish of an already-shipped version. Happy to add it if you'd rather have the simpler instruction.

Worth noting the pipeline in #552 doesn't inherit this hazard: the two publishes are separate steps with `--skip-duplicate`, a `vsce` failure stops the run before anything is tagged, and re-running is safe by construction.

PR #552, opened 2026-09-03
Publish releases from a manual pipeline

jgfoster:
Follows the review suggestion on the release instructions: make publishing something a (manual) pipeline can run.

> **Stacked on #551.** Based on `docs/publishing-a-release` because it rewrites the document that PR adds, so it targets that branch rather than `main`. Merge #551 first and this retargets to `main` cleanly.

## The split

The version bump and the changelog stay human, as you guessed — promoting `[Unreleased]` to a dated section, and sweeping `main` for changes that never wrote an entry, are editorial calls. Everything after that is mechanical:

1. **Release PR** (human) — `npm version`, changelog promotion and sweep. Merges through the queue like anything else.
2. **`Release` workflow** (`workflow_dispatch`) — verify, package, publish, wait, tag, announce. Only ever publishes a commit already on `main` and already green.

Inputs are `version`, `ref` (default `main`), and `dry-run`, which runs every check and builds the `.vsix` without touching either registry.

## What it refuses to do

Pre-flight, before any credential is touched or any reviewer is asked to approve: the ref must be an ancestor of `main`; `package.json` must be at the requested version; `CHANGELOG.md` must have a dated `## [X.Y.Z] - YYYY-MM-DD` section; no `vX.Y.Z` tag may exist; and the **newest** `ci-complete` check run on that exact commit must have concluded `success`.

The release deliberately does not re-run `lint`/`compile`/`test`. The merge queue already ran them on that commit across both platforms and every GemStone version in `.gemstone-integration-releases.json`, plus the Node floor, and ran `npm run package` — a stronger gate than one run here could be. What matters is that it passed, so the workflow checks that instead of re-deriving it.

## Two things the pipeline does that the manual process could not

**One artifact, everywhere.** The `.vsix` is built once and published as-is to both registries via `--packagePath`, then attached to the GitHub Release. The artifact you can download from the run, the two the registries serve, and the one on the Release are the same bytes. Run by hand, each publish command repackages from source, so the local `.vsix` is *not* what gets uploaded.

**It waits.** Both CLIs print success when the *upload* is accepted, but the version takes ~2–22 minutes to become queryable and either registry can lead — this cost a misdiagnosis during 1.8.11. `scripts/await-published-version.sh` polls both and states plainly that a timeout is propagation, not a failed publish, so re-publishing is not the fix.

Tagging happens only after both registries are serving the version, so a failed or half-finished publish leaves no tag to clean up and the run can just be repeated; `--skip-duplicate` on both commands makes that safe.

`scripts/changelog-section.sh` is used twice — as the pre-flight guard and as the source of the Release notes — so the check and the published text cannot disagree about which section belongs to a version.

## Needs one-time setup before the first real run

The workflow is inert until this exists:

1. A **`release` environment** (Settings → Environments).
2. **Required reviewers** on it. This is what makes the approval step a real gate and where the audit trail of who released what comes from. Without it, anyone who can run a workflow can publish.
3. `VSCE_PAT` and `OVSX_PAT` as **environment** secrets, not repository secrets, so no other workflow can reach them.

Worth flagging explicitly, since it's a governance change rather than a mechanical one: **this repo currently stores no third-party secrets** — only the automatic `GITHUB_TOKEN`. These would be the first, and because Azure DevOps PATs are personal and expire (a year at most), they are one person's identity acting for the org and they expire silently. The workflow verifies both before building anything, so an expired token fails in seconds rather than halfway through, but they still need an owner who rotates them.

Also new: **the repo has no GitHub Releases today.** This starts creating them, with the changelog section as the notes and the `.vsix` attached. Easy to drop that step if you'd rather not adopt the convention.

Releasing by hand stays documented as a fallback — the pipeline is the normal path, not the only one.

## Verification

Checked against the same three gates CI applies, all clean on the new files: `scripts/lint-workflow-timeouts.sh`, `actionlint` (with `shellcheck` on every `run:` block and both new scripts), and `zizmor` — no `${{ }}` reaches a `run:` script, actions are pinned to SHAs, `persist-credentials: false` throughout (the tag is created through the API rather than pushed), and permissions are `contents: read` at the top with `contents: write` only on the publish job.

Both scripts were exercised against the live registries and the real changelog: `changelog-section.sh` extracts `1.8.14`, stops at the next heading, and rejects an unpromoted version; `await-published-version.sh` confirms `1.8.14` on both registries and produces the propagation guidance on timeout. The `ci-complete` selection was checked against a real commit carrying two runs.

`lint`, `format:check`, `compile` and `npm test` pass (client 452 files / 7294 tests, server 322, mcp-server 92).

The workflow itself cannot be end-to-end tested without publishing a version; `dry-run` exists so its first real exercise costs nothing.


.github/workflows/release.yml
```diff
@@ -0,0 +1,262 @@
...
+      # under discussion if a publish goes wrong.
+      - name: Upload the .vsix
+        # zizmor(unpinned-uses): actions must be pinned to a commit SHA,
+        # not a mutable tag like @v6
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: gemstone-ide-${{ inputs.version }}-vsix
+          path: gemstone-ide-${{ inputs.version }}.vsix
+          if-no-files-found: error
+      - name: Stop here (dry run)
```
MatiasFernandez: 💭 This step doesn't actually stop anything — it echoes, and what keeps a dry run from publishing is the `if: '!inputs.dry-run'` repeated on each of the five steps below it. That makes the safety property depend on every future step remembering the guard: someone adding a step here later gets a real publish on a dry run, and the step named "Stop here" reads like it already prevented that.

Would it be worth splitting the publish half into its own job with a single job-level `if: !inputs.dry-run`? The `.vsix` is already uploaded as an artifact just above, so the second job could `download-artifact` rather than repackage, and the guard would live in exactly one place. If that's more machinery than it's worth, even renaming this to something like "Dry run: nothing further will run" plus a comment on the guard convention would stop it from reading as a hard stop.

scripts/await-published-version.sh
```diff
@@ -0,0 +1,85 @@
...
+# version whose upload landed but is not yet active is absent from it — which
+# is exactly the "already published, but currently isn't active" state.
+openvsx_has_version() {
+    curl -sf --max-time 30 "https://open-vsx.org/api/${namespace}/${extension}" 2>/dev/null |
+        jq -e --arg v "$version" '.allVersions | has($v)' >/dev/null 2>&1
+}
+
+marketplace_has_version() {
+    npx --no-install @vscode/vsce show "${namespace}.${extension}" --json 2>/dev/null |
+        jq -e --arg v "$version" '[.versions[].version] | index($v) != null' >/dev/null 2>&1
```
MatiasFernandez: 💭 Both probes collapse "the registry doesn't have this version yet" and "the query itself failed" into the same `false`. `curl -sf` swallows any HTTP error, `2>/dev/null` hides the CLI's own complaint, and `jq -e` returns non-zero for a shape it can't parse — so a DNS blip, a 500, an expired-token path in `vsce show`, or a future change to its `--json` output all look identical to slow propagation.

The consequence is in the timeout message: after 30 minutes of a broken query it tells the operator "The upload was already accepted, so this is propagation, not a failed publish" — confidently, on no evidence. Would it be worth distinguishing the two, even minimally? E.g. capture the response, and if the request failed rather than merely missed the version, print that on the `...still waiting` line so a persistent query error is visible in the log instead of reading as a slow registry.

.github/workflows/release.yml
```diff
@@ -0,0 +1,262 @@
...
+          if [ "$conclusion" != "success" ]; then
+            echo "error: ci-complete concluded '$conclusion' for $SHA, not success." >&2
+            exit 1
+          fi
+          echo "ci-complete passed on $SHA."
+
+  publish:
+    name: Publish to both registries
+    needs: preflight
+    timeout-minutes: 45
```
MatiasFernandez: 💭 Does the 45-minute budget leave enough room for the wait step? `await-published-version.sh` defaults to `TIMEOUT_SECONDS=1800`, so 30 of those 45 minutes can go to polling, leaving ~15 for checkout, `npm ci`, token verification, `npm run package` (which runs `bundle` through `vscode:prepublish`) and both publishes.

The reason it matters is which failure you get. If the job is cancelled by this timeout, it dies after both registries have been published to and before the tag and Release exist — and the operator sees a bare cancellation rather than the script's carefully-worded propagation message. Given the doc puts observed propagation at up to ~22 minutes, would it be better to raise this, or to lower `TIMEOUT_SECONDS` for the workflow's invocation so the script always gets to explain itself first?

.github/workflows/release.yml
```diff
@@ -0,0 +1,262 @@
...
+        env:
+          OVSX_PAT: ${{ secrets.OVSX_PAT }}
+        run: npx --no-install ovsx publish --packagePath "gemstone-ide-${VERSION}.vsix" --skip-duplicate
+      - name: Wait for both registries to serve the version
+        if: '!inputs.dry-run'
+        run: bash scripts/await-published-version.sh "$VERSION"
+      # Tagged only once both registries are actually serving the version, so
+      # a failed or half-finished publish leaves no tag to clean up and the
+      # run can simply be repeated. The tag still points at the vetted commit.
+      - name: Tag the release
```
MatiasFernandez: 💭 Tagging last is a nice property, but it leaves one narrow window that the workflow can't recover from itself: if the tag is created and `gh release create` then fails (transient API error, say), a re-run is refused by the "Require the tag to be free" pre-flight check, so the run can't finish the job it started. The recovery note in the doc covers the other case — tag still missing, so a re-run finishes cleanly — but not this one, where someone has to know to delete `vX.Y.Z` by hand first.

Worth a sentence in the doc's recovery paragraph? Or is it cheap enough to move the tag creation to after the Release is created, so the tag is genuinely the last thing that happens?

scripts/changelog-section.sh
```diff
@@ -0,0 +1,48 @@
...
+        if (v == want && $3 == "-") { inside = 1; next }
+    }
+    inside { print }
+' "$changelog")
+
+if [ -z "$section" ]; then
+    echo "error: $changelog has no dated '## [$version] - <date>' section." >&2
+    echo "Promote the [Unreleased] section to '## [$version] - $(date -u +%Y-%m-%d)' and merge that before releasing." >&2
+    exit 1
+fi
```
MatiasFernandez: 💭 Should this get a test? It's the same shape as `scripts/lint-supply-chain.mjs`, which is covered by [`client/src/__tests__/lintSupplyChain.test.ts`](client/src/__tests__/lintSupplyChain.test.ts) with black-box subprocess runs against throwaway fixtures — and for the same stated reason: it's a guard, so a regression that stops detecting what it claims to detect is invisible. Here it's worse than invisible, because the script has a second job: whatever the awk extracts becomes the published Release notes. A quiet drift in the parser either blocks a release or ships the wrong text under a version's name, and neither shows up until release day. The `CHANGELOG_PATH` override already makes it trivially testable against a temp file.

A couple of cases I'd want pinned down: `## [Unreleased]` never matching, the section stopping at the next `## [` heading, and the empty-section case — a dated `## [X.Y.Z] - YYYY-MM-DD` heading with no body currently trips the `-z` check and reports "has no dated section", which would send someone off to promote a section that is already promoted.

CONTRIBUTING.md
```diff
@@ -131,4 +131,4 @@ CI runs on **GitHub Actions**. The [Health Check workflow](.github/workflows/hea
 
 ## Publishing a release
 
-Publishing to the VS Code Marketplace and Open VSX needs a personal access token for the `gemtalksystems` publisher on each registry, so it falls to a maintainer rather than to contributors. The procedure — pre-flight token checks, the version-bump and changelog sweep, packaging, and what the registries do after a publish reports success — is in [docs/how-to/publishing-a-release.md](docs/how-to/publishing-a-release.md).
+Publishing to the VS Code Marketplace and Open VSX needs a personal access token for the `gemtalksystems` publisher on each registry, so it falls to a maintainer rather than to contributors. A release is a human-authored release PR (version bump and changelog) followed by the manual `Release` workflow, which tags, packages, publishes and announces it. Both halves are in [docs/how-to/publishing-a-release.md](docs/how-to/publishing-a-release.md).
```
MatiasFernandez: nit: "which tags, packages, publishes and announces it" has the order backwards, and the order is one of the design's deliberate safety properties — the workflow packages and publishes first and tags only once both registries are serving the version, precisely so a half-finished publish leaves no tag to clean up. "packages, publishes, tags and announces"?

docs/how-to/publishing-a-release.md
```diff
@@ -43,16 +68,25 @@ curl -s https://open-vsx.org/api/gemtalksystems/gemstone-ide | jq -r .version
...
+## Releasing by hand
+
+The pipeline is the normal path, not the only one. If it is unavailable, the same steps run locally, from a merged release commit, with credentials stored as below:
 
 ```sh
-npx @vscode/vsce login gemtalksystems   # VS Code Marketplace
-npx ovsx login gemtalksystems           # Open VSX
+npm run lint && npm run format:check && npm run compile && npm test
+npm run package                                    # gemstone-ide-X.Y.Z.vsix
+npm run publish:vsce                               # or: npm run publish
```
MatiasFernandez: nit: taken literally as a copy-paste block, the `# or: npm run publish` alternative followed by `npm run publish:ovsx` on the next line publishes to Open VSX twice — which is the exact thing the paragraph below warns against. Would it read better as just `npm run publish` here, with the split-halves advice left to the paragraph?

jgfoster:
**Allowlist hole found in review, and closed.**

The question was whether allowing `-----BEGIN PRIVATE KEY-----` allows a private key. It did. A real key split across string literals —

```js
const leaked = "-----BEGIN PRIVATE KEY-----\n" + "MIIEvQIBADANBgkq...";
```

— begins with exactly the prefix the delimiter allowlist matched, so it was suppressed. Reproduced against a planted key before the fix, and re-checked after.

RE2 has no lookahead, so "allow the header but not the contents" can't be one negative regex. Split into two concerns instead:

- Both allowlists now carry `targetRules`, so each can only suppress the rule whose false positive it exists for.
- New `gemstone-pem-key-material` matches a PEM header followed within a few characters by a 40+ character unbroken base64 run. **No allowlist targets it.** The false positives can't reach it — PEM-handling code has identifiers and dots between the markers, never a base64 run that long.

| Fixture | Before | After |
| --- | --- | --- |
| Real bundle false positives | not flagged ✅ | not flagged ✅ |
| Key inline in one literal | `private-key` ✅ | `private-key` + `gemstone-pem-key-material` ✅ |
| **Key split across literals** | **not flagged ❌** | `gemstone-pem-key-material` ✅ |
| Raw `.pem` file | ✅ | ✅ |
| Both historical `swordfish` forms | ✅ | ✅ |
| GitHub PAT | ✅ | ✅ |
| Real 1.9.0 package | clean ✅ | clean ✅ |

**And the limit that remains, now written into both the config and the how-to.** gitleaks matches text, not program semantics, so concatenation defeats every rule and allowlist here — `"-----BEGIN PRIVATE" + " KEY-----"` and `"sword" + "fish"` are both undetected, as is base64 key material with no header. Verified, not assumed.

That is the right trade for what this job is *for*: it predicts Open VSX's server-side verdict before an upload spends a version number, and Open VSX runs the same class of regex scanner — so a secret that evades these rules evades theirs too, and doesn't cause the rejection being guarded against. The scan is calibrated to an accidental commit, which is never obfuscated. Deliberate smuggling is answered by review, the merge queue and the release-approval gate. The config says explicitly not to chase concatenation patterns with looser regexes: each one costs false positives, and a scan that cries wolf is one somebody eventually turns off.

.github/workflows/release.yml
```diff
@@ -0,0 +1,500 @@
...
+  # built package and a clean scan already in hand to look at.
+  #
+  # Everything below this line is irreversible. The registries are immutable
+  # per (publisher, name, version), so a publish that goes wrong burns this
+  # version number for good — there is no retry, only the next version.
+  #
+  # This environment must hold NO secrets: the registry tokens live on the two
+  # publish environments instead, so approving a release does not hand the
+  # build a credential.
+  gate:
```
MatiasFernandez: 💭 `gate` is the only job below `scan` without `if: ${{ !inputs.dry-run }}` — `release`, both publish jobs and `verify` all carry it. So a dry run still parks at `release-approval` and pings a required reviewer, who then approves a run whose every remaining job skips.

The doc says "a first release of the day is worth doing as a dry run — it exercises validation, the build and the secret scan without touching a registry or creating a tag", and the PR description says it "costs nothing". Needing a human approval is the one thing that makes people stop reaching for it, which is a shame given `validate → package → scan` is exactly the part worth rehearsing.

Should `gate` take the same condition? Nothing after it runs in a dry run anyway, so skipping it changes no behaviour other than not waking a reviewer.

scripts/changelog-section.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+#
+# Prints the CHANGELOG.md body for one released version, and fails if that
+# version has no *dated* section yet.
+#
+# Used twice by the release workflow: as a pre-flight guard (an entry still
```
MatiasFernandez: 🚫 "Used twice by the release workflow … Kept as one script so the guard and the notes can never disagree" — but the guard doesn't call it. `validate` open-codes its own `grep -Eq "^## \[$esc\] - …"` plus an awk sweep of `[Unreleased]` (release.yml:141-168), and this script only ever runs in `release`. There are two independent implementations of "which text belongs to version X", and the invariant the header claims is precisely the one nothing enforces.

It already differs in a way that bites: `validate`'s grep is satisfied by a *heading*, while this script additionally requires a non-empty body. A dated `## [1.9.1] - 2026-09-11` with nothing under it passes validation, survives the approval gate, gets tagged — and then fails here, on the step right after the tag. That lands in the one row of the docs' failure table that needs manual cleanup ("Delete the tag if it was created, then re-dispatch"), for a mistake that was visible before anything was created.

`validate` checks nothing out on purpose, so it genuinely can't run this. But `package` does check out, and `release` runs it after tagging — would a `bash scripts/changelog-section.sh "$VERSION" >/dev/null` step in `package`, or simply ahead of "Tag the release", close it? Then the header is true and this whole class of failure happens while everything is still recoverable. If the duplication is deliberate instead, should the header say "kept in step with" rather than "can never disagree"?

scripts/publish-to-registry.sh
```diff
@@ -0,0 +1,95 @@
...
+        exit 2
+        ;;
+esac
+
+echo "Publishing $(basename "$vsix") to $registry..."
+
+# Capture rather than stream so the message can be classified; echo it back
+# either way, so the log still shows exactly what the CLI said.
+set +e
+output=$("$@" 2>&1)
```
MatiasFernandez: 💭 `output=$("$@" 2>&1)` holds everything until the command returns. The publish steps run under `timeout-minutes: 5` (release.yml:427, 471), so if the CLI hangs, GitHub kills the process and the log contains `Publishing gemstone-ide-X.Y.Z.vsix to marketplace…` and nothing else — no CLI output at all.

That is the failure the docs call out as the common one: "`vsce publish` does time out on the Azure DevOps Gallery API, and a timeout tells you nothing about whether the upload landed." It's also the moment when the CLI's own words are the only evidence of whether the upload went out, and a version number is riding on the answer. Losing the output in exactly that case seems like the wrong trade.

Could it tee rather than capture — `output=$("$@" 2>&1 | tee /dev/stderr)` — so classification still works but the log stays live? It needs `${PIPESTATUS[0]}` for the status, since the pipeline's exit code becomes `tee`'s. (The `printf` on line 68 could then go.)

scripts/publish-to-registry.sh
```diff
@@ -0,0 +1,95 @@
...
+printf '%s\n' "$output"
+
+if [ "$status" -eq 0 ]; then
+    echo "result: published"
+    exit 0
+fi
+
+# Order matters: the inactive message also contains "is already published",
+# so the more specific case is tested first.
+if printf '%s' "$output" | grep -qiF "isn't active and therefore not visible"; then
```
MatiasFernandez: 💭 The classification is the whole reason this script exists — the header says so, and it's the thing `--skip-duplicate` gets wrong. But nothing exercises it: these two greps and their ordering only ever run during a real release, which is the one place you can't afford to discover they're stale. If ovsx rewords `isn't active and therefore not visible`, the first signal is a red publish job on a version number that's already spent — and the messages have moved before (the `=` form failed 1.7.6, the `:` form 1.8.3 after their rule widened).

The repo already tests its scripts this way — `client/src/__tests__/lintSupplyChain.test.ts` drives `scripts/lint-supply-chain.mjs`. Is a small one worth it here? A stub `npx` on `PATH` that echoes each canned message and exits non-zero, asserting `published` / `awaiting-activation` / `already-published` / hard-fail, would also pin the ordering that the comment on line 75 depends on — which is currently a fact about the two regexes that only a careful reader notices.

scripts/publish-to-registry.sh
```diff
@@ -0,0 +1,95 @@
...
+# either way, so the log still shows exactly what the CLI said.
+set +e
+output=$("$@" 2>&1)
+status=$?
+set -e
+
+printf '%s\n' "$output"
+
+if [ "$status" -eq 0 ]; then
+    echo "result: published"
```
MatiasFernandez: nit: The header says the script "Prints one of `published`, `already-published`, `awaiting-activation` on success", but this one goes to stdout and the other two go to stderr (lines 78, 85). Anything reading the result off stdout only ever sees `published`. Worth putting all three `result:` lines on stdout and leaving just the explanatory prose on stderr?

.gitleaks.toml
```diff
@@ -0,0 +1,110 @@
...
+#
+# It mirrors client/src/__tests__/publishSecretScan.test.ts, which stays: that
+# test fails in seconds on a developer's machine and in every PR, long before
+# a release is dispatched. This is the backstop that also covers the bundle,
+# server/, mcp-server/ and resources/ — everything the .vsix actually ships
+# and the unit test never sees.
+[[rules]]
+id = "gemstone-password-literal"
+description = "A `password`-suffixed key assigned a short quoted literal — the form Open VSX's hashicorp-tf-password rule rejects"
+regex = '''(?i)password\w*\s*[:=]\s*["'][A-Za-z0-9]{7,20}["']'''
```
MatiasFernandez: 💭 This is character-for-character `secretAssignment` in `publishSecretScan.test.ts`, and the comment above frames that as the two mirroring each other — but they're two independent copies, and this is the pattern history says moves. Open VSX's rule already tightened once: the `=` form failed 1.7.6, then the `:` form failed 1.8.3 after they widened it. Next time it moves, whoever edits one has nothing pointing at the other, and the copy that silently goes stale is whichever they didn't touch. The unit test going stale costs a burned version number; this rule going stale costs the same.

Could the test read the pattern out of `.gitleaks.toml` rather than restating it, so there's one definition? `vscodeignoreShippedAssets.test.ts` already reads a root-level config file from `client/src/__tests__`, so the path walk exists. Then the "mirrors" claim is structural rather than a promise.

MatiasFernandez (approved):

scripts/changelog-section.sh
jgfoster: Right on all three counts, and I reproduced the third. With a fixture whose `## [1.9.1] - 2026-09-11` has an empty body:

```
validate heading grep : PASSES
validate [Unreleased] : PASSES (empty)
changelog-section.sh  : FAILS
```

Worth adding that it was one step worse than described: the script runs inside `Publish the Release`, which is *after* `Tag the release` — so the failure burned a reviewer's approval as well as leaving an orphan tag.

Taken the first option, in `package` rather than ahead of the tag. Ahead of the tag still spends the build, the scan and a human approval before failing; `package` is the earliest job with a checkout, so it fails while everything is recoverable, and the reviewer at the gate now knows the notes actually render — which is what the gate is for.

Kept `validate`'s grep. It is the checkout-free fast fail, and it is stricter in the other direction (this script accepts `## [1.9.1] -` with no date at all; validate requires a full ISO date). The two do diverge both ways — only one direction is harmful, and the guard closes that one.

Header rewritten to describe what actually happens, including why `validate` cannot run this.

scripts/publish-to-registry.sh
jgfoster: Agreed, and teed. One correction to the mechanics though: `${PIPESTATUS[0]}` does **not** work here.

The pipeline runs inside the command substitution's own subshell, so the parent shell only ever sees the status of the assignment itself:

```
$ x=$(false | tee /dev/stderr); echo "$? ${PIPESTATUS[0]}"
1 0
```

`$?` is already correct, because of the `pipefail` in the `set -euo pipefail` at the top — `set +e` turns off errexit but leaves pipefail alone. So it is just `output=$("$@" 2>&1 | tee /dev/stderr)` with the existing `status=$?`, and yes, the trailing `printf` is gone. I left a comment saying why PIPESTATUS is wrong here, since it is the obvious thing to reach for.

There is a regression test for the case you describe: the stub prints a line and then sleeps, the script is killed mid-run, and the assertion is that the line survived in the log. It fails against the old capture.

scripts/publish-to-registry.sh
jgfoster: Fixed — all three `result:` lines are on stdout now, and the explanatory prose stays on stderr.

Also moved the `Publishing ... to <registry>...` progress line to stderr, so stdout really is one line and the header can say so exactly: a caller can now read the outcome with `$(...)` and get the outcome, not the outcome plus a progress message. Pinned by a test over all three outcomes.

scripts/publish-to-registry.sh
jgfoster: Worth it, and added — `client/src/__tests__/publishToRegistry.test.ts`, built the way you describe: a stub `npx` on PATH plays back each canned message and exit status, and the assertions are on exit code, stdout and stderr. Fourteen cases covering `published` / `awaiting-activation` / `already-published` (both the ovsx and the vsce spelling) / hard-fail, plus the ordering pin — a separate case asserts the inactive message does *not* classify as `already-published`, since both are successes and a reordering would otherwise be silent.

It also pins which CLI each registry gets, that a missing .vsix or an unknown registry never reaches the CLI at all, and that the CLI's own exit status is propagated rather than flattened to 1.

Negative control: 8 of the 14 fail against the pre-change script.

Extended the same treatment to `changelog-section.sh` (`changelogSection.test.ts`), since the `package` guard added for your other comment now depends on its behaviour — in particular that a dated-but-empty section fails, which is the exact gap that comment found.

Both are POSIX-only via the existing `onSupportedPosixDescribe`, as the release workflow runs on `ubuntu-latest` and the stub relies on a shebang and a `:`-separated PATH.

scripts/publish-to-registry.sh
jgfoster: Follow-up: `tee /dev/stderr` broke the ubuntu legs, so the shape here has changed again — and my correction above about `${PIPESTATUS[0]}` is now only half right.

```
tee: /dev/stderr: No such device or address
```

`tee /dev/stderr` *opens* /dev/stderr, which on Linux is `/proc/self/fd/2` — and that cannot be opened when fd 2 is a socket rather than a pipe, which is what a parent capturing output hands it. macOS treats /dev/stderr as a dup of fd 2, so it works there; that is why it passed locally and on the macOS leg and failed on every ubuntu one.

Worse than a noisy log, too: tee died, `pipefail` took the status from tee, and the script then classified on that instead of the CLI's — a clean publish came back as a hard failure, and a CLI exit of 3 came back as 1.

Now:

```bash
log=$(mktemp)
trap 'rm -f "$log"' EXIT
set +e
"$@" 2>&1 | tee "$log" >&2
status=${PIPESTATUS[0]}
set -e
output=$(cat "$log")
```

`>&2` dups fd 2 rather than opening it, so it works whatever fd 2 is. And with the pipeline in this shell rather than inside a `$(...)`, `${PIPESTATUS[0]}` is now correct and is genuinely the better choice — it is the CLI's own status, not "whichever stage failed last".

Worth noting the tests you asked for in the other thread are what caught this: four of them failed on the ubuntu legs. Added a fifth asserting stderr carries no `tee:` error, so a regression of this shape says so rather than surfacing as four odd assertion diffs.

PR #555, opened 2026-09-03
File out and file in Topaz .gs files

### LINKED ISSUE #539: Need file out support
Gemstone Explorer needs file out support at the methods, class, and dictionary level. Class category & method protocol a nice to have.

ericwinger:
## What this does

Jasper can now write GemStone code out to a `.gs` file on your own machine, and read one back in. Neither half existed before: Jasper could compile a method and define a class, but it could not produce the Topaz file that GemStone code is shipped in, and it could not read one.

Both halves work on **your** machine, not the stone's host. File Out uses VS Code's save dialog, so the folder you pick is the one you're sitting at. File In reads the file here and runs it chunk by chunk over the connection, rather than handing a path to GemStone's `GsFileIn` — which reads a path on the *stone's* host and so cannot see a file on your laptop at all.

Running it on this side is also what makes the report possible. Every chunk's outcome is known, so a method GemStone refuses is named with its file and line **while the rest of the file still files in** — which is what you want when one method out of twenty is stale. Nothing is ever committed; the summary says so.

## Where to find it

| | |
|---|---|
| **File Out** | Right-click any GemStone Explorer row: dictionary, class category, class (Classes *and* Hierarchy panes), method category, or a multi-selection of methods |
| **File In** | Dictionaries pane toolbar and dictionary row menu; the ⤓ on a connected session in Logins & Sessions; Command Palette; right-click a `.gs`/`.tpz` in VS Code's Explorer; an open Topaz file's title bar, context menu, and a link on its first line |

---

### File Out

Writes the same artifact Jadeite writes — a `fileformat utf8` header naming the image, then the chunks — via GemStone's own `Class>>fileOutClass`, so a file-out from Jasper and one from Jadeite read back in identically. The folder you chose last time is where the next one opens, shared with File In.

Two cases needed care:

- **A dictionary file-out gets a preamble that creates the dictionary.** GemStone's `fileOutClassesAndMethodsInDictionary:on:` writes none, while every class definition in it says `inDictionary: <Name>` — so without one, the file fails on its very first chunk on a stone that lacks the dictionary and takes the rest with it. The preamble is composed client-side: building it in the doit needs a `$'` literal, which 3.6.x's compiler rejects with an internal `ComStrmSetCursor` error.
- **Method- and protocol-level file-outs carry no class definition.** That is the point of a method-level file-out — shipping one fix without redefining the class, which would drop every method it has.

A class category brings its sub-categories' classes with it, matching what the Classes pane shows, and stops there — a superclass filed under a different category is not swept in behind your back. A dictionary comes out superclass-first, so it reads back in as-is.

### File In

Honours class definitions, class comments, `removeAllMethods`, method categories, `set compile_env:` and `input`, so a class filed out and filed back in returns with its methods in their original protocols, and a loader naming sibling files pulls them in beside it.

**A hand-written `.tpz` script files in too.** Its preamble addresses the topaz *program* rather than the image — `set gemstone`, `login`, `output push`, `commit`, `logout` — and none of it runs: Jasper files in over the session you chose, not one the file names, and does not commit because a file said to. Those lines are **reported** rather than silently dropped, since they change what the file means; `exit` ends the file-in where topaz stops reading; and a `run` chunk ending in `^ something` is handled (the chunk wrapper uses `ensure:` so a non-local return can't skip past it).

### Session targeting

Every route that already names a session files straight into it with no "which session?" prompt — the session row knows its own, and the Explorer is showing exactly one. Routes with no such context (Command Palette, a file in VS Code's Explorer) ask, as any other write does.

The Explorer's File In sits in a group of its own **below** "File Out Dictionary…", not beside it: a file names its own dictionaries, so a file-in is not scoped to the row it was started from, and sharing the group would imply it was.

## Testing

Full suite against a live GemStone 3.6.2 stone: **463 files, 7475 passed, 0 failures.**

- **Integration** (32, live stone): class round trip returning every selector to its protocol; a dictionary removed entirely coming back with its classes; a method-only file-out filed back without touching the class definition; superclass ordering; two concurrent sessions proving a file-in lands only in the session it was handed; error paths.
- **Unit and manifest**: chunk parsing and topaz-script handling, file composition, entry-point placement, and session hand-off.
- **Manually verified** against a live stone: filing a fixture in, a class round trip, class-category ordering, and a method-category file-out carrying no class definition.

## Known limitations

- **Filing a class back onto one owned by another user fails**, with a `SecurityError` per refused chunk. This is GemStone authorization working correctly — the `GsRefactoring` classes, for instance, are installed over a transient SystemUser session, so a DataCurator session cannot modify them. It is reported honestly but not *explained*: sixteen near-identical authorization errors is a poor way to say "you don't own this class." Worth its own issue.
- A method whose source contains a line that is **only** `%` — the chunk terminator — is not covered by a test, since such a method can't be authored in a hand-written chunk file. It needs a `compileMethod:` fixture built from parts.

Fixes #539


client/src/gemstoneExplorer.ts
```diff
@@ -5302,6 +5306,209 @@ export class ExplorerController {
...
+    if (inCategory.size === 0) {
+      void vscode.window.showWarningMessage(`"${node.fullPath}" has no classes to file out.`);
+      return;
+    }
+    await this.runFileOut({
+      title: `File Out ${node.fullPath}`,
+      defaultFileName: fileOutFileName(node.fullPath),
+      label: node.fullPath,
+      build: () => {
+        const ordered = queries
```
MatiasFernandez: 💭 `getDictionaryClassFileOutOrder` answers `[]` when the dictionary doesn't resolve (`dict ifNil: [^ '']`), and this then filters an already-empty list — so `composeFileOut(header, [])` writes a header-only `.gs` and the user gets "Filed out Fauna" for a file with no code in it. The `inCategory.size === 0` guard above catches the empty *category*, but not a `state.dictIndex` that has gone stale (a dictionary removed in another session, a tree not yet refreshed).

The same thing happens more quietly for a partial mismatch: any class in `inCategory` that isn't in the order query's answer is dropped without a word, so the file is short a class and nothing says so.

Everywhere else this PR is careful about exactly this — `classFileOutBody` turns the not-found sentinel into a raise precisely so a `.gs` can't be written with the wrong contents, and `fileOutDictionary` raises rather than "writing an empty file". Should this path do the same, e.g. throw from `build` when `ordered.length !== inCategory.size` (or at least when it's 0), naming the classes that didn't resolve?

client/src/fileOut.ts
```diff
@@ -0,0 +1,130 @@
...
+// file-out queries.
+import * as fs from 'fs';
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { rememberDirectory, rememberedDirectory } from './fileTransferDirectory';
+
+/** Save-dialog file types. `.gs` first — a Topaz file-out is what these commands write. */
+export const FILE_OUT_FILTERS: Record<string, string[]> = {
+  'GemStone Files': ['gs'],
+  'Smalltalk Files': ['st'],
```
MatiasFernandez: 💭 Offering `.st` here breaks the round trip this PR is built around. What gets written is Topaz chunk format either way, but `.st` maps to `gemstone-tonel` in the manifest, so a file-out saved through this filter comes back as: wrong syntax highlighting, no "File In to GemStone" lens (`isFileInTarget` gates on `gemstone-topaz`), no entry in the editor title bar or context menu (both `resourceLangId == gemstone-topaz`), no entry in VS Code's Explorer menu (`.gs`/`.tpz` only), and not even listed by the File In open dialog, whose `FILE_IN_FILTERS` are `gs`/`tpz`.

So the one filter that isn't `.gs` is the one that makes the file unreachable from every route the other half of the PR adds. Should this be `['gs', 'tpz']` to match `FILE_IN_FILTERS`, and drop `.st` entirely?

client/src/extension.ts
```diff
@@ -2256,6 +2257,41 @@ export function activate(context: vscode.ExtensionContext) {
...
+    vscode.commands.registerCommand('gemstone.fileIn', async (item?: GemStoneSessionItem) => {
+      await fileInCommand(sessionManager, context.globalState, item?.activeSession);
+    }),
+
+    vscode.commands.registerCommand(
+      'gemstone.fileInFile',
+      async (uri?: vscode.Uri, selected?: vscode.Uri[]) => {
+        // VS Code hands an Explorer context command the clicked resource AND the whole
+        // selection; the editor title bar passes only the one resource. Falling back to
+        // the active editor covers the palette-shaped call with no argument at all.
```
MatiasFernandez: 💭 The comment says the active-editor fallback "covers the palette-shaped call with no argument at all", but `gemstone.fileInFile` is `"when": "false"` in `commandPalette` — and `fileIn.manifest.test.ts` asserts that on purpose ("it has no file to act on there"). Every remaining route passes a URI: the Explorer menu passes `(uri, selected)`, the title bar and editor menu pass the resource, and the code lens passes `document.uri` explicitly.

So as far as I can tell the fallback and the `'Open or select a .gs file to file in.'` warning are both unreachable today. No objection to keeping them as defence for a future keybinding or an `executeCommand` from elsewhere — but should the comment say that rather than naming the palette, so the next reader doesn't go looking for a palette entry that the manifest deliberately doesn't have?

client/src/fileIn.ts
```diff
@@ -0,0 +1,347 @@
...
+ */
+export async function fileInCommand(
+  sessionManager: SessionManager,
+  store?: vscode.Memento,
+  // The session to file into, when the caller already knows one — the File In button
+  // on a session row in Logins & Sessions, or the GemStone Explorer, which is showing
+  // exactly one session. Without it the user is asked, as any other write is.
+  session?: ActiveSession,
+): Promise<void> {
+  const uris = await vscode.window.showOpenDialog({
```
MatiasFernandez: 💭 Should the session be resolved before the open dialog rather than after it? As written, with nothing connected the user browses, picks their files, and only then gets "No GemStone sessions are active. Please log in first." from `resolveSession` — the picking was wasted.

File Out goes the other way round on purpose: `fileOutSession()` warns first, and the dialog only opens once there's a session (the ordering comment in `saveFileOut` is about the opposite hazard — not paying for expensive queries on a cancel — which doesn't apply here, since resolving a session is free). Doing the same here would also mean `fileInCommand` and `fileInUris` agree on when the question gets asked.

client/src/topazFileIn.ts
```diff
@@ -258,6 +266,225 @@ export function fileInClass(
...
+  for (const region of parseTopazDocument(text)) {
+    if (region.kind === 'topaz') {
+      region.text.split('\n').forEach((raw, offset) => {
+        const line = region.startLine + offset;
+        const trimmed = raw.trim();
+        // Blank lines and `!` comments carry the file's provenance banner and its
+        // section headings — nothing to run.
+        if (trimmed.length === 0 || trimmed.startsWith('!')) return;
+
+        const cat = trimmed.match(/^category:\s*'(.*)'\s*$/i);
```
MatiasFernandez: 💭 This is now the third `category:` scanner in this file, and the only one that un-doubles the quotes: `fileInClass` and `parseFileStructure` both use `/^category:\s*'([^']*)'/i`, which stops at the first inner quote and would file `Bob''s methods` into a category called `Bob`.

Those two are outside this PR's scope, but they got a sibling that behaves differently, which is the kind of divergence that's hard to spot later. Worth pulling the match into one small helper here (`parseCategoryDirective(line): string | undefined`) and pointing the older two at it? The `''` handling you added is the correct one, and the class mirror would quietly get it too.

client/src/__tests__/fileIn.test.ts
```diff
@@ -0,0 +1,431 @@
...
+
+    const outcome = fileInFile(SESSION, A_GS);
+
+    expect(outcome.executed).toBe(1);
+    expect(outcome.errors).toEqual([]);
+    expect(outcome.skipped).toHaveLength(1);
+    expect(outcome.skipped[0].message).toContain('wibble');
+  });
+
+  it('runs a topaz script s chunks and leaves its topaz commands alone', () => {
```
MatiasFernandez: nit: the apostrophes have been stripped out of a handful of test titles, which reads oddly in the runner output — "a topaz script s chunks" here, plus "the caller s session" (:337), "topaz s own" (topazScript.test.ts:158), "a binary selector s punctuation" (fileOut.test.ts:29), "the organizer s" / "the category s" (explorerFileOut.test.ts:117, :139), and "GemStone s own" / "GemStone s error" (fileIn.integration.test.ts:126, :464). Double-quoting the title is enough (`it("runs a topaz script's chunks…")`), and there are already tests in these files doing that.

client/src/__tests__/fileIn.manifest.test.ts
```diff
@@ -0,0 +1,176 @@
+import { describe, it, expect } from 'vitest';
```
MatiasFernandez: 💭 Not sure what would be a good folder name for grouping this "file in"/"file out" logic, but if you could find a concept to group all the related code into its own folder, that would help to keep the codebase more manageable/easier to follow

MatiasFernandez (approved):

client/src/gemstoneExplorer.ts
ericwinger: Fixed in 09de46ea. `build` now raises when the order query cannot account for every class in the pane, rather than writing the file:

```ts
if (ordered.length !== inCategory.size) {
  const missing = [...inCategory].filter((name) => !ordered.includes(name)).sort();
  throw new Error(...);
}
```

The message names them: `Could not file out class Cat — no longer in this dictionary. Refresh and try again.` Both cases you separated are covered by their own test — a stale `dictIndex` making the query answer nothing, and the quieter partial mismatch where the dictionary resolves but a class in it does not.

Verified against a live stone: removed a class from a dictionary, left the Explorer tree stale, filed the category out, and got the named error with nothing written to disk.

client/src/fileOut.ts
ericwinger: Fixed in 09de46ea. `FILE_OUT_FILTERS` is now `{ GemStone Files: [gs, tpz], All Files: [*] }`, matching `FILE_IN_FILTERS`, and `.st` is gone. A test asserts the two sets stay equal, so they cannot drift apart again.

Note the System Browser has its own separate copy of these filters at `systemBrowser.ts:187` with the same `.st` entry. Left alone deliberately — that file is frozen.

client/src/extension.ts
ericwinger: Fixed in 09de46ea. You are right that the fallback is unreachable today, and the comment now says why rather than naming a palette entry that is not there:

> Every route this command is wired to therefore arrives with a URI — the manifest keeps it out of the Command Palette (`"when": "false"`), since there it would have no file to act on. The active-editor fallback and the warning below are defence for a call from somewhere else — a user keybinding, or another extension's `executeCommand` — not for a palette entry.

Kept the fallback, as you suggested.

client/src/topazFileIn.ts
ericwinger: Fixed in 09de46ea. Pulled it into one exported `parseCategoryDirective(line)` in `topazFileIn.ts`, and pointed `fileInClass` and `parseFileStructure` at it as well, so the class mirror picks up the `''` handling. Tests pin `Bob''s methods` through all three readers now, not just `parseTopazScript`.

Side effect worth noting: the helper trims first, so an indented `category:` line is recognised where the older two required it at column 0.

client/src/__tests__/fileIn.test.ts
ericwinger: Fixed in 09de46ea — all nine restored via double-quoted titles, including one you did not list (`fileOutQueries.test.ts:134`, "the organizer s output").

client/src/fileTransfer/__tests__/fileIn.manifest.test.ts
```diff
@@ -0,0 +1,176 @@
+import { describe, it, expect } from 'vitest';
```
ericwinger: Done in 09de46ea. Added `client/src/fileTransfer/`, holding `fileIn.ts`, `fileOut.ts`, `directory.ts` (was `fileTransferDirectory.ts` — the stutter goes away inside the folder) and a co-located `__tests__/`.

Left where they were: the `queries/fileOut*` and `queries/fileIn*` modules, since every query in the codebase lives under `queries/` by convention, and the Explorer-level tests, which belong with the other Explorer tests.

client/src/fileTransfer/fileIn.ts
```diff
@@ -0,0 +1,347 @@
...
+ */
+export async function fileInCommand(
+  sessionManager: SessionManager,
+  store?: vscode.Memento,
+  // The session to file into, when the caller already knows one — the File In button
+  // on a session row in Logins & Sessions, or the GemStone Explorer, which is showing
+  // exactly one session. Without it the user is asked, as any other write is.
+  session?: ActiveSession,
+): Promise<void> {
+  const uris = await vscode.window.showOpenDialog({
```
ericwinger: Fixed, and it needed two goes.

09de46ea did what you asked: `fileInCommand` resolves before opening the dialog and hands the result down to `fileInUris`, so both routes ask at the same point.

That turned out not to produce a prompt at all. `resolveSession` answers with the *selected* session and only asks when there is none — so with several sessions logged in, one of them always current, File In silently inherited whichever the window happened to be showing. Your comment (and mine, and the PR body) all said the user is asked; none of us was right about the existing behaviour.

b2f6d1c9 adds `alwaysAsk` to `resolveSession`, passed by File In's two routes with no session of their own. This is deliberately more than any other write in Jasper does — a file-in can redefine classes across a whole dictionary, and a wrong stone is not something you find out about until afterwards. The current session leads the list marked `(current)` so the usual answer is Enter, and with one session logged in nothing is asked.

f571d96b replaces the tests. The first set stubbed `resolveSession` and asserted it had been called, which passes whether or not anyone is ever asked anything — the rule that went wrong lives *inside* `resolveSession`. They now drive a real `SessionManager` and assert the quick pick itself. Checked by reverting `alwaysAsk` and confirming four of them fail.

Verified in the IDE: palette File In with two sessions logged in now asks; the session-row button and the Explorer entry still ask nothing.

ericwinger:
Merged. Recording what the CI state was at merge, since it was forced.

**Green:** 24 of 26 checks, including every Linux and Windows cell and four of five macOS cells.

**Red:** `macos-latest 3.7.5`, which hit the **job-level** `timeout-minutes: 8` cap in `health-check.yml`. GitHub reports that as *cancelled* rather than *failed* — there was no test failure. Its sibling `macos-latest 3.6.8` passed on the same commit in 5m00, and `3.7.5` itself finished in 3m28 two runs earlier. macOS job durations swung 2m24–9m06 across five cells on identical code in a single run.

**On the refactoring test timeouts in this branch** (`fd446083`, `9db118a3`) — these are not part of the File Out / File In feature and are worth knowing about separately:

Fourteen `refactoring*.integration.test.ts` tests each run `GsFileIn` on `resources/refactoring/engine-tests.gs` (22 test classes, 414 KB, compiled in the stone) and then run one SUnit suite, in a single blocking GCI call. The harness aborts after every test, so the payload cannot be filed in once and shared — every test pays the whole compile. That put them at 5.6–5.9s against vitest's 5s default, so on a loaded macOS runner whichever test drew the busiest cell failed. Three different tests failed on three consecutive runs of this PR, which touches no refactoring code. All fourteen now carry `60_000`, with the reasoning recorded once on `fileInEngineTestsExpr` in `support/refactoring.ts`.

That budgets for the cost rather than removing it, and ~80s of redundant compilation per plugin pass per matrix cell is also why these macOS jobs sit near their 8-minute cap. Removing it means changing how the payload reaches the stone, which is a change to the integration harness and needs its own discussion — not folded in here.

PR #556, opened 2026-09-04
Register an existing database, and the fixes that testing it turned up

ericwinger:
Adds **Register Existing…**, which adopts a GemStone installation Jasper did not create — a customer's stone, a colleague's, one from another checkout — and everything that testing it uncovered.

## Added

**Register a database that already exists, instead of only creating new ones.** The panel could make a database out of an installed release and nothing else, so a stone somebody else set up could not be managed, or even seen, however plainly it was running. Register Existing… asks for the installation's product directory, reads the release out of its own `version.txt` (rather than asking you to type a version, which is how you end up with a login whose client library does not match the stone), and takes the stone and NetLDI names its servers run under. Its NetLDI port is recorded and used, because a NetLDI *name* only resolves through `/etc/services` and an installation Jasper did not set up commonly uses one that was never added there. Jasper writes nothing inside the installation: the record is one file in Jasper's own directory, **Delete Database** is refused, and **Unregister Database** drops the record leaving the installation and any running stone exactly as they were.

**Register Local Version is gone.** It symlinked a product tree into the versions folder, but refused the case people actually hit — a tree already sitting there — with "already exists" and no way forward. A tree placed in the versions folder is now recognised on its own, and a stone running from one elsewhere is adopted by Register Existing…, which records where it really lives.

## Fixed — all found while testing the above

- **A registered database's servers read Stopped while `gslist` listed them as plainly up.** `gslist -cvl` exits non-zero when it finds nothing, which is by definition what Jasper's own registration directory looks like on a machine whose databases are all registered elsewhere. That non-zero exit arrives as a throw, and it was caught as a failed read and abandoned the whole refresh before the registered directories were ever asked. An empty listing is now told apart from an unreadable one — by the message, on either stream, since releases differ about which.

- **Unregister Database could never complete.** It removed its record directory non-recursively, but registering creates a `log/` inside that directory, so the remove died on its own subdirectory with `EISDIR`.

- **A database registered on a machine with no installed release was invisible.** Registering worked and the record was written, but Versions led the panel whenever nothing was installed and the header read "No GemStone release on this machine yet" directly above the database just added. Databases now lead in every state.

- **An action the panel refused said nothing at all.** The host posted the reason — its own comment says the text goes into the panel so a long one can wrap and be read — and the webview dropped it, redrawing the unchanged state. A refusal was indistinguishable from a button that did nothing.

- **Logins.** A login matching no local database appeared nowhere in the panel, though it had saved: those now get an **Other Logins** section, with their own New Login and the sessions opened from them. Saving a login whose user, stone and host match one that already exists overwrote that login instead of adding a row — **New Login** on a database that already had that user hit it every time (the form is prefilled from the database, so every answer already matched), and an *edit* reached the same replace by changing those fields into another row's identity. Both are refused with the reason now. Rows carry the full label rather than the bare user name.

- **GemStone Search exhausted the gem's temporary object memory and took the session down with it** — `AlmostOutOfMemoryError (6022)`, "VM temporary object memory is full … too many markSweeps since last successful scavenge". Everything else in that session then answered with a broken connection instead of its own result, which is why it surfaced as a failed SUnit discovery and made discovery look like the culprit. `ClassOrganizer new` indexes every class in the image, so its cost follows the image rather than the question, and every search, senders, implementors and references query built its own — twelve sites, one of them twice in a single doit. One organizer is now built per session, per environment, and reused — by the hierarchy queries too, which had each been building their own. A captured class list is the one thing reuse cannot keep current, so it is dropped by class definition and deletion (in the same doit that changes the class), by a commit or abort, and by GemStone Search's ⟳. Clearing on every workspace doit is deliberately not done — most create nothing searchable — so the ⟳'s tooltip now names that case.

## Review round

All ten review comments are addressed in `a365e36c`. The two 🚫 findings first:

- **`discoverServersUnder` could never match on Windows.** It compared the folder dialog's `\\wsl$\…` UNC against `ps` output and a `gslist` path that are both guest-side, so discovery came back empty on *every* Windows registration and the form fell back to GemStone's defaults for the conf and registration directories — exactly the case that leaves Jasper unable to stop the stone it just adopted, and reading as "nothing is running there" rather than as a failure. The tree is converted first now, the way every other consumer of a registered product path already did.

  Two more of the same shape came out of it: a folder picked on the Windows side is refused at the pick (there is no Windows GemStone build, so it can never work), and trailing-separator trimming was POSIX-only.

- **`recordNetldiPort` wrote `confPath: "undefined"`** back into a record that had only named its product tree. One serializer for both writers now, writing the resolved record.

The rest: the NetLDI name is checked for a clash when registering (against the other *databases* only — the NetLDI being adopted is normally already up under that name); a refused registration keeps its answers and reopens the form; the sidebar carries the database's provenance so it offers Delete xor Unregister, and no longer offers Replace Extent or the online extent backup on a registered row; `clearClassOrganizerCode` is wired to commit, abort and the ⟳; the three hierarchy queries share the organizer; the login duplicate guard runs on the edit path too; and a shadowed `open` was renamed.

Chasing the sidebar comment turned up one more: `gemstone.onlineExtentBackup` resolved its extents as `<record dir>/data`, which does not exist for a registered database — it would have reported an empty backup as a successful one. Refused, with #562 filed for making it actually work, which needs the extent list off the running stone or the recorded conf.

## Testing

Full CI matrix green — 21/21 jobs, all five Windows legs included ([run 33918163797](https://github.com/GemTalk/Jasper/actions/runs/33918163797)). That is worth more than it was on the previous runs: `databasesPanel.test.ts` had pinned `needsWsl: () => false`, so every platform took the POSIX branch and the broken discovery was never executed. `needsWsl` is a `vi.fn` now, and the new tests drive a mocked guest while handing in the UNC the dialog would give — reverting the one-line fix makes them fail, which was checked rather than assumed.

The Windows legs provision WSL, install and start GemStone inside the guest, and run the native Windows suite against it, so the integration layer did run against a live stone across a real WSL boundary. (An earlier draft of this description said it had not — that was true of local runs only.)

One gap left, and it is manual: nobody has driven **Register Existing…** through the UI on a Windows machine against a stone someone else started, which is the scenario the feature exists for. Every Windows claim here rests on tests.

## For the reviewer

Worth a look in particular: `client/src/queries/classOrganizer.ts` (the cache, what invalidates it, and why clearing on every doit is deliberately not one of them), and `processManager.discoverServersUnder` (the host/guest path boundary the review caught).


client/src/manager/databaseManager.ts
```diff
@@ -133,6 +139,156 @@ export class DatabaseManager {
...
+  recordNetldiPort(db: GemStoneDatabase, port: number): DatabaseYaml {
+    if (!isRegisteredDatabase(db) || db.config.netldiPort === port) return db.config;
+    const updated: DatabaseYaml = { ...db.config, netldiPort: port };
+    wslWriteFileSync(
+      path.join(db.path, 'database.yaml'),
+      `---\nregistered: true\n` +
+        `version: "${updated.version}"\n` +
+        `stoneName: "${updated.stoneName}"\n` +
+        `ldiName: "${updated.ldiName}"\n` +
+        `netldiPort: ${port}\n` +
```
MatiasFernandez: 🚫 On a record that only recorded its product tree, this writes `confPath: "undefined"` and `globalDir: "undefined"` back into the yaml. That shape is supported on purpose — `databaseYaml.test.ts` pins it ("is still read when only the product tree was recorded") — and `registeredPaths` treats the literal string `"undefined"` as truthy, so `confDir`/`globalDir` come back as `undefined` instead of falling back to `defaultConfDir(productPath)` / `DEFAULT_GLOBAL_DIR`, and the next start hands the installation `GEMSTONE_SYS_CONF=undefined`. Observing a live port would quietly corrupt the record it was correcting, and the `recordNetldiPort` tests all start from a fixture that has both paths, so nothing catches it.

Would it be worth extracting the one serializer this and `registerExistingDatabase` share, and writing the *resolved* values from `registeredPaths(config)` (or omitting a line when there is nothing to write)? That would also keep the two copies of this template from drifting apart.

client/src/manager/databasesPanel.ts
```diff
@@ -1036,6 +1096,109 @@ export class DatabasesPanel {
...
+        this.failed('register the database', new Error(`No ${field} was given.`));
+        return;
+      }
+    }
+
+    const taken = this.deps.storage.getDatabases().map((d) => d.config);
+    if (taken.some((c) => c.stoneName === msg.stoneName)) {
+      this.failed(
+        'register the database',
+        new Error(`A stone called "${msg.stoneName}" is already registered.`),
```
MatiasFernandez: 💭 The re-check covers the stone name but not the NetLDI name — and the form's `registerProblems` only checks `stoneNames` too, while `createDatabase` refuses a clash on both. If two databases end up sharing an `ldiName`, they both match the same `gslist` row (`isNetldiRunning` matches on name + version), so each row reports the other's NetLDI as its own and `netldiPortFor` can hand one database's port to the other's login. Should the ldiName go through the same check here?

client/src/manager/databasesView.js
```diff
@@ -1198,9 +1531,58 @@
...
+        command: 'registerDatabase',
+        productPath: registerForm.productPath,
+        stoneName: registerForm.stoneName,
+        ldiName: registerForm.ldiName,
+        ...(port ? { netldiPort: Number(port) } : {}),
+        ...(registerForm.confPath ? { confPath: registerForm.confPath } : {}),
+        ...(registerForm.globalDir ? { globalDir: registerForm.globalDir } : {}),
+      });
+      registering = false;
+      resetRegisterForm();
```
MatiasFernandez: 💭 `submitRegister` resets the form as well as closing it, so when the host refuses the registration — a stone name taken between the redraw and the click, a root it cannot write — the new failure banner appears but every answer is gone, including the product directory, which has to be re-chosen through the folder dialog. `submitCreate` closes the form but deliberately leaves `createForm` filled for the same reason a mid-typing refresh doesn't discard answers. Should the answers survive here too, with the reset kept to `beginRegister` / `cancelRegister`?

client/src/manager/databaseTreeProvider.ts
```diff
@@ -85,7 +86,18 @@ export class DatabaseTreeProvider implements vscode.TreeDataProvider<DatabaseNod
         item.description = `${node.db.config.stoneName} (${node.db.config.version})`;
         item.contextValue = `gemstoneDb${this.databaseContext(node.db)}`;
```
MatiasFernandez: 💭 `contextValue` doesn't say whether the database is registered, and `package.json` gates `gemstone.unregisterDatabase` on `gemstoneDb(Running|Stopped|External)` — the same clause `deleteDatabase` uses. So a sidebar row offers Unregister on a database Jasper created and Delete on one it registered, and each then fails with an error message; the panel gets this right by greying Delete and showing the reason. Should the context value carry the distinction (a `Registered` suffix, or a separate when-clause context) so the menu only offers the action that can succeed?

client/src/queries/classOrganizer.ts
```diff
@@ -0,0 +1,61 @@
...
+ * Keyed by environment: an organizer collects its classes under one environment id,
+ * so environments cannot share one. `newForEnvironment:` sets that at collection
+ * time, which is what the callers setting `environmentId:` after `new` were
+ * reaching for — that only relabels an organizer whose classes were already
+ * gathered under environment 0.
+ */
+export function classOrganizerExpr(environmentId: number | string = 0): string {
+  const key = `#'JasperClassOrganizer_${environmentId}'`;
+  return `(SessionTemps current at: ${key} ifAbsent: [
+  SessionTemps current at: ${key} put: (ClassOrganizer newForEnvironment: ${environmentId})])`;
```
MatiasFernandez: 💭 Has `ClassOrganizer newForEnvironment:` been checked against a 3.6.2 stone? The generated Smalltalk under `client/src/queries` carries a pretty deliberate 3.6.x discipline (`searchSelectors.ts` notes `includesSubstring:` DNUs there, `util.ts` the Unicode7 literal problem), and this swaps every organizer construction on the search paths at once — so if the selector is 3.7-only, search, senders, implementors and references all DNU on the older stones rather than degrading. If it isn't there, would `new` + `environmentId:` be an acceptable fallback for those releases, or is setting the environment at collection time load-bearing enough that it needs a version check?

client/src/queries/classOrganizer.ts
```diff
@@ -0,0 +1,61 @@
...
+  // without `asString`, so a pattern match here silently cleared nothing and left
+  // a stale organizer behind. `isString` guards the send: session temps are a
+  // shared namespace and nothing promises every key in it is a String.
+  return `SessionTemps current keys asArray do: [:k |
+  (k isString and: [k asString beginsWith: 'JasperClassOrganizer_'])
+    ifTrue: [SessionTemps current removeKey: k ifAbsent: [nil]]].`;
+}
+
+/** The same, as a whole doit for callers with nothing else to run. */
+export function clearClassOrganizerCode(): string {
```
MatiasFernandez: 💭 `clearClassOrganizerCode` has no caller — was it meant to be wired to something? The two invalidation points are Jasper's own class definition and deletion, but a class created by *executing* `subclass:` in a workspace is the case the CHANGELOG itself calls out as announcing nothing, and a class another session added and committed is the same. The cached organizer then keeps its snapshot class list for the life of the session, and GemStone Search's ⟳ — which reloads its lists from the stone — wouldn't clear it, so a refresh would still answer from the old class set. Would hooking this into that refresh (or into commit/abort) close the gap?

client/src/queries/classOrganizer.ts
```diff
@@ -0,0 +1,61 @@
+/**
+ * The Smalltalk for reaching a `ClassOrganizer` — built once per gem session and
+ * reused, rather than constructed per query.
+ *
+ * `ClassOrganizer new` walks the whole symbol list and builds the class list,
+ * hierarchy and category index for the entire image, so its cost scales with the
+ * image rather than with the question being asked. Every search, senders,
+ * implementors and references query built its own; the Source scope built one per
```
MatiasFernandez: 💭 `getClassHierarchy.ts`, `getSiblingClassNames.ts` and `getClassDescendantNames.ts` still do a bare `ClassOrganizer new`, so they keep paying the per-image cost this file measures. Out of scope for the fix, or worth moving them onto `classOrganizerExpr` too while the reasoning is fresh? (If there's a reason not to — a hierarchy question wanting a guaranteed-current class list, say — that would be worth a line in this doc, since it reads as "one organizer per session" now.)

client/src/loginEditorPanel.ts
```diff
@@ -170,6 +177,26 @@ export class LoginEditorPanel {
...
+    // A login is identified by its user, stone and host, so a *new* one whose
+    // label already exists is that same login again. `saveLogin` reads a matching
+    // label as the row being edited and replaces it — which for a new login means
+    // the existing one is silently overwritten, no row is added, and the panel
+    // looks exactly as it did before the Save. That is what pressing New Login on
+    // a database that already has its DataCurator login did every time: the form
+    // is prefilled from the database, so its answers match the login already
+    // there unless something is changed by hand. Refused with the reason, and the
+    // form left open on the answer that has to change.
+    if (!originalLabel) {
```
MatiasFernandez: 💭 The guard is skipped whenever `originalLabel` is set, so an *edit* that changes the user, stone or host into another existing login's identity still reaches `saveLogin`'s replace path and silently overwrites that other login — the same vanishing row this fixes for New Login, just arrived at by editing. Should the check run on the edit path as well, excluding the row being edited (`loginLabel(l) === data.label && loginLabel(l) !== originalLabel`)?

client/src/manager/databasesView.js
```diff
@@ -829,15 +933,82 @@
...
+  function renderOtherLogins(logins, open) {
+    const rows = logins
+      .map((l) => {
+        // The label names the user, stone and host, the way every login row does.
+        // The NetLDI and release follow it dimmed: with no database row above to
+        // supply them, they are all that separates two logins to the same stone.
+        const extra = [l.netldi ? `via ${l.netldi}` : '', l.version]
+          .filter((part) => part)
+          .join(' · ');
+        const open = l.sessions || [];
```
MatiasFernandez: nit: `const open = l.sessions || []` shadows `renderOtherLogins`'s own `open` parameter (the section's open flag, still used by the `section(...)` call below). It's correct, but the two meanings in one function took a second read — `sessions` for the inner one?

client/src/manager/__tests__/registerUnregisterRoundTrip.test.ts
```diff
@@ -0,0 +1,115 @@
+// Register and unregister against a real directory, not a mocked filesystem.
```
MatiasFernandez: Thanks for this one ❤️ — a stubbed `wslFs` could never have caught the EISDIR, and a real-directory round trip read back through the same storage the panel uses is exactly the test that would have. Good call writing the reason at the top, too.

client/src/manager/processManager.ts
```diff
@@ -176,27 +214,236 @@ export class ProcessManager {
...
     try {
-      const gsPath = gslistPath.replace(/\/bin\/gslist$/, '');
-      const rootPath = needsWsl() ? this.storage.getWslRootPath() : this.storage.getRootPath();
-      const env = this.versionEnvironment(gsPath, rootPath);
-      const output = wslExecSync(`"${gslistPath}" -cvl`, env);
-      this.cachedProcesses = parseGslist(output);
-      // Reaching here at all is the signal that the read happened, so an empty
-      // parse is a real "nothing registered" rather than a failure to look.
-      this.gslistReadable = true;
+      hosts = this.hostServers().filter((p) => p.command.startsWith(`${tree}/`));
```
MatiasFernandez: 🚫 Not from the new commits — I missed this last time, and ecbf9b74's "Windows is the platform that most needs covering" is what sent me back to it. On Windows this filter can never match. `hostServers()` runs `ps -Ao pid=,command=` through `wslExecSync`, so every `command` is a Linux path — while the `productPath` the panel hands in is `uris[0].fsPath` (`databasesPanel.ts:1115`), i.e. `\\wsl$\Ubuntu\opt\theirs\product`. So `startsWith(`${tree}/`)` compares a UNC prefix against `/opt/theirs/product/sys/stoned` and finds nothing. `readGslist(tree, globalDir)` further down has the same problem from the other side: it builds `"\\wsl$\...\bin/gslist"` and runs that inside the guest.

Discovery then comes back empty and the form falls back to GemStone's defaults for `confPath` and `globalDir` — which is exactly the case this doc block warns about ("wrong for any other [layout], which would leave Jasper unable to stop the very stone it just adopted"), except it happens on *every* Windows registration rather than only on a non-stock one, and reads as "nothing is running there".

Every other consumer of a registered `productPath` converts first — `productPathFor`, `registeredProcesses` and `getEnvironment` all do `needsWsl() ? windowsPathToWsl(...)`. Should this one join them? And nothing in the suite can catch it as it stands: `databasesPanel.test.ts` pins `needsWsl: () => false` so every platform takes the POSIX path.

client/src/manager/__tests__/registerUnregisterRoundTrip.test.ts
```diff
@@ -0,0 +1,164 @@
...
+      }
+    }
+  });
+
+  it('resolves the configured Linux root to the directory it actually writes', () => {
+    // The setting names a place in the guest; the record has to land there and
+    // not somewhere the host invented from it. A root that resolves wrong writes
+    // a real directory to a real place, so nothing throws — only this catches it.
+    const { storage } = makeManager();
+    expect(storage.getRootPath()).toBe(HOST_ROOT);
```
MatiasFernandez: 💭 I don't think this assertion can fail on any platform. `HOST_ROOT` is `wslPathToWindows(ROOT)`, and `getRootPath()` computes `wslPathToWindows(raw)` from the same `ROOT` — which has no `~` for it to expand — so both sides are the same function over the same input; on Linux it reduces to `ROOT === ROOT`. The comment says "a root that resolves wrong writes a real directory to a real place, so nothing throws — only this catches it", but what actually catches that is line 98 (`wslExistsSync(HOST_ROOT)`, which does prove the UNC round-trips back to a real directory) and the next test's `expect(db.path).toBe(path.join(HOST_ROOT, db.dirName))`.

Worth either dropping line 97 and letting the comment point at the two that do the work, or asserting against something independent of `wslPathToWindows` — the `\\wsl$\<distro>` shape, say — so a change to that conversion actually shows up here?

client/src/manager/__tests__/registerUnregisterRoundTrip.test.ts
```diff
@@ -0,0 +1,164 @@
...
+  wslExistsSync,
+  wslMkdirSync,
+  wslReaddirSync,
+  wslRmSync,
+  wslWriteFileSync,
+} from '../../wslFs';
+
+// The configured setting: a Linux path, as a user's would be.
+const ROOT = needsWsl()
+  ? wslExecSync('mktemp -d /tmp/jasper-register-XXXXXX').trim()
```
MatiasFernandez: 💭 `needsWsl()` is just `isWindows()` — it doesn't check WSL is actually there — and this runs at import time, so on a Windows machine without WSL (or with a distro that won't start) `wsl.exe` fails and the whole file errors during *collection*: no test result, no skip, just a module-load crash. Same for `wslPathToWindows` on the next line, which calls `getWslInfo()`.

CI is fine — `setup-gemstone-in-windows` provisions Ubuntu-22.04 before `npm test` — so this is about `npm run test:client` on a plain Windows dev box, where the previous `describe.skipIf` at least degraded gracefully. Since the suite is genuinely POSIX-or-WSL-only, would it be worth gating on WSL answering rather than on the platform, and keeping the skip for the no-WSL case? That also keeps the reason visible in the skip report instead of arriving as a crash.

MatiasFernandez (approved):
I left some comments worth reviewing. I'm worried we may be introducing issues for windows users.

client/src/manager/databaseManager.ts
ericwinger: Done — extracted `registeredDatabaseYaml` in `registeredDatabase.ts`, and both writers go through it. It writes the *resolved* values from `registeredPaths(config)`, so a record that only named its product tree gets its actual defaults written down rather than the literal `"undefined"`, and the two copies of the template can no longer drift.

You were right that nothing caught it: the `recordNetldiPort` tests all started from a fixture with both paths. Added one that starts from a product-tree-only record and asserts the yaml contains no `undefined` at all.

client/src/manager/databasesPanel.ts
ericwinger: Added. One thing worth flagging, because it makes the register check deliberately *not* the same as `createDatabase`'s: the NetLDI being registered is normally already running under that very name — that is the thing being adopted. So checking against `ldiNames` (which folds in every NetLDI `gslist` reports, exactly right for Create) would refuse essentially every legitimate registration.

The check is against the other **databases**' NetLDI names only, which is the collision you describe. That needed a new `dbLdiNames` on the state, since `ldiNames` is the wider list and the form uses both. Covered on the host side and in the view, including the case that must still be allowed (a name only a live NetLDI holds).

client/src/manager/databasesView.js
ericwinger: Fixed, though not quite as suggested — keeping the reset in `beginRegister`/`cancelRegister` alone was not enough on its own. `beginRegister` resets, so "press Register Existing… again" would still have handed back an empty form; the answers had to survive *and* the form had to come back by itself.

`submitRegister` now closes without resetting and sets a `registerPending` flag. An `actionFailed` while that is set reopens the form on its answers, under the banner saying why. The next state clears the flag, and `beginRegister` empties the form after that, so a landed registration still starts clean next time. Both directions are tested.

client/src/manager/databaseTreeProvider.ts
ericwinger: Done — a `Registered` suffix on the context value, and the when-clauses take it as optional wherever the action works on either kind.

I extended it to the Stone and NetLDI rows too, since the same problem was sitting there: **Replace Extent** is refused for a registered database and was still being offered. And chasing that turned up a third one — `gemstone.onlineExtentBackup` resolves its extents as `path.join(db.path, "data")`, which for a registered database is Jasper's record directory with no `data/` in it. It would have run against a directory that does not exist and reported an empty backup as a successful one. Refused now, and withheld in both surfaces.

Backing up a registered database is worth having and is not just a matter of lifting the refusal — it needs the extent list read off the running stone or the recorded conf file. Filed as #562.

client/src/queries/classOrganizer.ts
ericwinger: Wired. Commit and abort clear it (`extension.ts`), which covers a class another session added, and GemStone Search's ⟳ clears it before reloading its corpora, which is the gap neither of those closes.

On the workspace-`subclass:` case: we decided to leave that to the button rather than clear on every doit. Most doits create nothing searchable, and paying a corpus reload plus an organizer rebuild for all of them gives back exactly what the per-session organizer bought. What was missing was that nothing *told* anyone — so the ⟳'s tooltip now names the case ("A class, global or method created by executing code in a workspace announces nothing, so it will not turn up in a search until this is pressed"), and the trade is written down in `classOrganizer.ts` and DESIGN.md as a decision to revisit if stale searches turn out to bite.

client/src/queries/classOrganizer.ts
ericwinger: Moved all three onto `classOrganizerExpr`. No reason not to that survived looking at it: they ask `subclassesOf:` and `allSuperclassesOf:`, which read the snapshot the searches read, and every refactoring that creates the class they would then ask about compiles it through `compileClassDefinition` — which clears the cache in the same doit that creates it. That reasoning is now in the doc comment, since you are right that it read as "one organizer per session" while three callers quietly opted out.

client/src/loginEditorPanel.ts
ericwinger: Fixed as suggested — the guard runs on both paths now, with `loginLabel(l) !== originalLabel` excluding the row being edited so re-saving it unchanged, or changing only its password, tags or NetLDI, is not a clash with itself. Tests for both: an edit that collides is refused with the panel left open, and an edit that changes only a non-identifying field still saves.

client/src/manager/databasesView.js
ericwinger: Renamed to `sessions`.

client/src/manager/processManager.ts
ericwinger: Good catch, and thank you for going back to it. Fixed — the tree is converted before anything reads it, the way `productPathFor`, `registeredProcesses` and `getEnvironment` all already did.

Two more of the same shape came out of chasing it. A folder picked on the *Windows* side can never work at all (no Windows GemStone build), so it is refused at the pick with the reason rather than registering happily and failing on every command afterwards. And the trailing-separator trimming was POSIX-only, so a UNC ending in a backslash kept it — both separators now, in all three places a product path is normalised.

On the coverage point: `needsWsl` is a `vi.fn` in `databasesPanel.test.ts` now, defaulting to false so the other tests are unchanged. The new discovery tests drive a mocked guest — Linux `ps`, a Linux environment, a `gslist` that answers only at a Linux path — while handing in the UNC the dialog would give. Reverting the one-line fix makes them fail with `[]`, which I checked rather than assumed.

client/src/manager/__tests__/registerUnregisterRoundTrip.test.ts
ericwinger: You are right, and it is worse than not failing — it passes however wrong `wslPathToWindows` becomes, because both sides are that function over the same input.

Rewritten to do both things you suggested. The shared test round-trips through the inverse (`windowsPathToWsl(root) === ROOT` on Windows, `root === ROOT` elsewhere) and keeps the `wslExistsSync` line that does the real work; the `\\wsl$\<distro>\` shape is pinned in a Windows-only `it.runIf` of its own, so a change to that conversion surfaces as a named test rather than as a silent behaviour change.

PR #557, opened 2026-09-04
Add a macOS leg to the Health Check matrix

npapagna:
Extends the platform axis alongside the existing ubuntu-latest and windows-latest legs, so health-check now exercises the same GemStone client/server setup on macOS as it already does on Linux and Windows.

### DIFF
--- a/.github/workflows/health-check.yml
+++ b/.github/workflows/health-check.yml
@@ -122,11 +122,13 @@ jobs:
         run: npm audit signatures
 
   health-check:
-    # The Windows leg needs extra headroom (WSL provisioning + GCI client
-    # download); the Linux legs still fit in 5. Scoped per-platform so a
-    # Linux-side regression that doubles the runtime still trips the guard
-    # instead of quietly passing under the Windows leg's larger budget.
-    timeout-minutes: ${{ matrix.platform == 'windows-latest' && 10 || 5 }}
+    # Each platform gets its own budget, scaled to its actual runtime plus
+    # headroom, so a regression on one platform still trips the guard
+    # instead of quietly passing under a slower platform's larger budget.
+    # Windows needs the most (WSL provisioning + GCI client download);
+    # macOS needs more than Linux (DMG mount + shared-memory setup); Linux
+    # is the baseline.
+    timeout-minutes: ${{ fromJSON('{"ubuntu-latest":5,"windows-latest":10,"macos-latest":8}')[matrix.platform] }}
     name: ${{ matrix.platform }} ${{ matrix.gemstone-version }}${{ matrix.node-version && format(' (Node {0})', matrix.node-version) || '' }}
     needs: prepare
     runs-on: ${{ matrix.platform }}
@@ -157,7 +159,7 @@ jobs:
       SKIP_REPORT_PLUGIN_PATH: ${{ github.workspace }}/test-results/skips-${{ matrix.platform }}-${{ matrix.gemstone-version }}-${{ matrix.node-version || 'nvmrc' }}-plugin.json
     strategy:
       matrix:
-        platform: [ubuntu-latest, windows-latest]
+        platform: [ubuntu-latest, windows-latest, macos-latest]
         gemstone-version: ${{ fromJson(needs.prepare.outputs.versions) }}
         node-version: ['']            # empty ⇒ dev jobs resolve Node from .nvmrc (node-version-file); floor include overrides
         # node-version's declared axis value is only '' (which falls back to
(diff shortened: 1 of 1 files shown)


.github/workflows/health-check.yml
```diff
@@ -122,11 +122,13 @@ jobs:
...
-    # Linux-side regression that doubles the runtime still trips the guard
-    # instead of quietly passing under the Windows leg's larger budget.
-    timeout-minutes: ${{ matrix.platform == 'windows-latest' && 10 || 5 }}
+    # Each platform gets its own budget, scaled to its actual runtime plus
+    # headroom, so a regression on one platform still trips the guard
+    # instead of quietly passing under a slower platform's larger budget.
+    # Windows needs the most (WSL provisioning + GCI client download);
+    # macOS needs more than Linux (DMG mount + shared-memory setup); Linux
+    # is the baseline.
+    timeout-minutes: ${{ fromJSON('{"ubuntu-latest":5,"windows-latest":10,"macos-latest":8}')[matrix.platform] }}
```
MatiasFernandez: 💭 The old ternary always produced a number, so an unmapped platform still got the 5-minute default. This map returns `null` for any platform that isn't a key, and `lint-workflow-timeouts.sh` only checks that the `timeout-minutes` key *exists* — it never evaluates the expression — so whoever adds the next platform to the axis and forgets this map gets a leg with no effective guard (GitHub's 360-minute default) and nothing tells them.

Would it be worth closing that with a fallback, e.g. `fromJSON('...')[matrix.platform] || 5`? Then a forgotten platform gets the strictest budget and trips loudly instead of quietly running for six hours.

.github/workflows/health-check.yml
```diff
@@ -157,7 +159,7 @@ jobs:
       SKIP_REPORT_PLUGIN_PATH: ${{ github.workspace }}/test-results/skips-${{ matrix.platform }}-${{ matrix.gemstone-version }}-${{ matrix.node-version || 'nvmrc' }}-plugin.json
     strategy:
       matrix:
-        platform: [ubuntu-latest, windows-latest]
+        platform: [ubuntu-latest, windows-latest, macos-latest]
```
MatiasFernandez: 💭 Putting `macos-latest` on the axis fans it out across every GemStone version, so each run now queues 5 macOS jobs (they all went green here in 3–5 min, so the leg itself looks solid 👍). My question is about concurrency rather than correctness: GitHub caps concurrent macOS jobs much lower than Linux ones, and 5 at a time is already at that ceiling for the free tier — with a couple of PRs and the merge queue in flight, the macOS legs would serialize and become the long pole of every run.

Given that what the macOS leg proves is mostly OS-specific (DMG install, shared memory, native GCI) rather than version-specific, does it need the full version fan-out? The file already has the pattern for a single pinned leg — the Node-floor `include` right below. Something like one `include` entry on the newest version would keep the platform coverage at a fifth of the runner pressure. Happy to be wrong if you'd rather have the full matrix from day one and narrow it only if it hurts.

MatiasFernandez: nit: the `SKIP_REPORT_*` comment further up ("so the Linux and Windows legs of the same version/node combo don't produce same-named files") now enumerates two of three platforms. The names themselves already include `matrix.platform` so nothing breaks — just the comment reads as if the axis were still a pair.

.github/workflows/health-check.yml
```diff
@@ -254,27 +256,40 @@ jobs:
...
+        # macOS's default kern.sysv.shmmax/shmall are far too small for it
+        # (Stone fails with "fork of cache monitor failed, Monitor process
+        # did not start"), and macOS doesn't apply /etc/sysctl.conf-style
+        # config on its own. resources/setSharedMemory.sh is the same script
+        # end users run once via the extension's setup wizard; applying it
+        # with sysctl -w takes effect immediately, no reboot needed, which
+        # is what a same-boot CI runner needs.
+        if: runner.os == 'macOS'
+        run: sudo ./resources/setSharedMemory.sh
       - name: Setup GemStone in Linux
```
MatiasFernandez: 💭 This step's condition is `runner.os != 'Windows'`, so as of this PR it also runs on the macOS legs — the name "Setup GemStone in Linux" is now inaccurate, and it's the step name that shows up in the macOS job logs. Should we rename it to something platform-neutral ("Setup GemStone", or "Setup GemStone natively") to match what the condition actually selects?

.github/workflows/health-check.yml
```diff
@@ -254,27 +256,40 @@ jobs:
...
+        # GemStone's Stone process attaches a System V shared page cache;
+        # macOS's default kern.sysv.shmmax/shmall are far too small for it
+        # (Stone fails with "fork of cache monitor failed, Monitor process
+        # did not start"), and macOS doesn't apply /etc/sysctl.conf-style
+        # config on its own. resources/setSharedMemory.sh is the same script
+        # end users run once via the extension's setup wizard; applying it
+        # with sysctl -w takes effect immediately, no reboot needed, which
+        # is what a same-boot CI runner needs.
+        if: runner.os == 'macOS'
+        run: sudo ./resources/setSharedMemory.sh
```
MatiasFernandez: Nice that this runs the very same `resources/setSharedMemory.sh` end users get from the setup wizard rather than an inlined `sysctl -w` pair — it means CI now exercises that script on every run, so it can't silently rot ❤️

MatiasFernandez (approved):

PR #558, opened 2026-09-04
Attribute the Enhanced Inspector payload to the holders its upstreams name

### LINKED ISSUE #453: Add a THIRD-PARTY.md section for the Enhanced Inspector vendored payload
## Summary
`resources/enhancedInspector/*.gs` ships seven third-party MIT-licensed files verbatim in the VSIX (`resources/**` is not excluded via `.vscodeignore`), but `THIRD-PARTY.md` currently only has a section for the Pharo AST-Core (Refactoring Browser AST) vendored source. Each shipped `.gs` file's header only points readers out to "LICENSE in the origin repository" rather than including the MIT notice/permission text the way the existing AST-Core section does.

## Where
`THIRD-PARTY.md` (needs a new section) and `resources/enhancedInspector/README.md` (could link to it once added).

## Suggested change
Per `gs-src/enhancedInspector/build/apply_jasper_transforms.sh`, six files (`Announcements.gs`, `STON.gs`, `patch-gemstone.gs`, `gtoolkit-wireencoding.gs`, `gt4gemstone.gs`, `gtoolkit-remote.gs`) come from feenk-affiliated upstream repos, and `RemoteServiceReplication.gs` comes from GemTalk's own repo. A new section modeled on the existing AST-Core one (What/Where/Origin/Obtained via/License/Modifications, then the license text) could cover these.

This needs a careful pass rather than a copy of the existing per-file attribution table: the upstream `gtoolkit-remote` repo's `LICENSE` file names an individual feenk gmbh developer as the copyright holder rather than "feenk gmbh" directly, so each origin repo's actual `LICENSE` file should be checked individually instead of assuming uniform wording across all six feenk-affiliated files.

jgfoster:
Fixes #453.

`resources/enhancedInspector/*.gs` — seven files, all third-party — ship verbatim in the VSIX, but `THIRD-PARTY.md` and `NOTICE` covered only AST-Core, and each file's header pointed readers at "LICENSE in the origin repository" instead of carrying a notice.

The issue asked for a careful pass over each origin repo's `LICENSE` rather than a uniform table, and that pass turned up more than a missing section.

## What the upstream LICENSE files actually say

| Payload file | Copied from | Upstream project | `LICENSE` copyright line |
|---|---|---|---|
| `gt4gemstone.gs`, `patch-gemstone.gs` | feenkcom/gt4gemstone | same | `Copyright (c) 2021 feenk` |
| `gtoolkit-remote.gs` | feenkcom/gtoolkit-remote | same | `Copyright (c) 2019 Juraj Kubelka` |
| `gtoolkit-wireencoding.gs` | feenkcom/gtoolkit-wireencoding | same | `Copyright (c) 2024 feenk` |
| `RemoteServiceReplication.gs` | GemTalk/RemoteServiceReplication | same | `Copyright (c) 2017-2024 GemTalk Systems` |
| `Announcements.gs` | feenkcom/gt4gemstone | **GemTalk/Announcements** | `Copyright (c) 2020-2021 GemTalk Systems` |
| `STON.gs` | feenkcom/gt4gemstone | **svenvc/ston**, GemStone port in **GemTalk/Rowan** | `Copyright (C) 2012 Sven Van Caekenberghe` / `Copyright (c) 2018 Dale Henrichs` |

Every header we ship named a holder that appears in no upstream `LICENSE`: `feenk gmbh` (upstream says `feenk`), `GemTalk Systems, Inc` (upstream says `GemTalk Systems`), and `feenk gmbh` for `gtoolkit-remote`, which is licensed by an individual.

Two files are not feenk's work at all. `gt4gemstone`'s [`scripts/convertRsrToGsFormat.topaz`](https://github.com/feenkcom/gt4gemstone/blob/main/scripts/convertRsrToGsFormat.topaz) exports `GemTalk/Announcements` and `GemTalk/RemoteServiceReplication` into gt4gemstone's own `src-gs/` — so `src-gs/Announcements.gs` is GemTalk's code, and its packages (`Announcements-Core-GemStone`, `Announcements-Core-GemStone-Test`, `Announcements-Extensions-GemStone`) match that repo's `src/` exactly. `src-gs/STON.gs` carries `STON-Core` and `STON-GemStone-Kernel`; the `STON` class comment is byte-identical to [svenvc/ston](https://github.com/svenvc/ston)'s, and those two package names are the ones in `GemTalk/Rowan` at `platforms/gemstone/projects/STON/src/`.

## Changes

- **`THIRD-PARTY.md`** — a new section modelled on the AST-Core one (What / Where / Obtained via / Modifications / License), plus the per-file provenance table and the full MIT notice under all seven copyright lines. I verified the permission-and-warranty text is word-for-word identical in all seven `LICENSE` files (they differ only in the copyright line, the presence of an `MIT License` title, and line wrapping), so it is reproduced once with that stated, rather than seven near-identical times.
- **`NOTICE`** — a matching bullet listing all seven holders.
- **`apply_jasper_transforms.sh`** — corrected attribution table; headers now carry an `Upstream:` line where the copied-from repo is not the copyright holder's, the holder as its `LICENSE` writes it, and a pointer to `THIRD-PARTY.md`/`NOTICE` (both of which ship in the VSIX).
- **`resources/enhancedInspector/*.gs`** — regenerated headers. **Bodies are untouched** (verified by diffing header-stripped files).
- **Two READMEs** — link to the new section and say to fix the table rather than a header.

The old code only prepended a header when the sentinel was absent, so correcting the table would never have reached the committed payload. It now regenerates the header block, and refuses to rewrite a file whose sentinel is present but whose closing rule is missing rather than stripping it to nothing.

## Verification

`npm run lint && npm run format:check && npm run compile && npm test` all pass (7294 + 322 + 92 tests). Re-running `apply_jasper_transforms.sh` reproduces the committed payload byte for byte, twice; the malformed-header guard was exercised on a copy and left the file untouched.

`apply_jasper_transforms.sh` is broken on macOS (#429, unrelated and left alone), so I ran it through a `sed -i` shim to produce these bytes; on Linux the script itself produces them.

## For review

Two calls a human should sign off on, since this is a license file:

1. **STON's lineage is inferred**, not documented anywhere upstream — from the identical class comment and the matching Rowan package names. Both candidate holders are listed, and both are MIT, so the notice is complete either way, but the wording "GemStone port: GemTalk/Rowan" is my inference.
2. **The MIT text is reproduced once** rather than per holder. The bodies are provably identical, but if you would rather see seven verbatim blocks, say so and I will expand it.

Also worth noting: unlike AST-Core (which pins RowanV3 3.7.5, build `cf61017e`), **this payload records no upstream revision** — the `.gs` files came from whatever was in the maintainer's `$ROWAN_PROJECTS_HOME` at the time. Provenance here is repo-level only. `THIRD-PARTY.md` says so and suggests recording per-file commits on the next re-vendor; that is out of scope here.


gs-src/enhancedInspector/build/apply_jasper_transforms.sh
```diff
@@ -34,9 +38,45 @@ REPO="$(CDPATH= cd -- "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)"
...
+    # carry more than one notice; print the rest as continuation lines.
+    while IFS= read -r c; do
+        [ -z "$c" ] && continue
+        if [ "$first" = 1 ]; then
+            echo "! License : MIT - $c"
+            first=0
+        else
+            echo "!           MIT - $c"
+        fi
+    done <<< "$(echo "$copyrights" | tr ';' '\n' | sed -E 's/^ +//; s/ +$//')"
```
MatiasFernandez: 💭 If a row's copyright field is ever left blank (or trims to empty), the loop skips every iteration, `first` stays `1`, and no `! License :` line is emitted at all — the file still gets `transformed: ...` and the script still exits 0. I ran `emit_header` with an empty 5th field:

```
! Jasper Enhanced Inspector vendored source
! ----------------------------------------------------------------------------
! Origin  : http://o
! Source  : src/x.gs
!           Full MIT notice and permission text: THIRD-PARTY.md and
```

That's the same class of silently-wrong attribution #453 is about, except now it fails open on a payload file we ship in the VSIX. Should we refuse the run when `first` is still `1` after the loop, the way the missing-closing-rule case below refuses rather than emptying a file?

gs-src/enhancedInspector/build/apply_jasper_transforms.sh
```diff
@@ -34,9 +38,45 @@ REPO="$(CDPATH= cd -- "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)"
...
+    echo "!           Full MIT notice and permission text: THIRD-PARTY.md and"
+    echo "!           NOTICE at the root of https://github.com/GemTalk/Jasper"
+    echo "!"
+    echo "! Vendored into Jasper and filed into the stone by the Enhanced Inspector"
+    echo "! installer. DO NOT EDIT BY HAND - regenerated from upstream by"
+    echo "! update_enhanced_inspector_support.sh, which re-applies this header and rewrites"
```
MatiasFernandez: 💭 This line still tells a reader that `update_enhanced_inspector_support.sh` "re-applies this header", but the new prose in both READMEs is explicit that this script owns them ("`apply_jasper_transforms.sh` also owns their attribution headers"). Since every header is regenerated now anyway, is it worth pointing the line at the script that actually writes it? As it stands, someone who reads only the header — which is the one place that says DO NOT EDIT BY HAND — gets sent to the wrong file.

THIRD-PARTY.md
```diff
@@ -69,3 +69,113 @@ SOFTWARE.
...
+- <https://github.com/svenvc/ston/blob/master/LICENSE> and
+  <https://github.com/GemTalk/Rowan/blob/masterV3.4/LICENSE>
+
+> Why the last two rows differ from the file they were copied out of:
+> `gt4gemstone`'s `src-gs/Announcements.gs` and `src-gs/STON.gs` are *exports of
+> other projects* that gt4gemstone's `scripts/convertRsrToGsFormat.topaz` writes
+> into its own `src-gs/`; the packages inside them
+> (`Announcements-Core-GemStone…`, `STON-Core` / `STON-GemStone-Kernel`) belong
+> to `GemTalk/Announcements` and to STON respectively, not to feenk. The
+> copyright holders above are the ones named in those projects' `LICENSE` files.
```
MatiasFernandez: 💭 You flagged in the description that STON's lineage is *inferred* — from the byte-identical class comment and the matching Rowan package names — but the blockquote states it flatly ("the packages inside them ... belong to `GemTalk/Announcements` and to STON respectively"). Since this is the file someone reaches for in a licensing question, should it record *how* we know? Something like "identified by package name (`STON-Core` / `STON-GemStone-Kernel`, matching `GemTalk/Rowan` at `platforms/gemstone/projects/STON/src/`) and a byte-identical `STON` class comment; gt4gemstone does not document the export".

Then a future re-vendor can re-check the inference instead of re-deriving it, and it stays honest that nothing upstream says so. No objection to the conclusion — both candidates are MIT, so the notice is complete either way.

THIRD-PARTY.md
```diff
@@ -69,3 +69,113 @@ SOFTWARE.
...
+  post-processed by `gs-src/enhancedInspector/build/apply_jasper_transforms.sh`.
+  The `.gs` files themselves are Rowan/topaz exports produced *by those upstream
+  projects* (their own `src-gs/` directories), not by Jasper.
+- **Modifications:** two deterministic, behaviour-preserving transforms applied
+  by `apply_jasper_transforms.sh` — (1) a per-file attribution header is
+  prepended, and (2) class placement is rewritten from upstream's `Globals` to
+  the dedicated `GsEnhancedInspector` dictionary, so the whole payload can be
+  removed by dropping one dictionary. Otherwise verbatim.
+- **License:** MIT throughout — but across **six** upstream repos with **seven**
+  distinct copyright holders. See the table and the notices below.
```
MatiasFernandez: nit: "**seven** distinct copyright holders" — there are seven copyright *notices*, but five distinct holders: `feenk` appears twice (2021 and 2024) and `GemTalk Systems` twice (2017-2024 and 2020-2021). In a file whose whole point is naming the holders correctly, "seven copyright notices across six upstream repos" would be exact. Same wording in `resources/enhancedInspector/README.md:40`.

NOTICE
```diff
@@ -11,5 +11,19 @@ This product includes vendored third-party software:
   License: https://github.com/pharo-project/pharo/blob/v12.0.0/LICENSE
   Location in this repo: gs-src/refactoring/vendor/rowanv3-ast/.
 
+- The Enhanced Inspector server-side payload -- feenk's GToolkit remote
+  inspector (GT4GemStone, RemotePhlow, wire encoding) plus the libraries it
+  needs on the stone (RemoteServiceReplication, Announcements, STON). MIT
+  License, across six upstream repositories:
```
MatiasFernandez: nit: "across six upstream repositories" is immediately followed by seven repo URLs — `GemTalk/Rowan` being the seventh. `THIRD-PARTY.md` resolves this by writing "six upstream repos (seven counting `GemTalk/Rowan` for the STON port)"; worth the same parenthetical here so the count and the list below it agree.

MatiasFernandez (approved):
I left some comments worth reviewing

PR #563, opened 2026-09-04
Npapagna/non blocking execute

npapagna:


client/src/__tests__/gciLibrary.test.ts
```diff
@@ -211,6 +229,140 @@ describe('GciLibrary', () => {
...
+
+      expectOopToBeTrue(await firstOperation);
+    });
+
+    /**
+     * Makes the next readiness check on the session's socket throw, forcing
+     * the raw-socket fallback path regardless of whether the connected
+     * GemStone version has GciTsNbPoll available.
+     */
+    function simulatePollFailure() {
```
MatiasFernandez: 🚫 I don't think this forces the raw-socket fallback path. `isNbResultReady` only reaches `nativeSocketLibrary.isReadable` when `isAvailable('GciTsNbPoll')` is false; on any 3.7+ stone (which is what CI logs into) the mocked `GciTsNbPoll` returns -1 first, `isNbResultReady` reports ready right there, and the `isReadable` spy is never called. So both tests below pass through the `result !== 0` branch, the `mockThrowOnce` is dead code, and the branch the PR description highlights — raw-socket check throws → `console.warn` → single blocking fetch — has no coverage at all on modern GemStone.

Could we drive that branch directly instead, e.g. stubbing `isAvailable('GciTsNbPoll')` to false so the socket path is the one under test, and keep a separate test for the `GciTsNbPoll` returning -1 case? At minimum the doc comment needs to stop claiming it forces the fallback "regardless of whether GciTsNbPoll is available".

client/src/gciLibrary.ts
```diff
@@ -2304,6 +2314,162 @@ export class GciLibrary {
...
+   *   only) if the readiness check itself failed; `false` if the check
+   *   succeeded and the call isn't ready yet.
+   */
+  private isNbResultReady(session: unknown) {
+    if (this.isAvailable('GciTsNbPoll')) {
+      const { result } = this.GciTsNbPoll(session, 0);
+      return result !== 0;
+    }
+
+    const fd = this.socketFor(session);
```
MatiasFernandez: 💭 If `socketFor` throws here, `executeAndFetchOop` rejects *after* `executeNb` already started a call, and nothing ever fetches the result — so the session stays "busy" and every later call on it fails with `session has a GciTsNb operation in progress` (the very error the test at line 288 pins down). That's the opposite of the tradeoff made three lines below, where a failing readiness check deliberately falls back to a blocking fetch rather than failing the call outright.

Should this one be inside the try as well, so any failure to *check* readiness — including "I can't even find the fd" — degrades to the blocking fetch instead of poisoning the session?

client/src/__tests__/gciLibrary.test.ts
```diff
@@ -211,6 +229,140 @@ describe('GciLibrary', () => {
...
+                true`,
+      );
+
+      expectOopToBeTrue(resultOop);
+    });
+
+    it('throws when the expression signals an error', async () => {
+      await expectToBeRejectedWithGciLibraryError(
+        gciLibrary.executeAndFetchOop(session, `self error: 'oops'`),
+        'a UserDefinedError occurred (error 2318), reason:halt, oops',
```
MatiasFernandez: 💭 The sync sibling for this case uses `expectToThrowExpectedGciLibraryError(...)` so the error text lives in one place. Now that the matcher supports `.rejects`, is it worth an async counterpart of that helper rather than hardcoding `'a UserDefinedError occurred (error 2318), reason:halt, oops'` here? Otherwise the two suites drift the next time the message or error number changes.

client/src/__tests__/support/timers.ts
```diff
@@ -0,0 +1,55 @@
...
+ *   milliseconds, `callback` must allow for.
+ * @param callback - the operation to run and await while sampling.
+ */
+export async function expectEventLoopToRemainResponsiveDuring(
+  pollTimeMs: number,
+  expectedIdleTimeMs: number,
+  callback: () => Promise<unknown>,
+) {
+  let totalIdleTimeMs = 0;
+  const timer = setInterval(() => (totalIdleTimeMs += pollTimeMs), pollTimeMs);
```
MatiasFernandez: 💭 `totalIdleTimeMs` isn't really idle time — it's `numberOfTicks * pollTimeMs`, i.e. how much interval time the timer was *scheduled* for, which over-counts if `setInterval` coalesces late ticks and under-counts partial intervals. Would accumulating real elapsed time between firings (`Date.now()` deltas) be closer to what the name promises, or should the name just say what it counts (`timerFiredForMs` / `sampledTicks`)?

Related: the one caller passes `(100, 800)` against a 1s `Delay`, which leaves ~2 ticks of slack. Any idea how that holds up on a loaded CI box, or should the margin be wider?

client/src/__tests__/support/timers.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Asserts that the event loop is blocked from doing other work for the
+ * entire duration of `callback`'s returned promise, by checking that a timer
+ * sampled every `pollTimeMs` never gets a chance to fire before `callback`
+ * settles.
+ *
+ * @param pollTimeMs - how often, in milliseconds, a responsive event loop
+ *   would let the sampling timer fire.
+ * @param callback - the operation to run and await while sampling.
+ */
+export async function expectEventLoopToBeBlockedDuringDefault(
```
MatiasFernandez: nit: `expectEventLoopToBeBlockedDuringDefault` — is the `Default` suffix intentional? It reads like a leftover; the counterpart is just `expectEventLoopToRemainResponsiveDuring`.

client/src/sockets/factory.ts
```diff
@@ -20,3 +21,17 @@ export function createNativeSocketLibrary(): NativeSocketLibrary {
...
+ * A `NativeSocketLibrary` stub for unit tests that construct a `GciLibrary`
+ * but never poll a socket's readiness. Passing it explicitly keeps
+ * `GciLibrary`'s constructor from calling `createNativeSocketLibrary()`,
+ * which loads the real platform library (and, on Windows, real `koffi` --
+ * unavailable once a test has mocked that module).
+ *
+ * @throws {Error} If `isReadable` is called -- no test using this stub
+ *   should be exercising socket readiness.
+ */
+export function fakeNativeSocketLibrary(): NativeSocketLibrary {
```
MatiasFernandez: 💭 This puts a test double in the production factory, so `FakeNativeSocketLibrary` ships in the extension bundle and is reachable from production code by autocomplete. Would `client/src/__tests__/support/` (next to the new `timers.ts`) be a better home, keeping `factory.ts` to the real platform selection?

client/src/sockets/FakeNativeSocketLibrary.ts
```diff
@@ -0,0 +1,15 @@
+import { NativeSocketLibrary, RawSocketConnection } from './nativeSocketLibrary';
+
+export class FakeNativeSocketLibrary extends NativeSocketLibrary {
```
MatiasFernandez: nit: every other module under `client/src` is camelCase-named (`nativePOSIXSocketLibrary.ts`, `gciLibraryError.ts`), even when it exports a single class. Should this be `fakeNativeSocketLibrary.ts` to match?

client/src/gciLibrary.ts
```diff
@@ -2347,8 +2513,8 @@ export class GciLibrary {
    * past its own return (e.g. by returning it, or by capturing it in
    * something that outlives the call): the oop is released the instant
    * `callback` returns, so any use of it afterwards operates on an already
-   * released oop. Async callbacks are rejected at the type level for this
-   * reason; there is no equivalent check for an oop returned directly.
+   * released oop. Only the synchronous part of that contract is enforced;
```
MatiasFernandez: 💭 This doc edit doesn't seem related to the PR, and it drops the concrete bit: the old text said async callbacks are rejected *at the type level* (which `NotPromise<T>` still does), while "only the synchronous part of that contract is enforced" leaves the reader guessing how. Was this intentional, or a leftover from another branch?

MatiasFernandez (approved):

client/src/__tests__/gciLibrary.test.ts
npapagna: Good catch on the doc comment, but I don't think there's a coverage gap: CI runs the full version matrix on every PR (`client/.gemstone-integration-releases.json`), including 3.6.2 and 3.6.8, both of which predate `GciTsNbPoll` (added in 3.7.2). On those legs `isAvailable('GciTsNbPoll')` is false, so `isNbResultReady` takes the raw-socket branch and the `isReadable` mock is genuinely exercised, not dead code. You can see it pass on this PR's own run, e.g. the `macos-latest 3.6.2` and `ubuntu-latest 3.6.8` jobs.

So each mock covers the version range where it's actually reachable, by design. I did tighten the doc comment though, since "regardless of whether GciTsNbPoll is available" was misleading: it now says only one of the two mocks fires per run, depending on the connected version.

client/src/gciLibrary.ts
npapagna: Confirmed, this is a real bug. Moved `socketFor(session)` inside the `try` so a failure to identify the socket degrades to the blocking-fetch fallback like every other readiness-check failure, instead of poisoning the session. Added a regression test that mocks `socketFor` to throw once; verified it fails against the unfixed code with exactly the symptom you predicted, since the test harness's own cleanup hook trips over `session has a GciTsNb operation in progress`. Also updated the two doc comments that referenced the old (throwing) behavior.

client/src/__tests__/gciLibrary.test.ts
npapagna: Good idea. Added `expectToBeRejectedWithExpectedGciLibraryError`, an async counterpart of `expectToThrowExpectedGciLibraryError`, and pointed this test at it. The expected error text now lives in one place instead of being duplicated as a hardcoded string.

client/src/__tests__/support/timers.ts
npapagna: Good point on both counts. Switched `totalIdleTimeMs` to accumulate real elapsed time between timer firings (`Date.now()` deltas) instead of a flat `pollTimeMs` per tick. That makes the name accurate again, and it also fixes the CI-margin worry: with the old scheme a tick delayed under load only credited the nominal `pollTimeMs`, so a slow box would *undercount* real elapsed time and could fail exactly when it shouldn't. With real deltas a late tick still credits what actually elapsed, so the existing `(100, 800)` margin against the 1s `Delay` should hold up fine without widening it.

client/src/__tests__/support/timers.ts
npapagna: Leftover, no meaning behind it. Renamed to `expectEventLoopToBeBlockedDuring` to match its counterpart.

client/src/sockets/factory.ts
npapagna: Good call, moved it. `FakeNativeSocketLibrary` (and its `fakeNativeSocketLibrary()` factory) now live in `client/src/__tests__/support/fakeNativeSocketLibrary.ts`, so `factory.ts` is back to just real platform selection and the double no longer ships in the extension bundle.

client/src/sockets/FakeNativeSocketLibrary.ts
npapagna: Resolved by the same move as the sibling comment on `factory.ts`: the file's new home is `client/src/__tests__/support/fakeNativeSocketLibrary.ts`, already lowercase-camelCase to match the rest of that directory.

client/src/gciLibrary.ts
npapagna: Leftover, not intentional. Thanks for flagging it. Reverted to the original wording; `NotPromise<T>` is still exactly the enforcement mechanism, so the old text was correct.

Issue #564, opened 2026-09-04
Tabbed Inspector follow-ups: structural editing, filter, multi-select, dynamic instVars, key inspection, status line


ericwinger:
The tabbed Inspector that replaced the side-panel tree (#508) ships read-only structure plus
value editing. These are the items from that issue's wish list that it deliberately left out,
collected here so they can be picked off one at a time rather than tracked across six issues.

Each is independent — none blocks another, and none blocks anything already shipped.

### Structural editing of collections and dictionaries

Add an object to a collection, add a key to a dictionary, remove the selected elements, remove a
key. Today the Inspector can rewrite a slot or an element in place (`instVarAt:put:`, `at:put:`,
a dictionary key) but cannot change how many there are. Jadeite's inspector does all four.

Needs a decision about what the revert affordance means once the shape of the object has changed
— the current revert restores the value a slot held before the first edit, which has no analogue
for a removed element.

### Search / filter within a tab

Matters most on a large collection, where paging gets you to the elements but not to the one you
want. Should filter what the tab is already showing without pulling more pages, and say so.

### Multi-select across rows

Select several instance variables or elements and have the row actions — copy, remove, browse
class — operate on the whole selection. A prerequisite for "remove selected elements" above.

### Dynamic instance variables as their own group

Shown separately from the declared slots, and removable. The Slots tab renders
`allInstVarNames`, which does not include them.

### Inspect a dictionary key separately from its value

The Entries tab dives into the value. The key is often the more interesting object and today
there is no way to reach it.

### A status line

Reports what is selected and what just happened — a write committed, a revert applied, a page
loaded, an evaluation failed. Errors currently reach the output log, which is not where the eye
is.

### Also considered, not proposed here

Two items from #508's wish list are answered rather than deferred, and are recorded so they are
not re-raised:

- **Computed vs. real entries** (`self`, an "as oop" entry, with actions disabled on the computed
  ones). The Evaluate pane's scope list serves the same purpose from a place where it cannot be
  confused for a writable slot.
- **Explicit paging controls.** Load more / Load all, bounded at 50 pages a click, is the
  deliberate replacement for Jadeite's "display more" / "display all".

ericwinger:
## Internal cleanups deferred from the review of #565

Four items from the code review of #565 that are structural rather than user-facing, so they do not
belong in the wish list above but should not be lost either. None changes what the Inspector does;
each was declined during the review round whose job was to fix two correctness bugs, and the reasons
[...]

PR #565, opened 2026-09-04
Replace the side-panel Inspector tree with a tabbed Inspector that needs no server support

### LINKED ISSUE #508: Replace the side-panel Inspector with a tabbed inspector that needs no server support
## Summary

Replace the classic Inspector tree in the GemStone side panel with a tabbed inspector that
looks and behaves like the Enhanced Inspector, but is built entirely on GCI primitives so it
needs **no server code installed**.

The Enhanced Inspector is the better tool, and it is out of reach for a large share of
sessions. It needs its support classes installed in the image *and* a stone at or above its
minimum version (`ENHANCED_INSPECTOR_MIN_VERSION = '3.7.5'` in
`client/src/enhancedInspector/enhancedInspectorInstall.ts`), so on a 3.6.x stone it can never
route there no matter what is installed — `routeInspect` falls back to the classic tree every
time. Those users get a plain read-only tree while everyone else gets tabs, tables and
drill-in.

The classic inspector already proves the data can be had without server support: it renders
entirely from GCI calls in `client/src/debugQueries.ts` (print strings, class names, instance
variable names, indexed slots, dictionary entries). This issue is about putting a modern shell
on that same foundation.

## Everything the classic inspector does today

All of this is current behavior in `client/src/inspectorTreeProvider.ts`,
`client/src/inspectRouter.ts`, and the view/command contributions in `package.json`. It is the
baseline a replacement has to preserve.

**Placement and entry**
- A `TreeDataProvider` tree view, id `gemstoneInspector`, titled *Inspector*, in the GemStone
  sidebar container.
- `Inspect It` (`gemstone.inspectIt`, `$(eye)`) from the editor context menu and bound to
  `Ctrl+K I` / `Cmd+K I`; gated on an active session and no execution in flight.
- `routeInspect` is the single routing point behind every Inspect surface — editor, global and
  debugger — choosing the Enhanced Inspector when the session has support and this tree
  otherwise.

**Roots**
- Holds many inspected objects at once; each Inspect It appends a root rather than replacing.
- Roots carry a caller-supplied label, and can be looked up by it.
- Remove a single root: `gemstone.removeInspectorItem` (`$(close)`), inline, root rows only.
- Clear every root: `gemstone.clearInspector` (`$(clear-all)`), in the view title bar.

**Row rendering**
- Row label is the slot name or index; the row's description is the object's `printString`,
  truncated at 200 characters.
- Tooltip is `ClassName: printString`.
- Distinct icons per row kind — root `eye`, named instance variable `symbol-field`, indexed
  slot `symbol-array`, dictionary association `symbol-key`, range group `symbol-array`.
- A row is expandable only when the object actually has named instance variables or indexed
  slots; `nil` and special (immediate) oops are leaves.

**Structure it can walk**
- Named instance variables, listed by name.
- Indexed elements, labeled `[i]`.
- `SymbolDictionary` gets a dedicated path: children are key/value associations labeled by key.
- Paging for large collections — more than 100 indexed elements are grouped into `[start..end]`
  rows of 100, each showing an "N items" count. Those range rows are virtual: rendering one
  costs no GCI traffic, and the slice is only fetched when it is expanded.

**Session awareness**
- Every node records the session it came from.
- Logging a session out removes just that session's roots and leaves other sessions alone.
- A row whose session is gone renders as `<session disconnected>` with a warning icon rather
  than erroring.

**Arrival feedback**
- On a new root the view gets a badge (`Inspector updated`) and a status-bar item
  (`$(eye) Inspector updated`, warning background) appears; both clear themselves after five
  seconds.
- The view is focused, then focus is handed back to the editor so typing is not interrupted.

**What it deliberately does not do** — the gap this issue closes: no tabs, no tables, no
drill-in/miller columns, no editing of any value, no evaluation pane, no per-type views, no
sorting, filtering or search, no copy action, and no way to get from a value to its class in
the browser.

## Acceptance criteria

- The side-panel inspector presents an object as **tabs**, in the visual idiom of the Enhanced
  Inspector.
- It works against a stone with **no server support installed**, including 3.6.x, using only
  GCI primitives.
- Every capability in the list above still works, or has a stated, deliberate replacement.
- The Enhanced Inspector keeps working exactly as it does today, on the sessions that can
  reach it — this work must not disturb it.

## The Enhanced Inspector stays as it is

Decided: the Enhanced Inspector is **kept as-is** for now, and `routeInspect` keeps its current
rule — the Enhanced Inspector on a session whose image supports it, the new tabbed side-panel
inspector everywhere else. It is expected to be enhanced further in its own right later, so
nothing here should constrain it.

Two consequences for this work:

- This is not a migration. Nothing about the Enhanced Inspector's behavior, routing or install
  path changes, and no part of it should be refactored in service of the new inspector.
- Borrowing from it is one-directional. Reusing its look, its stylesheet or its column model is
  welcome; changing those shared pieces to suit the side panel is not, since it will keep
  evolving separately.

Folding server-provided views into the new inspector as extra tabs — one inspector whose tab set
grows with the image's capability — was considered and is explicitly **not** in scope. If it
ever happens it belongs to the Enhanced Inspector's own future, not here.

## Feature wish list

Drawn from the JadeiteForPharo inspector (`Jadeite-Inspector/JadeiteInspector.class.st` and
`RowanInspectorService`) and from Pharo's own inspector. Offered as a menu to pick from, not a
scope commitment — several of these are worth their own follow-up issues.

**From the Jadeite inspector — things it has that Jasper's classic tree does not**
- **Editing values in place.** Jadeite can save a named instance variable, an indexed element,
  a dictionary entry, an unordered-collection element, or a dynamic instance variable, with the
  usual dirty/revert/discard handling around it. This is the single biggest gap.
- **An evaluation pane** on the object, so you can type an expression against the selected
  value and run it, rather than only reading.
- **Dive with history.** Drilling into a value replaces the inspected object in place, with
  previous/next navigation across the objects visited — a back/forward trail rather than an
  ever-deeper tree.
- **Inspect in a new window**, including inspecting a dictionary *key* separately from its
  value.
- **Dynamic instance variables** shown as their own group, and removable.
- **Computed vs. real entries distinguished**, with `self` and an "as oop" entry alongside the
  real slots, and actions correctly disabled on the computed ones.
- **Structural editing of collections and dictionaries** — add an object, add a key, remove
  selected elements, remove a key, remove from a dictionary.
- **Inspect bytes** — a byte-level view of the object.
- **Browse the class** of the selected value, straight from the inspector.
- **Multi-select** across instance variables, with actions that operate on the whole selection.
- **Explicit paging controls** — "display more" / "display all" against a large collection,
  rather than only fixed-size buckets. Jadeite's server-side cap is 65000.
- **Special handling for string objects**, and a guard for protected items that must not be
  edited.
- **A status line** on the inspector reporting what is selected and what happened.

**From Pharo's inspector — ideas rather than parity**
- **Per-type tabs contributed by the object itself.** A Pharo class declares extra inspector
  tabs with `<inspectorPresentationOrder: N title: '…'>`, which is where tabs like *Items*,
  *Bytes*, *Tree*, *Source*, *Definition*, *Details*, *Icons* and *Colors* come from. The
  interesting idea for Jasper is the shape: a small set of built-in tabs plus a mechanism for
  type-specific ones, so a `Dictionary`, an `OrderedCollection` and a `GsProcess` each present
  themselves well without the inspector hardcoding every case. Without server support that
  mechanism has to live client-side — a table of known GemStone classes to tab builders — which
  is a real design question worth its own discussion.
- **A raw/all-slots tab** that is always present regardless of what other tabs an object offers,
  so there is never an object the inspector cannot show.
- **A meta tab** for class, hierarchy and methods of the inspected object.
- **An evaluation pane bound to the inspected object**, the same idea as Jadeite's.
- **Search/filter within a tab**, which matters most on large collections.
- **Copy the print string / copy the value** as first-class actions.

## Implementation notes

Possible starting points, not prescriptions:

- The data layer already exists and needs no server code: `client/src/debugQueries.ts` supplies
  print strings, class names, instance variable names and oops, indexed size and slices, special
  oop detection, and dictionary entries. That is enough for a raw tab, an items tab and a meta
  tab.
- A side-panel tabbed UI means a `WebviewViewProvider` rather than the current
  `TreeDataProvider`, since a tree view cannot render tabs. The Enhanced Inspector is a
  `WebviewPanel` opened beside the editor; its stylesheet, tab bar and table rendering in
  `client/src/enhancedInspector/enhancedInspector.ts` are the closest reference for making the
  new panel look like it belongs to the same tool.
- The Enhanced Inspector's miller-column model is already factored out into
  `client/src/enhancedInspector/enhancedInspectorColumns.js`, which may be reusable if drill-in
  goes that way rather than Jadeite's in-place dive — read-only reuse, or a copy, rather than
  reshaping a file the Enhanced Inspector still owns.
- Paging deserves care. The existing virtual range rows are cheap precisely because they cost no
  GCI traffic until expanded; whatever replaces them should keep that property, since the
  inspector is often pointed at very large collections.
- Editing values is the largest single item here and could reasonably be split into its own
  issue once the read-only shell exists.

ericwinger:
Fixes #508

The Enhanced Inspector needs its server-side payload installed *and* a stone at 3.7.5 or later,
so on a 3.6.x stone `Inspect It` fell back to a read-only tree in the sidebar no matter what was
installed. That tree is removed outright, along with its Remove and Clear All commands, and
replaced by a basic Inspector that opens as an editor tab in the Enhanced Inspector's own idiom —
tabs over a miller-column strip — built entirely on GCI primitives and kernel sends, so it needs
no server support and works back to 3.6.2.

The Enhanced Inspector is untouched: `routeInspect` keeps its existing rule, and the only file it
shares in this change is the Meta-tab scroll fix, which it had the same bug in.

## What it does

An object gets only the tabs its structure warrants: **Slots**, **Items** or **Entries**,
**Bytes**, **Print**, **Meta** and **Evaluate**.

- Double-click a row to open that object to the right; Enter dives in place. Each column carries
  its own Back/Forward.
- A row's context menu copies its printString or OOP, browses its class, and edits its value
  through `instVarAt:put:`, `at:put:` or a dictionary key, with a revert to the value from before
  the first edit.
- Slots list alphabetically, sorted by clicking the Name column. Each row keeps the
  `instVarAt:` index it was read at, so sorting cannot misdirect a write, and declaration order —
  what `allInstVarNames` answers — is that index ascending.
- Meta shows the class name and an info bar of superclass, package and OOP, then sub-tabs for
  instance methods, class methods, definition and comment.
- The Evaluate pane answers to the editor's own Ctrl+K D / E / I for Display, Execute and Inspect
  It, and lists the names in scope beside the expression: the receiver, then every instance
  variable grouped under the class that declares it. Debug It is absent — it works by compiling
  with the single-step flag set, and the pane's `evaluateInContext:symbolList:` runs to
  completion, leaving no halted process to attach to.
- Every paged tab offers Load all beside Load more, bounded at 50 pages a click so a
  million-element collection cannot hold the session; the tab keeps showing a remainder and
  another click carries on.
- Panels are tracked per session and disposed together when that session logs out — the
  replacement for the tree's per-session root removal.

The column strip both inspectors were growing separately is now one module,
`client/src/webview/millerColumns.js`, which owns order, insert-right drilling, close, focus and
width; each inspector supplies what is drawn inside a column.

## Acceptance criteria

- [x] The side-panel inspector presents an object as tabs, in the Enhanced Inspector's visual
      idiom.
- [x] Works against a stone with no server support installed, including 3.6.x, using only GCI
      primitives.
- [x] Every capability of the classic tree still works or has a stated replacement — multiple
      roots become multiple editor tabs, per-session logout disposal replaces per-session root
      removal, and the arrival badge/status-bar flash is replaced by the tab opening.
- [x] The Enhanced Inspector keeps working exactly as it does today.

## Follow-ups

The wish-list items in #508 that are not in this change are collected in #564 — structural
editing of collections and dictionaries, search/filter within a tab, multi-select, dynamic
instance variables, inspecting a dictionary key separately from its value, and a status line.

## Testing

- `npx vitest run src/basicInspector` — 185 tests across 3 files, all passing.
- `tsc -p client --noEmit` and `eslint client/src/basicInspector` both clean.
- Coverage was measured by recording every function entry across a run rather than assumed; the
  two paths no test reached — `dive` and the `fetchObjectMeta` wrapper — now have tests.
- Exercised by hand against a live stone.


ericwinger:
CodeQL's 30 alerts were one finding reached by 30 paths, and it was right. Fixed in bb8425c6.

The column strip indexed its columns in a plain `{}`, and `Columns.get(id)` is handed ids that
come from outside the model — a host message's `columnId`, and `colEl.dataset.colId` in
`columnOf`. Ask a plain object for `__proto__` and it hands back `Object.prototype`, so the very
next line in `populateColumn` — `col.oop = …`, `col.title = …`, and twenty-eight more — would
have written a column's worth of fields onto the prototype of every object in the frame. Each of
those assignments is what CodeQL flagged, which is why one map produced thirty alerts.

`columnsById` is now `Object.create(null)`, which answers `undefined` for those keys; every
caller already treats `undefined` as "no such column" and returns. The three per-column maps
keyed off the same messages get the same treatment — `tabData` and `loadedRows` by tab name,
`methodSource` by selector — in both the place that seeds them and the place that resets them.

No behavior change: these are lookup tables, nothing iterates them expecting inherited keys, and
the module's only `hasOwnProperty` is the guarded `Object.prototype.hasOwnProperty.call` form
against a different object.

Three tests pin it in `millerColumns.test.ts`, at the model rather than in this inspector, so it
holds for the Enhanced Inspector's strip too: a prototype key answers `undefined`, a child
message naming `__proto__` as its source appends at the far right and leaves bystander objects
clean, and a real id still finds its column.

Worth noting for anyone reading the check history: this push is also the first time the
**Health Check** workflow ran on this PR. The branch was conflicting with `main`, so GitHub could
not build the merge ref the `pull_request` trigger needs, and CodeQL — which analyzes
`refs/pull/565/head` instead — was the only thing that ran. The merge with `main` in bb8425c6
conflicted only in `CHANGELOG.md`, where both sides had appended to Added and to Fixed; both sets
of entries are kept.

client/src/basicInspector/basicInspector.ts
```diff
@@ -0,0 +1,943 @@
...
+  private disposables: vscode.Disposable[] = [];
+  /** Monotonic id handed to each miller column; the root column is 0. */
+  private nextColumnId = 1;
+
+  /**
+   * Slots the user has edited, keyed by {@link slotKey}: the OOP the slot held
+   * before its FIRST edit, so revert always restores the original rather than
+   * the previous edit. Mirrors the debugger's variable-revert bookkeeping.
+   */
+  private undoOriginals = new Map<string, bigint>();
```
npapagna: **Blocker**: BasicInspector's slot-revert undo tracker (undoOriginals/undoDirty/undoPinned) pins and releases OOPs from the session's non-ref-counted GCI export set completely independently of DebuggerPanel's identical tracker on the same session, with no coordination between them.

Failure scenario: A user debugging a frame opens Inspect on a variable (routed to a BasicInspector on the same session via openedInspectors), and both the debugger and the inspector end up tracking the same aliased OOP as an 'original' to revert. Stepping/resuming/restarting in the debugger calls clearUndoState(), releasing that OOP from the session's export set even though BasicInspector's own undoOriginals map still depends on it staying pinned. If the OOP is scavenged and reused before the inspector's revert runs, revertSlot writes the stored OOP number back into whatever unrelated object now holds it, silently corrupting data instead of reverting.

client/src/basicInspector/basicInspector.ts
```diff
@@ -0,0 +1,943 @@
...
+        : row,
+    );
+  }
+
+  /**
+   * Refuse a write while the session is mid-call. One GCI call is in flight per
+   * session at a time, so a write issued now would fail deep in the binding with
+   * a much less helpful message than this one.
+   */
+  private busyError(): string | null {
```
npapagna: **Blocker**: busyError() itself calls session.gci.GciTsCallInProgress unguarded before setSlot/revertSlot/evaluate enter their own try/catch; if that call throws, the exception escapes to handleMessage's outer catch, which only logs and returns without posting any reply to the webview.

Failure scenario: If the session's native GCI handle becomes invalid while a BasicInspector panel is still open (e.g. connection drops before disposeForSession runs) and the user clicks Evaluate or edits/reverts a slot, the webview never receives a result message and is left showing a permanent pending/loading state with no visible error, unlike the old tree provider which explicitly surfaced a disconnected-session warning.

client/src/extension.ts
```diff
@@ -2426,19 +2412,7 @@ export function activate(context: vscode.ExtensionContext) {
     vscode.commands.registerCommand(
```
npapagna: **Blocker**: The old sidebar Inspector's reveal-existing-root dedup for gemstone.inspectGlobal was dropped with no replacement; the new handler unconditionally calls codeExecutor.inspectExpression, and neither routeInspect nor BasicInspector.create/EnhancedInspector.create has any oop/label-keyed lookup of already-open panels.

Failure scenario: A user repeatedly inspects the same global from the GemStone Explorer's Globals view. Previously this focused the one existing inspector root for that global; now every click opens a brand-new editor tab, cluttering the editor with duplicate Inspector tabs for the same object.

client/src/basicInspector/basicInspectorView.js
```diff
@@ -0,0 +1,1322 @@
...
+    });
+  }
+
+  function cancelEdit(col) {
+    col.editing = null;
+    renderTab(col);
+  }
+
+  /** A write landed: every printString and OOP on the tab is stale, so refetch. */
+  function refetchActiveTab(col) {
```
npapagna: **Blocker**: refetchActiveTab clears tabData/loadedRows and calls activateTab, which re-fetches with from:1 and no 'all' flag, so a paged tab that had loaded extra pages via 'Load more'/'Load all' collapses back to page 1 after any successful slot edit or revert.

Failure scenario: On a 500-element Array, click 'Load all' to view rows past index 100, then edit or revert a value at index 150. The write succeeds but the tab redraws showing only rows 1-100 again, so the just-edited row disappears from view until the user re-triggers Load more/Load all.

client/src/basicInspector/basicInspectorView.js
```diff
@@ -0,0 +1,1322 @@
...
+  function handleHostMessage(msg) {
+    var col;
+    switch (msg.command) {
+      case 'addRoot':
+        Columns.addRoot(msg);
+        return;
+      case 'addChild':
+        Columns.addChild(msg);
+        return;
+      case 'replaceColumn':
```
npapagna: **Blocker**: The 'replaceColumn' host-message handler (used for in-place navigation: Enter, Back/Forward, 'Dive Here') calls populateColumn() but never Columns.focus(), which is the only code path that posts 'setTitle' to the extension host.

Failure scenario: Selecting a row and pressing Enter (or using Back/Forward/'Dive Here') to navigate to a different object in the same column recomputes the internal title but the VS Code editor tab keeps showing the previous object's caption. Only opening a brand-new column via double-click updates the tab title correctly.

client/src/basicInspector/queries/basicInspectorQueries.ts
```diff
@@ -0,0 +1,531 @@
...
+  className: string;
+}
+
+/**
+ * Resolve a value's class and the symbol dictionary holding it, for "Browse
+ * Class" on a row. Mirrors `fetchMethodBrowseLocation` in the Enhanced
+ * Inspector's queries, minus STONJSON and minus the method category — this one
+ * browses to the class, not to a selector.
+ */
+export function fetchBrowseLocation(execute: QueryExecutor, oop: bigint): BrowseLocation | null {
```
npapagna: fetchBrowseLocation reimplements the class/dictionary resolution logic (obj class theNonMetaClass, then dictionariesAndSymbolsOf:) already present in getEnhancedInspectorViewSpecs.ts's fetchMethodBrowseLocation, its own doc comment admitting it 'mirrors' that function instead of sharing it, making this a third independent implementation of dictionary-name resolution (debugQueries.ts's getBrowseTarget is the second).

Failure scenario: This is the same class of bug the project has already shipped once (class-dict mismatch in onClassDefinitionCompiled URI): a future fix to how the owning symbol dictionary is chosen has to be found and applied in three separate places, and can easily be fixed in two while silently leaving the third one wrong.

client/src/debugQueries.ts
```diff
@@ -1400,6 +1402,23 @@ function resolveGlobalOop(session: ActiveSession, name: string): bigint | null {
...
+ * evaluation sees the globals the user would see in a workspace.
+ *
+ * Every `evaluateInContext:` send passes one. The single-argument
+ * `String>>evaluateInContext:` looks like the leaner choice for an evaluation
+ * that needs no extra bindings, but it does not exist before GemStone 3.7: on a
+ * 3.6.x stone that perform fails with NameError 2404, "There is no Symbol with
+ * the specified value". The two-argument form is present on every supported
+ * release, so it is the only one used.
+ */
+function sessionSymbolListOop(session: ActiveSession): bigint {
```
npapagna: sessionSymbolListOop() performs three fresh GCI round trips (resolveSymbol 'System', perform 'myUserProfile', perform 'symbolList') on every call even though the result never changes for the life of the session, and it is invoked on the hot path of BasicInspector's evaluate() and setSlot().

Failure scenario: Clicking Display It or editing an inspector slot five times pays 15 wasted synchronous GCI round trips fetching the same OOP each time, adding real, avoidable latency to the interactive eval/edit loop.

client/src/basicInspector/queries/basicInspectorQueries.ts
```diff
@@ -0,0 +1,531 @@
...
+  oop: bigint,
+  from: number,
+  count: number,
+): InspectorRow[] {
+  if (!Number.isInteger(from) || from < 1) return [];
+  if (!Number.isInteger(count) || count < 1) return [];
+  const code = `| obj out ${DUMP_PAYLOAD_TEMPS} keys k v |
+obj := Object _objectForOop: ${oop}.
+out := WriteStream on: String new.
+${dumpPayloadPrelude()}keys := [obj keys asSortedCollection asArray]
```
npapagna: fetchEntries re-sorts a dictionary's entire key set from scratch on every page request, then discards everything but the requested slice.

Failure scenario: 'Load all' on a large SymbolDictionary can trigger up to 50 sequential fetchEntries calls, each redoing the full O(m log m) sort just to extract 100 rows, so the same sort is repeated up to 50 times instead of once.

client/src/basicInspector/queries/basicInspectorQueries.ts
```diff
@@ -0,0 +1,531 @@
...
+out := WriteStream on: String new.
+${dumpPayloadPrelude()}${ROW_BLOCK}last := ${from} + ${count} - 1.
+(obj isKindOf: SequenceableCollection)
+  ifTrue: [
+    ${from} to: (last min: obj size) do: [:i |
+      row value: '[', i printString, ']'
+          value: ([obj at: i] on: Error do: [:e | nil])
+          value: i]]
+  ifFalse: [
+    (obj isKindOf: Collection)
```
npapagna: fetchItems' fallback for a non-sequenceable Collection (Set, Bag, ...) enumerates the entire collection with obj do: on every call, with no early exit once the counter passes the requested page's end index.

Failure scenario: Reading a 10,000-element Set via 'Load all' does roughly 50 full O(10,000) enumerations from the start instead of one streaming O(10,000) pass, since each page re-walks the whole collection from the beginning.

client/src/basicInspector/basicInspector.ts
npapagna: **Blocker**: BasicInspector's slot-revert undo tracker (undoOriginals/undoDirty/undoPinned) pins and releases OOPs from the session's non-ref-counted GCI export set completely independently of DebuggerPanel's identical tracker on the same session, with no coordination between them.

Failure scenario: A user debugging a frame opens Inspect on a variable (routed to a BasicInspector on the same session via openedInspectors), and both the debugger and the inspector end up tracking the same aliased OOP as an 'original' to revert. Stepping/resuming/restarting in the debugger calls clearUndoState(), releasing that OOP from the session's export set even though BasicInspector's own undoOriginals map still depends on it staying pinned. If the OOP is scavenged and reused before the inspector's revert runs, revertSlot writes the stored OOP number back into whatever unrelated object now holds it, silently corrupting data instead of reverting.

client/src/extension.ts
npapagna: **Blocker**: The old sidebar Inspector's reveal-existing-root dedup for gemstone.inspectGlobal was dropped with no replacement; the new handler unconditionally calls codeExecutor.inspectExpression, and neither routeInspect nor BasicInspector.create/EnhancedInspector.create has any oop/label-keyed lookup of already-open panels.

Failure scenario: A user repeatedly inspects the same global from the GemStone Explorer's Globals view. Previously this focused the one existing inspector root for that global; now every click opens a brand-new editor tab, cluttering the editor with duplicate Inspector tabs for the same object.

client/src/basicInspector/basicInspectorView.js
npapagna: **Blocker**: refetchActiveTab clears tabData/loadedRows and calls activateTab, which re-fetches with from:1 and no 'all' flag, so a paged tab that had loaded extra pages via 'Load more'/'Load all' collapses back to page 1 after any successful slot edit or revert.

Failure scenario: On a 500-element Array, click 'Load all' to view rows past index 100, then edit or revert a value at index 150. The write succeeds but the tab redraws showing only rows 1-100 again, so the just-edited row disappears from view until the user re-triggers Load more/Load all.

client/src/basicInspector/basicInspectorView.js
npapagna: **Blocker**: The 'replaceColumn' host-message handler (used for in-place navigation: Enter, Back/Forward, 'Dive Here') calls populateColumn() but never Columns.focus(), which is the only code path that posts 'setTitle' to the extension host.

Failure scenario: Selecting a row and pressing Enter (or using Back/Forward/'Dive Here') to navigate to a different object in the same column recomputes the internal title but the VS Code editor tab keeps showing the previous object's caption. Only opening a brand-new column via double-click updates the tab title correctly.

client/src/basicInspector/queries/basicInspectorQueries.ts
npapagna: fetchBrowseLocation reimplements the class/dictionary resolution logic (obj class theNonMetaClass, then dictionariesAndSymbolsOf:) already present in getEnhancedInspectorViewSpecs.ts's fetchMethodBrowseLocation, its own doc comment admitting it 'mirrors' that function instead of sharing it, making this a third independent implementation of dictionary-name resolution (debugQueries.ts's getBrowseTarget is the second).

Failure scenario: This is the same class of bug the project has already shipped once (class-dict mismatch in onClassDefinitionCompiled URI): a future fix to how the owning symbol dictionary is chosen has to be found and applied in three separate places, and can easily be fixed in two while silently leaving the third one wrong.

client/src/debugQueries.ts
npapagna: sessionSymbolListOop() performs three fresh GCI round trips (resolveSymbol 'System', perform 'myUserProfile', perform 'symbolList') on every call even though the result never changes for the life of the session, and it is invoked on the hot path of BasicInspector's evaluate() and setSlot().

Failure scenario: Clicking Display It or editing an inspector slot five times pays 15 wasted synchronous GCI round trips fetching the same OOP each time, adding real, avoidable latency to the interactive eval/edit loop.

client/src/basicInspector/queries/basicInspectorQueries.ts
npapagna: fetchEntries re-sorts a dictionary's entire key set from scratch on every page request, then discards everything but the requested slice.

Failure scenario: 'Load all' on a large SymbolDictionary can trigger up to 50 sequential fetchEntries calls, each redoing the full O(m log m) sort just to extract 100 rows, so the same sort is repeated up to 50 times instead of once.

client/src/basicInspector/queries/basicInspectorQueries.ts
npapagna: fetchItems' fallback for a non-sequenceable Collection (Set, Bag, ...) enumerates the entire collection with obj do: on every call, with no early exit once the counter passes the requested page's end index.

Failure scenario: Reading a 10,000-element Set via 'Load all' does roughly 50 full O(10,000) enumerations from the start instead of one streaming O(10,000) pass, since each page re-walks the whole collection from the beginning.

npapagna (changes requested):
Thanks @ericwinger!
I believe blockers are worth fixing, let me know your thoughts!

client/src/basicInspector/basicInspector.ts
ericwinger: Fixed in ca119c52.

Export-set pins are now ref-counted per session, in a new `client/src/exportSetPins.ts`: `GciTsSaveObjs` runs when the first holder claims an OOP, `GciTsReleaseObjs` only when the last one lets go. Both trackers hold their pins through it — `DebuggerPanel`'s `captureUndoOriginal`/`clearUndoState` and `BasicInspector`'s `captureOriginal`/`releasePins` — and logout drops the session's bookkeeping.

Worth recording that this was reachable before this PR: two debugger panels on one session hit the same thing, which is why `releaseObjs`' own comment already said "a session may host more than one debugger panel". So it is a fix to shipped behaviour, and it has a CHANGELOG entry of its own.

Nine tests in `exportSetPins.test.ts`, including your scenario directly: two holders on one OOP, the first release must not unpin it, and the second must.

client/src/extension.ts
ericwinger: Fixed in ca119c52.

`BasicInspector.revealExisting(session, label)` focuses a panel already open on that name; `revealInspect` in `inspectRouter.ts` is the routing-aware entry point, and `gemstone.inspectGlobal` asks it before executing anything.

The scope is exactly where the tree's rule was — only that command, and the Enhanced Inspector stays excluded from it, as it was excluded before. An editor **Inspect It** still opens a panel per invocation, since two inspectors on one object side by side is a reasonable thing to want.

Since c63d2a2e the dedup follows `gemstone.inspector.preferred`, so it keeps working on a session that has the Enhanced Inspector but has been told to use the basic one, rather than quietly switching itself off.

client/src/basicInspector/basicInspectorView.js
ericwinger: Fixed in ca119c52.

The refetch now carries a `through` row count — the length the tab already had — and `readPages` keeps reading pages until it reaches it, under the same ceiling a "Load all" uses. Your 500-element Array case comes back 300 rows long instead of 100, with the edited row still there.

Two follow-ons from testing this by hand:

- The ceiling itself was silent, so a **Load all** that stopped at 5,000 rows looked like a bug. It now says so in the toolbar and names the setting that raises it (`gemstone.inspector.loadAllPageLimit`, new in the same commit).
- A rebuilt table still started at offset zero with nothing selected, so the row just written scrolled out of sight even though it was loaded. 2eea6c3d notes the scroll offset and the row on the way into a write and puts both back — on a failed write too, since that redraws the table as well.

client/src/basicInspector/basicInspectorView.js
ericwinger: Fixed in ca119c52.

The `replaceColumn` handler now calls `Columns.focus(col, true)` after `populateColumn`. `focus` is the only path that posts `setTitle`, and an in-place dive doesn't change which column has focus — hence the forced call.

Pinned by a view test: a column that dives in place from an `Account` to a `ScaledDecimal` must post the new title.

client/src/basicInspector/queries/basicInspectorQueries.ts
ericwinger: Fixed in ca119c52 — and it was four implementations, not three.

`homeDictionaryNameExpr(classVar)` in `queries/util.ts` is now the single answer to "which dictionary owns this class", built on the `symbolListIndexOfClassExpr` that already lived there and already documented itself as the one home for the identity idiom. Callers: `fetchBrowseLocation` (this one), `fetchMethodBrowseLocation` (Enhanced Inspector), `getBrowseTarget` and `getReceiverClassChain` (debugger Browse and Implement), and `classDefiningDictionaryName` (a class rename's scope).

Two of those were wrong in the way you predicted:

- `(… dictionariesAndSymbolsOf: cls) first first name` answers *every* binding, alias entries included — Python's `#object` → `Object` sorts ahead of `Globals` — so its first pair can name a dictionary that merely references the class; and `first` on an empty result raised, so Browse failed outright for a class in no dictionary instead of saying where it could not go.
- The debugger's own loop matched on the class's *name* alone and kept the **last** dictionary that matched.

The shared rule is the one the rename scope already used: the first symbol-list slot binding the class object by identity, under its own name, or '' when nothing does. That change to released behaviour has its own CHANGELOG entry. Guard tests live with the expression, and both inspectors' query tests now assert they no longer emit `dictionariesAndSymbolsOf:`.

Related, from testing: `Browse Class` on a row now navigates the **GemStone Explorer** rather than the System Browser (277edfac), passing that resolved dictionary to `gemstone.explorer.findClass` — which needed a dictionary argument of its own, since it had been picking the first entry of a given name.

client/src/debugQueries.ts
ericwinger: Fixed in ca119c52.

`sessionSymbolListOop` is memoized on the `ActiveSession` (`symbolListOop`), so the three performs happen once instead of once per Display It, Execute It, inspector evaluation and slot edit.

It is safe to hold for the life of the session: the profile's `SymbolList` is a persistent object reached from the profile, and adding or removing a dictionary mutates that object rather than replacing it. Only `UserProfile>>symbolList:` installing a whole new list would leave the memo stale, and the memo dies with the session anyway. That reasoning is in the doc comment rather than left implicit.

A test drives two evaluations and asserts one `myUserProfile` and one `symbolList` perform across both.

client/src/basicInspector/queries/basicInspectorQueries.ts
ericwinger: Fixed in ca119c52, together with the `fetchItems` walk below — both go through one new `PAGE_SNAPSHOT` block.

The sorted key Array is built once and kept in a session temp (`JasperInspectorKeys`), so the pages after the first index into it instead of re-sorting. A "Load all" over a large `SymbolDictionary` now sorts once rather than up to fifty times.

Two guards on reuse, so the memo can't hand back an ordering that outlived what it was taken from:

- the **first page of every read always rebuilds** it, so re-opening a tab reads the dictionary as it is now — the memo holds one read's ordering still while its later pages arrive, it is not a cache of the object between reads;
- a later page reuses it only when it was built for this same object at this same size.

One slot per tab bounds what a session keeps alive; two columns paging different objects turn by turn simply miss and pay what they used to.

client/src/basicInspector/queries/basicInspectorQueries.ts
ericwinger: Fixed in ca119c52 — same `PAGE_SNAPSHOT` block as the Entries sort above.

An unordered Collection is now paged through an Array snapshot of its `do:` order (`obj asArray`, session temp `JasperInspectorItems`), so reading a 10,000-element Set is one pass plus indexed reads rather than fifty walks from the start. Rows are still numbered by position in that order and still carry no write index, since there is no positional write to make.

Same two guards as above: the first page of a read always rebuilds, later pages reuse only for the same object at the same size.

Verified by hand on a 50,000-element `Set` and a `Bag` — which also turned up that `Load all` stopping at its 5,000-row ceiling said nothing at all, so that now announces itself and has a setting.

client/src/basicInspector/basicInspectorView.js
```diff
@@ -0,0 +1,1430 @@
...
+    tabs.push({ id: 'eval', label: 'Evaluate' });
+    return tabs;
+  }
+
+  /** Total rows behind a paged tab, so the toolbar can say "N of M". */
+  function totalFor(col, tab) {
+    if (!col.header) return 0;
+    if (tab === 'items') return col.header.itemCount;
+    if (tab === 'entries') return col.header.entryCount;
+    if (tab === 'bytes') return col.header.itemCount;
```
npapagna: **Blocker**: The Bytes tab's total count uses the header's character count instead of the physical byte count, so a multi-byte String's Load more/Load all toolbar can disappear before every byte is readable.

Failure scenario: `fetchObjectHeader` computes `itemCount` as `obj size` (character count) for any Collection, while `fetchBytes` reads the physical byte count via `_basicSize`/`_basicAt:`. For a GemStone DoubleByteString/QuadByteString, character count is smaller than the physical byte count (e.g. 300 characters stored as 600 bytes). `totalFor` returns that character count as the Bytes tab's total, and the toolbar only offers Load more/Load all when `total > bytes.length`. Once 300 bytes have been read, `total (300) > bytes.length (300)` is false even though 600 real bytes exist, so the remaining 300 bytes become permanently unreachable through the UI with no error or truncation notice.

client/src/basicInspector/basicInspectorView.js
```diff
@@ -0,0 +1,1430 @@
...
+        return;
+    }
+  }
+
+  /** Merge a page into the tab's accumulated data, then draw it. */
+  function applyTabData(col, msg) {
+    // Whether THIS read stopped at the ceiling, and what one click is worth —
+    // replaced by every reply, so the note disappears as soon as a later click
+    // reaches the end of the object.
+    col.stoppedAtLimit = msg.stoppedAtLimit === true;
```
npapagna: **Blocker**: `applyTabData` stores `stoppedAtLimit`/`loadAllRows` on the column instead of per-tab, so switching tabs on the same column can show stale load-limit state from whichever tab fetched last.

Failure scenario: Load pages on a column's Items tab until Load all stops at its ceiling (`col.stoppedAtLimit = true`, `col.loadAllRows` set), then switch to that column's Bytes tab, a fresh fetch since `col.tabData['bytes']` was still undefined, whose reply overwrites `col.stoppedAtLimit`/`col.loadAllRows` with the Bytes tab's own ceiling. Switching back to Items hits the cache path and renders straight from the cached rows without re-fetching, so the ceiling note and load-more buttons show the Bytes tab's stale numbers on the Items tab: a wrong, user-visible "Load all stopped at N" message.

client/src/basicInspector/queries/basicInspectorQueries.ts
```diff
@@ -0,0 +1,590 @@
...
+} from '../../queries/dumpPayload';
+
+/** Rows fetched per page of Items, Entries or Bytes. */
+export const PAGE_SIZE = 100;
+
+const VALID_SELECTOR =
+  /^[a-zA-Z_][a-zA-Z0-9_]*:?$|^([a-zA-Z_][a-zA-Z0-9_]*:)+$|^[+\-*/<>=~&|@%?,]{1,2}$/;
+
+/** Guards a selector before it is interpolated into a doit as `#'...'`. */
+export function isValidSelector(selector: string): boolean {
```
npapagna: **Nice to fix**: `VALID_SELECTOR`/`isValidSelector` here duplicate the existing export of the same name from `enhancedInspector/queries/getEnhancedInspectorViewSpecs.ts` verbatim instead of importing it.

This regex guards a selector before it's interpolated unescaped into a doit as `#'...'`, a selector-injection guard now living in two places. If an edge case in the guard is found later and only one copy gets patched, the other inspector stays exploitable via a crafted selector string.

client/src/basicInspector/basicInspector.ts
```diff
@@ -0,0 +1,1065 @@
...
+      selector: string;
+      isClassSide: boolean;
+    }
+  | { command: 'browseClass'; oop: string }
+  | { command: 'copyText'; text: string; what: string }
+  | { command: 'setTitle'; title: string }
+  | { command: 'closePanel' };
+
+export class BasicInspector {
+  private static panels = new Map<number, Set<BasicInspector>>();
```
npapagna: **Nice to fix**: `BasicInspector`'s per-session webview-panel registry (`static panels = new Map<number, Set<BasicInspector>>()`, `create()`/`close()`/`disposeForSession()`) re-implements `EnhancedInspector`'s near-identical bookkeeping (enhancedInspector.ts:112-166) instead of sharing one implementation.

A leak/double-dispose bug was already fixed once in one of these copies (commit c63d2a2e); a future fix to the same class of bug in one inspector won't automatically propagate to the other.

client/src/basicInspector/basicInspectorView.js
```diff
@@ -0,0 +1,1430 @@
...
+      keyOop: row.keyOop,
+      editable: !!kind && (kind === 'entry' ? !!row.keyOop : !!row.index),
+    };
+  }
+
+  function showCtxMenu(x, y, target) {
+    ctxTarget = target;
+    var edit = ctxMenu.querySelector('[data-action="edit"]');
+    if (edit) edit.style.display = target.editable ? '' : 'none';
+    ctxMenu.style.display = 'block';
```
npapagna: **Nice to fix**: Unlike `enhancedInspector.ts`'s context menu (which clamps to the viewport with `Math.min(e.clientX, window.innerWidth - w - 4)` and closes on Escape), `showCtxMenu` here places the menu at the raw cursor position with no viewport clamp, and there's no Escape handler wired to `hideCtxMenu` (only the outside-click listener closes it).

Right-clicking a row near the right or bottom edge of the window can open the context menu partially or fully off-screen, and a keyboard-only user has no way to dismiss it short of clicking elsewhere.

client/src/basicInspector/basicInspector.ts
```diff
@@ -0,0 +1,1065 @@
...
+    const busy = this.busyError();
+    if (busy) {
+      this.postSlotResult(msg.columnId, false, busy);
+      return;
+    }
+    const oop = BigInt(msg.oop);
+    try {
+      // Evaluate BEFORE capturing the original: a bad expression must leave the
+      // slot — and the undo bookkeeping — completely untouched.
+      const valueOop = debug.evaluateWithReceiverToOop(this.session, oop, msg.expression);
```
npapagna: **Nice to fix**: `setSlot` performs 3 sequential GCI round trips (`evaluateWithReceiverToOop`, then a separate `captureOriginal` read, then a separate `writeSlot` write) on every edit, where 2 would do.

`captureOriginal`'s read and `writeSlot`'s write could be combined into one doit (read-then-write-then-return-old), cutting a full round trip out of the most common edit path on a slow stone.

client/src/basicInspector/queries/basicInspectorQueries.ts
```diff
@@ -0,0 +1,590 @@
...
+      and: [(memo at: 1) == obj2 and: [(memo at: 2) = sz]]])
+    ifTrue: [memo at: 3]
+    ifFalse: [ | arr |
+      arr := build value.
+      SessionTemps current at: key put: (Array with: obj2 with: sz with: arr).
+      arr]].
+`;
+
+/** Emits one `label \t value \t oop \t class \t index` record. Needs `out`, `tab`, `esc`, `psOf`. */
+const ROW_BLOCK = `row := [:lbl :obj2 :idx |
```
npapagna: **Nice to fix**: Three near-identical row parsers (`parseSlotRows`, `parseRows`, `parseEntries`) and two near-identical Smalltalk row templates (`ROW_BLOCK`, `SLOT_ROW_BLOCK`) duplicate the same five-field record shape, differing by only one bonus field.

A future change to the common record shape (e.g. adding escaping or a new column) has to be applied by hand in three parsers and two Smalltalk templates; missing one silently drops or misparses data on just that one tab instead of failing loudly.

client/src/basicInspector/basicInspector.ts
```diff
@@ -0,0 +1,1065 @@
...
+  ): void {
+    const exec = this.makeExecutor();
+    const payload: Record<string, unknown> = { command: 'tabData', columnId, tab, from };
+    const maxPages = loadAllPageLimit();
+
+    switch (tab) {
+      case 'slots':
+        payload.rows = this.stampRevertible(oop, fetchSlots(exec, oop), 'instvar');
+        break;
+      case 'items': {
```
npapagna: **Nice to fix**: `postTabData`'s items/entries/bytes switch branches repeat the same `readPages` -> stamp/assign -> `stampCeiling` shape three times, varying only the page size and fetch function.

A shared helper like `pagedTab(pageSize, fetch, wrap)` would cut roughly 30 lines of repetition, and a fix to the ceiling/`through` bookkeeping (e.g. the bytes case's `PAGE_SIZE * 4`) currently has to be applied identically in three places.

npapagna (changes requested):
Reviewed the latest merge. Two correctness bugs I'd call blockers, plus a handful of duplication/efficiency items that are worth fixing but shouldn't hold up the PR. Let me know your thoughts!

client/src/basicInspector/basicInspectorView.js
ericwinger: Fixed in fb04423a — and you were right to call it a blocker; it is worse than one class.

I checked the premise on a live 3.7.5 stone rather than reasoning about it, and `size` diverges from
`_basicSize` for every wide character type, not just `DoubleByteString`:

| class | `size` | `_basicSize` |
|---|---|---|
| `String`, `ByteArray` | 3 / 4 | 3 / 4 |
| `DoubleByteString`, `Utf16`, `Unicode16`, `DoubleByteSymbol` | 3 | **6** |
| `QuadByteString`, `Unicode32`, `QuadByteSymbol` | 3 | **12** |

`_basicAt:` answers plain 0-255 bytes throughout, so the dump's values were never wrong — only the
total was.

`ObjectHeader` now carries `byteSize`, read with the same `_basicSize` send `fetchBytes` pages by
and guarded by the same `isBytes` test that decides whether the tab exists at all, so the total
cannot disagree with what a page can fetch whatever the class does with `size`.

Reproduction, `(DoubleByteString new: 300) atAllPut: $a; yourself` → Bytes:

- now: `Showing 400 of 600 bytes`, Load more and Load all both offered; one Load more reaches
  `600 of 600`.
- before: `Showing 400 of 300 bytes` and **neither button**, with 200 bytes unreachable — the count
  was already visibly nonsense at that point.

Covered from the query side (`byteSize` parsed apart from `itemCount`; the doit asks with
`_basicSize`), from the webview side (a `QuadByteString` whose total must be 1200 and not 300), and
by a live integration case that builds a real `DoubleByteString` and pins `itemCount` 3 against
`byteSize` 6.

While looking at the dump I also found the hex values were space-padded, so a zero byte printed
` 0` and a wide String read `61  0 61  0` instead of `61 00 61 00`, losing the pairing that shows
two bytes to a character. Fixed in the same commit, tested in both radices, with the blank cells
past the end of a short line staying blank rather than filling with zeroes.

client/src/basicInspector/basicInspectorView.js
ericwinger: Fixed in fb04423a. `stoppedAtLimit` and `loadAllRows` are now prototype-less maps keyed by tab name,
written against `msg.tab`, and `moreButtons`/`ceilingNote` take the tab they are drawing for.

Your trace is exactly right, including the detail that the return trip renders from cache and never
refetches, so stale state is all the tab has to draw from. Reproduced two ways:

`(DoubleByteString new: 9000) atAllPut: $a; yourself` — sized so Items overruns its 5,000-row
ceiling while Bytes (18,000 bytes, 45 pages of 400) finishes inside its 50-page one. Items → Load
all → Bytes → Items:

- now: `Showing 5000 of 9000` with `Load all stopped at 5000 — click it again for the next 5000`.
- before: the note **disappeared** — the Bytes reply carried `stoppedAtLimit: false` and cleared the
  flag Items had earned.

`(DoubleByteString new: 200000)` with a Load all on Bytes too gives the wrong-number version you
described: Items came back saying `Load all stopped at 20000 — click it again for the next 20000`,
and its Load all tooltip `one click reads 20000`, against its own `Showing 5000 of 200000`.

Both are unit-tested. Values above were read off the real view driven in jsdom, once with the fix
and once with the previous version swapped back in, rather than computed.

client/src/basicInspector/queries/basicInspectorQueries.ts
ericwinger: Done in fb04423a. This was the nice-to-fix worth taking outright — an injection guard is the one
case where "patched in one copy is patched in neither" actually costs something.

It has one definition now, in `client/src/queries/util.ts`, which both inspectors' queries already
imported `escapeString` and `homeDictionaryNameExpr` from. Both local copies are gone. Its test
moved with it, to `queries/__tests__/isValidSelector.test.ts`, since the old filename claimed it
belonged to the Enhanced Inspector; the basic inspector's queries test keeps the case that pins
`fetchMethodSource` actually going through the guard, and drops its duplicate of the guard's own
cases.

An unrecognised selector still refuses without reaching the stone, same as before.

client/src/basicInspector/basicInspectorView.js
ericwinger: Done in fb04423a. Clamped with the same `Math.min(..., innerWidth - w - 4)` shape
`enhancedInspector.ts` uses, and Escape folded into the existing document keydown listener rather
than a second one.

Two things worth flagging since they were not obvious from the outside:

**The cell editor's Escape still wins.** Its input handler calls `stopPropagation`, so Escape inside
an open editor cancels the edit and does not reach the menu handler.

**The clamp needed a `Math.max(0, ...)` too.** A menu measuring taller than the window would
otherwise clamp to a negative top and put its first items off the top edge, unreachable — the same
failure the fix was for, at the other end.

On testing it: my first version of the clamp test was near worthless. jsdom lays nothing out, so
`offsetWidth` is 0 and it only ever exercised the 80x30 fallback. It now stubs a real 200x120 menu
so the measured arithmetic runs, plus the oversized-menu case above. Escape is tested directly.

client/src/basicInspector/basicInspector.ts
ericwinger: Agreed on the substance, and not taken in this change.

Deduplicating it means editing `enhancedInspector.ts`'s panel lifecycle, and leaving that file's
behaviour alone is the claim this PR rests on — it is what makes "the Enhanced Inspector keeps
working exactly as it does today" checkable rather than something a reviewer has to take on trust.
Rewriting its dispose path in the last review round trades a real risk for a structural improvement
that will be just as available next week.

You are right about the direction of the risk, though: the leak/double-dispose fix landed in one
copy only, so the other still has whatever that fix was for. Worth doing as its own change, where
it can be reviewed as a lifecycle change rather than as a footnote to an inspector.

Not yet added to #564 — say the word and I will, or fold it in when you next touch that list.

client/src/basicInspector/basicInspector.ts
ericwinger: Not taken in this change, and this is the one I am least willing to do under review.

The saving is real — one round trip of three — but `setSlot`'s read-then-write is the sequence the
revert affordance depends on: `captureOriginal` records the OOP the slot held before its *first*
edit, so a revert restores the original rather than the previous edit. Collapsing the read and the
write into one doit changes when that capture happens relative to the write, on the one path in
this PR where getting it wrong silently loses a user's original value rather than showing an error.

It is worth doing, with its own tests for the first-edit and repeat-edit cases, and not worth doing
alongside two correctness fixes. Not yet added to #564.

client/src/basicInspector/queries/basicInspectorQueries.ts
ericwinger: Fair, and not taken in this change.

The three parsers and two templates are genuinely one record shape with one bonus field each, and
your failure mode is the right one to worry about: a missed parser drops or misparses data on one
tab silently rather than failing loudly.

But it is the wire format for every tab in the inspector, and the change would touch Slots, Items
and Entries at once — during the review round whose job is to fix two correctness bugs. The blast
radius is every tab; the payoff is future edits to a shape that is not currently changing.

Worth doing before the next feature that adds a column, which is when the duplication actually
starts costing. Not yet added to #564.

client/src/basicInspector/basicInspector.ts
ericwinger: Agreed, and this is the closest call of the four I left out — it is contained in one file, well
covered by the panel tests, and directly adjacent to the ceiling bookkeeping I was already fixing.

I left it because the fix in this round changed the *webview* side of that bookkeeping, and I did
not want the diff a reviewer has to check for correctness to also carry a 30-line restructuring of
the host side. `pagedTab(pageSize, fetch, wrap)` is the right shape and I would take it as the first
follow-up.

One correction to the premise, for whoever picks it up: `stampCeiling`'s argument is not identical
across the three, which is part of why it reads as repetitive but is not quite copy-paste — items
and entries pass `maxPages * PAGE_SIZE`, bytes `maxPages * PAGE_SIZE * 4`, because a byte page is
four times a row page. The helper needs the page size threaded through, not hardcoded.

Not yet added to #564.

npapagna (approved):
🚢

client/src/basicInspector/basicInspector.ts
ericwinger: Tracked now, so resolving this does not lose it: #564 carries all four of the deferred cleanups in a
separate section from that issue's user-facing wish list, each with the reasoning from these threads
and a link back here — https://github.com/GemTalk/Jasper/issues/564#issuecomment-5606173308

Resolving to unblock the merge, since `main` requires every thread resolved. Reopen any of them if
you would rather it were done here.

client/src/basicInspector/basicInspector.ts
ericwinger: Tracked now, so resolving this does not lose it: #564 carries all four of the deferred cleanups in a
separate section from that issue's user-facing wish list, each with the reasoning from these threads
and a link back here — https://github.com/GemTalk/Jasper/issues/564#issuecomment-5606173308

Resolving to unblock the merge, since `main` requires every thread resolved. Reopen any of them if
you would rather it were done here.

client/src/basicInspector/queries/basicInspectorQueries.ts
ericwinger: Tracked now, so resolving this does not lose it: #564 carries all four of the deferred cleanups in a
separate section from that issue's user-facing wish list, each with the reasoning from these threads
and a link back here — https://github.com/GemTalk/Jasper/issues/564#issuecomment-5606173308

Resolving to unblock the merge, since `main` requires every thread resolved. Reopen any of them if
you would rather it were done here.

client/src/basicInspector/basicInspector.ts
ericwinger: Tracked now, so resolving this does not lose it: #564 carries all four of the deferred cleanups in a
separate section from that issue's user-facing wish list, each with the reasoning from these threads
and a link back here — https://github.com/GemTalk/Jasper/issues/564#issuecomment-5606173308

Resolving to unblock the merge, since `main` requires every thread resolved. Reopen any of them if
you would rather it were done here.

